PR #51, opened 2019-07-27
Add Docker entrypoint to handle db migration

siomiz:
This adds a new entrypoint for the Docker image where the db migration will be applied on container creation. It makes the stack easier to run on Docker swarm.

This also slightly reduces the final Docker image size by using `COPY --chown` command.
`COPY` then `RUN chown -R` takes 2x space for the layer.
(Directory `.` still needs to be `chown`ed for tmp stuff...)

Thank you for your hard work!

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,22 @@
+#!/bin/sh
+set -e
+
+: ${DATABASE_HOST:='127.0.0.1'}
+
+# wait until the postgres port opens
+sleep 1s
+p=1
+while : ; do
+  set +e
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+set -e
+
+: ${DATABASE_HOST:='127.0.0.1'}
```
adriankumpf: Another PR added a `DATABASE_PORT` env variable. Could you please add that here?

adriankumpf:
Awesome! I really like that solution. Thanks!

PR #52, opened 2019-07-27
Support custom database port through environment DATABASE_PORT env var

mvaal:
Based on example from https://github.com/elixir-ecto/ecto


mvaal:
I have never used elixir, so please double check this is correct.

adriankumpf:
Well done. Thanks!

Issue #49, opened 2019-07-26
How to find $car_id?


ChromoX:
```
teslamate_1  | 21:55:04.269 [info] Found car 'Tesla'
teslamate_1  | 21:55:04.278 [info] Running TeslaMateWeb.Endpoint with cowboy 2.6.3 at :::4000 (http)
teslamate_1  | 21:55:04.279 [info] Access TeslaMateWeb.Endpoint at http://homeservices
teslamate_1  | 21:55:04.940 [info] Start / :online
```
I am attempting to get Tasker setup properly by using the given links from the readme:
```
PUT https://teslamate.your-domain.com/car/$car_id/logging/resume
PUT https://teslamate.your-domain.com/car/$car_id/logging/suspend
```
I've tried filling `$car_id` with `1`, `0`, `Tesla`, `tesla`.
I've tried looking in the HTML to figure out what the ID is, but couldn't find mention of it.
With all the above I continually get 404 Not Found errors.

```
teslamate_1  | 02:53:02.327 [info] PUT /car/1/logging/resume
teslamate_1  | 02:53:02.328 [info] Sent 404 in 299µs
teslamate_1  | 02:53:02.328 [info] Converted error Phoenix.Router.NoRouteError to 404 response
```

Does this application not work immediately upon setup? Do I need to drive the car once or something like that to get all the state in the right place?

Rfvgyhn:
Looks like the docs are incorrect. The car route is [defined](https://github.com/adriankumpf/teslamate/blob/b8bd368c4794a1cad5dd1a8eb377c65c01de36ca/lib/teslamate_web/router.ex#L23) under an `api` route. So you should send your requests to teslamate.your-domain.com/**api**/car/$car_id/logging/resume

ChromoX:
That fixed it! Thanks!

PR #53, opened 2019-07-27
Update README.md to fix resume and suspend logging PUT requests.

ChromoX:
This fixes the README.md HTTP paths for API resume, and suspend.

This update would have prevented #49.


adriankumpf:
Thanks you!

PR #54, opened 2019-07-27
Removing DATABASE_PORT from docker-compose example

mvaal:
DATABASE_PORT is useful when running manually without docker-compose but
the DATABASE_HOST db is referring to the docker-compose postgres which
is on the network created in the docker-compose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,6 @@ services:
     environment:
       - POSTGRES_USER=teslamate
       - POSTGRES_PASSWORD=secret
-    ports:
-      - 5432:5432
     volumes:
       - teslamate-db:/var/lib/postgresql/data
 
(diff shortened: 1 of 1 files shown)


mvaal:
This brings up the question as to whether we need to expose the port at all on the postgres. I am assuming grafana is using it.

adriankumpf:
Nope, exposing the database port is not required in order to run the application. It may be useful if one would want to access the db from outside the docker container, though.

adriankumpf:
But I'm all for keeping the `docker-compose.yml` simple. Let's remove the `DATABASE_PORT` from the example.

mvaal:
True, but for the sake of the example probably not needed.

adriankumpf:
Yeah, you're right. Let's also remove it. Do you want to include it in this PR?

mvaal:
Done.

adriankumpf:
Cool, thanks!

PR #69, opened 2019-07-30
add TZ environment variable to sample docker-compose

normalfaults:
If TZ is not set in container Tesla API auth fails


adriankumpf:
Thanks for the PR! 

Which error did you get? I'm running the container without `TZ` and everything seems works fine.

normalfaults:
looks like its fixed in the latest container ... maybe add a instruction for `docker-compose down; docker-compose pull; docker-compose up -d` for updating the stack 

closing the pull request

PR #70, opened 2019-07-31
Introduce custom teslamate/grafana Docker image

adriankumpf:


Rfvgyhn:
Would introducing an `ARG` for the language be preferrable to getting rid of the translations? I also got started on a PR for this (but got side-tracked with some docker hub stuff). The plan was to have separate images for each language. Then users could just update with a `pull`.

https://github.com/Rfvgyhn/teslamate/blob/auto-provision-grafana/grafana/Dockerfile

adriankumpf:
Oh, I didn't know that you also started on a PR! Regarding the translations: maintaining the German translations turned out to be really error-prone and time consuming. I also hope that focusing on one language makes contributing changes to the dashboards easier.

PR #104, opened 2019-08-19
Disable basic auth in Grafana

olexs:
Allows hiding Grafana behind a basic-auth-secured reverse proxy without interfering with Grafana auth.


olexs:
Issue: when a basic-auth-enabled proxy is used in front of Grafana, Grafana will recognize the forwarded "Authentication" header and attempt to login a Grafana user with the submitted username and password. Since a basic auth username/password usually has nothing to do with Grafana users, this fails, and Grafana becomes unusable (showing "Incorrect username or password").

I worked around this by setting the environment variable in my docker-compose.yml. However, disabling basic auth by default in the container would be simpler since it's not used anyway, and would be more consistent with the recommendation in the project's README ("I strongly recommend to use a reverse-proxy with HTTPS and basic access authentication when exposing TeslaMate to the public internet").

adriankumpf:
Thank you!

PR #140, opened 2019-09-03
Add support for MQTT SSL

virtualm2000:
I have a linux container install for TeslaMate and I also have a separate MQTT broker which uses SSL connection. 
I have added an ENV var MQTT_SSL as true/false to specify if MQTT connection is using SSL or not.

Please feel free to make the proper corrections as this is my first attempt at Elixir.

Best regards,
Mihai


adriankumpf:
Awesome! Thank you. I will look into it soon.

lib/teslamate/mqtt/mqtt.ex
```diff
@@ -27,13 +27,25 @@ defmodule TeslaMate.Mqtt do
...
+        handler: {Tortoise.Handler.Logger, []},
+        subscriptions: []
+      ]
+    else
+      [
+        user_name: Keyword.get(auth, :username),
+        password: Keyword.get(auth, :password),
+        server:
+          {Tortoise.Transport.SSL,
+           host: Keyword.get(auth, :host), port: 8883, verify: :verify_none},
```
adriankumpf: I think one would expect security be default i.e. certificates should be verified when `MQTT_SSL` is enabled.

Do you use a self-signed certificate? Or could we add a CA store with the `:cacertfile` option?

virtualm2000: Yes, I use a self-signed certificate, hence the verify_none option. But a CA store option would work as well.

adriankumpf: Gotcha! Then let's add a CA store and make`:verify_none` optional e.g. with another env var? I'll try to integrate  https://github.com/ericmj/castore  later.

virtualm2000: Perfect! Looking forward to test.

Thanks!

adriankumpf:
Make sure to check the box to allow commits from me on your PRs in the future. I'll add the CA store after merging. Thanks!

PR #146, opened 2019-09-05
Datasource variable on all dashboards

andrewhunt12:
Currently the Grafana dashboards don't specify a datasource, so when imported they use the default. This is fine for new installations off of the docker compose file, however in my situation (and I'm sure for others) I had other data sources already defined in my Grafana installation. This changes adds a variable to each dashboard that will allow users to pick which Postgres datasource to use, if they only have one it will default to that and should function exactly the same as it does before this change.

### DIFF
--- a/grafana/dashboards/charge-details.json
+++ b/grafana/dashboards/charge-details.json
@@ -19,6 +19,7 @@
   "links": [],
   "panels": [
     {
+      "datasource": "$Datasource",
       "aliasColors": {},
       "bars": false,
       "dashLength": 10,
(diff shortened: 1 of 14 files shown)


adriankumpf:
Thank you!

I'd rather hide the dropdown menu by default to keep it simple for the majority of users. Or alternatively, why not use a hard-coded name for the data source?

adriankumpf:
With v1.9.0 the data source is hard-coded to "TeslaMate" and hence doesn't need to be the default data source anymore. Feedback is welcome. Closing this for now.

PR #175, opened 2019-09-22
Fix charge counter when using with multi vehicles

abhinavWOT:
Added a check for car_id for charge count and charge sum.


adriankumpf:
Thanks!

PR #195, opened 2019-10-07
Initial Documentation Commit

ngardiner:
Commit of initial documentation structure and a full guide for configuring Tasker to send resume requests when Bluetooth is connected (quite a common request). Not all topics are complete currently, will follow up with further PRs to complete these documents and tutorials.

Next commit will focus on completing current MQTT documentation, HomeAssistant integration and DigitalOcean guide.


adriankumpf:
Thanks so much Nathan! I really appreciate your help.

Do you want me to merge the PR right now or do you want to complete the HomeAssistant integration first?

docs/MQTTBrokerIntegration.md
```diff
@@ -0,0 +1,39 @@
...
+The MQTT function within TeslaMate allows useful values to be published to a MQTT broker. This is useful in allowing other automation platforms to consume data from TeslaMate.
+
+Unless the MQTT feature is disabled data is published to the following topics
+(`$car_id` usually starts at 1):
+
+## MQTT Topics
+
+| MQTT Topic                                  | Example | Description                        |
+| ------------------------------------------- | ------- | ---------------------------------- |
+| teslamate/cars/$car_id/battery_level        | 88      | Battery Level Percentage           |
```
adriankumpf: Great idea to provide an example value!

docs/TeslaSleepConfiguration.md
```diff
@@ -0,0 +1,118 @@
...
+  * If this value is set too high, the time after a drive during which the car is continually actively polled and is unable to go to sleep will be longer than necessary, consuming a greater amount of battery power.
+
+## Attempt to Sleep
+
+Once the Time to Try Sleeping idle timer is reached, TeslaMate will try to allow the car to go to sleep. It does this by halting all polling for the configured time period, which is **21 Minutes** by default. 
+
+  * You will know that the vehicle is in this state when the status in TeslaMate is *falling asleep for x minutes*.
+  * At the end of this polling window, TeslaMate will poll the vehicle status. If the vehicle status is offline, this indicates that the sleep process succeeded and the car is no longer directly queryable.
+  * Polling will continue once per configured interval (*by default, 21 minutes*), meaning TeslaMate's resolution of polling interval for vehicles which have gone into sleep mode is 21 minutes - unless otherwise notified, TeslaMate will not know that the vehicle is awake for a total of 21 minutes (*by default*).
```
adriankumpf: One minor correction: the resolution of the polling interval for vehicles which have gone into sleep mode is __1 minute__ not 21. That means if the car is in sleep mode TeslaMate will notice when it wakes up with a short delay of up to 60 seconds.
The worst-case, i.e.  missing 21 minutes of a drive session. only occurs if the vehicle wakes up from sleep mode right after TeslaMate halts polling.

docs/TeslaSleepConfiguration.md
```diff
@@ -0,0 +1,118 @@
...
+
+_Note: Substitute your IP address for the commands_
+```
+  sleep_teslamate: curl -X PUT http://192.168.1.1:4000/api/car/1/logging/suspend
+  wake_teslamate: curl -X PUT http://192.168.1.1:4000/api/car/1/logging/resume
+```
+
+### Time of Day Hints
+
+_Are these available? I think this would be a good addition to TeslaMate, if not_
```
adriankumpf: Could you elaborate on this?

ngardiner: Just a thought I'd had while writing the doc - a feature request that I was considering would be the ability to specify times (for example, 7am - 8am daily) where TeslaMate would resume polling in anticipation of the vehicle waking up. This could help for daily commuters who aren't able to get a stable setup using Tasker, etc. What are your thoughts?

docs/TeslaSleepConfiguration.md
ngardiner: Thanks for the clarification Adrian! One of the reasons I wanted to go into this level of detail in the document was because I wasn't 100% sure that I understood the mechanism and wanted to break it down to do so.

Just so that I can better elaborate the process in this document, would this be accurate from a high level perspective:

- If TeslaMate polls the vehicle and it is one of (eg): driving, charging - it is online and will be polled regularly (possible to get the exact resolution of polling?)
- If the vehicle is online but not otherwise active, it is polled once per minute until we reach the idle period (which is 15 minutes by default)
- *My assumption at this point was that it would then revert to a poll once per (configured sleep period) to allow the car to attempt to sleep, sounds like however it is still polling every minute during this period?*
- Once the car is confirmed offline, it will be polled once per configured sleep polling period (once per 21 minutes by default). If it remains offline, this polling remains at this resolution, however as soon as it is polled as online, the polling interval reverts to one per minute

Also, just so I fully understand the resume/suspend mechanism:

- Resume would cause TeslaMate to revert to once per minute polling (for a given period of time?) even if the car currently shows offline, to allow it to pick up an imminent online event.
- Suspend would cause TeslaMate to revert to the offline polling schedule of (by default 21 minutes)?

adriankumpf: Yeah, sure. I think it's worth to explain it in detail!

> If TeslaMate polls the vehicle and it is one of (eg): driving, charging - it is online and will be polled regularly (possible to get the exact resolution of polling?)

Exactly. When driving it polls every 2.5 seconds. I'd consider this an implementation detail and it might change slightly in the future. But in general the vehicle is being polled  _every few seconds_. 

When charging, the polling interval depends on the charger power: The higher the power, the higher the polling frequency. The resolution can range from a few seconds to 30 seconds.

>  If the vehicle is online but not otherwise active, it is polled once per minute until we reach the idle period (which is 15 minutes by default)

If it is online but not otherwise active, the polling frequency depends on a few variables e.g. 

- is a driver still present
- is the vehicle unlocked
- is sentry mode active
- ...

In all cases the vehicle is being polled regularly every 10 to 30 seconds. This process takes 15 minutes by default or until we send the suspend command.

> My assumption at this point was that it would then revert to a poll once per (configured sleep period) to allow the car to attempt to sleep, sounds like however it is still polling every minute during this period?

I was a little ambiguous here. After the idle period of 15 minutes, polling is halted (for 21 minutes by default). AFAIK the vehicle can only go into sleep mode if it's not being polled. This is the most "critical" part which can potentially cause a data gap if a drive session is started during that period because TeslaMate simply cannot know. Basically, this period is the reason why there is the `resume` command. At this point, we have to instruct TeslaMate to continue polling if we start to drive (or charge).

> Once the car is confirmed offline, it will be polled once per configured sleep polling period (once per 21 minutes by default). If it remains offline, this polling remains at this resolution, however as soon as it is polled as online, the polling interval reverts to one per minute

Once it is offline (or asleep) it will be polled every minute. You may be wondering now why the heck the car continues to sleep although we start polling again :-) 

The reason is that there are two different kinds of API requests to fetch vehicle data. The one that gets us all data (position, speed, charger power etc.) is only possible if the car is online. This kind of request prevents the vehicle from falling asleep. 
Then there's another one that can be made even though the car is offline/asleep. This request returns not that much useful information but it contains the vehicle status. This makes it possible to know when the car is waking up.

> Resume would cause TeslaMate to revert to once per minute polling (for a given period of time?) even if the car currently shows offline, to allow it to pick up an imminent online event.
    Suspend would cause TeslaMate to revert to the offline polling schedule of (by default 21 minutes)?

Like mentioned above `resume` "reverts" the halted polling or in other words it instructs TeslaMate to start polling again, if it was halted earlier.

Now, to make it even more complex, `resume` was given a second task:  if the vehicle is offline/asleep (= polling resolution of 1 minute) and we send the `resume` command it will increase the polling frequency so that TM can almost immediately know when the vehicle wakes up. The use case is that my smartphone is often already connected to the car (which as a result is online) but TM would only notice this with a short delay. Sending `resume` will speed up the process.

docs/HomeAssistantIntegration.md
```diff
@@ -0,0 +1,339 @@
...
+If your intention is to only use read-only sensor values, those provided by TeslaMate via MQTT are sufficient, and you do not need to utilise the official Tesla compoent. If however you would like to be able to write values to the Tesla API (Lock/Unlock Doors or automate Climate), there is a solution which involves configuring an extremely high polling interval for the Tesla component and using automation to populate the values from the TeslaMate MQTT parameters.
+
+## Screenshot
+
+Coming soon
+
+## Current Status
+
+  * Sensors: All existing sensors available except for:
+    * Park Brake
```
adriankumpf: This sensor can be created by checking if the `shift_state` (sensor.tesla_shift_state)  is `"P"

ngardiner: Perfect. Whilst it does seem redundant given that, it would at least allow a drop-in replacement for all of the existing sensors (which is my plan) plus the addition of quite a few that didn't exist before. I'll update this shortly.

adriankumpf: Makes perfect sense 👍

docs/HomeAssistantIntegration.md
```diff
@@ -0,0 +1,339 @@
...
+  * Locks: Not implemented
```
krezac: There is Locket: yes on the teslamate dashboard (localhost:4000), so it must be available.

krezac:
Just FYI to prevent duplication - I'm working on Shortcuts instructions for iPhone. The pull request should be available tomorrow or so.

docs/TeslaSleepConfiguration.md
ngardiner: That's 100% clear, thanks Adrian. I'll do some work to adjust the wording around this but I think with that we'll all understand the mechanism and the reasons for the resume/suspend interface

PR #200, opened 2019-10-10
Initial version of Shortcuts guide

krezac:


adriankumpf:
Looks good. Thank you!

I'll merge this after @ngardiner's documentation PR :)

vkarthikbalaji:
Hi @krezac , the ios shortcuts automation does not work. I used your API suggestion https://teslamate.mydomain.com/wake 
But this seems to be not working correctly. Thanks btw for the screenshots. I learnt how to make an automation myself.

krezac:
Hi @vkarthikbalaji , thanks. what does it mean "does not work"? DO you get HTTP error? Doesn't it reach the inner server? Something completely different?

The URLs are just an example, they will be different based on what approach you use to get into your internal network (traefik/nginx/dyndns/...) and how you deploy (docker/native).

> Hi @krezac , the ios shortcuts automation does not work. I used your API suggestion https://teslamate.mydomain.com/wake
> But this seems to be not working correctly. Thanks btw for the screenshots. I learnt how to make an automation myself. By the way, this line is not very clear to me :
> 
> > You have wake up endpoint exposed as https://mytm.myweb.com/wake and proxied to sth like http://teslamate:4000/api/car/car_id/logging/resume where car_id is 1 if you have only one car.

vkarthikbalaji:
Thanks @krezac  for getting back. I get '401 unauthorized' on my phone when I run the shortcuts automation. I used docker (advanced setup) and hosted on digital ocean droplet.

krezac:
401 - despite the text - means "not authenticated" (bad credentials), not  "unauthorized" (you are not allowed). So either you haven't provided the credentials properly - like typo in the auth header (or header not at all) or the server doesn't know the credentials (typo in password file, server not configured properly to use the file,...). If you can get to server logs, they should tell you more. 
If I remember well, someone else mentioned making it work on digital ocean here, I'd try to find and check (I'm not familiar with digital ocean).

vkarthikbalaji:
IFor eg, I can see in docker logs of teslamate container when I do a curl request like this:

`karthik@kv:~/>curl --request PUT --header 'Authorization: Basic 'Base64encoded$TM_USER:TM_PWD$' 'https://teslamate.mydomain.com/api/car/1/logging/suspend'
`
docker logs teslamate_container on digitalocean does get the PUT request. However, when I use the same API on iphone shortcuts it returns '401 Unauthorized'. That is why I am not sure. Btw, double checked the credentials. 

`10:49:49.449 [info] PUT /api/car/1/logging/suspend
10:49:49.455 [info] Sent 204 in 5ms
`

krezac:
That's weird. I'd try to use some debugging proxy (i.e. we used Fiddler on windows in my previous job) running locally to capture the raw request and point phone to it (assuming it's on the same "home" wifi as your proxy). Once you have the raw request, you can try to decode the header or compare it with working call from curl.
One thing to check - don't you have extra new line in the edit box for user:passwd?

vkarthikbalaji:
Got it. It was my mistake. Missed the space between 'Basic' and 'base 64 encoded' as mentioned in this screenshot. Now it works. Thanks for your 'extra new line' check idea.
[https://github.com/krezac/teslamate/blob/feature/shortcuts_doc/images/shortcuts/create_26_headers_values_var.png](url)

krezac:
Congrats to solve find the mystery (these are the worst ones). Have a nice day.

PR #229, opened 2019-10-30
Consistent language for label of charging events

dopeghoti:
The rest of the interface is English, the German word seems out of place
here.


adriankumpf:
Thank you!

PR #268, opened 2019-11-24
Charge Costs integration - Grafana and API

ngardiner:
This PR isn't ready for merging yet, however I did want to consult on the API routing so that I understand the overall best direction before working on the controllers and functions.

At the moment, we don't have a web controller for object classes represented in the API other than on a per-Car basis. Although charges could be associated to cars, I don't really see a reason to do so (this could be filtered out/in of the dataset, and in my view the more relevant pivot is the address rather than car), so I have represented these 2 tables separately - one controller for charges and another for addresses.

Does that make sense, or would you see this arranged in another fashion?


ngardiner:
I should add, I also modified the column headings in the charges.json grafana dashboard to stop them bunching up on each other after adding the Cost column. The purpose is to make some a little less verbose but for the headings to still make sense, if any are unclear I can fix them up

adriankumpf:
I could imagine that due to the potentially large number of stored addresses the `GET /address` route would need to be paginated.  What do you say we embed the address in the charges (instead of returning an `address_id`) and only add the charges_controller? The embedded address could contain only the most important attributes e.g. `display_name`, `latitude` and `longitude`. That way one requests could return all information and it would save additional work.

ngardiner:
Sounds good, @adriankumpf . Makes sense to me.

I am closing this PR off as there's a merge conflict with the charge stats grafana template and I want to avoid it being an issue again if there are any more updates, so I'll split the grafana template changes from the API changes.

Issue #269, opened 2019-11-25
Extend Charge Stats Dashboard


marcogabriel:
I'd like to see the Charge Statistics Dashboard extended by the following values/graphs:

* Discharge Stats - How many times was the battery discharged to what level?
* Charge Delta - What delta from discharged to charged point (for a batteries lifespan, it's better to just stay between 20 and 80, and even better to stay between 35 and 65% SOC.
* Charge Heatmap - A combination of Charge Stats, Discharge Stats, combined on a certain timeframe

I extended the standard dashboard accordingly. please see the pull request if this is useful to be integrated upstream.
![image](https://user-images.githubusercontent.com/12827079/69553298-44b33280-0fa0-11ea-9575-18cd90b024a5.png)

adriankumpf:
Closed with #270

PR #270, opened 2019-11-25
Update charging-stats.json

marcogabriel:
extends charging-stats.json according to #269


adriankumpf:
This is really useful information and the Charge Heatmap & Dela look amazing. Thanks!

vkarthikbalaji:
@marcogabriel, a minor detail -> Discharge stats should have '# discharges' instead of '# charges' . Thx for making the heatmaps and deltas. This is very handy.

PR #278, opened 2019-11-27
Using timerange in "Charging Stats" dashboard

nnoally:
Hi,
I think Charging stats dashboard would be more usefull if it use the timerange filter (and even more when charge cost will be enabled).

### DIFF
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -264,7 +264,7 @@
       ],
       "valueName": "avg"
     },
- {
+    {
       "cacheTimeout": null,
       "colorBackground": false,
       "colorValue": false,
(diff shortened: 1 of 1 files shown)


adriankumpf:
Thank you! Will merge soon.

nnoally:
You're welcome. Thanks for your great work.

adriankumpf:
I've made one minor change and increased default time range to 'Last 10 years' because most people presumably want to see the long-term statistics first.

bahree:
I pulled the latest and sorry if this is obvious - where do I enter the charging cost for it to show in the dashboard? E.g. most of my charging is at home and I can't figure out where do I add the cost per KWh from my utility? 

Thanks.

PR #314, opened 2019-12-22
Direct people to new RTD site

gundalow:
* Add Shield Badge
* Ensure anyone going into `/docs` is directed to RTD

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,35 +1,4 @@
-# TeslaMate Documentation
+## Documentation
 
-## Installation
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -1,35 +1,4 @@
-# TeslaMate Documentation
+## Documentation
```
gundalow: FYI it's `README.rst` (rather than this file), that's uses as the index page for https://teslamate.readthedocs.io/en/latest/

PR #326, opened 2019-12-29
Update simple docker documentation

dirkjanm:
add `POSTGRES_DB` parameter. If not specified, this name is taken from `POSTGRES_USER` which only works as long as the username equals the database name in the teslamate/grafana configs.


adriankumpf:
Thank you!

Issue #321, opened 2019-12-27
battery_level vs. usable_battery_level


ctraber:
Sorry, just saw that I added my first post regarding this topic to an already closed issue #257  
 
Recently I changed my setting in the car to % and was wondering why the Tesla App showed different % values than TeslaMate and other 3rd party apps.

After looking at the raw values of the API, it seems that the Telsa App and the display in the car are showing `usable_battery_level` and nor `battery_level`. In cold weather `usable_battery_level` is lower and once the difference is higher (>3%?) The app and car show the blue snowflake.

I also noticed that whenever I had the snowflake I also saw a spike down in degradation Grafana report.

I think the reason is that the degradation report is using  `battery_level` to calculate the projected 100% rated range:
`battery_range / battery_level * 100`

But I think that the range in the API is also affected by cold weather as you can see it going down/up when temperature is changing. So the behavior seems to be more related to `usable_battery_level` than to `battery_level`. Which would also explain the spikes in degradation report.

To get a better overview I would love to see `usable_battery_level` and the projected 100% ranges in addition to the `battery_level` on the status page.
Maybe it is also possible to add a second graph to the degradation report showing the line using the `usable_battery_level` for calculation.

If I find time I'll try to make a development setup to try it by myself...

**Thank you very much for all your effort! I love TeslaMate and I hope I'll be able to contribute in some way in the future.**

ctraber:
I was able to setup the dev environment.

Now I see that usable_battery_level is not yet in the schema.
Would it be high effort to add additional api fields to the positions table?

adriankumpf:
That would indeed be interesting!

> Would it be high effort to add additional api fields to the positions table?

The field would have to be added by a migration: https://devhints.io/phoenix-migrations I will look at it in the future (if you haven't prepared a PR by then :)

ctraber:
Thanks for the hint!
Never used Elixir, Phoenix, erlang, ... But I'll try.

PR #338, opened 2020-01-03
battery_level vs. usable_battery_level #321

ctraber:
As a first step I added usable_battery_level to positions table and summary page.

Sorry, but I'm not familiar with the Elixir yet and I'm sure there are still things missing (translation, test,?)...

Bildschirmfoto 2020-01-03 um 12 35 51


ctraber:
Tests passing on my dev env now.

Here I get:
12:49:44.528 [error] Postgrex.Protocol (#PID<0.1069.0>) disconnected: ** (DBConnection.ConnectionError) owner #PID<0.3853.0> exited

lib/teslamate_web/templates/car/summary.html.leex
```diff
@@ -158,7 +158,29 @@
...
+                if @settings.unit_of_length == :mi do
+                  "#{Float.round(@summary.rated_battery_range_km/@summary.battery_level * 100 / 1.60934, 2)} mi"
+                else
+                  "#{Float.round(@summary.rated_battery_range_km/@summary.battery_level * 100,2)} km"
+                end
+ 
+              %> 100%)</td>
+          </tr>
+          <% end %>
+          <%= unless is_nil(@summary.usable_battery_level) do %>
```
adriankumpf: Does it make sense to hide the 'Usable State of Charge' row if `usable_battery_level` == `battery_level`? What do you think?

adriankumpf:
Looks really good! After rerunning the tests everything passes :)

ctraber:
How can you restart the test without commit? Just in case i happens again.

adriankumpf:
There is a "restart build" button on the Travis page. But I'm not sure if it allows you to do that. Ideally, it shouldn't be required to rerun that often. Some tests are sensitive to timing though and therefore can be flaky.

lib/teslamate_web/templates/car/summary.html.leex
ctraber: Sure makes sense. There is no additional information when both are the same.
I'll also try to make a snowflake when difference is higher than 2% I think thats what the TeslaApp does.

adriankumpf: > I'll also try to make a snowflake when difference is higher than 2% I think thats what the TeslaApp does.

Yeah, that would be great!

adriankumpf:
Excellent. Will merge soon!

ctraber:
Finally it was cold enough for a snowflake. By using usable_battery_level, the projected range is not changing that much with temperature. I think it gives a better degradation overview.

Bildschirmfoto 2020-01-08 um 11 51 33

Bildschirmfoto 2020-01-08 um 12 02 38

adriankumpf:
It took a little longer, but I could finally take a closer look at the PR. I made a just few adjustments:

* Replaced the existing degradation dashboard with your 'projected range' dashboard
* Moved the "(xxx km 100%)" text into a tooltip
* Added a test for the snowflake icon

Thanks again! :)

adriankumpf:
Resolves #321

ctraber:
Will I see the tooltip in iOS Safari?
Most of the time I view it on my phone.

adriankumpf:
Should work with 3904a18e2406089b50fc2c933cfa9baf5f180704

ctraber:
Thanks a lot!
Just tried. Works perfectly.

ctraber:
I tested your new projected range dashboard.
In the second panel the coalesce for usable_battery_level is missing. I would add it but for some reason I do not get the dashboard, I changed in filesystem, after building a new grafana image. 

Do you have an idea why it is not showing the changed dashboard after docker build and restart of container?

adriankumpf:
You are referring to the following query (line 3), right?

```SQL
SELECT
	$__timeGroup(date, '6h') AS time,
	convert_km(sum([[preferred_range]]_battery_range_km) / sum(usable_battery_level) * 100, '$length_unit') AS "Projected Range (using usable_battery_level) [$length_unit]",
	convert_km(sum([[preferred_range]]_battery_range_km) / sum(battery_level) * 100, '$length_unit') AS "Projected Range (using battery_level) [$length_unit]"
FROM
	positions
WHERE
	$__timeFilter(date) and
	car_id = $car_id
GROUP BY
	1
ORDER BY
	1,2  DESC
```

I don't see why `sum(coalesce(usable_battery_level,battery_level))` should be necessary. Could you explain that?

> Do you have an idea why it is not showing the changed dashboard after docker build and restart of container?

Does it show up if you add the dashboard under a new name (and with a different `uid`)?

ctraber:
The null values mess up the graph at the point where database starts having values for usable_battera_level. Maybe sum() with some null values, not sure. 

see the spike with app. 1700 km.
Bildschirmfoto 2020-01-18 um 14 33 55

I can add new dashboards.
But, I copied yours with "save as" changed it and copied the json to the existing file, changed names... But if I re-build the container its again your previous version..
=> changing in file, build... and still the version before file change.

adriankumpf:
Hmm, seems to work fine for me. Anyway, if `coalesce` helps, let's add it back in.

> But, I copied yours with "save as" changed it and copied the json to the existing file, changed names... But if I re-build the container its again your previous version..
=> changing in file, build... and still the version before file change.

Very strange. Sounds like it takes the previous file when creating the image? Maybe try doing the following:

1. Stop the container
2. delete the grafana volume
3. build the image
4. run the new image

ctraber:
Deleting the volume did the trick.
Thanks!

PR #361, opened 2020-01-15
Add documentation for docker installation with apache and backup/restore

DrMichael:
Add documentation for docker installation with apache and backup/restore


adriankumpf:
Thanks so much!

One small tip: with docker-compose you don't have to look up the ID every time and can use the name instead: `docker-compose exec -it db pg_dump -U teslamate > teslamate.bck`

PR #364, opened 2020-01-18
usable_battery_level for charges and new charge level dashboard

ctraber:
- adds usable_battery_level to charges
- new charge level dashboard

### DIFF
--- a/grafana/dashboards/charge-level.json
+++ b/grafana/dashboards/charge-level.json
@@ -1,0 +1,232 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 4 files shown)


adriankumpf:
Thanks Christian!

Delta-99:
@adriankumpf I don't see these in my Grafana? Is there something needed to get these included vs just a regular update?

ctraber:
Are you missing the whole "Charge Level" Dashboard or only the usable_battery_level values?

Usable_battery_level is only logged since 1.15. You will not see values for the past. It will only start showing different values for usable_battery_level for the time since you upgraded to 1.15.
Also the values are only different to battery_level values when the battery is cold.

Here an example (I had a pre-release version logging it since longer 3 weeks):
Bildschirmfoto 2020-01-25 um 07 38 04

Delta-99:
Ignore my comment. For some reason I thought these graphs were released in the previous update I just did. I just updated again to 1.15 I think it is and they are there!!

PR #367, opened 2020-01-18
enhanced projected range dashboard

ctraber:
- new panel with projected range and battery level
- added charges datapoints (union) to all panels
- coalesce for usable_battery_level
- variable to select time resolution for $__timeGroup (default 6h)
(I did not find a way to automatic change group interval when zoomin in, yet.)

### DIFF
--- a/grafana/dashboards/projected-range.json
+++ b/grafana/dashboards/projected-range.json
@@ -103,7 +103,7 @@
           "hide": false,
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


adriankumpf:
This is great; thanks! The customizable time resolution is really neat.

Just a small note: I think it makes sense to coalesce `battery_level` and `usable_battery_level` in the first panel in the 'Projected Range' dashboard (like it was before). This allows to see the battery degradation at a glance. The other two panels distinguish between the two values anyway :)

ctraber:
Ups, sorry... I copied the same query with union to all three panels.
You are right. It's good to have one clean panel with only one graph.

PR #369, opened 2020-01-19
added 'tesla' tag

TechForze:
added 'tesla' tag for finding car related information between other dashboards.


adriankumpf:
Thanks!

Issue #392, opened 2020-01-26
Projected Range Dashboard missing Charges selection


khorton:
In the past, the previously named Degradation dashboard (recently renamed to Projected Range) had the ability to optionally select the display of times when a charge had been completed.  Today I noticed that this was no longer present in v1.15.1.  One possible variable is that there is a charge ongoing now - I suspect all the previous time I had looked at this dashboard was charging was not in progress.

**Relevant entries from the logs**

```
teslamate_1  | 19:18:06.385 [info] Already up
teslamate_1  | 19:18:16.646 [info] Version: 1.15.1
teslamate_1  | 19:18:16.825 [info] Refreshed api tokens
teslamate_1  | 19:18:16.825 [info] Scheduling token refresh in 40d
teslamate_1  | 19:18:16.833 [info] Running TeslaMateWeb.Endpoint with cowboy 2.7.0 at :::4000 (http)
teslamate_1  | 19:18:16.835 [info] Access TeslaMateWeb.Endpoint at http://XX.XX.Xx.XX
teslamate_1  | 19:18:16.874 [info] Adding elevation to 2 positions ...
teslamate_1  | 19:18:16.975 [info] Querying location for elevation takes longer than 100ms ...
teslamate_1  | 19:18:17.572 [info] Starting logger for 'BlueBird'
teslamate_1  | 19:18:17.582 [info] MQTT connection has been established
teslamate_1  | 19:18:17.789 car_id=1 [info] Start / :online
teslamate_1  | 19:18:18.368 [info] Adding elevation to 1 positions ...
teslamate_1  | 19:18:18.666 car_id=1 [info] Charging / SOC: 28% / Home
teslamate_1  | 19:18:26.480 [info] GET /
teslamate_1  | 19:18:26.499 [info] Sent 200 in 19ms
teslamate_1  | 19:36:15.079 [info] GET /
teslamate_1  | 19:36:15.084 [info] Sent 200 in 4ms
teslamate_1  | 20:06:12.480 [info] GET /settings
teslamate_1  | 20:06:12.502 [info] Sent 200 in 22ms
teslamate_1  | 20:09:08.415 [info] GET /
teslamate_1  | 20:09:08.419 [info] Sent 200 in 4ms```

**Operating environment**

- TeslaMate Version: v1.15.1
- Type of installation: Docker
- OS: Ubuntu 19.10

ctraber:
You are right. The Charged annotation is missing in the new "Projected Range" dashboard.
[...]

PR #393, opened 2020-01-26
added Charged annotation to projected range dashboard. #392

ctraber:
adds Charged annotation to projected range dashboard. #392 

Prepared to show region after grafana update. (grafana#20752)


adriankumpf:
Thanks Christian! First time I've heard from region annotations. Pretty cool.

PR #397, opened 2020-01-26
Add French Translation

tomS3210:
Add French Translation to Teslamate


adriankumpf:
Thank you, Thomas!

tomS3210:
Thank for the merge. Where can we choose the langage ?

Thomas

adriankumpf:
If you're running the master branch, it'll automatically pick the preferred language of your browser. To overwrite this setting add `?locale=fr` to the URL.

tomS3210:
Hello Adrian,

Ho how are you ?

I can see that you did a very good job on TeslaMate ! This Streaming API seems to be a very good idea !  
What is the principle ? Teslamate subscribe to the API and then the car send the informations ?
So there is no more pulling every 5 seconds ?

I just send you an update of the French Translation.

How can I update to the new 1.9 Dev version ?

Thanks again for your work.

Regards

Thomas
www.club-tesla.fr <http://www.club-tesla.fr/>

PR #405, opened 2020-01-28
update docs to include wake proxy example for apache

DP19:
Just finished setting this up from scratch and it was hard to tell you actually needed to set this up if you didn't read the docs to carefully. I looked through the docs for traefik to see if there was an easy label to set for the http router to proxy this as well but didn't see anything obvious. If someone wants to add that, it would also be useful and I imagine most people would like this enabled by default.

As a side note it looks like the docs for the wake integration are pretty different as the ios docs are the only ones that mention proxy'ing an endpoint like this.

### DIFF
--- a/docs/configuration/guides/shortcuts.md
+++ b/docs/configuration/guides/shortcuts.md
@@ -8,8 +8,8 @@ Unfortunately, only couple of the useful events (like NFC on NFC enabled phones)
 
 These values are used in the screenshots. Whenever you see them, replace them with the actual values used on your system.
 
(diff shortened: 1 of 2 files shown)


adriankumpf:
I wonder if we shouldn't change the URL mentioned in the iOS guide instead, to keep it simple:

    `https://teslamate.example.com/wake` -> `https://teslamate.example.com/api/car/1/logging/resume`

Proxying should not be necessary, it was just used as an example by the original contributor of the guide IRC.

PR #412, opened 2020-02-01
Docs/faq update

wishbone1138:
FAQ entry for adding token directly into the database instead of using username/password


adriankumpf:
Thank you!

wishbone1138:
Happy to contribute, thank you for an AMAZING tool.

PR #416, opened 2020-02-03
Improve standalone install documentation

Niek:
See: https://github.com/adriankumpf/teslamate/issues/411


adriankumpf:
Thank you!

PR #438, opened 2020-02-10
Update backup_restore.md

AlwindB:
Update to solve the issue where the service could not be found, in some configurations the database is defined as 'db'.
If not configured in an explicit way the user might get a "ERROR: No such service: database" when running the command


adriankumpf:
Thank you for your contribution. The docker-compose files have recently been updated so that all installations use the same service names. The backup instructions refer to the latest version (`database`).

Issue #316, opened 2019-12-23
Import from TeslaFi


rikbrown:
It would be nice to have an automated import from TeslaFi (which I hope to migrate from). I think it can export all the data, but not sure on the format.

(This is something I may try to contribute once I’ve set up Teslamate, just creating this to track it or see if it is on your radar).

bahree:
I have seen this request a few times and happy to take a crack at it and write a tool. The one thing though - I don't use TeslaFi - is there somewhere one can read up on its schema used and is there a export feature? Need to see what's the data structure is like.

Somewhere one of you can point me to or share this?

Cheers.

tucopup:
TeslaFi allows export of data as a csv by month.   Attached is a clip of an export.

Let me know if you need more/different.

[TeslaFi82017 edit.zip](https://github.com/adriankumpf/teslamate/files/3995570/TeslaFi82017.edit.zip)

bahree:
Thanks @tucopup ; let me try and make sense of this, and see how to map it to the schema of TeslaMate.

bahree:
I started to look at this and wanted to create a documentation of the existing schema, so then to understand how to map TeslaFi's to this. I took a stab at this [here](https://dbdiagram.io/d/5e03cf2fedf08a25543f67cb). @adriankumpf , does this look directionally correct?

Tigwin:
This is my most important issue to migrate off of TeslaFi. I don't want to lose all my historical data. 

Speaking of which, one thing that I really like for this historical data is the lifetime map and calendar view (which shows the map of drives for the month displayed)

mboerma:
Yes, this would be a great feature!

adriankumpf:
> I started to look at this and wanted to create a documentation of the existing schema, so then to understand how to map TeslaFi's to this. I took a stab at this here. @adriankumpf , does this look directionally correct?

Looks good to me, @bahree!

I have never used TeslaFi and therefore have no access to exported data. The CSV data from @tucopup seems to be quite complete though.

There are currently no plans to support an automated import from TeslaFi. Contributions are of course welcome though. So if anyone wants to contribute, that would be great!

mel0en:
Is someone still working on this? I would like a one-time conversion as well.

adriankumpf:
[...]

PR #454, opened 2020-02-15
Add docs for tesla-apiscraper import

olexs:
As mentioned in [my comment](https://github.com/adriankumpf/teslamate/issues/316#issuecomment-583761153) on #316, I've cleaned up and pushed the CSV extraction and conversion toolkit (put it up in a separate repository, since it's not directly/not only TeslaMate related: https://github.com/olexs/tesla-apiscraper-to-teslafi-export), and documented the import process for tesla-apiscraper data to TeslaMate. Hope there's more than just me who actually needed this :)


adriankumpf:
Fantastic work! Thanks for sharing :) I'm sure quite a few will be very happy about this.

PR #460, opened 2020-02-17
Add cost overview to charge dashboard

Niek:
This adds a small panel next to the energy added/energy used ones.


adriankumpf:
Thanks!

PR #482, opened 2020-02-24
Added magic number to charge details grafana board

codingoverdrive:


adriankumpf:
I don't understand. Could you explain why this "magic number" is needed?

codingoverdrive:
Many older Teslas have been rate limited when Supercharging. Owners see this as a slower charge rate, faster taper and longer charge times.

Prior to update 2019.16.x the average “magic number” for a Model S85 was 120-130, following the update, the number is typically 100-110.

Owners want to know if future updates are reducing the magic number.

Please see https://teslamotorsclub.com/tmc/posts/4504468/ and the messages below it.

adriankumpf:
I see. Since this metric seems to be useful for supercharging sessions and only some owners know how to interpret it I'd suggest to hide it by default, e.g. with a series override on line 91:

```json
        {
          "alias": "Magic Number",
          "hiddenSeries": true
        }
```

With this approach, shift-clicking on "Magic Number" in the legend would display the series.

codingoverdrive:
Sounds sensible.

codingoverdrive:
Maybe a different name for the metric would be better than "magic number". 

What about "SC Factor"?

adriankumpf:
Yeah, that's a good idea 👍

PR #484, opened 2020-02-25
Show stats in desired units in 'Drive Stats' dashboard

pichalite:
Update 'Drive Stats' dashboard to display 'Median distance driven per day' , 'Extrapolated monthly mileage' & 'Extrapolated annual mileage' in desired units.


adriankumpf:
Good catch. Thank you!

PR #485, opened 2020-02-25
Swedish translation

tobiasehlert:
Adding po files for swedish


adriankumpf:
That's awesome! Thank you.

PR #488, opened 2020-02-27
Documentation fix for upgrading manual installs

Niek:
The existing documentation doesn't work, this fix actually checks out the latest available tag.

### DIFF
--- a/docs/upgrading.md
+++ b/docs/upgrading.md
@@ -21,7 +21,7 @@
 
    ```bash
    git pull
-   git checkout $(git describe --tags)
+   git checkout $(git describe --tags `git rev-list --tags --max-count=1`)
 
    mix deps.get --only prod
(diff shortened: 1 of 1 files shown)


docs/upgrading.md
```diff
@@ -21,7 +21,7 @@
 
    ```bash
    git pull
-   git checkout $(git describe --tags)
+   git checkout $(git describe --tags `git rev-list --tags --max-count=1`)
```
adriankumpf: The manual installation guide would probably have to be changed as well, right?

https://github.com/adriankumpf/teslamate/blob/d39a5e4a99fb8706af8f69c7e2e67a0bd9d0ed39/docs/installation/debian.md#L95

adriankumpf: Gotcha!

Issue #524, opened 2020-03-21
Dropdown values and displayed text swapped


tobiasehlert:
**Describe the bug**

HTML code of dropdown Range (Preferred range) has two values (ideal and rated) and the english values should be the value inside the options dropdown and the displayed text should be the translated one.

This results in an error showing up right now, saying that the option is invalid. So you can not change the value right now :)

**To Reproduce**

1. Go to Settings
2. Click on Range dropdown
3. Select one of the values

**Expected behavior**

As it is now:

```
<div class="select is-fullwidth">
    <select id="global_settings_preferred_range" name="global_settings[preferred_range]">
        <option value="idealisk" selected="">ideal</option>
        <option value="nominell">rated</option>
    </select>
</div>
```

As it should be:

```
<div class="select is-fullwidth">
    <select id="global_settings_preferred_range" name="global_settings[preferred_range]">
        <option value="ideal" selected="">idealisk</option>
        <option value="rated">nominell</option>
    </select>
</div>
```

**Relevant entries from the logs**

no

**Screenshots**

![image](https://user-images.githubusercontent.com/4417850/77235610-b8b0c080-6bb7-11ea-91d9-fc2f7680b47c.png)

**Data**

no

**Operating environment**

- TeslaMate Version: v1.18.0
- Advanced Docker Setup with Traefik
- OS: CentOS 8

tobiasehlert:
created PR #525 for this :)

PR #525, opened 2020-03-21
fixing settings dropdown issue and translation

tobiasehlert:
PR contains:
- fixing #524 dropdown issue
- fixing translation of words in swedish


adriankumpf:
Good catch! I just fixed a minor syntax error to make the build pass. Thanks.

PR #543, opened 2020-03-30
Display vampire drain range logg per hour in the correct units.

### LINKED ISSUE #542: Vampire drain reports TR loss / h as being in KM instead of preferred units
**Describe the bug**

The vampire drain grafana report does display the preferred units to report range loss / hr.

**To Reproduce**

1. Update length units to Miles
2. Open Vampire Drain
3. Notice that TR loss / HR is listed as being in KM even though it is calculated in the preferred length unit

**Expected behavior**

TR Loss / h should use the `[[length_unit]]` variable. Note: It is calculated in miles just not labelled as such.

**Relevant entries from the logs**

This is purely a JSON change

**Screenshots**

image

**Operating environment**

- TeslaMate Version: v1.19.0-dev
- OS: Ubuntu I think

ograff:
Closes #542.


adriankumpf:
Good catch, thanks!

PR #555, opened 2020-04-06
Send credentials with manifest request

### LINKED ISSUE #554: Cookie header not sent when requesting site.webmanifest
**Describe the bug**

I'm hosting TeslaMate behind an [auth proxy](https://github.com/authelia/authelia) which uses the Cookie header. Since the request for the site.webmanifest resource doesn't have the required cookie header, the request fails.
I've looked around a bit and found [this StackOverflow thread](https://stackoverflow.com/questions/51777346/cookies-not-sent-with-request-for-web-app-manifest-json).
According to the accepted answer, replacing the line `<link rel="manifest" href="/site.webmanifest?v=5AB53N3ALo">` with `<link rel="manifest" href="/site.webmanifest?v=5AB53N3ALo" crossorigin="use-credentials">` in [this](https://github.com/adriankumpf/teslamate/blob/master/lib/teslamate_web/templates/layout/root.html.eex) file should do the trick. I haven't been able to test the change myself since I didn't want to spin up a new environment (I'm using the Docker image) and (if it's even possible) haven't been able to locate the file containing the line to change in the Docker container to test it live.

**Operating environment**

- TeslaMate Version: 1.18.2
- OS: Ubuntu 16.04

MaxG88:
Fixes #554.
Please let me know if I missed something.


adriankumpf:
LGTM. Thank you!

Issue #561, opened 2020-04-06
Show VIN on home page


MaxG88:
**Describe the solution you'd like**

I would like to have the VIN on the home page. I would've loved to do a pull request myself and I feel like it should be a pretty simple thing to add considering that the data is already available in the database, but I've never used Elixir before. However, if I'm wrong and/or you don't want to do it but are open to a PR, let me know and I'll try to make a PR for it sometime in the next few weeks.

MaxG88:
Actually ended up being way easier than I thought.

PR #566, opened 2020-04-07
Add VIN under Model

MaxG88:
PR for #561.
I added the VIN right under the model (Model 3/S/X).
UI design has never been my strong suit, so please let me know if you have a better idea.

### DIFF
--- a/lib/teslamate_web/templates/car/summary.html.leex
+++ b/lib/teslamate_web/templates/car/summary.html.leex
@@ -18,7 +18,11 @@
       <div class="media-content">
         <p class="title is-5"><%= @summary.display_name %></p>
         <%= unless is_nil(@car.model) do %>
-        <p class="subtitle is-6 has-text-weight-light">Model <%= @car.model %> <%= @car.trim_badging %></p>
+        <p class="subtitle is-6 has-text-weight-light">
+          <span>Model <%= @car.model %> <%= @car.trim_badging %></span>
+          <br>
+          <span><%= @car.vin %></span>
+        </p>
         <% end %>
       </div>
        <div class="icons ml-5">
--- a/lib/teslamate_web/templates/car/summary.html.leex
+++ b/lib/teslamate_web/templates/car/summary.html.leex
@@ -21,7 +21,7 @@
         <p class="subtitle is-6 has-text-weight-light">
           <span>Model <%= @car.model %> <%= @car.trim_badging %></span>
           <br>
-          <span><%= @car.vin %></span>
+          <span class="is-size-7"><%= @car.vin %></span>
         </p>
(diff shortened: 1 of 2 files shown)


MaxG88:
I tried to add a test for the VIN <span> but couldn't find how to assign a VIN. If you have any pointers that'd be great.

adriankumpf:
I wonder if the VIN should be displayed so prominently as it will not change and you probably won't need this information so often. Maybe it would make sense to hide it in a tooltip (see below for an example) that becomes visible when hovering over the name or model?

https://github.com/adriankumpf/teslamate/blob/231583b48f91875f0262347dabd6dc9d7844e23d/lib/teslamate_web/templates/car/summary.html.leex#L62-L64

MaxG88:
Is the tooltip text selectable? The main reason I was looking to add the VIN to the home page was so that I could easily copy it if I needed to write it somewhere. Currently I have to either login to my Tesla account or to my inactive TeslaScope account and go to my car's details to find my VIN and copy it. For some reason the Tesla website has always been really slow for me, so the only reason my TeslaScope account still exists is for easy and faster access to my VIN.
It's not something I need often, but I've needed it often enough to want to add it here.

MaxG88:
I tried playing with the font-size of the VIN and I found that with `class="is-size-7"` it looked better. Much smaller but still readable. Might be a tad too small though I'm not sure.

MaxG88:
I tried with the VIN and don't really like how it looks.
image
What do you think?

adriankumpf:
I think the tooltip looks okay, but you are right, the text must be easy to copy. Therefore a tooltip is not a good solution.

But somehow I am not happy with the other solution either. I would prefer if the VIN was not displayed so prominently on top or if it was hidden by default.

jmiverson:
What if you just add it to the end of the list of other data on the summary page? Below “Version” for example. It would be copyable then, but also more out of the way and not the very first thing you always see?

MaxG88:
I'm going to close this PR since it's getting a little bit messy and open a new one with @jmiverson's suggestion.

adriankumpf:
An additional modal or view in which general vehicle data, such as exterior color, wheel type and also the VIN are displayed would be ideal.

MaxG88:
Agreed. I don't think I'm going to be able to submit a PR for that though.

jmiverson:
@MaxG88 — As time allows, I may be able to throw something together to at least get a modal or some screen started to show Vehicle Information and will tag you in a PR to help out fill it in, if you’d like.

MaxG88:
@jmiverson Sure! As long as there is a basic outline/layout in place, I should be able to help (I hope).

PR #578, opened 2020-04-10
Updated Labels to Title Case

jmiverson:
Some labels were using title case and others weren't.  Attempted to make it consistent across the board.

New to Elixir, but would like to start helping out if I can.  Let me know if you'd prefer these changes be made a different way.  I didn't want to modify the other languages, because I don't know the standard in those languages, but at least for English, it makes sense for the labels to be title cased.


adriankumpf:
Looks great to me! Thank you.

PR #581, opened 2020-04-12
Add portainer to the documentation

DrMichael:


adriankumpf:
The `<yourhost>` had to be removed because it was interpreted as an HTML tag. Otherwise, it looks great. Thank you!

PR #583, opened 2020-04-12
Add VIN under Version

MaxG88:
Follow-up to PR #566.


Kevga:
All the other info on that page is dynamic, while the VIN never changes. I don't see the point in displaying it there. If you need to copy it from time to time, create a text file with your VIN or something.

MaxG88:
You're right. Closing this.

PR #584, opened 2020-04-13
Add Danish translation

MartinNielsen:


adriankumpf:
Awesome. Thank you!

jmiverson:
@adriankumpf — Is there a template for language support somewhere that folks are using?  I noticed that this latest language file doesn’t have the new “msgid” values that I updated in my PR last week, which I think will cause the translation to not take for those labels.

If there’s a template, I probably need to update that as well.  Otherwise, I can run through this new Danish file and update them like I had done for the others.  Just don’t want it to become a perpetual problem.

Here’s my PR for reference: #578

DrMichael:
Well, @adriankumpf probably extracts the messages with "xgettext". Then the result of that has to be merged with the existing translations with "msgmerge".
I have that automated in one of my projects through a Makefile, but cannot access that currently.

adriankumpf:
Yes, gettext ([`mix gettext.extract --merge`](https://hexdocs.pm/gettext/Mix.Tasks.Gettext.Extract.html#content)) is used to extract/remove translations. Merging translations works quite well most of the time. I've done this after merging this PR so it should have picked up the updated labels:

https://github.com/adriankumpf/teslamate/blob/a99f6c4a3acf9d14f595d748199a8475891504ce/priv/gettext/da/LC_MESSAGES/default.po#L14-L572

larskochhansen:
Hi. I know that this translation issue may be a bit "old" but there are 2 words that should be changed if possible. I do not know how to do this but they are actually easy to locate in the above code.

Line 42: Opladet
instead of "lader"

Line 346: Oversigter
instead of oversikter - spelling error.

Is this possible to implement in a future update :) Thanks in advance and thanks for this amazing software development.

adriankumpf:
If you want, you can propose those changes directly using the GitHub web interface @larskochhansen . The file with the Danish translations is located here: https://github.com/adriankumpf/teslamate/blob/master/priv/gettext/da/LC_MESSAGES/default.po

And here is a guide on how to edit files and open PR: https://docs.github.com/en/github/managing-files-in-a-repository/editing-files-in-your-repository

PR #592, opened 2020-04-15
Add Title to Grafana table to allow sorting

MaxG88:
I was trying to sort my trips by distance on the Trip dashboard while seeing the total distance for a set period. 
I know I could've went to the Drives dashboard to do the sorting, but I felt like it would be nice to be able to do it in the Trip dashboard since I was already in it.

### DIFF
--- a/grafana/dashboards/trip.json
+++ b/grafana/dashboards/trip.json
@@ -1718,7 +1718,7 @@
           ]
         }
       ],
-      "title": "",
+      "title": "Drives",
       "transform": "table",
       "type": "table"
     },


adriankumpf:
Absolutely! Sorting should be possible on the Trips dashboard as well. I find it rather interesting that Grafana only allows sorting if a title is available ...

MaxG88:
Unless I missed something, this is how I was able to make sorting work.
Have you been able to sort that particular table even without a title?

adriankumpf:
Nope, I just hadn't noticed that you can't sort that table. I briefly checked if there is already a GitHub issue on this topic in the Grafana repo, but did not find one. But it looks to me like there is a CSS problem that Grafana has.

PR #602, opened 2020-04-18
Update upgrading.mdx

DrMichael:
Changed backup_restore.md to backup_restore


adriankumpf:
The extension was not even the problem, but the path (`..`). Thanks anyway for the quick PR!

PR #614, opened 2020-04-20
Add Korean translation

dongbum:
Adding .po files for Korean translation.


kocsenc:
We should probably find a way to have community help with translation. The other day I stumbled upon [crowdin](https://crowdin.com/) which is a little overkill I think for the current scale but worth checking out. Other than that @dongbum I think having some documentation on how to help contribute with languages would be awesome.

PR #625, opened 2020-04-22
Add Chinese(Simplified) translation

edward4hgl:
Add Chinese(Simplified) translation

### DIFF
--- a/priv/gettext/zh_Hans/LC_MESSAGES/default.po
+++ b/priv/gettext/zh_Hans/LC_MESSAGES/default.po
@@ -9,7 +9,7 @@
 ## effect: edit them in PO (`.po`) files instead.
 msgid ""
 msgstr ""
-"Language: chs\n"
+"Language: zh_Hans\n"
 "Plural-Forms: nplurals=2\n"
 
 #, elixir-format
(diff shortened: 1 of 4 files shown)


priv/gettext/zh_Hans/LC_MESSAGES/default.po
```diff
@@ -0,0 +1,573 @@
...
+## `msgid`s here are often extracted from source code.
+## Add new translations manually only if they're dynamic
+## translations that can't be statically extracted.
+##
+## Run `mix gettext.extract` to bring this file up to
+## date. Leave `msgstr`s empty as changing them here as no
+## effect: edit them in PO (`.po`) files instead.
+msgid ""
+msgstr ""
+"Language: zh_Hans\n"
```
adriankumpf, line 12: Here, an An IETF BCP 47 language tag is required, which is why I  replaced `chs` with `zh_Hans`. Hope that's okay :) The same applies to the folder name.

edward4hgl: Brilliant!

chinaoldcat: How to set my panel to Chinese? Thank you!

PR #633, opened 2020-04-23
Add Chinese (Traditional) translation

occultsound:
Add Traditional Chinese (taiwan) translation

### DIFF
--- a/priv/gettext/zh_Hant/LC_MESSAGES/default.po
+++ b/priv/gettext/zh_Hant/LC_MESSAGES/default.po
@@ -8,7 +8,7 @@
 # # to merge POT files into PO files.
 msgid ""
 msgstr ""
-"Language: zh_TW\n"
+"Language: zh_Hant\n"
 
 #: lib/teslamate_web/templates/car/summary.html.leex:93
 msgid "Status"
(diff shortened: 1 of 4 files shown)


priv/gettext/zh_Hant/LC_MESSAGES/default.po
```diff
@@ -0,0 +1,478 @@
...
+# #
+# # Do not add, change, or remove `msgid`s manually here as
+# # they're tied to the ones in the corresponding POT file
+# # (with the same domain).
+# #
+# # Use `mix gettext.extract --merge` or `mix gettext.merge`
+# # to merge POT files into PO files.
+msgid ""
+msgstr ""
+"Language: zh_Hant\n"
```
adriankumpf, line 11: Here, an An IETF BCP 47 language tag is required, which is why I had to replace it with a tag from the [IANA Language Subtag Registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry). The CLDR library we use only supports a subset: https://github.com/elixir-cldr/cldr/tree/master/priv/cldr/locales Hence "zh_Hant". Hope that's okay :) The same applies to the folder name.

occultsound: Thanks~

PR #642, opened 2020-04-24
Update charging-stats.json

Dulanic:
Updated to address https://github.com/adriankumpf/teslamate/issues/528#

Added case statement to adjust to show kWh at non decimal level and MWh at the 3 decimal point level. Cannot address directly within grafana so used case statement. Removed grafana sorting so can be sorted properly in SQL since not a direct # now for grafana to sort.


Dulanic:
Forgot to show what it looks like for quick reference.

image

SigLinJo:
I think it looks great and does exactly what I was after! Thank you @Dulanic :+1: I would be really happy if this is merged!

PR #665, opened 2020-04-29
Add option HTTP_BINDING_ADDRESS to control the bound IP address

dyxyl:
I couldn't find a way to force the http server to listen only on 127.0.0.1, so I've created this PR to add an environment variable `HTTP_BINDING_ADDRESS` to control the bound IP address. This is useful for extra security when running behind a reverse proxy.

The code overrides the `:inet6` option because you can control the protocol by using the appropriate string, e.g. `::1` for IPv6.


DrMichael:
Well, you can restrict the address in the docker-compose.yml:

    ports:
      - 127.0.0.1:4000:4000

restricts it to 127.0.0.1. I am not sure about IPv6 though.

Is that, what you meant?

dyxyl:
@DrMichael That works for docker, but I'm running on FreeBSD, based on the manual setup.

DrMichael:
Ah, I see...

dyxyl:
Yes, in theory, but the prebuilt containers are all Linux-based so they will run under ABI emulation.
It would be better to use the FreeBSD native containers (jails) directly.

PR #666, opened 2020-05-01
Updated charging stats to show charging map

Dulanic:
Added map to show frequently used chargers. The 2 pie charts were rather large so placed between and shrunk down. Added plugin for worldmap for grafana also.

image

This addresses and closes https://github.com/adriankumpf/teslamate/issues/476

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -8,7 +8,7 @@ ENV  GF_ANALYTICS_REPORTING_ENABLED=FALSE \
      GF_SECURITY_ALLOW_EMBEDDING=true \
      GF_SECURITY_DISABLE_GRAVATAR=true \
      GF_USERS_ALLOW_SIGN_UP=false \
-     GF_INSTALL_PLUGINS=pr0ps-trackmap-panel,natel-discrete-panel,grafana-piechart-panel,grafana-worldmap-panel \
+     GF_INSTALL_PLUGINS=pr0ps-trackmap-panel,natel-discrete-panel,grafana-piechart-panel,https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip;grafana-worldmap-panel-ng \
      DATABASE_PORT=5432
 
 COPY datasource.yml /etc/grafana/provisioning/datasources/
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -8,7 +8,7 @@ ENV  GF_ANALYTICS_REPORTING_ENABLED=FALSE \
      GF_SECURITY_ALLOW_EMBEDDING=true \
      GF_SECURITY_DISABLE_GRAVATAR=true \
      GF_USERS_ALLOW_SIGN_UP=false \
-     GF_INSTALL_PLUGINS=pr0ps-trackmap-panel,natel-discrete-panel,grafana-piechart-panel \
+     GF_INSTALL_PLUGINS=pr0ps-trackmap-panel,natel-discrete-panel,grafana-piechart-panel,grafana-worldmap-panel \
      DATABASE_PORT=5432
 
 COPY datasource.yml /etc/grafana/provisioning/datasources/
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -15,7 +15,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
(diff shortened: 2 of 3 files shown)


DrMichael:
@Dulanic I installed it locally, but I do not get any data points in the map, although the query gives nice lat/long values as result. Any idea?

Dulanic:
- @DrMichael it needs a additional plugin , you can go into the grafana commissioner and run grafana-cli plugins install grafana-worldmap-panel

DrMichael:
@Dulanic Yes, did that already. It turns out, that the data is on the map, but the map is centered at 0,0. How could it be centered at the biggest circle?

Dulanic:
@DrMichael Good question, let me look into it further, i thought it was but maybe it cached my local zoom.

Dulanic:
So the best option is to add a geohash to the query, and I can use the Last Geohash as the default option...  My downside is I have little to no experience /w geohash :( Let me see what I can figure out there.

DrMichael:
Hmm, is there no possibility to center it on the most used charging station? The info is in the result of the query...

Dulanic:
I can't find one, which is unfortunate. I changed the zoom, wondering if this may work.

Dulanic:
I found a work around, there is a fork that does focus on the data... Looking into this!

image

Dulanic:
@DrMichael OK, done /w my changes. This looks to work better and more function using this fork of world maps by panodata. Thoughts?

DrMichael:
Excellent, that looks terrific!

DrMichael:
@Dulanic I tried to understand the percentages... It shows how much percent one location has charged of the location with the max charge total, right?

Wouldn't it be clearer, if it would be the percentage of the total amount being charged?

So instead of
`chg_total * 1.0 / (SELECT MAX(chg_total) FROM charge_data) * 100`
better
`chg_total * 1.0 / (SELECT SUM(chg_total) FROM charge_data) * 100`?

Dulanic:
Yeah I had it that way but it made you have a bunch of low percents.Visually it looked funny. I found this looked better if you had multiple spots you idea for charging.

DrMichael:
Well, I was irritated, that the biggest charging station shows 100% ...

Dulanic:
What I don't like about the panel is the restriction to 1 metric. Let see
if I can come up with a solution.

Dulanic:
for now ill go back to sum, I had it that way originally.

kongomongo:
AND charge.car_id = '1'

this doesn't sound right.

kongomongo:
I agree that the percentages of the heatmap should add up to 100% not more ...

spacecosmos:
Where exactly do you need to run this command?:

grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip

Dulanic:
It was intended to be labelled % of highest charger. I'll just remove the
label and just have it adjust size. I get it it confuses.

tucopup:
I'm wondering what is happening - I have 3 cars in the account, all show the same big green and small red dots on the map.  Places I've not been but one of the other cars has.

I have the latest teslamate, 1.19.3, and think that I have correctly loaded the needed map plugin (since I see the map and circles).  Synology, docker.

Dulanic:
https://github.com/adriankumpf/teslamate/pull/680

DrMichael:
> Where exactly do you need to run this command?:
> 
> grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip

Shell: docker-compose exec grafana grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip

DrJohnM61:
> > Where exactly do you need to run this command?:
> > grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip
> 
> Shell: docker-compose exec grafana grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip

To be complete, the full command is:
docker-compose exec grafana grafana-cli --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.9.0/grafana-map-panel-0.9.0.zip plugins install grafana-worldmap-panel-ng

PR #672, opened 2020-05-02
Correct gap in efficency.json for mi only.

Dulanic:
Updated distance to convert_km(distance, '$length_unit') to correct gap in original SQL which wasn't converting the distance and only the min distance.

Fixes https://github.com/adriankumpf/teslamate/issues/671


dyxyl:
I think your patch is wrong: ```convert_km(distance, '$length_unit') >= convert_km($min_distance, '$length_unit')``` is just a slower equivalent to ```distance >= $min_distance```. The ```distance``` column is always in km, but the ```$min_distance``` parameter comes from the user, so I think TeslaMate should assume it is in the user's configured units.

Therefore the test should be ```convert_km(distance, '$length_unit') >= $min_distance```.

The function ```convert_km``` has a confusing name and should be called something like ```convert_distance_from_km_to_unit```.

Dulanic:
Agreed, will fix. It just needs the conversion removed from the filter. I know the function name is slightly confusing, but I know what it does :)

Issue #698, opened 2020-05-07
Notification when the car or window is open for an extended period of time


rtega:
I don't know why the tesla-app doesn't do this in the first place. I guess they don't want to be liable when people get their car or items from it stolen when the app by accident doesn't notify them.

The idea is to have teslamate send an email or some other notification when it notices that one of the following states happens for "an extended period of time":
- the car is unlocked
- one or more doors is open
- one or more windows is open
- the frunk and/or the trunk is open

The extended period of time in minutes would be an option in the settings.

Is this doable?

cleanev:
There is an app on iOS named stats that does this exact thing that you mention. Unfortunately there is no android equivalent

baylanger:
Configuring mail service can be tricky.  Most people would configure their ISP or VM provider's mail server and that often requires a username/password.  The SMPT service might be on port 25 and if it's using encryption that's another port.

I'd too like TM to be able to send me notification or automate some tasks.  I believe there are a few online services that can probably notify you via a smartapp, seems I recall seeing something in the past but can't remember the name.  This requires also to install and configure some software in a docker instance.  It's definitely possible to do.  Hopefully some people will comment below ideas how to do this "easily" for everyone.

ngardiner:
The window state is exported by MQTT, so any system which can consume MQTT topics could likely do this. In homeassistant for example, you could set an automation which triggers a script with a delay every time the window state is down. If the window state is set back to up, it would cancel the script. Once the delay has passed, you can use the notification components such as Amazon SNS (for email), Telegram, even Facebook Messenger to send a message.

If OP is using HASS outside of TeslaMate, I would be happy to put together an example automation for this, as I think it would be handy to have myself. We already have some HASS integration examples as part of the doco so if I get a chance to do this, I'll contribute it.

rtega:
Didn't know it could be that simple with mosquitto_sub. I hashed a simple bash-script to do this which could be further improved upon. I run it each 10 minutes with crontab so it is bound to catch some false positives but it works.
[checkscript.zip](https://github.com/adriankumpf/teslamate/files/4605269/checkscript.zip)

rtega:
You could also use something like Wirepusher to send a notification via curl
`
curl "https://wirepusher.com/send?id=your_id&title=Alert&message=`cat /tmp/doorswindowstate|tr '\n' ' '`&type=default"
`

JakobLichterfeld:
[...]

PR #705, opened 2020-05-09
Update HomeAssistant documentation

### LINKED ISSUE #695: Home Assistant device_tracker always shows not_home
**Describe the bug**

The Home Assistant device_tracker never knows the car is home.  Even if proximity.home_tesla is 0, the device_tracker entities never show the car as home.

**Expected behavior**

If the car is parked at home, the device_tracker should say 'home'.

**How to reproduce it (as minimally and precisely as possible)**:
Normal usage.

 
**Environment**

- TeslaMate version: v.1.19.3
- OS TeslaMate is installed on: Docker installed on Intel NUC, RHEL
- HA v0.109.4

ngardiner:
- Removed the static not_home location from the device_tracker definition which fixes #695 
   - Added Garage Door automation
   - Added automation for doors and windows open which would address #698 (for HomeAssistant users at least)

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -397,7 +397,7 @@ We add the random 30 second interval after each notification to avoid clobbering
 
 #### automation.yaml
 
-```
+```yml title="automation.yaml"
 - alias: Set timer if teslamate reports something is open to alert us
   initial_state: on
   trigger:
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -42,7 +42,6 @@ The following provides an automation to update the location of the `device_track
     - service: device_tracker.see
       data_template:
         dev_id: tesla_location
-        location_name: not_home
         gps:
           [
             "{{ states.sensor.tesla_latitude.state }}",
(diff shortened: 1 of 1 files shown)


website/docs/integrations/home_assistant.md
```diff
@@ -366,3 +365,139 @@ The below is the Lovelace UI configuration used to make the example screenshot a
...
+
+The following set of automations and scripts will detect when a Tesla door, frunk, trunk or window is left open. The script will notify you after the defined time period (by default, 5 minutes). If you would like to customize how the notification is performed, you can edit the ```notify_tesla_open``` script which is called by all of the four notifications.
+
+By default, the script will repeatedly notify every 5 minutes. Remove the recursive ```script.turn_on``` sequence in the ```notify_tesla_open``` script if you'd only like to be informed once.
+
+We add the random 30 second interval after each notification to avoid clobbering the notification script when we have multiple things open at once. For example, opening the door will open the door and the window. If we don't delay the calls, we will only get a message about the window (as it is the last call to the script) and if we then close the window, we won't get notifications about other things left open. This results in more notifications but less chance on missing out on knowing something was left open.
+
+#### automation.yaml
+
+```
```
adriankumpf, line 400:
```suggestion
```yml title="automation.yaml"
```

website/docs/integrations/home_assistant.md
```diff
@@ -366,3 +365,139 @@ The below is the Lovelace UI configuration used to make the example screenshot a
...
+
+## Useful Automations
+
+The below automations leverage TeslaMate MQTT topics to provide some useful automations
+
+### Garage Door Automation based on Tesla location
+
+This automation triggers when the Tesla transitions from not_home to home. This means that the vehicle would have had to have been outside of the home zone previously, and returned home. You may want to add conditions here to improve accuracy, such as time of day.
+
+```
```
adriankumpf, line 377:
```suggestion
```yml title="automation.yaml"
```

website/docs/integrations/home_assistant.md
```diff
@@ -366,3 +365,139 @@ The below is the Lovelace UI configuration used to make the example screenshot a
...
+      payload: 'false'
+  action:
+    - service: script.turn_off
+      data_template:
+        entity_id: script.notify_tesla_{{trigger.topic.split('/')[3]}}
+```
+
+#### script.yaml
+
+```
```
adriankumpf, line 444:
```suggestion
```yml title="script.yaml"
```

PR #712, opened 2020-05-12
Ensure dashboards are restored into the same folder as they currently belong to

sumnerboy12:
~added optional env variable `FOLDER_UID` to support restoring dashboards into a specific Grafana folder, instead of 'General' by default.~


sumnerboy12:
I see the Docker install will automatically create folders, but since I did a manual install I never got those provisioned. Hence the need for this override.

sumnerboy12:
Simplified this and removed the new env variable (`FOLDER_UID`). Instead the restore script will ensure that any dashboards which already exist, will be restored into the same folder as they currently belong to.

Issue #709, opened 2020-05-11
allow base MQTT topic to be configurable (instead of hardcoded to `teslamate/`) 


sumnerboy12:
**What would you like to be added**:
Allow (optional) configuration of the base MQTT topic, instead of being hardcoded to `teslamate/`

**Why is this needed**:
I would like teslamate data to live under an existing base topic, along with all my other home automation data/feeds.

**Additional context**:

sumnerboy12:
Just had a look at the code, one solution could be to remove `teslamate` from the hardcoded topic (line 79 in [here](https://github.com/adriankumpf/teslamate/blob/master/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex)) and just default `namespace` to `teslamate` if there is nothing explicitly defined in the env vars. This would keep the current behaviour (so long as `namespace` isn't set) and would mean no new config variables are needed.

This would break for anyone who currently uses `namespace` however, as they would need to prepend `telsamate/` to keep current behaviour.

PR #713, opened 2020-05-12
Allow configuration of base MQTT topic (defaulting to `teslamate`)

sumnerboy12:
Implements #709.

*NOTE* Breaking change for anyone with an explicit `MQTT_NAMESPACE` defined. They would need to prepend with `teslamate/` to ensure values are published to the same topics as currently.

Shouldn't effect anyone who doesn't use `MQTT_NAMESPACE`.

### DIFF
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -83,7 +83,7 @@ if System.get_env("DISABLE_MQTT") != "true" do
     password: System.get_env("MQTT_PASSWORD"),
     tls: System.get_env("MQTT_TLS"),
     accept_invalid_certs: System.get_env("MQTT_TLS_ACCEPT_INVALID_CERTS"),
-    namespace: System.get_env("MQTT_NAMESPACE") |> Util.validate_namespace!()
+    namespace: System.get_env("MQTT_NAMESPACE", "teslamate") |> Util.validate_namespace!()
 end
 
 config :logger,
--- a/config/releases.exs
+++ b/config/releases.exs
@@ -5,16 +5,6 @@ defmodule Util do
     :crypto.strong_rand_bytes(length) |> Base.encode64() |> binary_part(0, length)
   end
 
-  def validate_namespace!(nil), do: nil
-  def validate_namespace!(""), do: nil
-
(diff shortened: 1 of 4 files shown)


config/releases.exs
```diff
@@ -83,7 +83,7 @@ if System.get_env("DISABLE_MQTT") != "true" do
     password: System.get_env("MQTT_PASSWORD"),
     tls: System.get_env("MQTT_TLS"),
     accept_invalid_certs: System.get_env("MQTT_TLS_ACCEPT_INVALID_CERTS"),
-    namespace: System.get_env("MQTT_NAMESPACE") |> Util.validate_namespace!()
+    namespace: System.get_env("MQTT_NAMESPACE", "teslamate") |> Util.validate_namespace!()
```
adriankumpf, line 86: `validate_namespace/1` raises if `MQTT_NAMESPACE`contains a `/` which would make it impossible to use e.g. `teslamate/account_0`.

sumnerboy12:
It is common practice for anything publishing MQTT to allow configuration of the base topic. Users might have existing topic structures/standards they wish to adhere to.

config/releases.exs
sumnerboy12: My bad, didn't catch that - was there a reason for adding this validation? I will remove in this PR for now.

sumnerboy12:
I can't see why my change to remove the namespace validation is causing the `\vehicle\suspend_test.exs` to fail. Sorry, this is my first attempt at Elixir...

sumnerboy12:
@adriankumpf is there anything else you want me to do here, or would you prefer this wasn't merged?

adriankumpf:
At the moment, I am not in favour of introducing a breaking change, because such a feature is "common practice". However, this will make sense when releasing a version 2.0.

sumnerboy12:
Fair enough - will just leave this here then - unless you want me to close the PR?

adriankumpf:
Good question … let's close this for now. We still have #709 as a reminder to implement this.

PR #714, opened 2020-05-12
Add 'tesla' tag to internal dashboards for consistency

sumnerboy12:
Unless there was a reason for these being left out?


ngardiner:
These two dashboards are called by the other dashboards and offer nothing valuable on their own. I believe the reason they were not tagged was due to the number of issues being raised by people asking why the dashboards did not work (as they were clicking on them directly).

sumnerboy12:
Thanks @ngardiner, makes sense, I will close this PR.

PR #723, opened 2020-05-17
Update to show kWh even if still charging

Dulanic:
Will allow user to see the the amount charged even if still charging by using charges table instead of charge_processes.

### DIFF
--- a/grafana/dashboards/internal/charge-details.json
+++ b/grafana/dashboards/internal/charge-details.json
@@ -234,7 +234,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT charge_energy_added from charging_processes where id = $charging_process_id;",
+          "rawSql": "SELECT charge_energy_added from charges where id = $charging_process_id order by date desc fetch first 1 row only ;",
           "refId": "A",
           "select": [
             [


Dulanic:
Didnt mean to close this,

grafana/dashboards/internal/charge-details.json
```diff
@@ -234,7 +234,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT charge_energy_added from charging_processes where id = $charging_process_id;",
+          "rawSql": "SELECT charge_energy_added from charges where id = $charging_process_id order by date desc fetch first 1 row only ;",
```
adriankumpf, line 237: `id` should be `charging_process_id`

adriankumpf: In rare cases `charge_energy_added` does not start a zero so that the total energy added is less than the last reported `charge_energy_added` (`end_charge_energy_added - start_charge_energy_added` to be precise). Therefore we should still prefer to display the value reported in the `charging_processes` table, for example:

```SQL
WITH completed_charge AS (
	SELECT charge_energy_added
	FROM charging_processes
	WHERE id = $charging_process_id
),
in_progress_charge AS (
	SELECT charge_energy_added
	FROM charges
	WHERE charging_process_id = $charging_process_id
	ORDER BY date DESC FETCH FIRST 1 ROW ONLY
)
SELECT COALESCE(completed_charge.charge_energy_added, in_progress_charge.charge_energy_added)
FROM completed_charge
JOIN in_progress_charge on true
```

Dulanic: Thanks, I messed up my repor and had closed out this branch... so I couldnt fix it. I will close this PR as I made a new one.

https://github.com/adriankumpf/teslamate/pull/744

PR #734, opened 2020-05-20
Add supercharger dashboard

Dulanic:
Adding a supercharger dashboard. I added to doc also, wasn't positive if this would also add to the teslamate page? Please direct if another change needs to be added to create a new dashboard. Credit to Pankey on discord for this.

image

### DIFF
--- a/grafana/dashboards/supercharges.json
+++ b/grafana/dashboards/supercharges.json
@@ -534,7 +534,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "Supercharger history",
+      "title": "Supercharger History",
       "transform": "table",
       "type": "table"
     },
--- a/grafana/dashboards/supercharges.json
+++ b/grafana/dashboards/supercharges.json
@@ -534,7 +534,7 @@
       ],
       "timeFrom": null,
       "timeShift": null,
-      "title": "SuperCharges",
+      "title": "Supercharger history",
       "transform": "table",
       "type": "table"
     },
--- a/grafana/dashboards/supercharges.json
+++ b/grafana/dashboards/supercharges.json
@@ -1,0 +1,1067 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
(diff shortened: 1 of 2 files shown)


adriankumpf:
Cool idea! 

>  I added to doc also, wasn't positive if this would also add to the teslamate page? 

Yup, it's added. At compile time all dashboards in `grafana/dashboards` are automatically included. So adding/removing/renaming dashboards does not need additional changes.

grafana/dashboards/supercharges.json
```diff
@@ -0,0 +1,1067 @@
...
+        "ymin": null
+      },
+      "tableColumn": "",
+      "targets": [
+        {
+          "format": "time_series",
+          "group": [],
+          "metricColumn": "none",
+          "rawQuery": true,
+          "rawSql": "SELECT start_date as \"time\", charge_energy_added\nFROM charging_processes\nWHERE \n  $__timeFilter(end_date) AND \n  car_id = $car_id AND     \n   (charge_energy_added * 60 / NULLIF (duration_min, 0)) > 20 AND\n  ('${geofence:pipe}' = '-1' OR geofence_id in ($geofence))\nORDER BY start_date;",
```
adriankumpf, line 614: As I see, "supercharges" are determined by the charging speed:

```SQL
(charge_energy_added * 60 / NULLIF (duration_min, 0)) > 20
```

Due to the name of the dashboard I would have expected that only SuperCharges and no charges at other fast chargers would be displayed. If we wanted to include only the former we could filter for charging sessions where `fast_charger_brand` is `Tesla`.

Dulanic: As I think about this.... maybe I will think over a ability to use fast and/or super chargers and then set a filter. I'll work on it this weekend. Maybe modifying the charging dashboard is the best route too... I had thought about it too.

Dulanic:
Sp I have a small dataset... do you think just setting it a filter for fast_charger_brand or fast_charger_present is best? I've never used any fast chargers outside of tesla....

adriankumpf:
> Sp I have a small dataset... do you think just setting it a filter for fast_charger_brand or fast_charger_present is best? I've never used any fast chargers outside of tesla....

Yup, that sounds good to me. A filter for charger power would be interesting as well. We could also postpone this and create a new PR for it :)

Dulanic:
Yeah I was thinking any fast charger is > 300 v. Maybe that's the simplistic way to do it. I'll close it this PR until then.

PR #768, opened 2020-05-28
Optimize conversion helpers

### LINKED ISSUE #740: SQL function enhancement
Please remove IMMUTABLE STRICT from the convert functions. This prevents sql inlining.  The convert functions should look like so. I also took the time to replace it as the true data type it was changed to a while back for the database enhancements.

As can be seen here, IMMUTABLE prevents sql inlining which is more efficent for these types of case statements. https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions

I'd do it, but I have no idea how to code it in elixir format.

```

--DROP FUNCTION public.convert_km(double precision, text);

CREATE OR REPLACE FUNCTION public.convert_km(
	n numeric(6,2),
	unit text)
    RETURNS numeric(6,2)
    LANGUAGE 'sql'

    COST 100
    VOLATILE 
    
AS $BODY$
  SELECT
    CASE $2 WHEN 'km' THEN $1
         	WHEN 'mi' THEN $1 / 1.60934
    END;
$BODY$;

ALTER FUNCTION public.convert_km(numeric(6,2), text)
    OWNER TO teslamate;
```

```
-- FUNCTION: public.convert_celsius(double precision, text)

--DROP FUNCTION public.convert_celsius(double precision, text);

CREATE OR REPLACE FUNCTION public.convert_celsius(
	n numeric(4,1),
	unit text)
    RETURNS numeric(4,1)
    LANGUAGE 'sql'

    COST 100
    VOLATILE   
    
AS $BODY$
  SELECT
    CASE $2 WHEN 'C' THEN $1
         	WHEN 'F' THEN ($1 * 9 / 5) + 32
    END;
$BODY$;

ALTER FUNCTION public.convert_celsius(numeric(4,1), text)
    OWNER TO teslamate;
```

adriankumpf:
Closes #740


Dulanic:
Would you mind updating the drop statements to be IF EXISTS? Since I already did this for testing, it fails for me.

adriankumpf:
Sure! 7c7494a

PR #790, opened 2020-06-13
Update of upgrade-documentation

roadrash2108:
added "docker-compose up -d" as default for bringing containers up after upgrade (to match the description of installation-documentation).


cwanja:
Is there any benefit to doing a `docker system prune` during an upgrade? For example, here is my upgrade script:
`docker-compose down && docker-compose pull && docker system prune && docker-compose up -d`. Sorry for the newbie docker questions 😊

PR #880, opened 2020-08-25
Tweak some Dutch translations

fatbasstard:
Tweaked some of the Dutch translations (and added the new translation that was missing):

* `Geo-Fences`: Has no Dutch equivalent., just removed the `-` to make it "Dutch"
* `Trunk`: We call that the "kofferbak"
* `Per Minute`: Was missing so added that


adriankumpf (approved):
Thank you!

GtHbbrr:
Hi, also from NL. In the home screen it says in the first line “Status”: “online voor 3h”
My suggestion would be: “online sinds 3h”

fatbasstard:
@GtHbbrr I'll take a look

PR #881, opened 2020-08-25
More Dutch translation tweaks

fatbasstard:
As suggested here https://github.com/adriankumpf/teslamate/pull/880#issuecomment-680023633


adriankumpf (approved):
That was fast! Thanks again :)

PR #890, opened 2020-08-31
"Customize" Grafana logo

fatbasstard:
Couldn't help it, had to do it (close the PR if you want).

This PR replaces the Grafana logo (and favicon) with the Teslamate logo:

image

I always get confused with the Grafana I'm looking at (working with multiple Grafana's), let's make the Teslamate Grafana look like Teslamate. 😛 

(Images are "copies" from the assets/static folder)


adriankumpf (approved):
Fantastic idea! Thank you.

gimmespam:
Hey, I have saved the teslamate grafana site as "shortcut" to my iOS home screen. Looks like this change did not have any impact to it even though I recreated it to the home screen. Is there anyone who has the ability dig around where the Safari pulls the icon? Or it might be the case that the old icon is still in my phone's cache.

image

adriankumpf:
I also get the original Grafana icon. Safari seems to require an icon marked as `apple-touch-icon`:

https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

gimmespam:
That kind of approach did not cross my mind so I didn’t even bother to google for solution. Now when googling around the approach,  looks like similar change would also work for androids.

gimmespam:
So the solution would be to resize the existing teslamate icon to the needed pixel sizes, then in graphana’s Dockerfile copy them to right place and echo the safari ”Web Clip” and similar android details to the web template which loads for example the favicon image.

adriankumpf:
We should already have the icons: https://github.com/adriankumpf/teslamate/tree/master/assets/static If you have found the "right place", a PR would be very much appreciated :-)

gimmespam:
Ah, looks like all the pieces needed exist already. The grafana even has the "apple specific" label in the page source, it is just a matter of copying the apple-touch-icon.png from the teslamate assets source to the public/img/ directory of graphana in graphanas Dockerfile. I'll try to carry out the PR. ..I am a newbie with this github.

Issue #921, opened 2020-09-14
Allow map-zoom on car overview


leewillis77:
**What would you like to be added**:
Currently the overview of a car in Teslamate (http://teslamate.server.address:4000/) shows a small square map, and a summary of key information. If the car is driving, the map auto-updates and follows the car on its journey. Unfortunately, the zoom of that map is fixed, and often the small size, and fixed zoom mean there's not enough context for the map to actually be useful.

For example, the current map view I'm looking for my car looks like this:
![Screenshot 2020-09-14 at 15 57 56](https://user-images.githubusercontent.com/1097338/93102730-a9f92100-f6a3-11ea-9b1b-f3118cd0177a.png)

If I adjust the zoom I could get enough context about the position for this to be much more useful.

PR #924, opened 2020-09-15
Allow zoom controls in car live view

longzheng:
Addresses #921 

Copied `data-zoom="1"` from https://github.com/adriankumpf/teslamate/blob/master/lib/teslamate_web/live/charge_live/cost.html.leex#L17 which seems to enable zooming according to the JS code https://github.com/adriankumpf/teslamate/blob/master/assets/js/hooks.js#L162

(I have not tested this locally since I don't have a dev environment set up)


leewillis77:
My install is via docker currently so I can't reasily just apply the change I don't think - it doesn't look like I can run a shell on the teslamate docker instance. Is there any way I can help test - or would I need to set up a separate manual instance?

PR #944, opened 2020-09-25
Put logo in readme.md

douglasevaristo:


adriankumpf (changes requested):
I am not a fan of how the image is aligned on the page.  Is there any way to place the logo to the left of the headline?

PR #964, opened 2020-10-01
Change colours of pie chart

DrMichael:
As asked for in https://github.com/adriankumpf/teslamate/issues/608

### DIFF
--- a/grafana/dashboards/trip.json
+++ b/grafana/dashboards/trip.json
@@ -1,4 +1,70 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_TESLAMATE",
+      "label": "TeslaMate",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "postgres",
+      "pluginName": "PostgreSQL"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "panel",
+      "id": "bargauge",
+      "name": "Bar gauge",
+      "version": ""
+    },
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "7.2.0"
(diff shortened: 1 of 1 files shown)


grafana/dashboards/trip.json
```diff
@@ -46,7 +113,7 @@
   "panels": [
     {
       "collapsed": false,
-      "datasource": "TeslaMate",
+      "datasource": "${DS_TESLAMATE}",
```
adriankumpf, line 116: The `datasource` must not be changed. It would break other setups.

grafana/dashboards/trip.json
```diff
@@ -2806,6 +2740,6 @@
   },
   "timezone": "",
   "title": "Trip",
-  "uid": "FkUpJpQZk",
-  "version": 1
-}
+  "uid": "T_dr9MFMk",
```
adriankumpf, line 2743: Please revert this chagne to the `uid`.l

grafana/dashboards/trip.json
```diff
@@ -1,4 +1,70 @@
 {
+  "__inputs": [
```
adriankumpf, line 2: I'm not sure where these changes are coming from. It would be nice if you could remove them.

adriankumpf (changes requested):

grafana/dashboards/trip.json
DrMichael: Aaaah, I think those strange changes came from switching this on
image

I will make another PR...

DrMichael:
I made a new PR https://github.com/adriankumpf/teslamate/pull/970. I tried to make minimal changes...

PR #965, opened 2020-10-01
Added Statistics dashboard

DrMichael:
for yearly, monthly, weekly, daily stats... As asked for in https://github.com/adriankumpf/teslamate/issues/664

### DIFF
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -1,0 +1,719 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_TESLAMATE",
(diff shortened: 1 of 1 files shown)


DrMichael:
The lins don't work yet, waiting for Grafana 7.3 to be able to fix that.

adriankumpf:
Great dashboard! You can tell how much time you've put in.

Regarding the timezone issue, I think we can first update the documentation so that people know how to configure the correct timezone. I will also add a text label to the dashboard that shows the currently used timezone.

adriankumpf (approved):

PR #968, opened 2020-10-03
Allow negative cost_per_unit for geofences

### LINKED ISSUE #939: Allow a negative Cost per kWh
**What would you like to be added**:
I would like to be able to set a negative "Cost per kWh" on the geo-fence page.

**Why is this needed**:
We pay a monthly subscription to have a charging station at home and then get reimbursed for the amount used. However, the amount reimbursed is slightly higher than what we pay per kWh on our electrical bill.

**Additional context**:
While the amount is relatively low, it's still a part of our Tesla economy and we would like to be able to track it.

ayonix:
Closes #939


adriankumpf (approved):
Thank you!

PR #973, opened 2020-10-05
Added links Statistics again

DrMichael:
Some smaller changes:

- the should be invisible fields moved to the far right and made them small
- added panel title to be able to sort by clicking the column header
- timezone back to current_setting, given that you can set it via PGTZ in the Grafana container

### DIFF
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -15,7 +15,6 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": 31,
   "iteration": 1601887874542,
   "links": [
     {
@@ -38,7 +37,7 @@
   "panels": [
     {
       "collapsed": false,
-      "datasource": "TeslaMate",
+      "datasource": "${DS_TESLAMATE}",
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -419,8 +418,8 @@
         ]
       },
       "gridPos": {
-        "h": 17,
-        "w": 21,
+        "h": 18,
+        "w": 23,
         "x": 0,
         "y": 1
       },
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -547,7 +547,7 @@
           "id": "organize",
           "options": {
             "excludeByName": {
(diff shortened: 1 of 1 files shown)


grafana/dashboards/statistics.json
```diff
@@ -696,16 +748,16 @@
           "value": "GMT"
         },
         "datasource": "TeslaMate",
-        "definition": "select name from pg_timezone_names where name like 'posix/%';",
+        "definition": "SELECT current_setting('TIMEZONE');",
```
adriankumpf, line 751: Subsequently changing the timezone via `PGTZ` does not work for me. 

Could both approaches be combined? Then a meaningful timezone could be preselected, but one would still have the possibility to change the timezone.

adriankumpf (commented):
Using invisible fields is a simple workaround 👍

grafana/dashboards/statistics.json
DrMichael: > Subsequently changing the timezone via `PGTZ` does not work for me.
> 
You have set it for the Grafana container?

> Could both approaches be combined? Then a meaningful timezone could be preselected, but one would still have the possibility to change the timezone.
Well, "select current_setting('TIMEZONE'), name from pg_timezone_names where name like 'posix/%';" should do it. It does for my changed version, it does not for the "official" version. Whyoever...

adriankumpf: > You have set it for the Grafana container? (Not the Database one!)

Oh, that worked! Okay, let's make the time zone configurable then but default to the current time zone setting.

DrMichael: Ok, that's fine... And links as soon as the hidden fiels issue is fixed?

adriankumpf: Yeah, that'd be great. I hope they will fix it soon!

DrMichael: Just saw, you have the links in. So, then it is only needed to hide the two fields, when it is fixed.

adriankumpf: Ah, sorry that's what I meant. Those almost invisible fields are fine for now.

PR #975, opened 2020-10-05
Fix tooltips in car overview being hidden by .card

ayonix:
As mentioned in https://github.com/adriankumpf/teslamate/issues/943#issuecomment-703233419 the tooltips weren't visible in the car overview.

Setting `overflow: visible` on `.car` is a workaround, there might be a better way, but it doesn't seem to break anything and it works.


adriankumpf (approved):
Thanks!

PR #976, opened 2020-10-05
Display update version in the homescreen update tooltip

### LINKED ISSUE #943: Update SW version in Teslamate homescreen?
**What would you like to be added**:
TM shows nicely, when a new update is available. Would it be possible to also show the version of the update?

**Why is this needed**:
Currently you have to go to the car to see the SW version, which is up to installing.

Of course a question, whether the API provides that information...

ayonix:
When an update is available the version should be known and shown in the icon tooltip, which closes #943. 

Maybe we should not show the parantheses if the version is not available for some reason

If there is anything wrong with the formatting/localization/anything else, just let me know.

### DIFF
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -11,7 +11,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
     speed outside_temp inside_temp is_climate_on is_preconditioning locked sentry_mode
     plugged_in scheduled_charging_start_time charge_limit_soc charger_power windows_open doors_open
     odometer shift_state charge_port_door_open time_to_full_charge charger_phases
-    charger_actual_current charger_voltage version update_available is_user_present geofence
+    charger_actual_current charger_voltage version update_available update_version is_user_present geofence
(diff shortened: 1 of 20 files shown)


lib/teslamate/vehicles/vehicle/summary.ex
```diff
@@ -11,7 +11,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
     speed outside_temp inside_temp is_climate_on is_preconditioning locked sentry_mode
     plugged_in scheduled_charging_start_time charge_limit_soc charger_power windows_open doors_open
     odometer shift_state charge_port_door_open time_to_full_charge charger_phases
-    charger_actual_current charger_voltage version update_available is_user_present geofence
+    charger_actual_current charger_voltage version update_available update_version is_user_present geofence
```
adriankumpf, line 14: Fields added here are automatically published via MQTT. Could you add the new `update_version` topic to the MQTT docs (https://github.com/adriankumpf/teslamate/blob/master/website/docs/integrations/mqtt.md)?

adriankumpf (commented):
Nice work!

adriankumpf (approved):
Thanks!

PR #978, opened 2020-10-06
Add a docker-compose for development

jrester:
This adds the ability to use docker-compose for development with hot reloading. Now one only has to type `docker-compose -f docker-compose.dev.yml up -d` to get started with development.


docker-compose.dev.yml
```diff
@@ -0,0 +1,39 @@
...
+      - db
+      - mqtt
+    environment: 
+      MIX_ENV: dev
+      DATABASE_HOST: db
+      MQTT_HOST: mqtt
+    ports:
+      - 4000:4000
+    volumes:
+      - .:/app
```
adriankumpf, line 30: If the local directory is used as volume for the "/app" directory, then the installed Elixir/npm dependencies will be lost, won't they? With a fresh repo it fails for this reason.

config/dev.exs
```diff
@@ -33,12 +33,12 @@ config :teslamate, TeslaMate.Repo,
   username: "postgres",
   password: "postgres",
   database: "teslamate_dev",
-  hostname: "localhost",
+  hostname: System.get_env("DATABASE_HOST", "localhost"),
```
adriankumpf, line 36: This is no longer necessary with the latest changes. The env variables are now respected in development too! :-)

Dockerfile.dev
```diff
@@ -0,0 +1,18 @@
+FROM elixir:1.10-alpine
```
adriankumpf, line 1: On the master branch Elixir 1.11 is now required. Since the official Docker image is not yet available, this could be used instead: ` FROM hexpm/elixir:1.11.0-erlang-23.1-alpine-3.12.0 AS builder `.

adriankumpf (changes requested):
I'm a bit in doubt if it is a good idea to use Docker for development. I think it makes things overly complicated. But if it helps some, there is no reason not to offer it as an additional option. Dependency caching must work properly though in order to merge this.

docker-compose.dev.yml
jrester: Yea, you are right! I changed it, but you now have to set the correct permission by supplying your UID and GID.

jrester:
@adriankumpf I believe it provides a much more easy starting point. To get started developing one only has to spin up a container. The dependencies are cached on the host so, when running teslamate without the container it should still work.

adriankumpf:
I see. Once `docker-compose -f docker-compose.dev.yml up -d ` works on a fresh repo and (and conflicts are resolved) this can be merged :-)

jrester:
For me it works perfectly fine on a new repo. You just have to set the environment variables `UID` and `GID` to the ids of your user to have it work correctly. 
I will add documentation for this, once the PR is merged.

adriankumpf:
Something must have gone wrong during merging. The "Files changed" tab shows 5k changed lines 😱  Can you take a look?

akac:
I can attest that in my environment, this would make dev easier.

PR #980, opened 2020-10-06
Always show 0% and 100% on state of charge diagram

mbertheau:
In the beginning of recording data with teslamate it's likely that the SoC hasn't gone down close to 0. However the minimum value in the diagram is determined by the minimum SoC recorded in TeslaMate. The resulting diagram is arguably misleading.

This PR fixes that by fixing the minimum and maximum values for the SoC diagram to 0 and 100 respectively.

Caveat: I didn't test this change.

### DIFF
--- a/grafana/dashboards/charge-level.json
+++ b/grafana/dashboards/charge-level.json
@@ -167,8 +167,8 @@
           "format": "percent",
           "label": "Charge Level",
           "logBase": 1,
-          "max": null,
-          "min": null,
+          "max": "100",
+          "min": "0",
           "show": true
         },
         {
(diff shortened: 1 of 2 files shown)


grafana/dashboards/charge-level.json
```diff
@@ -167,8 +167,8 @@
           "format": "percent",
           "label": "Charge Level",
           "logBase": 1,
-          "max": null,
-          "min": null,
+          "max": 100,
```
roadrash2108, line 170: I believe this should be "100" (with the quotes) - when editing in Grafana directly it also makes quotes around the numbers...

roadrash2108 (changes requested):
line 170 and 171: the numbers should be with quotes.

grafana/dashboards/charge-level.json
mbertheau: Thanks, fixed.

adriankumpf (approved):
Thanks!

Issue #722, opened 2020-05-17
Add export gpx files


mrwolf78:
I would like the possibility to export drive details to a gpx file. 

Why? Because sometimes I drive a nice route, and I'd like to share or ride it again later, which would be easier if I had a gpx file (or other shareable format)

PR #993, opened 2020-10-10
Gpx export

ayonix:
The current prototype exposes an endpoint to download a drive as a gpx file.
The drive details dashboard would have a link to the endpoint, so you could directly download the file from the dashboard.

As a note: I don't exactly know the details and features of GPX but as a minimal solution a list of trackpoints should be sufficient.

This is an effort to implement #722 

Todos:
- [x] Test the endpoint/view

### DIFF
--- a/lib/teslamate_web/controllers/drive_controller.ex
+++ b/lib/teslamate_web/controllers/drive_controller.ex
@@ -1,0 +1,22 @@
+defmodule TeslaMateWeb.DriveController do
+  use TeslaMateWeb, :controller
+
+  require Logger
+
+  alias TeslaMate.Log.Drive
+  alias TeslaMate.Repo
+
+  def index(conn, %{"id" => id}) do
+    drive = Repo.get(Drive, id) |> Repo.preload(:positions)
+    filename = drive.start_date
+
+    conn
+    |> put_resp_content_type("application/xml")
+    |> put_resp_header("content-disposition", ~s(attachment; filename="#{filename}.gpx"))
+    |> render("gpx.xml", drive: drive)
+  end
+end
+
+defmodule TeslaMateWeb.DriveView do
+  use TeslaMateWeb, :view
+end
--- a/lib/teslamate_web/controllers/drive_controller.ex
+++ b/lib/teslamate_web/controllers/drive_controller.ex
@@ -6,17 +6,21 @@ defmodule TeslaMateWeb.DriveController do
   alias TeslaMate.Log.Drive
   alias TeslaMate.Repo
 
-  def index(conn, %{"id" => id}) do
+  def gpx(conn, %{"id" => id}) do
     drive = Repo.get(Drive, id) |> Repo.preload(:positions)
-    filename = drive.start_date
+
+    case drive do
+      nil -> conn |> send_resp(404, "Drive not found")
+      drive -> send_gpx_file(conn, drive)
+    end
+  end
+
+  defp send_gpx_file(conn, drive) do
+    filename = "#{drive.start_date}.gpx"
 
     conn
     |> put_resp_content_type("application/xml")
-    |> put_resp_header("content-disposition", ~s(attachment; filename="#{filename}.gpx"))
+    |> put_resp_header("content-disposition", ~s(attachment; filename="#{filename}"))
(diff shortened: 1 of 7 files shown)


adriankumpf:
Cool! I really like the simplicity of the integration with the dashboard.

ayonix:
I think it might be tested enough now, but as always, if you have any concerns, let me know :) 

In the future it might be better to have a real gpx library/implementation, since i'm really not sure what I'm breaking with this.
Also the newlines take some unneeded space, but this should be easy to fix

ayonix:
Is there anything I should do or you want to improve? :)

lib/teslamate_web/controllers/drive_controller.ex
```diff
@@ -0,0 +1,26 @@
+defmodule TeslaMateWeb.DriveController do
+  use TeslaMateWeb, :controller
+
+  require Logger
+
+  alias TeslaMate.Log.Drive
+  alias TeslaMate.Repo
+
+  def gpx(conn, %{"id" => id}) do
+    drive = Repo.get(Drive, id) |> Repo.preload(:positions)
```
adriankumpf, line 10: Here we cannot rely on the standard sorting:

```elixir
  import Ecto.Query

  drive =
    Drive
    |> Repo.get(id)
    |> Repo.preload(positions: from(p in TeslaMate.Log.Position, order_by: p.date))
```

test/teslamate_web/controllers/drive_controller_test.exs
```diff
@@ -0,0 +1,99 @@
...
+      elevation: 100
+    })
+
+    Log.insert_position(drive, %{
+      date: DateTime.utc_now(),
+      latitude: 10.0,
+      longitude: 10.0,
+      elevation: 200
+    })
+
```
adriankumpf, line 32: To test the date-based sorting, a third `Position` with a date in the past could be inserted here.

adriankumpf (changes requested):
Sorry for the delay. The changes looks very good! 🙌  I only noticed one small thing regarding the sorting, which can cause drives to look like this:

Screenshot 2020-12-22 at 16 45 54

test/teslamate_web/controllers/drive_controller_test.exs
ayonix: Done as suggested :)

adriankumpf (approved):
Thank you! 🙏

Dustin781:
Hello, is it also possible to export all trips to a GPX file - i.e. my complete TeslaMate. Preferably a GPX file for every trip. The way I see it, this is currently only possible manual?

ayonix:
Yes, currently this would only be possible manually/programmatically. You could write a script that iterates over ids `1` to `n`, where `n` is the currently highest id and downloads the file under `/drive/$n/gpx`.
Then you'd also have one file per drive.

Dustin781:
The the answer. Unfortunately I can't program, so I hope that an export function will come up.

DrMichael:
That caught me. Pls find attached a small python3 script. You need to install python3-psycopg2, either with `sudo apt install python3-psycopg2` or with `pip install psycopg2`.

Then you need to call it with the drive id... It writes the route to out.xml.
[GetGpsTrack.zip](https://github.com/adriankumpf/teslamate/files/9175722/GetGpsTrack.zip)

DrMichael:
Ok, I tweeked it a bit. Now it asks for From, To and CarID. If you give an empty input, then it defaults to last 24 hours and car no 1. It now produces a track for every drive.
[](url)
Output now is out.gpx. You still need to set the database parameters either by environment or manually in the code.
[GetGpsTrack.zip](https://github.com/adriankumpf/teslamate/files/9179941/GetGpsTrack.zip)

Dustin781:
Unfortunately I don't understand the file. I can neither query database nor program. Wouldn't it be great to have a button to download all the tracks directly from the website with the push of a button?

For example, I downloaded all my training sessions from Apple Health as a ZIP file and then put them together and have everything directly in Google Earth. What would I like in TeslaMate.

It would also be an idea that this is installed under TRIP? You can already select the start date and end date and then the card is there. Only a new button "Export GPX" would have to be installed here.

PR #994, opened 2020-10-11
fix version info on development guide & minor spelling fix

techgaun:
I was trying to setup local dev and noticed that the recent updates to upgrade elixir 1.11.0 has resulted in this project not working with elixir<1.11 (such as use of `config_env/0`. I fixed that information here.

Additionally, I wanted to fix the debian guide too since erlang-solutions package site has not yet received 1.11.0 uploaded. I wasn't sure which one of various approaches you would want to take such as using version managers or downloading pre-compiled binaries from github release. Happy to create a PR to fix debian installation guide.


adriankumpf (approved):
Thanks! 

Regarding the debian guide: I would just wait a little bit until the package is out. Even the official docker image is not yet available :-)

Issue #989, opened 2020-10-08
Drives: How to add odometer to dashboard?


Eriktesla:
It would be great to have the possibility to view odo_start and odo_end in the **drives** dashboard:

This information is needed for the kilometer administration for the Dutch tax return ;-)

Any help to add odo_start and odo_end in the JSON panel is greatly appreciated.

cwanja:
I think it used to be on the Drivers dashboard, but to reduce the scrolling on the dashboard it was removed. I think it makes more sense on the Drive Details, under the smaller blocks in my opinion.

![Screen Shot 2020-10-08 at 12 27 01 PM](https://user-images.githubusercontent.com/54645561/95493071-93bc4900-0961-11eb-82f2-b196ccf64bed.png)

Eriktesla:
The tax authorities impose the following requirements on the kilometer administration:

Drive date, drive ID, the **start and end position of the odometer** and the departure and arrival address. 

The only items what I miss in the Drives dashboard is odometer start and odometer end.

If those items are available then it is a peace of cake to export all car trips (each trip is a row) to csv format.

wooter:
The fields **start_km** and **end_km** are available in the table **drives**.

I'd have to look into the source to see how to integrate a new query and JSON in the necessary Drives dashboard, but basically you need to 

- add "**start_km, end_km,**" in both **select** queries.
- add 2 Column styles, one for **start_km** and the other for **end_km**

roadrash2108:
I'm on it. Are above columns all you need?

Then I would suggest to build a new dashboard soley for the purpose of "Tax reporting" - what do you think?

wooter:
Yeah, I made it myself already; copying the Drives dashboard into a separate dashboard, and modifying the columns.

I can do a push request, but it's already a crowded table.

roadrash2108:
This would be my proposal for the table:
![Screenshot 2020-10-09 002646](https://user-images.githubusercontent.com/26000761/95519807-2bec1980-09c6-11eb-9690-350e24e922cd.png)

Eriktesla:
Wow I am impressed! That is exact the information what is needed.
Maybe it is possible to add the column with the end date/time of the ride and a column with the duration of the ride. 
That makes it complete.

roadrash2108:
New version:
![Screenshot 2020-10-10 193832](https://user-images.githubusercontent.com/26000761/95661507-3296a000-0b30-11eb-9d82-e10afa15bee6.png)

[...]

Issue #814, opened 2020-06-28
Texts/labels and the graphs in Charging stats does not match


petariga:
**Describe the bug**
It is just a nuance. When you look at the “Charging stats” in Grafana the circle graph AC/DC split has the AC value on the right (green) and DC value on the left (yellow), however the name of the graph states “AC/DC - kWh”, so from the first look you expect the AC on the left and DC on the right. The same vice versa expectation is for the “AC/DC - Duration”.

**Expected behavior**
It the graph label states AC/DC, then the circle graph should have AC charges on the left and DC charges on the right (regardless of the color chosen).

**Screenshots**

![54752521-3EDF-4754-A374-C065794D11A4](https://user-images.githubusercontent.com/64809507/85946682-500b7980-b946-11ea-8712-26b67e95a175.jpeg)
![99AF5E45-26D4-4084-A3FE-085D61FB6A3E](https://user-images.githubusercontent.com/64809507/85946684-5568c400-b946-11ea-8238-3d5f4ae5aaec.jpeg)

**Environment**

- TeslaMate version: 1.19.3
- Type of installation: Docker
- OS TeslaMate is installed on: Linux
- User OS & Browser: any
- Others:

petariga:
I don't want to be too picky, but I have just seen the result in the version 1.20.1 and I think that there is still a way to improve.

Previously, **AC was green and DC was yellow**, which was correct in my opinion (AC is better for the battery live, then I would expect it to use the green color). Now **the colors are vice versa**. The title and the order in the circle graph match, this is correct (left AC, right DC), however when you look under the circle graphs, **the two rows are now vice versa** - the first row is DC, the second is AC. The previous order was correct - **first AC and then DC** (the same is in the title of the graph and in the circle graph), now it has changed to the exact opposite.

![image](https://user-images.githubusercontent.com/64809507/97220581-f91a9180-17cb-11eb-8d7b-7387e62a542e.png)
![image](https://user-images.githubusercontent.com/64809507/97220662-18b1ba00-17cc-11eb-82fd-a1a866982cf1.png)

roadrash2108:
I will look into it. @petariga in case you have a docker-installation of your TeslaMate consider changing the grafana-part to "edge", so that you get the latest changes before the release and can give early feedback.

roadrash2108:
So, here the summary of my investigation:
- Grafana uses the plugin [piechart-panel](https://github.com/grafana/piechart-panel) for the pie-chart
[...]

PR #998, opened 2020-10-12
new dashboard for reporting to Dutch tax & matching of Titles/labels now match pie-charts in Charging stats

roadrash2108:
solves #989 and #814 - sorry for mixing into one PR. Seems that I have to learn more about Git -.-


tonypilborg:
Should we collaborate to make a new table hold hold a "classification" of drives (personal vs. business), as well as a "note" (could be costumer, if geo-fencing is not applicable in rare cases..??

grafana/dashboards.yml
```diff
@@ -21,3 +21,13 @@ providers:
   updateIntervalSeconds: 86400
   options:
     path: /dashboards_internal
+- name: 'teslamate_reports'
```
adriankumpf, line 24: Good idea to put those specialized dashboards in a separate folder 👍

grafana/dashboards/reports/dutch-tax.json
```diff
@@ -0,0 +1,509 @@
...
+            {
+              "name": "$__timeFilter",
+              "params": [],
+              "type": "macro"
+            }
+          ]
+        }
+      ],
+      "title": "Drive",
+      "type": "table"
```
adriankumpf, line 346: I migrated this table to the new table component. It comes with a bunch of new features and looks better on mobile.

adriankumpf (approved):
Thanks!

grafana/dashboards/reports/dutch-tax.json
roadrash2108: Thanks!

grafana/dashboards/reports/dutch-tax.json
```diff
@@ -320,7 +320,7 @@
...
...rt_geofence.name, CONCAT_WS(', ', COALESCE(start_address.name, nullif(CONCAT_WS(' ', start_address.road, start_address.house_number), '')), start_address.city)) AS start_address,\n    COALESCE(end_geofence.name, CONCAT_WS(', ', COALESCE(end_address.name, nullif(CONCAT_WS(' ', end_address.road, end_address.house_number), '')), end_address.city)) AS end_address,\n    duration_min,\n    distance\n  FROM drives\n  LEFT JOIN addresses start_address ON start_address_id = start_address.id\n  LEFT JOIN addresses end_address ON end_address_id = end_address.id\n  LEFT JOIN geofences start_geofence ON start_geofence_id = start_geofence.id\n  LEFT JOIN geofences end_geofence ON end_geofence_id = end_geofence.id\n  LEFT JOIN cars car ON car.id = drives.car_id\n  WHERE $__timeFilter(start_date) AND drives.car_id = $car_id\n  ORDER BY drive_id DESC\n)\nSELECT\n    drive_id,\n    start_date_ts,\n    convert_km(start_km::numeric, '$length_unit') as start_$length_unit,\n    start_address,\n    end_date_ts,\n    convert_km(end_km::numeric, '$length_unit') as end_$length_unit,\n    end_address,\n    duration_min,\n    convert_km(distance::numeric, '$length_unit') AS distance_$length_unit\nFROM data;",
```
roadrash2108, line 323: It was intentionally that I didn't show geofences as the requirement in #989 was to have the address - so I anticipated, that a geofence or location-name wouldn't be sufficient....

adriankumpf: My bad. I assumed that the same behaviour as in the Drive dashboard would be useful.

tonypilborg: @adriankumpf : Did you see my concerns regarding "missing start of drives", and the issue where a late start of logging has the consequence that start_km of drive x is not equal to end_km for drive x-1....

It can (will) result in missing km in a tax report - similar to how the current "trip" report goes from start_km of first drive to end_km of last drive... And that is obviously not always the sum of the logged drives...
- So perhaps a way to cluster subsequent drives with the same tag (same logical trips)???

adriankumpf: You always have to expect that trips can be recorded incompletely or even missing completely. A manual follow-up check will always be necessary imo.

tonypilborg: That's my point - adding the end_km from the previous drive could be of value... It probably shouldn't be visible in a standard report - but when exporting to excel it could be used to quickly do this validation - and adjust in there, since adjust the start_km directly in the database is probably not without implications (or perhaps it is fine to adjust the start_km - and we should aim to allow for that in the frontend?)

PR #1004, opened 2020-10-13
Replace default Graphana Web App logo with Teslamate logo

gimmespam:
Replace existing grafana logo with teslamate logo for apple devices when creating an "app" to home screen. Enhances existing #890 PR. Still have to check how this behaves with android devices, at least with the preliminary testing the results vary based on the used browser in android device.


adriankumpf (approved):
Thanks!

PR #1039, opened 2020-10-25
Improve garage door automation example

andrewfoster:
The current automation example incorrectly triggers based on the state of the garage door switch, rather than on the state of the Tesla's location. Also changed the action to turn on the garage door switch to be consistent with the description of the automation.


adriankumpf (approved):
Thanks!

PR #1016, opened 2020-10-16
add more filtering capabilities

Kosta-Github:
* charges: allow to filter for minimum charging `duration` in minutes (defaults to `0`, so no change in behaviour for the initial page view)
* drives: allow to filter for minimum `distance` and `speed` (both default to `0`, so no change in behaviour for the initial page view)

Note: this has only been tested with `km` (metric system); this also needs some testing for `miles` (imperial system).

PR #1045, opened 2020-10-25
Add minimum duration time

RafaG:
Add minimum duration time in minutes to main filter. For example, I don't want to see charges below 60 minutes.

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -511,7 +511,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


roadrash2108 (approved):
I really like this one! IMHO there should be a maximum filter as well.

RafaG:
> I really like this one! IMHO there should be a maximum filter as well.

In my opinión, current "min voltage" filter has no sense in this dashboard, I think should be removed. Add a maximum duration filter is so easy, if you want it, I can do it

roadrash2108:
> > I really like this one! IMHO there should be a maximum filter as well.
> 
> In my opinión, current "min voltage" filter has no sense in this dashboard, I think should be removed. Add a maximum duration filter is so easy, if you want it, I can do it

I must admit I didn't get the point of having "min voltage", but when looking in the explanation of #857 then it makes sense for me ("Added filter for voltage to allow user to filter down to for example, superchargers.").

In my opinion a feature shouldn't be removed because it doesn't make sense for yourself, as the feature was added for a specific reason. It could be discussed within an issue to remove it, if it really should be removed - for this specific feature I don't see the problem of having more filter as (at least for my devices where I use the dashboards) enough space is left...

Regarding the "maximum voltage" - either you or me can add it. As you prefer!

RafaG:
> I must admit I didn't get the point of having "min voltage", but when looking in the explanation of #857 then it makes sense for me ("Added filter for voltage to allow user to filter down to for example, superchargers.").

Great, I understand now why this filter exists, thanks you
 
> Regarding the "maximum voltage" - either you or me can add it. As you prefer!

I have limited access now, I prefer if you add it

Regards

adriankumpf (approved):
Looks good to me. Thanks!

A filter for maximum duration would make sense too. The labels should then perhaps be updated (shorter description) to make it look a bit more organized.

roadrash2108:
I just wanted to add the maximum duration filter and streamline the labels, but realized, that I don't have the latest changes as this PR isn't merged. Wouldn't it be easier if @RafaG adds it to his repo and then add the commit to this PR? If not, what would be the right approach for me? (forking RafaG:patch-1?)

Thanks for clarification and sorry for possible git-noob-question ;)

adriankumpf:
Ah, I just realized that #1016 already added the minimum duration filter. So the changes are now on the master branch. I'll close this one. Nevertheless, thanks for the contribution, @RafaG!

PR #1051, opened 2020-10-27
Update addresses in dutch-tax

roadrash2108:
Remove names (geofence, address) to only display street+number+city.


grafana/dashboards/reports/dutch-tax.json
```diff
@@ -320,7 +320,7 @@
...
+          "rawSql": "WITH data AS (\n  SELECT\n    drives.id as drive_id,\n    round(extract(epoch FROM start_date)) * 1000 AS start_date_ts,\n    round(extract(epoch FROM end_date)) * 1000 AS end_date_ts,\n    start_km,\n    end_km,\n    CONCAT_WS(', ', CONCAT_WS(' ', start_address.road, start_address.house_number), start_address.city) AS start_address,\n    CONCAT_WS(', ', CONCAT_WS(' ', end_address.road, end_address.house_number), end_address.city) AS end_address,\n    duration_min,\n    distance\n  FROM drives\n  LEFT JOIN addresses start_address ON start_address_id = start_address.id\n  LEFT JOIN addresses end_address ON end_address_id = end_address.id\n  LEFT JOIN cars car ON car.id = drives.car_id\n  ORDER BY drive_id DESC\n)\nSELECT\n    drive_id,\n    start_date_ts,\n    convert_km(start_km::numeric, '$length_unit') as start_$length_unit,\n    start_address,\n    end_date_ts,\n    convert_km(end_km::numeric, '$length_unit') as end_$length_unit,\n    end_address,\n    duration_min,\n    convert_km(distance::numeric, '$length_unit') AS distance_$length_unit\nFROM data;",
```
adriankumpf, line 323: I now understand the need to display only the street names etc. But there are some adresses without this data. Wouldn't it be useful to display other data as a fallback (e.g. `adress.name`)?

adriankumpf (changes requested):

grafana/dashboards/reports/dutch-tax.json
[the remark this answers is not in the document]
roadrash2108: No, I'm sorry - I tested the SQL directly in pgAdmin and deleted it, because it isn't set there...

I will add this of course!

roadrash2108: Oh, I didn't come across such a constellation - do you have an example? (I only had stops at addresses with only road+city, but I think this is fine as well)

adriankumpf: Giga Berlin would be an example. At least when I visited the OSM data looked like this:

```csv
display_name,latitude,longitude,name,house_number,road,neighbourhood,city,county,postcode,state,state_district,country,osm_id,osm_type
"Grünheide, Grünheide (Mark), Oder-Spree, Brandenburg, 15537, Deutschland",52.389753,13.791737,,,,Grünheide,Grünheide (Mark),Oder-Spree,15537,Brandenburg,,Deutschland,way
```

In such a case perhaps only the `display_name` could be used as a fallback?

Anyway, it's not a blocker. Let's merge this.

adriankumpf (approved):
Thanks!

PR #1059, opened 2020-10-28
Update charge-level.json

roadrash2108:
Set green bars for 20/80% in graph - to match style of graph "Charge Delta" in Charging Stats
(only a minor "cosmetical" change)


adriankumpf (approved):
Thanks!

PR #1061, opened 2020-10-29
Fix unit of measurement for charge added shown in overview, documentation to be correct

landler:
The grafana overview dashboard and a few places in the documentation were using a wrong unit of measurement for charge energy added. They were incorrectly using a unit of power (kW) to describe a change in energy (kWh). This PR should fix the issue.


adriankumpf (approved):
Good catch! Thank you.

Issue #1074, opened 2020-11-04
Monthly Statistics do not show the right numbers


JNB-Stravinsky:
**Describe the bug**

Statistics dashboard does not show the right numbers in Monthly view. The sum of the daily data of a month does not correspond to the total value shown.

![Captura de pantalla 2020-11-04 a las 14 09 38](https://user-images.githubusercontent.com/73945797/98115639-6e125900-1ea7-11eb-9366-6c4837e232f5.png)

![Captura de pantalla 2020-11-04 a las 14 10 59](https://user-images.githubusercontent.com/73945797/98115771-a154e800-1ea7-11eb-9b11-affb599e13a1.png)
![Captura de pantalla 2020-11-04 a las 14 11 17](https://user-images.githubusercontent.com/73945797/98115776-a3b74200-1ea7-11eb-91ba-82cef18b7599.png)

**Expected behavior**

The Statistics monthly view must show the sum of the daily data of each month.

**How to reproduce it (as minimally and precisely as possible)**:

1. Go to 'Statistics'
2. Click on 'Period -> Month'
3. Compare the values against 'Period -> day' and sum the data of the days of the month.

**Environment**

- TeslaMate version: 1.20.1
- Type of installation: Docker
- OS TeslaMate is installed on: MacOS 10.15.7
- User OS & Browser: MacOS 10.15.7 - Safari

DrMichael:
This looks indeed very strange. I checked some of my months, all are correct. Don't know, how to pin that  down without having the database at hand.

(I assume, you did not change the display time in the upper right or the timezone...)

JNB-Stravinsky:
Just noticed the error is related with the Timezone. My current Timezone is Europe/Madrid and the statistics are wrong:

![Captura de pantalla 2020-11-04 a las 16 23 49](https://user-images.githubusercontent.com/73945797/98130157-3e208100-1eba-11eb-9d69-61a728cd3e4c.png)

But if I change the Timezone to UTC, it seems to correct the statistics by adding +1 to each month:

![Captura de pantalla 2020-11-04 a las 16 25 26](https://user-images.githubusercontent.com/73945797/98130286-5db7a980-1eba-11eb-9db6-76e7f24d56c3.png)

DrMichael:
How did you set the time zone to Europe/Madrid?

JNB-Stravinsky:
I didn't set it or I don't remember setting it. It is in Europe/Madrid by default. I think Grafana takes it from the server:

[...]

PR #1082, opened 2020-11-12
Resolve issue with month groupings on statistics dashboard

leewillis77:
This is an attempt at fixing #1074 

Reading the Grafana queries it looks (to me, but I'm not a grafana "expert" or Postgres lover) that there is an extraneous `AT TIME ZONE` conversion going on. Editing the queries in Grafana to remove the extraneous `AT TIME ZONE...`) seems to work for me, resolving the missing months in my statistics, but not sure if there is a knock-on issue with other periods or indeed if this change causes time zones not to be accounted for properly down at the detail level.


adriankumpf (approved):
I have not been able to reproduce the error so far. With your changes, no difference is visible with my time zone (which is good, I guess?). Since the problem seems to be fixed for you, I'll merge this. Let's see what others will report.

PR #1097, opened 2020-11-25
Adjusting documentation info about TeslaMateAgile text to mention Tibber support

tobiasehlert:
TeslaMateAgile previously only supported calculation of cost for Octopus Agile supplier.
Now the support for Tibber has been added as well, thanks to @MattJeanes!


MattJeanes:
You should probably rebase this branch onto master, loads of commits here you probably don't want to include!

adriankumpf (approved):
Cool!

Issue #1060, opened 2020-10-29
[cold weather] charge port latch behaviour change -> MQTT: plugged_in = false


LelandSindt:
**Describe the bug**

It appears Tesla has changed the charge port latch behaviour [?in cold weather?]. Historically, after plugging in the charge cable the charge port latch would engage and stay engaged until the power cable button is pressed. Recently [in cold weather] the charge port latch will engage after the charge cable is plugged in, and then disengage. 

This is causing teslamate to set `plugged_in = false`

https://github.com/adriankumpf/teslamate/blob/master/lib/teslamate/vehicles/vehicle/summary.ex#L131

**Preferred [Expected] behavior**

Teslamate use `charging_state != Disconnected` instead of `charge_port_latch == Engaged` to determine if the car is plugged in.

**How to reproduce it (as minimally and precisely as possible)**:

1. Be in cold climate.
2. Plug in charge cable.
3. Observe MQTT: plugged_in = false

LelandSindt:
Additional observation....
In the last hour `charge_port_cold_weather_mode` flipped from `true` to `false` and `charge_port_latch` changed to `Engaged`.

bpankz:
I believe the threshold is 4C / 40F when the lock remains disengaged so it doesn't freeze

LelandSindt:
closed by 5ed9131e1fb572f97fd4c097d894e92f63f2a3b2

PR #1098, opened 2020-11-26
set plugged_in == true if charging_state != 'Disconnected'

LelandSindt:
I don't know that this is an acceptable solution, but I figured it was worth a try. :smiley: 

addresses #1060

### DIFF
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -133,6 +133,12 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
     true
   end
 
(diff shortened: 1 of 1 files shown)


adriankumpf:
I'm sure this works but it does feel a little hacky 😄 Maybe we could use this approach if `charge_port_cold_weather_mode` is `true`, as you have described here https://github.com/adriankumpf/teslamate/issues/1060#issuecomment-718323754. I've tried this in 5ed9131e1fb572f97fd4c097d894e92f63f2a3b2

LelandSindt:
Works for me... That said the more I think about, testing for `charging_state != 'Disconnected'` could cause a false positive if `charge_state = 'some failed state we are not aware of' `

I went looking for a list of valid/expected values for charging_state but I have not found anything that looks definitive. :disappointed:

PR #1113, opened 2020-12-09
Average power and speed in the drive-details over 10 last data entries

mluiten:
For your consideration.

This will smoothen the graph and take out the insignificant spikes. This looks much more like Tesla's own power graph, and provides me better insight and less "noise", personally. Curious to see what you think... I'm not sure if 10 is "smooth" enough for longer trips -- maybe there is a way to have a bigger moving average set for longer trips for example? But this is a start and you won't lose much detail averaging over 10 data points.

Before:

Before

After:

Screenshot 2020-12-09 at 12 05 01

And here's the new query:

Screenshot 2020-12-09 at 12 02 45

### DIFF
--- a/grafana/dashboards/internal/drive-details.json
+++ b/grafana/dashboards/internal/drive-details.json
@@ -16,7 +16,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 1,
-  "iteration": 1601722850623,
+  "iteration": 1607507903288,
   "links": [
     {
       "icon": "dashboard",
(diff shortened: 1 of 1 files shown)


adriankumpf:
This is an interesting idea! 

I noticed that short trips with few data points (because streaming was not possible), for instance, are displayed in a misleading way:

#### before
Screenshot 2020-12-29 at 19 14 13

#### after
Screenshot 2020-12-29 at 19 14 19

The overall smoothing effect tends to be less noticeable for longer trips. And in general, the non-smoothed graphs don't bother me too much. 

Therefore I think that this could be an optional setting which can be activated if desired. Wdyt?

pkuehnel:
I moved to TeslaMate because of all the details you can see in the graphs. I would not be happy with the 'smoothing'.

adriankumpf:
I'm going to close this for inactivity. If desired, we can open the PR again.

PR #1154, opened 2020-12-30
Added state nil for charge port cold weather mode

virtualm2000:
I have made some tests with my Model X, 2017 model. It doesn't have the charge_port_cold_weather_mode state. So the state nil was not handled to detect plugged in.


adriankumpf (approved):
Thank you!

PR #1169, opened 2021-01-03
Fix colors in Updates dashboard

fatbasstard:
The panels in the Update dashboards are configured with a white font, which makes it invisible in the "light" theme of Grafana.

This PR changes the color to the same as used in e.g. the Overview dahboard:

OLD vs NEW schrrenshot: 
image


adriankumpf (approved):
Thanks!

PR #1170, opened 2021-01-03
Add Docker update documentation

fatbasstard:
This PR adds some minor documentation about updating TeslaMate when installed with Docker.

This to help out with updating it when people are not too familiar with Docker.

### DIFF
--- a/website/docs/installation/docker.md
+++ b/website/docs/installation/docker.md
@@ -82,9 +82,17 @@ This setup is recommended only if you are running TeslaMate **on your home netwo
    docker-compose up -d
    ```
 
+### Update
+
+To update the running Teslamate condiguration to the latest version, run the following commands:
+
+```bash
+docker-compose pull
+docker-compose up -d
+```
+
 ## Usage
 
 1. Open the web interface [http://your-ip-address:4000](http://localhost:4000)
 2. Sign in with your Tesla account
 3. The Grafana dashboards are available at [http://your-ip-address:3000](http://localhost:3000).
-


adriankumpf (approved):
Thank you 👍

cwanja:
Consider compressing it into a single line? 
`
docker-compose down && docker system prune && docker-compose pull && docker-compose up -d
`
Not sure the prune is necessary, but I include it.

cwanja:
Might want to also edit this one. 
https://docs.teslamate.org/docs/upgrading

DrJohnM61:
> Consider compressing it into a single command? 
> 
> `
> 
> docker-compose down && docker system prune && docker-compose pull && docker-compose up -d
> 
> `
> 
> Not sure the prune is necessary, but I include it.

The prune should logically go after the pull as it will clean up and redundant images created by pulling new (updated) images

cwanja:
Excellent, thanks for that feedback. Will update my own script. 

Sent from my iPhone
Chris Wanja
chris.wanja@mac.com
713.444.6999

binaryPUNCH:
Isn't the prune risky to run in between docker-compose down and docker-compose up -d? Won't it delete your data? I ran prune as a final step instead.

cwanja:
> Isn't the prune risky to run inbetween `docker-compose down` and `docker-compose up -d`? Won't it delete your data?
> 
> Use this instead: `docker-compose down && docker-compose pull && docker-compose up -d && docker system prune`

Never had a data loss issue with the `docker system prune` erasing data. But I am open to more suggestions and advice. Am not a Docker expect by any stretch.

roadrash2108:
It's already merged, but I wonder why the information is duplicated as it is in https://github.com/adriankumpf/teslamate/blob/master/website/docs/installation/docker.md and https://github.com/adriankumpf/teslamate/blob/master/website/docs/upgrading.mdx !?

IMHO this should be referenced in the installation-documentation, but not duplicated...

PR #1185, opened 2021-01-08
Fix typo in the link

rogiervandergeer:
I noticed this typo in the link to TeslaMate; so here is a fix!


adriankumpf (approved):
Good catch :) Thanks!

PR #1188, opened 2021-01-09
Add link to Home Assistant addon

matt-FFFFFF:
Would you be good enough to include a link to my HA addon? Now that it has Ingress support it's in a position to be shared more widely.


adriankumpf (approved):
Absolutely. That is awesome!

PR #1191, opened 2021-01-09
Update drive duration query

fatbasstard:
Fixes: https://github.com/adriankumpf/teslamate/issues/1189

Simplified the query to use the `duration_min` from the drive instead of calculating it


adriankumpf (approved):
Thanks 👍

PR #1213, opened 2021-01-16
Reenable temperature in charges dashboard

DrMichael:
@adriankumpf sorry, I have zero idea, why github shows all lines as different. For me it is neither in notepad++ nor with Unix diff...


adriankumpf:
It's because of the DOS-style line endings. I cleaned up the file. Will take a look later.

DrMichael:
> It's because of the DOS-style line endings. I cleaned up the file.

Thanks. Didn't find in notepad++ how to change it...

> Will take a look later.

Ooooh, you removed the nice colours of the temperatures?:-)

And, you're right, it should not have been kw/h, but rather kwh/h, which is perhaps clearer than kw. It tells, how much energy is used per hour.

DrMichael:
:-)

adriankumpf:
> Ooooh, you removed the nice colours of the temperatures?:-)

Oops :-) I thought these were added by accident. Sorry! Re-added them.

Regarding `kWh/h` vs `kW`: I prefer the latter. But maybe it's just me :D

adriankumpf (approved):
Looks good to me! Thanks.

PR #1214, opened 2021-01-16
Update overview.json

SergioRius:
Temperature stats in Overview Dashboard panel often shows no data on installations with different time zones than UTC. The NOW statement induces a comparison between UTC and ZULU timestamps. The affected queries don't need a timestamp specification as they only get the last register. Although could be unnecessary, a restriction using the dashboard interval is added as a means of limiting the records in the search, in case it impacts in DBs with bigs amounts of data.

See: https://github.com/adriankumpf/teslamate/issues/1210#issuecomment-761615400

### DIFF
--- a/grafana/dashboards/overview.json
+++ b/grafana/dashboards/overview.json
@@ -1255,7 +1255,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


adriankumpf:
Closing this in favour of https://github.com/adriankumpf/teslamate/pull/1213. Nonetheless, thank for the PR!

PR #1215, opened 2021-01-16
Update charge-details.json

SergioRius:
The graph shows registers just some seconds past the end, that is supposedly the time the Api delays the notification of the end of the charging session. That makes that the min range values in the legend for voltage, current and power shows zero and become useless. By adding a zero restriction to power or current, we avoid those zero registers and better restrict the charging session times. As a side effect, the right y-axis can show a better interval and the graph has more detail.


adriankumpf (approved):
Thank you!

PR #1254, opened 2021-01-26
Show "Starting at" and fix timezone issue.

### LINKED ISSUE #1040: Order in Statistics dashboard wrong
**question**

@adriankumpf did you hide "Starting at" in Statistics on purpose? The order is so a bit wierd:
image

**Expected behavior**
image

Apparently Grafana does not order on hidden fields?!?!

**Environment**
- TeslaMate version: 1.21.0-dev
- OS TeslaMate is installed on: Ubuntu 20.04
- User OS & Browser: Chrome
- Others: -

DrMichael:
Needs to add PGTZ for the home timezone for the grafana container.

Fixes #1040


adriankumpf:
> Needs to add PGTZ for the home timezone for the grafana container.

Is this a requirement now? What happens if you don't set it?

DrMichael:
No, not really. Without it, the periods start at midnight UTC. When it is set, it starts at midnight in the given timezone.

This is all pretty tricky here anyway: when e.g. the timezone of the database is set to something else than UTC, the period would start midnight in that timezone (can be overridden via PGTZ and and and...)

So, having "Starting at"at least makes clear, when the period starts.

adriankumpf:
Indeed, pretty tricky :) Thanks for the explanation.

adriankumpf (approved):

PR #1273, opened 2021-01-31
Call out Debian update more obviously

RickyRomero:
My healthcheck script broke when updating TeslaMate (as `wget` is no longer available). This broke access through Traefik. I'm adding a note here for others who encountered this issue.


adriankumpf (approved):
Thanks!

PR #1289, opened 2021-02-04
Adjust log level for debug streaming messages

brianmay:
I think these messages are important to show, because it confirms the stream connection was successfully re-established.

### DIFF
--- a/lib/tesla_api/stream.ex
+++ b/lib/tesla_api/stream.ex
@@ -54,14 +54,14 @@ defmodule TeslaApi.Stream do
 
   @impl true
   def handle_connect(_conn, state) do
-    Logger.debug("Connection established")
+    Logger.info("Connection established")
     send(self(), :subscribe)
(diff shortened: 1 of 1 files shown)


adriankumpf:
This will lead to even more log spam though. I am not sure if this is a good idea. Maybe we should rather go in the opposite direction and minimise the streaming logs?

brianmay:
Looking at the discussion reports, it appears I am not the only one encountering issues with streaming not working exactly as it should work. So it would make sense to increase logging here. Or at least make it run-time configurable.

adriankumpf:
Yeah, that's right. On the other hand, people regularly ask whether all those messages are problematic or what they mean. Making the log level run-time configurable would be nice!

adriankumpf:
I'll close this here, since the debugs logs are not purged anymore from the release. The log level can be changed at runtime via `Logger.configure(level: :debug)`.

PR #1321, opened 2021-02-07
HomeAssistant documentation update

jschollenberger:
I'm adding TeslaMate into my HomeAssistant instance and realized the example doesn't have all the topics that MQTT publishes. 

This PR updates the sensors.yaml example to:
- add the missing MQTT topics as sensors
- normalizes sensor names to the MQTT topic documentation (charge_limit vs charge_limit_soc)
- add templates which provide imperial sensors
- add / update material design icons
- add missing units of measure

I'd like to make some contributions to the lovelace dash in the future (and a new screenshot), but for now I just made a few edits:
- use the stock button card instead of a custom button card (doesn't exist so it throws an error)
- add dark mode and a zoom level to the map for aesthetics
- add all of the sensors available so the user can easily delete the ones they don't want


adriankumpf (approved):
Great work. Thank you!

PR #1350, opened 2021-02-15
Updating docs with following..

tobiasehlert:
- adding [TeslaMateApi](https://github.com/tobiasehlert/TeslaMateApi) project
- adding index to document page top
- reordering projects in alphabetical order

### DIFF
--- a/website/docs/projects.md
+++ b/website/docs/projects.md
@@ -2,11 +2,6 @@
 title: Projects using TeslaMate
 ---
 
-## [MMM-Teslamate](https://github.com/denverquane/MMM-Teslamate)
-
-A [Magic Mirror](https://magicmirror.builders/) Module for TeslaMate.
-
-LINK: [github.com/denverquane/MMM-Teslamate](https://github.com/denverquane/MMM-Teslamate)
 
 ## [Gaussmeter](https://github.com/gaussmeter/gaussmeter)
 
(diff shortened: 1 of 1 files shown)


adriankumpf (approved):
Thanks Tobias. Great work on TeslaMateApi! It would be nice to have the same functionality directly in TeslaMate at some point. I'll keep an eye on it :-)

tobiasehlert:
> Thanks Tobias. Great work on TeslaMateApi! It would be nice to have the same functionality directly in TeslaMate at some point. I'll keep an eye on it :-)

Thanks @adriankumpf 

My initial thought was that there are only a few who want to interact with the data collected my TeslaMate so a separate container would be better, so you can add the API functionality if interested. But I can also see it as a good point to have it included in TeslaMate directly and let the web GUI talk to a backend instead as doing queries itself.

Anyhow.. I'm not skilled in Elixir to be able to build this and Golang was the way forward for me. Your time is maybe better to resolve other requests that have a broader interest =D

I guess some queries could be enhanced as well and data formatted a bit different, but I will see what responses will come when people start using the container.

adriankumpf:
Yeah, agreed. There have been requests for an API from time to time though. A GraphQL API could also be interesting. 

You also correctly point out that authentication should be considered, so that the data is not unprotected. This is of course also an important point, since some installations are – against all warnings – exposed directly to the internet without any authentication. Therefore, it might make sense to disable such an API by default if it were built directly into TeslaMate.

I think the approach is perfectly fine to write this in your preferred programming language and have it as an add-on. I'm also curious to see what feature requests will come. In any case, it's great that the option to have an API now exists.

PR #1355, opened 2021-02-16
Set bucket size on charging heatmap. Fixes #1354

### LINKED ISSUE #1354: Charge heatmap can be confusing due to auto bucket sizing.
**Describe the bug**

The Charge Heatmap (on the Charging Stats dashboard) shows charges > 100% even if no such data exists in the charging_processes table. 

At first I assumed this was just down to the bucket size (which is set to auto - might be better set to bucket size 10?), but changing that doesn't make any difference. 

See screenshot, and data extract below:
Screenshot 2021-02-16 at 17 43 11
Screenshot 2021-02-16 at 17 46 38

**Expected behavior**

Charge heatmap doesn't show charges > 100 unless the car actually manages to charge that far :)

**Environment**

- TeslaMate version: 1.21.2 
- OS TeslaMate is installed on: RaspberryPi
- User OS & Browser: Chrome / MacOSX

leewillis77:
Sets bucket size to 10.00001 to make buckets more intuitive (subjective opinion).

### DIFF
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -518,7 +518,7 @@
       },
       "yBucketBound": "auto",
       "yBucketNumber": null,
-      "yBucketSize": null
+      "yBucketSize": 10.00001
     },
     {
       "aliasColors": {},


leewillis77:
Before:
annie-before

After:
annie-after

adriankumpf (approved):
Good idea. Thank you!

PR #1363, opened 2021-02-18
Fix: "Gross" Panel under Overview looks smaller than others

DrMichael:
Fixes https://github.com/adriankumpf/teslamate/issues/1362


adriankumpf (approved):
Thanks @ DrMichael!

PR #1422, opened 2021-03-09
Add health check

brianmay:
See https://blog.jola.dev/health-checks-for-plug-and-phoenix


brianmay:
I am using `/` for this purpose, but it occurred to me that regularly polling a live view may not be such a good idea.

adriankumpf (approved):
Thanks @brianmay

PR #1428, opened 2021-03-11
Add battery heater info into charging dashboards

ToniA:
Add battery heater info into dashboards Overview / Charging Details and Charge Details

image

### DIFF
--- a/grafana/dashboards/internal/charge-details.json
+++ b/grafana/dashboards/internal/charge-details.json
@@ -109,7 +109,7 @@
           "hide": false,
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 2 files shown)


ToniA:
Battery heater is only shown if it was active during the charging process. This is because of the `"hideZero": true` setting on the dashboard. This could of course also be set to false to always show series which only have zero values. Just in case if you were wondering why it is not shown, it seems to activate only below ~0C degrees.

adriankumpf (approved):
Looks good. Thank you @ToniA!

timguy99:
I'm not seeing this or actually 3 things on my dashboard. Is there something that needs to be done to update dashboards. I think I notice them changing from time to time during updates but I'm only showing 6 out of the 9 things on this dashboard. I am running v1.22

PR #1439, opened 2021-03-12
Make Statistics dashboard look the same on both kilometers and miles

ToniA:
The order of columns 'Distance' and 'Consumption' is not the same if distance unit is changed to miles. Also, in kilometers view the column is 'Avg consumption' while in miles view the column is 'Consumption'.

Old:
image

New:
image

### DIFF
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -333,11 +333,14 @@
             "properties": [
               {
                 "id": "displayName",
-                "value": "Consumption"
+                "value": "Avg consumption"
               },
               {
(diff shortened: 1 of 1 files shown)


DrMichael:
Did you try to switch between km and mi? I think to remember, that there is a problem with the new table in Grafana, which causes the unused column to be moved to the end...

ToniA:
Yes I did, the problem is that the order of fields sum_distance_mi and efficiency_net_mi was not defined.

adriankumpf (approved):
LGTM!

PR #1445, opened 2021-03-15
Calculate efficiency from charged energy

ToniA:
Add a second efficiency column into 'Statistics' dashboard:
* Column 'Avg Consumption' renamed to 'Avg Consumption (drives)'. This is the average consumption during drives, i.e. the figure which tells how far you could typically drive with certain amount of energy in the battery
* Added column ' Avg Consumption (charges)'. This is the average consumption, calculated the same way as you would calculate the consumption of an ICE car. I.e. the amount of electricity you need to pay for versus the distance you have driven. Especially during winter the difference to 'Avg Consumption (drives)' could be huge. This will include vampire drain and preheating (at least when preheating without being plugged in)

The driven distance is now calculated as a difference between the smallest start_km and the highest end_km. This is a higher number than the sum of distances from drives _if_ you are missing any drives data.

Also notice that if your car is plugged in during preconditioning, I think the energy taken from the grid during that time is not recorded.

Works also when the units are in miles.

image

### DIFF
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -333,7 +333,26 @@
             "properties": [
               {
                 "id": "displayName",
-                "value": "Avg consumption"
+                "value": "Avg consumption (drives)"
+              },
+              {
+                "id": "unit",
+                "value": "Wh/mi"
+              },
+              {
+                "id": "custom.width"
+              }
+            ]
+          },
+          {
+            "matcher": {
+              "id": "byRegexp",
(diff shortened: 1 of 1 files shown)


adriankumpf:
Nice! 🙌 Depending on how much you drive, the `Avg Consumption (charges)` could vary greatly if you have an empty battery at the beginning of the period and a nearly full one at the end (or vice versa), right? But it is definitely useful for an approximate assessment.

adriankumpf (approved):

PR #1453, opened 2021-03-16
Explain the asleep mode with MCU1 and the non-streaming mode

ToniA:


adriankumpf (approved):
Great explanation. Thanks a lot!

PR #1494, opened 2021-03-30
Update the priorities of city aliases

### LINKED ISSUE #1030: City aliases are not precise enough
City aliases are not precise enough since #489 is resolved.

When i drive to a small town / village, Teslamate seems to use the municipality as city alias.

Some exemples:
https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=122005&class=boundary
is aliased as 
https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=1686640&class=boundary

https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=120585&class=boundary
is aliased as
https://nominatim.openstreetmap.org/ui/details.html?osmtype=R&osmid=1693119&class=boundary

The #489 reporter was right: hamlet is not a good alias for city. But i think it would be better to remove it from city alias (as locality and croft). Perhaps it could be used as last resort for neighbourhood_aliases ?
But according to https://wiki.openstreetmap.org/wiki/Key:place, the city alias order seems to be in wrong order. I think village / town / city / municipality order would be more precise.

If this problem is fixed, is there a way to force Teslamate to update all addresses with the new algorithm ?

shanbs:
To fix #1030, by moving the priority of "village" higher than "municipality".

### DIFF
--- a/lib/teslamate/locations/geocoder.ex
+++ b/lib/teslamate/locations/geocoder.ex
@@ -102,8 +102,8 @@ defmodule TeslaMate.Locations.Geocoder do
   @city_aliases [
     "city",
     "town",
-    "municipality",
     "village",
+    "municipality",
     "hamlet",
     "locality",
     "croft"


adriankumpf:
Thanks for the investigation of the issue and the fix. I've merge it as #1589.

PR #1496, opened 2021-03-31
Bump traefik to v2.4

tobiasehlert:
We could also rely on **_latest_** tag instead for traefik.. what you think @adriankumpf? :)

### DIFF
--- a/website/docs/guides/traefik.md
+++ b/website/docs/guides/traefik.md
@@ -108,7 +108,7 @@ services:
       - mosquitto-data:/mosquitto/data
 
   proxy:
-    image: traefik:v2.3
+    image: traefik:v2.4
     restart: always
     command:
       - "--global.sendAnonymousUsage=false"


tobiasehlert:
And even if not related to traefik version.. but **eclipse-mosquitto** version **1.6** is currently used but version **2** is out.

Maybe something to bump as well @adriankumpf?

adriankumpf:
> We could also rely on latest tag instead for traefik.. what you think @adriankumpf? :)

v3 will, at some point ,certainly introduce breaking changes. Let's stay on v2 :)

adriankumpf:
> And even if not related to traefik version.. but **eclipse-mosquitto** version **1.6** is currently used but version **2** is out.
> 
> Maybe something to bump as well @adriankumpf?

Last time I tried v2 it required custom changes via a config file, which would complicate the installation. Maybe it's easier now?

tobiasehlert:
> Last time I tried v2 it required custom changes via a config file, which would complicate the installation. Maybe it's easier now?

I have no clue.. I just noticed that there was a new release coming out around 4 month ago, but I didn't check more into what needs to be done to get it to work :)

PR #1504, opened 2021-04-02
add drive state battery power to published mqtt topics

mnadvornik:
I added the current battery power output to the vehicle summary definition in order for it to be published via MQTT.
This data is needed in my use case to send data to the ABRP (https://abetterrouteplanner.com/) API since battery power is a mandatory field for the ABRP API.

I hope this is useful for others and would welcome if this could be merged


adriankumpf (approved):
Nice work! Thanks.

kimmilde:
Great update! Is it possible to increase the update rate published to mqtt, so it would be the same as vehicle speed?

thunder2k:
> I added the current battery power output to the vehicle summary definition in order for it to be published via MQTT.
> This data is needed in my use case to send data to the ABRP (https://abetterrouteplanner.com/) API since battery power is a mandatory field for the ABRP API.
> 
> I hope this is useful for others and would welcome if this could be merged

Hey,
could you please elaborate on how are you using it? Are you getting real-time data from the car to ABRP to tweak your route in real-time? If you could point me in the right direction it would be great!
Thanks!

PR #1470, opened 2021-03-22
Update statistics.json

Dulanic:
Changed timezone as posix is not available for all. Added regex to restrict to abbrev that are 3 char (normal timezones) and removed posix as not everyone has it. Fix for  https://github.com/adriankumpf/teslamate/issues/1467

PR #1559, opened 2021-04-20
Complete my prior fix that didn't fix statistics.json

Dulanic:
Shaking my head in shame, I changed the def, not the SQL :(


adriankumpf:
Oops, I should have noticed that as a reviewer. Thanks for the fix!

xnaas:
@adriankumpf: This should be included in the notes for https://github.com/adriankumpf/teslamate/releases/tag/v1.23.0 as this ended up being the actual fix, not just a "display" fix like #1470 was.

PR #1562, opened 2021-04-23
Standardise date formats + other fixes

Ed-M72:
Added Date format (American/European) to settings screen and standardised date formats on dashboards that show them.  This gets around the fact hat some of the dashboards were using the distance unit to set the date format.

There are two exceptions to this.

 Dashboard "States" - Not possible to calculate a format on the fly due to a restriction on the Stats panel in Grafana.
Dashboard "Statistics" - New variable for date_format created, however the SQL needs reviewing by somebody who understands it. Aka, not me.
Dashboard "Dutch Tax" - This uses the local setting so will always render the date in the users local format.

Once the Grafana plugins have updated  fully by the Grafana team to the new standard it should be possible to use date time local for all dashboards.  Based on the progress of the Grafana team this could be a while away.

Added a fix for issue 1064 - cost fields too short.


.gitignore
```diff
@@ -54,3 +54,10 @@ config/*.env
 .eclipse/
 .elixir_ls/
 /.project
+assets/package-lock.json
```
adriankumpf, line 57: Could you revert these changes? I don't think they're necessary.

grafana/Dockerfile
```diff
@@ -1,3 +1,4 @@
+
```
adriankumpf, line 1: We don't really need this new line :-)

grafana/dashboards/vampire-drain.json
```diff
@@ -50,8 +50,8 @@
       "repeat": "car_id",
       "scopedVars": {
         "car_id": {
-          "selected": true,
-          "text": "1",
+          "selected": false,
+          "text": "Terry",
```
adriankumpf, line 54: It happens very easily that the default values are replaced by the own values. Can you revert this please?

grafana/dashboards/vampire-drain.json
```diff
@@ -547,11 +568,12 @@
         "allValue": null,
         "current": {
           "selected": false,
-          "text": "http://localhost:4000",
-          "value": "http://localhost:4000"
+          "text": "http://jupiter.dalestech.net:4000",
```
adriankumpf, line 571: Same here.

adriankumpf (commented):
Thanks for you work! I only looked over it briefly and added a few comments. Will take a closer look soon.

.gitignore
Ed-M72: Done. Learning as I go :)

grafana/dashboards/vampire-drain.json
Ed-M72: Yep, updated with the correct values.

grafana/dashboards/vampire-drain.json
Ed-M72: Yep, fixed here too

grafana/Dockerfile
Ed-M72: Aww, I liked the line. Even though it was by accident it was one of my favourites. Removed :-)

priv/repo/migrations/20210429220710_charges_increase_cost_size.exs
```diff
@@ -0,0 +1,15 @@
+defmodule TeslaMate.Repo.Migrations.ChargesIncreaseCostSize do
+  use Ecto.Migration
+
+  def change do
+    alter table(:charging_processes) do
+      modify(:cost, :decimal, precision: 8 , scale: 2, from: :decimal, precision: 6 , scale: 2)
+    end
+  end
+
+  def down do
```
adriankumpf, line 10: This is not necessary if you already defined `change`. You can either define `up` and `down` functions separately or write a [change callback](https://hexdocs.pm/ecto_sql/Ecto.Migration.html?#module-change).

priv/repo/migrations/20210429164504_geofences_increase_cost_size.exs
```diff
@@ -0,0 +1,17 @@
+defmodule TeslaMate.Repo.Migrations.GeofencesIncreaseCostSize do
+  use Ecto.Migration
+
+  def change do
+    alter table(:geofences) do
+      modify(:cost_per_unit, :decimal, precision: 7 , scale: 4, from: :decimal, precision: 6 , scale: 4)
+			modify(:session_fee, :decimal, precision: 7 , scale: 2, from: :decimal, precision: 6 , scale: 2)
```
adriankumpf, line 7: Formatting looks off here :)

lib/teslamate_web/live/settings_live/index.html.leex
```diff
@@ -228,6 +228,27 @@ end %>
...
+      <h2 class="title is-4"><%= gettext("Dates") %></h2>
+      <div class="field is-horizontal">
+        <div class="field-label is-normal">
+          <%= label f, :date_format, gettext("Date Format"), class: "label" %>
+        </div>
+        <div class="field-body">
+          <div class="field">
+            <div class="control">
+              <div class="select is-fullwidth">
+                <%= select(f, :date_format, [:European, :American]) %>
```
adriankumpf, line 242: Since date formats vary [between countries](https://en.wikipedia.org/wiki/Date_format_by_country), I wonder if these dropdown options should be labeled differently. For example `YYYY-MM-DD` and `MM/DD/YY`.

adriankumpf: It would also make sense to decouple the option value from the actual text, and to make it translatable:

```suggestion
                <%= select(f, :date_format, [{gettext("European"), :European}, {gettext("American"), :American}]) %>
```

lib/teslamate_web/live/charge_live/cost.html.leex
```diff
@@ -115,7 +115,7 @@
         </p>
         <p class="control is-expanded">
           <%= text_input f, :cost, class: "input",
-                                   type: :number, inputmode: :decimal, step: 0.01,
+                                   type: :number, inputmode: :decimal, step: 0.0001,
```
adriankumpf, line 118: The `step` does not match the `scale` of 2 defined in the migration below.

lib/teslamate/settings/global_settings.ex
```diff
@@ -6,6 +6,8 @@ defmodule TeslaMate.Settings.GlobalSettings do
     field :unit_of_length, Ecto.Enum, values: [:km, :mi]
     field :unit_of_temperature, Ecto.Enum, values: [:C, :F]
 
+  field :date_format, Ecto.Enum, values: [:European, :American]
```
adriankumpf, line 9: Formatting 🚔 (you can format all code via `mix format`)

adriankumpf (changes requested):

lib/teslamate_web/live/settings_live/index.html.leex
Ed-M72: Yep good idea on the text translations.  For the date formats, that is where I started, however the / & - don't play well with Grafana.  So I could make them YYYYMMDD & MMDDYYYY.  I tried that and eventually settled on European/American.  Am happy to go back to YYYYMMDD & MMDDYYYY if you think that is the correct thing to do?

adriankumpf: Got it. I mean, it's just a small detail. We do not need to rename the internal Postgres data type. Changing only the labels on the settings page would be fine. Wdyt?

albanfellows: While I am glad to see some efforts are being made to make the date more flexible, I think the date format issue is a bit more complicated. 
As a UK user the American format drives me nuts. We use miles and DD/MM/YYYY as the standard formats. Neither the "American" nor the "European" (which is really ISO) formats are really any good. 

Ideally users should be able to specify their own date format or it should pick it up from the browser locale.
There may be people who want words e.g. 13-May-2021 (DD-MMM-YYYY)

Or you could have a selection of common standard ones:
- American: MM/DD/YYYY
- British: DD/MM/YYYY
- European or ISO: YY-MM-DD

Ed-M72: Yes you are so right. I am also a UK user, thats why i started the change, the dates annoyed me enormously. I have to say that learning about the ins and outs of  Grafana was probably the biggest disappointment I had.  The problem is that Grafana does not support the UK date format across any of its plugins or panels.  Ideally you would want to use browser locale, however this is only supported on one of the dashboards we have.

Grafana is so inconsistent with its product that its ridiculous.  I think there will be a time when it will catch up, but I can't see that being for at least a couple of years.  The only consistent formats available are DateTime_ISO and DateTime_US.

The alternative would be to replace Grafana with a product which is more feature complete, but that is a whole heap of work.

priv/repo/migrations/20210429164504_geofences_increase_cost_size.exs
Ed-M72: Fixed

lib/teslamate_web/live/settings_live/index.html.leex
Ed-M72: I quite liked the idea of the texts,

- American: MM/DD/YYYY
- European or ISO: YY-MM-DD

I decided to update the date_formats type for consistency, whch now contains "ISO" & "US".

Dashboards had to be updated as a result.

lib/teslamate_web/live/charge_live/cost.html.leex
Ed-M72: So the problem here is the HTML field is used as calculation field and also used to store the result.

The calculation requires four decimal places to satisfy the requirement of bug 1540, wheras the result uses 2 decimal places.  For the time being I am going to increase the size of the cost field in the charging_processes table to compensate.

lib/teslamate_web/live/settings_live/index.html.leex
adriankumpf: >  Ideally you would want to use browser locale, however this is only supported on one of the dashboards we have.

The `Updates` dashboard uses the "new" table component. The only reason we haven't upgraded the other dashboards is this issue I created a year ago: https://github.com/grafana/grafana/issues/24092

I don't want to get my hopes up, but currently it looks like it will finally end up in the next beta of Grafana v8. So maybe we should wait for v8.0.0-beta2 before merging this PR.

PR #1571, opened 2021-04-27
typo in code_challenge_method fixed

tuxbox:
the code_challenge_method should be S256 I believe.
Seems to be a typo without any effect as of now


adriankumpf (approved):
Good catch! Thank you.

PR #1576, opened 2021-04-28
postgres to alpine and runs on to latest ubuntu

tobiasehlert:
- appending _-alpine_ to postgres in various doc-files and [.github/workflows/elixir.yml](https://github.com/adriankumpf/teslamate/blob/master/.github/workflows/elixir.yml) (ref #1569)
- switching to `ubuntu-latest` instead of `ubuntu-18.04` for [.github/workflows/elixir.yml](https://github.com/adriankumpf/teslamate/blob/master/.github/workflows/elixir.yml)

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -16,7 +16,7 @@ jobs:
 
     services:
       db:
-        image: postgres:13
+        image: postgres:13-alpine
         ports: ["5432:5432"]
         env:
           POSTGRES_PASSWORD: postgres
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -12,7 +12,7 @@ jobs:
           - elixir: "1.11.1"
             otp: "23.x"
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     services:
       db:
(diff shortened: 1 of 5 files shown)


adriankumpf:
I've reverted the commit that bumps the ubuntu image to `latest`. It does not seem to be compatible with the `setup-elixir` action and I'd prefer to use `20.04` instead of `latest`. Will change this in a separate commit.

adriankumpf (approved):
👍

tobiasehlert:
@adriankumpf looks like the _ubuntu-latest_ was reverted to _ubuntu-18.04_ but the _postgres:13-alpine_ part for postgres was not applied.
Want me to create a separate PR for that? :)

adriankumpf:
Hi @tobiasehlert, unfortunately the alpine images don't work on the Raspberry Pi because its package
sources contains an old version of libseccomp, which causes the time to be at 1970. I had to revert it in 6cfb2a24cf2a04d75f941f35c3950d9055cad919.

tobiasehlert:
> Hi @tobiasehlert, unfortunately the alpine images don't work on the Raspberry Pi because its package
> sources contains an old version of libseccomp, which causes the time to be at 1970. I had to revert it in [6cfb2a2](https://github.com/adriankumpf/teslamate/commit/6cfb2a24cf2a04d75f941f35c3950d9055cad919).

Ah alright! I thought it was only the builder part that broke things.. but oke :)

Issue #1516, opened 2021-04-05
Supercharger charges - showing even when free


Ed-M72:
**Describe the bug**

If I click "Free Supercharging" within the settings screen, the charge row within the charging_processes table always has a value assigned to it after charging with a Tesla Supercharger.

The code that derived the cost looks like this,  I assume the fast_charger_type is pulled in over the API, however if I look in the charging processes table both fast_charger_brand and fast_charger_type are always saved as <invalid>

cost =
      case {stats, charging_process} do
        {%{fast_charger_type: "Tesla" <> _},
         %CP{car: %Car{settings: %CarSettings{free_supercharging: true}}}} ->
          0.0

Not sure why this is happening, perhaps it is not getting passed through from the API.  I am looking to add a provider to the geofences table, so if there is no workaround for this it could revert to checking against the geofence.

**Expected behaviour**

The charge should be saved with a zero cost if charging at a Tesla Supercharger and "Free Supercharging" is set to true in the settings screen.

**How to reproduce it (as minimally and precisely as possible)**:

**Relevant entries from the logs**

```
paste logs here
```

**Screenshots**

**Data**

**Environment**

- TeslaMate version: 1.22.0
- Type of installation: Docker or Manual: Docker
- OS TeslaMate is installed on: Synology NAS
- User OS & Browser: Mac OS, Safari
- Others:

adriankumpf:
What values are stored for the `fast_charger_type` in the `charges` table for the affected charging sessions?

Ed-M72:
Hey Adrian. Every single row in the charging table has <invalid> set against fast_charger_brand and fast_charger_type.

adriankumpf:
Hmm interesting. Usually, all supercharges are clearly identifiable via these fields. The values come directly from the API. Unless there is another way to detect the charges, there is not much we can do 🤔

Ed-M72:
How about I add a Tesla Supercharger bool to the geofence table?  I have all the UK superchargers already in there.  It could be used as a fallback in case there is nothing passed back from the API.  I was thinking of adding a "Network" field to the table anyway to allow people to write dashboards that group charging costs by the charging network.

QQ- Does the cost and currency get passed back by the API too?  I struggled to find the piece of code that handled the API call.  Any pointers would be great.

adriankumpf:
Alternatively, tags could be added to geofences. This would potentially enable other use cases as well. It would be nice if Tesla could provide accurate information via the API though 😄

[...]

PR #1591, opened 2021-05-07
Geofence updates

Ed-M72:
Enhanced the geofences table, added the following fields:

Active - Is this record active
provider - Network provider of the charger
Supercharger - Is this record a supercharger, this is to overcome the issue in #1516
country code - Specifies the country of the geofence
currency code - specifies the charge currency.

In addition to this four new master data tables have been created to support this currencies & countries, with accompanying language tables for each.

The foreign key between charging_processes and geofences has been altered as there was an issue that allowed geofences that were already in use to be deleted, and therefore affecting historical data.  The foreign key constraint is now set to restrict, and therefore the record can only be deleted if it is unused.  To support this further the active flag has been introduced to distinguish between active and deactivated geofence records.

The teslamate_web live page has been updated for geofences, along with the schema and changeset to allow the new data to be validated and saved.  Error handling has been added to the frontend fields to ensure the user is prompted of any issues.

@adriankumpf - I need you to do one piece of coding to fully bring this to life.  The logic for the active and supercharger fields needs adding into log.ex.  I have tried to understand the code but it is beyond me.  Please could you update it on my behalf.

Active flag
Effectively the logic should be that geofences are only used if the active = true, that is a new charging process should only ever map to an active geofence

Supercharger
the case statement should read something like:
case (if fast_charger_type = "Tesla" and free_supercharging = true)
               then cost = 0.0
       else  (if geofence = supercharger and free supercharging = true)
               then cost = 0.00
      else calculate the cost from the geofence

The new fields can then be used to improve the existing dashboards and bring a bit more granularity to them.
Ed


Ed-M72:
@adriankumpf  - Ready for re-checking. Ed

adriankumpf:
I see that you have merged the changes from #1562 into this branch. Can we keep these two PR's separate? It's hard to track what this PR changes and what comes from the other.

Ed-M72:
> I see that you have merged the changes from #1562 into this branch. Can we keep these two PR's separate? It's hard to track what this PR changes and what comes from the other.

This is a legacy to me learning Github, this is a separate branch that is kept up to date from the master.  Ideally once #1562 is checked in I will no longer use the master for any further development, and just use branches.

Ed-M72:
@adriankumpf - I found an issue last night, for some reason I can only edit geofences not create/change them.  Will check later to see if it is caused by a code error or by a DB authorisation error after a Postgres upgrade at the weekend.  I expect to do one final commit to this later this week.

adriankumpf:
All right, let's get #1562 merged first then.

Ed-M72:
Re-checked and it was a database schema permissions error stemming from a data restore from backup.  Re-tested and no coding changes needed.  Ready for you to re-check.

Ed-M72:
Closed to allow the fork to be re-adjusted.

PR #1597, opened 2021-05-09
Add device classes and binary sensors to documented Home Assistant config

flacjacket:
For the Home Assistant config, for all of the sensors which are binary on/off sensors, add a section which configures these as a binary sensor type. In addition, add device classes to both the binary sensors as well as some of the regular sensors. The device class makes it so the UI will show, for example "Open" or "Closed" for the door, rather than "On" or "Off.


adriankumpf (approved):
Nice work! Thank you.

PR #1621, opened 2021-05-14
Add Timeline dashboard

DrMichael:


grafana/dashboards/timeline.json
```diff
@@ -0,0 +1,718 @@
...
... T2.car_id\r\nWHERE\r\n\t$__timeFilter(T1.end_date)\r\n\tAND T1.car_id=$car_id \r\n\tAND T2.car_id=$car_id \r\n\tAND TP2.car_id=$car_id \r\n\tAND TP1.car_id=$car_id \r\n\tAND TP2.odometer - TP1.odometer > 0.5\r\n\tAND T1.end_address_id <> T2.start_address_id AND T1.end_geofence_id <> T2.start_geofence_id\r\n\tAND '❓ Missing' in ($action_filter)\r\n\tAND ('Start Address' like '%$text_filter%' or 'End Address' like '%$text_filter%')\r\n\r\nUNION\r\nSELECT\r\n\tstart_date AS \"Start\",\r\n\tstart_date AS \"End\",\r\n\tROUND(EXTRACT(EPOCH FROM start_date))*1000 AS start_date_ts, \r\n\tROUND(EXTRACT(EPOCH FROM start_date))*1000 AS end_date_ts, \r\n\t'💾 Update' AS \"Action\",\r\n\tnull AS distance_$length_unit,\r\n\tnull AS start_path,\r\n\tnull AS end_path,\r\n\tversion AS \"Start Address\",\r\n\t'' AS \"End Address\",\r\n\tnull AS odometer_$length_unit,\r\n\tnull AS \"Distance\", \r\n\tnull AS \"kWh\",\r\n\tnull AS range_diff_$length_unit,\r\n\tnull AS \"SoC\",\r\n\tnull AS \"SoC Diff\",\r\n\tnull AS outside_temp_avg_$temp_unit\r\nFROM updates\r\nWHERE \r\n\t$__timeFilter(start_date)\r\n\tAND car_id = $car_id \r\n\tAND '💾 Update' in ($action_filter)\r\n\r\nORDER BY \"Start\" DESC\r\n",
```
adriankumpf, line 405: Nitpick: The actions are named somewhat inconsistently. How about this:

"🔋 charges" -> "🔋 Charging"
"🚗 Drive" -> "🚗 Driving"
"💾 Update" -> "💾 Updating"

grafana/dashboards/timeline.json
```diff
@@ -0,0 +1,718 @@
...
+        "current": {
+          "selected": false,
+          "text": "",
+          "value": ""
+        },
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "label": "Text Filter",
+        "name": "text_filter",
```
adriankumpf, line 611: Filtering addresses does not work. It matches the strings `'Start Address'` etc instead of the columns :)

grafana/dashboards/timeline.json
```diff
@@ -0,0 +1,718 @@
...
+        },
+        "datasource": null,
+        "definition": "SELECT name AS __text, id AS __value FROM cars;",
+        "description": null,
+        "error": null,
+        "hide": 0,
+        "includeAll": true,
+        "label": "Car",
+        "multi": false,
+        "name": "car_id",
```
adriankumpf, line 504: I think it would make sense to add a separate timeline per car* instead of having this filter. Right now, it's possible to select "All" cars which breaks if there is more than one car.       

* Add panel > Add a new row > Row options > repeat for: `car_id`

adriankumpf (commented):
This dashboard turned out real nice! 👏 Hopefully the longstanding Grafana issue gets included in v8.0.0-beta2 so we can finally hide those utility columns.

grafana/dashboards/timeline.json
```diff
@@ -0,0 +1,718 @@
...
+            }
+          }
+        }
+      ],
+      "type": "table"
+    }
+  ],
+  "schemaVersion": 27,
+  "style": "dark",
+  "tags": [],
```
adriankumpf, line 486: If you add the `tesla` tag to the dashboard, it will be selectable in the drop-down list. You can also add the link to the TeslaMate page and the dashboard picker via the settings to be consistent with the other dashboards.

grafana/dashboards/timeline.json
DrMichael: OK

grafana/dashboards/timeline.json
DrMichael: Hmmm, no idea currently, how to do it. Disabled for now.

grafana/dashboards/timeline.json
DrMichael: Added the tag. If you could add the link to the TeslaMate page and the dashboard picker? I don't dare to.:-)

adriankumpf:
I noticed two more things:

- Joining the next drives by adding `+ 1` to the drive id cannot be relied on. Drives will be deleted if the distance driven is 0, for instance, or if deleted manually.
- The query should not assume that the ideal range is the default. Better: use `[[preferred_range]]_battery_range_km` to take the configured preferred range into account.

I'll try to fix it when I have time.

adriankumpf (approved):

DrMichael:
> * Joining the next drives by adding `+ 1` to the drive id cannot be relied on. Drives will be deleted if the distance driven is 0, for instance, or if deleted manually.

That is a pity. I found "missing" the most interesting part. Maybe it would work with "lag(id) OVER (ORDER BY id)"? Or similar? PSQL wizard needed!!! :-)

adriankumpf (approved):

PR #1645, opened 2021-05-25
Update dashboards.sh

tuxbox:
make the script portable e.g. to BSD.

This small change allow to use the shell script also on BSDs where bash might be installed only as an option in /usr/local/bin


adriankumpf (approved):
Thanks!

PR #1646, opened 2021-05-25
Documentation update: FreeBSD added

tuxbox:
It enhances the documentation to explain the installation and upgrading on FreeBSD straight forward incl. some specialties required like compiling Elixir from source.


adriankumpf (approved):
Thank you :)

Issue #1383, opened 2021-02-23
Proposal: new index on positions


morten-holm:
Okay, so I've noticed that after running Teslamate on my Raspberry Pi 4 for over a year especially the 'Km driven' on the 'Drive Stats' has become sluggish.

I propose the following index:

create index 'name that you pick' on positions ((date_trunc('day', date)::date));

I have done some timings on my Pi for 7 days and 365 days with/without the index, here are the results:

7 days | Before | After |
------| ------| -----|
1st run | 0m52.566s |  0m11.859s |
2nd run | 0m41.985s |  0m11.745s |
3rd  | 0m32.045s | 0m11.647s |

365 days | before | After
---|---|---|
1st run | 0m31.581s | 0m12.100s |
2nd run | 0m21.928s | 0m11.793s |
3rd run | 0m27.231s | 0m11.565s | 

It will improve response times, but I can't tell if/how this will impact logging the positions. The rest of the dashboard is pretty instant and the index will only get it some of the way.

leewillis77:
The fact that both the 7 days and 365 days take ~11s seems strange. Surely the 7 day version should be quicker than the 365 day one?

Dulanic:
Indexing isn't necessary, the report shouldnt be built off positions as that table is just going to get slower and slower over time. I redid this query to use drives instead.

Dulanic:
My previous comment was wrong after speaking to @adriankumpf . I will look at the indexing and come up /w a batter plan.

Dulanic:
@morten-holm So I dont see the index helping too much, the problem is it needs to scan hundreds of thousands of rows. 

1st, can you test this query in grafana? This should be SLIGHTLY faster.

The other option is 100% up to @adriankumpf and that would be to create a new table where the distance travelled could be kept and he would need to create the code in elixir to do so. Honestly, the only reason it is as slow as it is for you is because of the Rasp Pi hardware. SD cards aren't great for databases.  

```

SELECT
    date::date as time,
    convert_km(COALESCE(max(odometer) - min(odometer), 0)::numeric, '$length_unit') as "distance_$length_unit" 
FROM
    positions
WHERE
    car_id = $car_id
    AND $__timeFilter(date)
GROUP BY
    date::date
ORDER BY
    1
```

PR #1676, opened 2021-06-03
Update drive-stats.json

Dulanic:
Fix for #1383 as reporting shouldn't use positions due to # of rows causes slowness.

### DIFF
--- a/grafana/dashboards/drive-stats.json
+++ b/grafana/dashboards/drive-stats.json
@@ -240,7 +240,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


grafana/dashboards/drive-stats.json
```diff
@@ -240,7 +240,7 @@
...
+          "rawSql": "SELECT date_trunc('day', dd)::date AS time,\n       convert_km(drives.distance,'$length_unit') as distance\n  FROM generate_series ((SELECT min(end_date) FROM drives), now()::DATE, '1 day'::interval) dd\n  LEFT JOIN (\n        SELECT car_id ,\n               end_date::date AS date,\n               distance\n          FROM public.drives\n       ) drives\n    ON dd.date = drives.date\n  JOIN settings\n    ON drives.car_id = settings.id\nWHERE $__timeFilter(dd.date)\nAND car_id = $car_id\nORDER BY 1;",
```
adriankumpf, line 243: The reason why this query uses the positions is that trips can sometimes not be logged. The greater the selected period, the larger the potential discrepancy. In my case, it only differs by 100km, but that can be different depending on the user.

Perhaps instead of looking for min and max odometer values, the earliest and most recent positions (where odometer != NULL) could be used.

adriankumpf (changes requested):

grafana/dashboards/drive-stats.json
Dulanic: So I have looked at ways to speed this up if we must use postiions... I have some ideas but the best I can do is save 20% off my run time. It works fine on my i5 server, but I can see Raspberry Pi's being slow. I will create a ticket about indexing then.... Let me rteview because I know some can be removed ex:

on positions
positions_drive_id_index (drive_id)
is covered by
positions_drive_id_elevation_idx (drive_id, elevation)

Dulanic: @adriankumpf  question.... if i backup and restore it completely sets the indexing back to what it was before. If you modify indexing and someone runs the restore, will they lose the indexing that was added?

Dulanic:
Closing for now due to your find that drives dont match positions all the time.

PR #1678, opened 2021-06-03
Update projected-range.json

Dulanic:
Fix for https://github.com/adriankumpf/teslamate/issues/1460. Shouldn't happen, but API errors happen so added wraping /w nullif and added to having not null to exclude bad data.


adriankumpf (approved):
Looks good! Thank you.

PR #1682, opened 2021-06-04
Standardise dashboard date formats

Ed-M72:
Split out from the previous change to facilitate further development

Added Date format (American/European) to settings screen and standardised date formats on dashboards that show them. This gets around the fact hat some of the dashboards were using the distance unit to set the date format.

There are two exceptions to this.

Dashboard "States" - Not possible to calculate a format on the fly due to a restriction on the Stats panel in Grafana.
Dashboard "Statistics" - New variable for date_format created, however the SQL needs reviewing by somebody who understands it. Aka, not me.
Dashboard "Dutch Tax" - This uses the local setting so will always render the date in the users local format.

Once the Grafana plugins have updated fully by the Grafana team to the new standard it should be possible to use date time local for all dashboards. Based on the progress of the Grafana team this could be a while away.

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -15,7 +15,8 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1583945409960,
+  "id": 24,
+  "iteration": 1616436965144,
   "links": [
     {
       "icon": "dashboard",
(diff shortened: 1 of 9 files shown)


Ed-M72:
Looks like the move to Grafana 7.4 removes the need for this. This can be achieved directly in the dashboards directly.

PR #1684, opened 2021-06-04
Add supercharger & currency to geofences

Ed-M72:
Enhanced the geofences table, added the following fields:

Active - Is this record active
Supercharger - Is this record a supercharger, this is to overcome the issue in #1516
currency code - specifies the charge currency.

The teslamate_web live page has been updated for geofences, along with the schema and changeset to allow the new data to be validated and saved. Error handling has been added to the frontend fields to ensure the user is prompted of any issues.

Active flag
Effectively the logic should be that geofences are only used if the active = true, that is a new charging process should only ever map to an active geofence Supercharger.


Ed-M72:
Fixes #1516

adriankumpf:
> The foreign key between charging_processes and geofences has been altered as there was an issue that allowed geofences that were already in use to be deleted, and therefore affecting historical data. The foreign key constraint is now set to restrict, and therefore the record can only be deleted if it is unused. To support this further the active flag has been introduced to distinguish between active and deactivated geofence records.

I don't understand the issue with deleting geofences that are referenced by charging sessions. Historical data should not be affected because the cost is stored on the charge itself, right?

Overall, while I'm sure this approach works for the specific use case you have in mind, I feel like it introduces too many fields to geofences that most users don't need. Charger codes, providers are only useful if you can find a publicly available dataset and are able to populate the database. I also have doubts about whether it is necessary to insert countries and currencies into the database.

Therefore, I think a more generic solution like [dynamic user-generated tags](https://github.com/adriankumpf/teslamate/issues/1516#issuecomment-813571617) would be preferable. This would also allow drive categories (https://github.com/adriankumpf/teslamate/discussions/505) for example.

Ed-M72:
@adriankumpf - I will break this down so we can address each point.  You have some great questions though, which leads me onto how we effectively work together moving forward.

The questions you ask are absolutely valid, however these are the questions we should have discussed and agree on the solution before build had started.  We seem to struggle with communication and all my requests for help, guidance and brainstorming are ignored.  We then get into a situation like this where I have spent weeks developing a solution and we are then back to discussing requirements and wanting to re-write the solution.  It needs to be more effective, perhaps using Discord or something to allow the developers to discuss requirements/solutioning.  I have tried the Github discussions but I don't seem to get any responses to developer specific questions, e-mail responses from yourself are sporadic. It is also not clear which contributor is leading which area of development to allow advice to be sought..

Don't get me wrong, I am happy to get the correct solution, however it needs to be agreed before any code is written, otherwise it wastes an awful lot of development time.  Thoughts?

So to answer your questions:

1) I don't understand the issue with deleting geofences that are referenced by charging sessions. Historical data should not be affected because the cost is stored on the charge itself, right?  

The Charging Dashboard has a panel called "Top charging station by cost".  If you delete a geofence, it will cascade delete form the geoofences table, this will result in the dashboard failing back to the address stored in the address_id.  And you will lose the historical information relating to the charger, there is then no way for a user to re-add the charger.  You will then end up with the panel showing a mixture of addresses and geofences and there will be no way for the user to correct that.

2) Overall, while I'm sure this approach works for the specific use case you have in mind, I feel like it introduces too many fields to geofences that most users don't need. Charger codes, providers are only useful if you can find a publicly available dataset and are able to populate the database.  

Dave W over at https://www.teslaev.co.uk, hosts a compiled list of all Tesla Superchargers which he sources from chargemap.org and compiles into standardised lists for each country.  As per my previous e-mail a charger_code is required to allow a previously uploaded list to be re-uploaded with updates, as in its current for there is no unique field in the current geofences to facilitate this, and runs  huge risk of duplicate records being created.

I have already discussed with Dave on the possibility of altering his dataset to provide the unique charger_codes, currency, country and provider.

In the longer term it may be possible to use the API from chargemap.org to pull in the data directly, however they don't seem to store a unique ID against each charger either, which is puzzling.  The API longer term solution will take me considerable time to implement.

3) Provider Field - This has been asked for a few times in the discussions, and would provide further granularity to the charger dashboards.  Am sure that this would be useful to somebody like @DrMichael, who seem sot be the dashboard guru .  It could be removed, thoughts?

3) I also have doubts about whether it is necessary to insert countries and currencies into the database. I will break this into two, as I am unclear of your question.  Hopefully one of my answers will answer your question :)

a) There is a requirement to hold this currency data against a geofence, for example if you live in a country like Ireland, you will have access to chargers in GBP and EUR.  I also have this need as I travel form the UK, through France and into Austria, so some of my charges are in GBP, and some in EUR.  Currently the dashboards costs don't make sense, as they just add the GBP values to  the EUR ones and give you a random total which makes no sense.  The totals should be split down by currency to make any sense.

The country field is an addition to add further granularity to future dashboards, something I thought @DrMichael could make use of.  It could be removed, thoughts?

b) If you meant. should we store master data for currency and countries, rather than pull it for an API, yes that is an interesting one.  We could effectively populate a pull down on the web screen at runtime to fill the list, hence no need for a master data table.  WDYT?  Do yu have an example of how to do that?

TAGS
----
We talked about that before, I get the concept, am not quite sure how Grafana allows you to restrict your dataset if one of the values is in a tag.  If you remember I asked you for some help with implementing the tags option as it was new to me, if you couldn't help me I said I would have to with the non-tags option.  You never replied, so I moved ahead with the non-tags option.  Now you say you want the tags option....I am happy to move forward with this, but am finding it difficult to work like this. 

As i can see it I can evaluate the tags option but yet again it will need some time and effort from yourself.  In addition to that I could do with you providing an example of it in use somewhere else that I can look at.  I assume it will need:

1) Some form of UI update to allow the tags to be defined and added.  
2) An agreement on what should become a tag and what should be a DB field,
3) Some Elixir code to manage the load, interpret and save the tags. 

I can split this pull request down into multiple pull requests once we agree on the right solution.

Ed-M72:
I have not had a response on the above, so I have, moved forward with the alternative option I presented.  And have taken to remove the following fields, "active", "country_code" & "provider".  That leaves just two new fields, "supercharger" & "currency_code". 

I will update this pull request shortly with the changes, once you have given me feedback on the log.ex change I made.

I will then raise a follow up pull request to add a "consistency" button on the Geofence overview page.  That way this pull request can be promoted first, without dependencies.

I will also raise a follow up tags pull request when you come back to me with some examples.

Ed-M72:
@leewillis77 - Am wondering if you could give me some inspiration?

I can fix all the test errors by removing any reference to the billing_type field on the geofences screen.  This is perplexing as I do not believe I have touched this field in my code changes.

  1) test New validates changes when creating a new geo-fence (TeslaMateWeb.GeoFenceLiveTest)
     test/teslamate_web/live/geofence_live_test.exs:206
     match (=) failed
     code:  assert "Per kWh" = field_cost_per_unit |> Floki.find("#geo_fence_billing_type option[selected]") |> Floki.text()
     left:  "Per kWh"
     right: ""

is the current issue. There are three more related to the same field.  This field is an odd one as it exists in the HTML and the Schema only.

Ed

PR #1685, opened 2021-06-04
Change some dashboards to browser local time

Ed-M72:
With the migration to Grafana 7.4 we now have the option of using the browser locale, and therefore catering for peoples local date/time format, aka they should now support the UK date format. #1692 

The following dashboards have been updated to reflect this:
Updates,
States
Stastistics.

There are three further dashboards which require updating, they are:
Charges
Drives
Vampire Drain,

However these are more complicated and require the panels to be upgraded to a newer version, so will be dealt with soon.


Dulanic:
Am I wrong or didn't grafana already convert epoch time to local browser time /w $__time(column)? I'm not questioning your approach, I am genuinely curious :)

Ed-M72:
> Am I wrong or didn't grafana already convert epoch time to local browser time /w $__time(column)? I'm not questioning your approach, I am genuinely curious :)

Yes it did, however Grafana was inconsistent and didn't use this approach across all its plugins.  If you look at my previous fix for this I was restricted to Date|Time_US & DateTime ISO.  Luckily this approach is no longer needed with the move to the latest version of Grafana.

adriankumpf (approved):
Looks good to me. Thank you!

PR #1693, opened 2021-06-06
Update charge-level.json

Dulanic:
Simplified query, charges not needed as position still captures during charges. No need to access 2 tables when 1 fits the need. Positions is a good fit due to short default timeframe.


adriankumpf (approved):
That makes sense. Thanks!

PR #1696, opened 2021-06-07
Convert dashboard table panels

Ed-M72:
Convert dashboards to the new table panel type.

Like for like conversion with the exception of dates are now using Datetime local, rather than being overridden to either US or ISO.

This means that more localised date formats are supported.


Ed-M72:
Closing until I have found a workaround for the data link variables

PR #1708, opened 2021-06-10
Adding Missing, some adjustments

DrMichael:
Now Odometer, Range and SoC should always have the values at the end of the slot.


adriankumpf (approved):
🚀🚀

Issue #1694, opened 2021-06-06
Grafana Headers Don't Scroll Horizontally


tcm0116:
Using version 1.22.0, the take headers in Grafana's mobile view don't scroll with the data. I believe it may be [this issue](https://github.com/grafana/grafana/issues/4157)  that should be resolved by using an updated table panel.

adriankumpf:
Yup, this is a known issue. However, we cannot migrate to new table component is blocked until this is implemented: https://github.com/grafana/grafana/issues/24092

PR #1729, opened 2021-06-15
Preview dashboards

Ed-M72:
In the spirit of DevOps I have migrated all of the old table panels into the new panel types, added the new dashboards into a "Preview" sub folder which gives us the following benefits:

1. People can start to get value early, using them now, getting around issues such as #1694.
2. The Dates/times are now displayed based on the browser locale rather than being linked to the unit of length, which means they now support UK date formats. #1692
3. We have a set of dashboards that can easily be tweaked when we migrate to Grafana 8.x.

These changes superseded the previous Erlang based change I crafted as it is a more strategic fix.

The dashboards have their reference fields visible and pushed out to the right of the display, as such that means the dashboards will have to stay in preview status until the the following Grafana feature has been implemented. [Feature request 24092](https://github.com/grafana/grafana/issues/24092).  Once it has been implemented we can then hide the reference fields.

Changes include:

- Standardisation of the column names across the dashboards
- Converted hard coded links to use variable names rather than hard cell references
- Handling of length and temperature settings correctly.
- Missing colors on some columns re-added.

@adriankumpf  - The docker image needs a new "Preview" folder adding to the database.  Unless you can point me in the direction of the Grafana image build script?


adriankumpf:
Nice work!

I am very much looking forward to the day when we can migrate all the tables. 

The problem with having old and preview dashboards at the same time is the increased maintenance burden: every small change needs to be synced. I am not willing to spend this extra time.

I will leave this PR open. So everyone who is interested can download and use the preview versions. As soon as Grafana allows to hide fields but still reference them in data links, all tables will be migrated to the new component. Let's hope that version 8.1 will finally ship this feature 🤞

Ed-M72:
@adriankumpf - OK am happy with that.  Am wondering if we should have a grafana-8 branch for this rather than pushing to the master branch.  Then they can be merged and everyone can prep for the Grafana 8.1 release.  Or is that what you don't want to do?

PR #1730, opened 2021-06-17
Fix Timeline Dashboard missing datasources

nickbock:
timeline.json was missing the datasources. Dashboard did not work without those.


adriankumpf (approved):
Good catch! Thanks.

PR #1748, opened 2021-06-24
Make kWh in Charging more exact

DrMichael:
Change kWh in Charging to charging_processes.charge_energy_added


adriankumpf (approved):
Thanks!

PR #1756, opened 2021-06-25
Confusing typo for binary_sensor heading in home_assistant.md

mrzeldaguy:
Fixed heading of the **binary_sensor** config. 

Old one caused some slight confusion while doing my own setup, as I tried adding it in the stated **"sensor:"** section first.


adriankumpf (approved):
Thanks!

PR #1758, opened 2021-06-26
website: Add guide for using tailscale

sit:
TailScale is a secure personal VPN that can allow setting up
TeslaMate for access from anywhere without exposing it to the
public internet.

Hope this is helpful, would love any feedback on how to improve clarity of the write-up. Thank you for all the hard-work on Teslamate!


adriankumpf:
Hi @sit, thanks for the detailed write-up! 

I have to say that I'm a little hesitant to merge this. This has nothing to do with the guide itself, but simply with the fact that I don't like to prominently promote a payed service (even though they offer a free tier). In addition, the sidecar container uses an unofficial image which is not ideal security-wise.

On the other hand, it would be very useful for some to have an easy to set up VPN. I'll have to think about this again.

sit:
Good points both, feel free to close without merging. 

I think Tailscale is working on an official image, but that won't change the fact that they have a paid service.

I suppose it would be possible to use WireGuard directly and then a user could use whatever WireGuard client they want to connect. That guide would be more complex for sure.

Dulanic:
This is outside the scope of this project.

Issue #1795, opened 2021-07-05
Grafana - permissions of /var/lib/grafana-plugins - Unable to use GF_INSTALL_PLUGINS


letienne:
**Describe the bug**

I would like to be able to extend grafana using GF_INSTALL_PLUGINS to add plugins so that I could then add a dashboard.

However when using GF_INSTALL_PLUGINS grafana does not start as /var/lib/grafana-plugins belongs to root.

**Expected behavior**

GF_INSTALL_PLUGINS should install additional plugins at startup time.

**How to reproduce it (as minimally and precisely as possible)**:

- add GF_INSTALL_PLUGINS=someplugin in the compose file
- start the container

**Data**

My guess is that /var/lib/grafana-plugins should belong to id 472 instead of root.

**Environment**

- TeslaMate version: 1.23.4
- Type of installation: Docker 
- OS TeslaMate is installed on: Container
- User OS & Browser: NA
- Others: NA

letienne:
#1814 should fix the issue

PR #1814, opened 2021-07-08
Update permissions to the grafana plugin directory

letienne:
Based on the existing docker image where plugins is owned by grafana, chmod 777.
(lines 77 and 78 of https://github.com/grafana/grafana/blob/main/Dockerfile)

This then allows the use the base GF_INSTALL_PLUGINS environment variable.

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -16,7 +16,9 @@ USER root
 RUN grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install pr0ps-trackmap-panel 2.1.2 && \
     grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install natel-discrete-panel 0.1.0 && \
     grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install grafana-piechart-panel 1.6.1 && \
-    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.15.0/grafana-map-panel-0.15.0.zip plugins install grafana-worldmap-panel-ng
+    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.15.0/grafana-map-panel-0.15.0.zip plugins install grafana-worldmap-panel-ng && \
+    chown -R "grafana" "${GF_PATHS_PLUGINS}" && \
+    chmod -R 777 "${GF_PATHS_PLUGINS}"
 
 USER grafana
 
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -13,14 +13,15 @@ ENV GF_ANALYTICS_REPORTING_ENABLED=FALSE \
 
 USER root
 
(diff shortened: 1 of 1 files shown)


letienne:
fixes #1795

grafana/Dockerfile
```diff
@@ -16,7 +16,9 @@ USER root
 RUN grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install pr0ps-trackmap-panel 2.1.2 && \
     grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install natel-discrete-panel 0.1.0 && \
     grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install grafana-piechart-panel 1.6.1 && \
-    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.15.0/grafana-map-panel-0.15.0.zip plugins install grafana-worldmap-panel-ng
+    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.15.0/grafana-map-panel-0.15.0.zip plugins install grafana-worldmap-panel-ng && \
+    chown -R "grafana" "${GF_PATHS_PLUGINS}" && \
+    chmod -R 777 "${GF_PATHS_PLUGINS}"
```
adriankumpf, line 21: The [custom Dockerfile ](https://github.com/grafana/grafana/blob/main/packaging/docker/custom/Dockerfile) suggests to change the ownership of the pluginsDir even before installing the plugins:

```bash
USER root

RUN mkdir -p "$GF_PATHS_PLUGINS" && \
    chown -R grafana "$GF_PATHS_PLUGINS"

USER grafana

RUN grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install pr0ps-trackmap-panel 2.1.2 && \
    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install natel-discrete-panel 0.1.0 && \
    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" plugins install grafana-piechart-panel 1.6.1 && \
    grafana-cli --pluginsDir "${GF_PATHS_PLUGINS}" --pluginUrl https://github.com/panodata/grafana-map-panel/releases/download/0.15.0/grafana-map-panel-0.15.0.zip plugins install grafana-worldmap-panel-ng

COPY logo.svg /usr/share/grafana/public/img/grafana_icon.svg
# ...
```

It also seems unnecessary to change the permissions that way. What do you think? Would that work for you as well?

letienne: Both should produce identical results regarding the owner of the folder and files, so yes it will work for me.

The only difference will be the missing chmod 777, but I don't think it's needed in that case. I only suggested it because it is included in the "default" Dockerfile [default Dockerfile](https://github.com/grafana/grafana/blob/main/Dockerfile) line 78:
`    chmod -R 777 "$GF_PATHS_DATA" "$GF_PATHS_HOME/.aws" "$GF_PATHS_LOGS" "$GF_PATHS_PLUGINS" "$GF_PATHS_PROVISIONING"`

Thanks !

adriankumpf: All right, I see. Let's try this approach for now then 👍

PR #1818, opened 2021-07-09
Show also missing drives with no geofence id

DrMichael:


adriankumpf (changes requested):
There's a typo: `pq: column "fence_id" does not exist`. I assume `fence_id` should be reverted to `T1.end_geofence_id` but I am not sure.

DrMichael:
Right, typo. Don't know why though... T1.end_geofence_id is right.

BTW, could you perhaps change "Parking", so that short stops don't show up? Perhaps smaller than 2-3 minutes?

adriankumpf:
> BTW, could you perhaps change "Parking", so that short stops don't show up? Perhaps smaller than 2-3 minutes?

Let's do that separately. It would also be good to make this value adjustable. Otherwise, some people will wonder why their short parking sessions were not recorded :)

adriankumpf (approved):
Looks good. Thanks!

DrMichael:
> > BTW, could you perhaps change "Parking", so that short stops don't show up? Perhaps smaller than 2-3 minutes?
> 
> Let's do that separately. It would also be good to make this value adjustable. Otherwise, some people will wonder why their short parking sessions were not recorded :)

Right, I filed an issue: https://github.com/adriankumpf/teslamate/issues/1853

PR #1826, opened 2021-07-11
Minor correction docker-compose.yaml

zerodivide1:
Commented the `volumes` configuration for the `teslamate` container since the following error will appear if you do not have an `import` folder in the same directory as the `docker-compose.yaml` file:

```
Error response from daemon: invalid mount config for type "bind": bind source path does not exist: <path to docker-compose.yaml>/import
```

I only commented rather than removed so that it is easier to restore if a new user is importing data from another system.

An alternative approach is to add a step to the instructions after creating the `docker-compose.yaml` to create the `import` folder. However, I chose this approach instead since its a bit harder to deploy that way if you're using Portainer to add the stack (which is what I was doing).

Recreated error on both Ubuntu Server 20.04 using Portainer and Docker for OSX.

### DIFF
--- a/website/docs/installation/docker.md
+++ b/website/docs/installation/docker.md
@@ -32,8 +32,9 @@ This setup is recommended only if you are running TeslaMate **on your home netwo
          - MQTT_HOST=mosquitto
        ports:
          - 4000:4000
-       volumes:
-         - ./import:/opt/app/import
+# Uncomment the following lines if you have an "import" folder in the same folder as the docker-compose.yaml file
+#       volumes:
+#         - ./import:/opt/app/import
        cap_drop:
          - all
 


adriankumpf:
With Docker version 20.10.7 + docker-compose v2.0.0-beta.6 (macOS) and Docker version 19.03.15 + docker compose v1.24.1 (Raspbian) the `import` directory gets created automatically.  This is also what previous versions did. If Portainer works differently, that's not a reason to make that change for all Docker users imo. The sample docker-compose.yaml should be as easy to use as possible.

zerodivide1:
Just wanted to follow-up since I saw you close this issue. Its been a while, but I suspect that I may have run into [this issue due to how I started the stack](https://www.reddit.com/r/docker/comments/nkvnff/comment/gzfkq7v/?utm_source=share&utm_medium=web2x&context=3) not realizing that there was (at the time) a difference. There is now no difference between them and using both forms `docker-compose` and `docker compose` result in the `import` directory being created.  That is the only thing I can think of that led to my initial issues on both Docker under Portainer on Ubuntu and Docker Desktop on OSX.

Either way, thanks!

Issue #1839, opened 2021-07-16
Projected Range -Battery Level chart (right axis) max should be 100%


toneus:
**Describe the bug**

Project Range Dashboard - Projected Range - Battery Level chart axis max value should be 100% not 200%

**Expected behavior**

This chart is only displaying Battery Level on the right axis, it should present a range of 0 to 100%.

**Steps to reproduce the behavior:**

1. Go to Projected Range Dashboard
2. Scroll down to Projected Range - Battery Level (2nd chart)
3. Observe Chart - Battery Level series limited to bottom 1/2 of chart due to right axis max range parameter.

**Screenshot**

![image](https://user-images.githubusercontent.com/12435503/125995516-68a9c1de-c9a0-4e0b-8c80-e64ece30df0f.png)

**Environment**

- TeslaMate version: v1.22
- Grafana v7.5.8
- Type of installation: Docker
- OS TeslaMate is installed on: Ubuntu
- User OS & Browser: Chrome
- Others:

PR #1840, opened 2021-07-16
Update projected-range.json

toneus:
Changes chart right y-axis battery level range max from 200% to 100%


toneus:
PR for Issue #1839

adriankumpf (approved):
Thanks!

PR #1857, opened 2021-07-23
Some minor documentation clarifications

billerby:
Ran into some memory issues when installing on a AWS Lightsail instance and thought it would be worth mentioning memory requirements in the docs.


adriankumpf (approved):
Thank you!

PR #1904, opened 2021-08-12
Car display priorities (pass 1)

leewillis77:
Hi; 

Opening this for discussion at this stage, not yet final. This goes some way to addressing the requirement in #1751 

So far, the PR:

- Adds a "display priority" column to the cars table
- Updates the grafana dashboards to order the cars based on that field, falling back to the car name if the priorities are equal

Not included in this PR:
- Make Teslamate web respect the priorities - no idea where to start looking for that
- A UI for users to actually set the priority (requires manual DB updates currently)

Feedback welcome.


priv/repo/migrations/20210812173700_car_priorities.exs
```diff
@@ -0,0 +1,9 @@
+defmodule TeslaMate.Repo.Migrations.CarPriorities do
+  use Ecto.Migration
+
+  def change do
+    alter table(:cars) do
+      add(:display_priority, :int, default: 1)
```
adriankumpf, line 6:
```suggestion
      add(:display_priority, :smallint, default: 1)
```

Not a big deal, but `smallint` should be enough :)

adriankumpf (changes requested):
Hi @leewillis77, thanks for the PR!

The changes look good to me. There are only three other dashboards that would need to be updated: 
- reports/dutch-tax.json
- internal/charge-details.json
- internal/drive-details.json

The necessary changes to the UI could be made in another PR as experienced users could already adjust the priority manually 👍

As for the UI I could imagine a new view on the [settings page](https://github.com/adriankumpf/teslamate/tree/master/lib/teslamate_web/live/settings_live) that allows renaming and reordering cars.
The code  for the main TeslaMate view can be found [here](https://github.com/adriankumpf/teslamate/blob/master/lib/teslamate_web/live/car_live/index.ex#L14).

leewillis77:
Thanks for the feedback. All of those changes (s/int/smallint and the missed dashboards) are updated now. 

Thanks for the pointers for phase 2 :)

adriankumpf (approved):

adriankumpf:
Thank you!

BriCo84:
How does one implement this order feature?  I can't find it anywhere.

I have 2 cars right now, and we just sold the first one.  I want to keep the data for the time being but would prefer it be put in to 2nd priority and show under our other car.  I can't figure out how to do that.

leewillis77:
> How does one implement this order feature? I can't find it anywhere.

There is no user-interface for car-ordering right now. To change the display order of cars you need to run some (relatively simple) database queries to set the desired order in the database. 

**Note: Before doing this, be sure that you have a backup that you're happy you can restore from if anything goes wrong.**

1. Connect to your database.
On a docker setup that will be something like `$ docker-compose exec database psql teslamate teslamate`

2. Get a list of the cars in the database
```SELECT id, vin, name, display_priority FROM cars;
 id |        vin        |     name     | display_priority 
----+-------------------+--------------+------------------
  1 | 5YJ3xxxxxxxxxxxxx | MyTesla      |                3
  3 | 5YJ3xxxxxxxxxxxxx | AnotherTesla |                1
  2 | 5YJ3xxxxxxxxxxxxx | Tessie       |                2
```

3. Update the "display_priority" field for each car to set your desired order, e.g. to have the cars in this example ordered AnotherTesla (car ID 3), MyTesla (car ID 1), then Tessie (car ID 2),  you could run the following:
```
UPDATE cars SET display_priority = 1 WHERE id = 3;
UPDATE cars SET display_priority = 2 WHERE id = 1;
UPDATE cars SET display_priority = 3 WHERE id = 2;
```

PR #1905, opened 2021-08-13
Switch to built-in grafana piecharts

mattbnz:
Is there any reason not to switch to the built-in grafana pie-chart panel types now that they're available?

They appear to works equivalently to the plugin and avoids the need for extra plugin installation steps.

Plus, I've found the plugin versions to be unreliable at rendering.

### DIFF
--- a/grafana/dashboards/charge-level.json
+++ b/grafana/dashboards/charge-level.json
@@ -36,7 +36,7 @@
   "panels": [
     {
       "collapsed": false,
-      "datasource": "TeslaMate",
+      "datasource": "${datasource}",
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -61,7 +61,7 @@
       "bars": false,
       "dashLength": 10,
       "dashes": false,
-      "datasource": "TeslaMate",
+      "datasource": "${datasource}",
       "decimals": 2,
       "fieldConfig": {
         "defaults": {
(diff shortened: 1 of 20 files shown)


adriankumpf (changes requested):
Nice catch! The only reason the plugin has been used after Grafana v7.5 was released is that no one has noticed or bothered to change it :) 

That means the `grafana-piechart-panel` does not need to be installed anymore: https://github.com/adriankumpf/teslamate/blob/master/grafana/Dockerfile#L23

The docs for manual installations would also need to be updated with the new Grafana requirements:
- https://github.com/adriankumpf/teslamate/blob/master/website/docs/installation/debian.md
- https://github.com/adriankumpf/teslamate/blob/master/website/docs/installation/freebsd.md

In addition, the changes made in a12df90 should be discussed in a separate PR.

mattbnz:
a12df90 was definitely not meant to be in this pull request sorry... I've stuffed something up there. 

I'll try and look at the requested changes and get rid of that 2nd committ tomorrow.

tobiasehlert:
> [a12df90](https://github.com/adriankumpf/teslamate/commit/a12df90e1416bf80ce7e9f4c25b906b1f54d56e5) was definitely not meant to be in this pull request sorry... I've stuffed something up there.
> 
> I'll try and look at the requested changes and get rid of that 2nd committ tomorrow.

@mattbnz do you have time to update your PR with the feedback from @adriankumpf, so that we can get those improvements ready? There was a separate PR popping up some days ago with some of your changes 😃

adriankumpf:
The master branch now uses Grafana 8 and was also switched to the built-in piechart panels.

PR #1931, opened 2021-08-27
Add note with custom TM_DB_USER when backing up

kyleawayan:
I was trying to backup my TeslaMate instance, and tried using the command given in the documentation. I kept getting this error:

```
pg_dump: error: connection to database "teslamate" failed: FATAL:  role "teslamate" does not exist
```

However when I was setting up TeslaMate with [this guide](https://docs.teslamate.org/docs/guides/traefik), I changed the values in the `.env` file, specifically `TM_DB_USER`. Since I changed that, I needed to change it in the backup command also.

I added another little note box for others in this scenario.


adriankumpf (approved):
Thank you!

The number of note boxes has grown so much over time that it's time to unify them 😃

PR #1937, opened 2021-08-29
Update grafana rule to limit to teslamate host

benoitm974:
useful if you share an already running traefik or have multiple domain


adriankumpf (approved):
Thanks!

PR #2093, opened 2021-10-24
DC charge curve scatter graph

ToniA:
New dashboard, scatter graph of charge power vs. SOC %. The tooltip shows the charge location and date.

image


adriankumpf (approved):
Awesome! Thank you.

PR #2152, opened 2021-11-14
Charging curve on Charging Stats and Charge Details dashboards

fmossott:
Added a charging curve on Charge Details and Charging Stats dashboards

image
image


parkr:
I have a question! It seems really useful (and straightforward) to have this available on the "Charging Details" page. It's a bit less straightforward for "Charging Stats" since aggregating a charging curve from Level 1, 2, and 3 chargers may not make sense. I see the title of the graph is "DC Fast Charging Speed" – is the intention to display an aggregate Level 3 charging curve over all your drives regardless of State of Charge and the available power at the Fast Charger (150kW vs 250kW vs 350kW)? A user might see their charging curve drop if they start using 150kW chargers instead of 250kW chargers, for example, and be confused about why that is. What is the intention with the aggregate graph? Thank you!

adriankumpf (approved):
Thank you @fmossott! I've combined your charging curve implementation with the one from  #2093 and made a few small tweaks to the curve on the Charge Details dashboard.

PR #2184, opened 2021-11-23
Enable zoom on live view of the car.

bogosj:


silverl:
@adriankumpf This is a great feature to add. Can you accept this PR?

adriankumpf:
Would it be possible to show the zoom buttons only when hovering (or tapping on mobile devices)? This would allow to preserve the clean appearance while offering advanced features, such as the map zoom.

bogosj:
I'm not sure how to accomplish that. Instead I've changed the options to allow for dragging the map, zooming with scrolling, and zooming with pinch on mobile. Thoughts?

adriankumpf:
I like that better.  The catch is that it becomes harder to scroll on mobile or small touch devices. It may also surprise you if you are scrolling with the mouse wheel and have just accidentally positioned the cursor over the map.

I can imagine that displaying the controls while hovering might be a good middle ground. For example, you could do it like this:

```js
    const map = createMap({
      // ...
      zoomControl: true,
      // ...
    });


    map.removeControl(map.zoomControl);

    map.on('mouseover', function(e) { map.addControl( map.zoomControl ); });
    map.on('mouseout', function(e) { map.removeControl( map.zoomControl ); });
```

bogosj:
Agreed, it's pretty nice this way.

adriankumpf (approved):
Cool! Thank you.

PR #2238, opened 2021-12-26
Add native SSL support to simplify the setup to expose to internet

howard0su:
1. Use cowboy SSL support to listen on HTTPS at 4040
2. Use reverse proxy to support Grafana dashboard


howard0su:
reverse proxy doesn't work. I am not familiar with Phoenix. If someone can give some hints, it will be great.

howard0su:
@adriankumpf what do you think about this direction? do you think I should continue? the limitation I can see is that the reverse proxy plug doesn't support websocket well. and new version of Grafana has some features depends on websocket.

tobiasehlert:
Hi @howard0su,

I can't see how this would be a simpler setup in terms that you actually have to sign certs and stuff yourself.

I would definitely thing that it's easier (even if it's called _Advanced installation with Traefik, Let's Encrypt & HTTP Basic Auth_ in the docs page) to just go with like docker-compose and usage of Traefik.
On that way, you can actually get Traefik to handle the TLS stuff and you really don't need to worry at all.

Also by doing this, you could just set the `GF_SERVER_ROOT_URL` and `GF_SERVER_SERVE_FROM_SUB_PATH` parameter in Grafana by using environment flags and you are good to go.

Here is a link to that page that I'm referring to:
https://docs.teslamate.org/docs/guides/traefik

So to be honest, I would rather aim for that instead :)

Kind regards,
Tobias

adriankumpf:
I agree with Tobias. Native SSL support gives the impression to users that it would be okay to expose TeslaMate to the internet without any other means of protection, like what the _Advanced installation with Traefik_ offers.

PR #2242, opened 2021-12-27
Separate Updater server side logic from GenServer implementation

oskar1233:
Not sure if that's any helpful, this is something I've done primarily as an exercise.

1) Create `TeslaMate.Updater.Worker` and move there all GenServer & scheduling related logic from `Updater`
2) Change `UpdaterTest` to not start processes nor use `Process.sleep`
3) Adjust rest of the application to use Worker instead of plain Updater module where appropriate


adriankumpf:
Hi @oskar1233, what do you think are the advantages of splitting the GenServer implementation from the updater logic?

I see that this causes the tests to be not as close to reality i. e. the GenServer is only tested implicitly by the _settings_test.exs_. I also find it a bit inconvenient that the rest of the code had to be adapted and now references a submodule of the updater. 

It is definitely desirable to get rid of `Process.sleep` in the test. This could also be achieved by less intrusive ways, e.g. by having the mocked HTTP call send a message to the test process.

PR #2254, opened 2021-12-31
Fix 'mix test' failures

howard0su:


lib/teslamate/vehicles/vehicle.ex
```diff
@@ -394,7 +394,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
     case stream_data do
       %Stream.Data{} when stale_stream_data? ->
-        Logger.warn("Received stale stream data: #{inspect(stream_data)}", car_id: data.car.id)
```
adriankumpf, line 397: This `inspect` call was essential to debug an issue with the stale streaming responses we had experienced in the past. It should not be removed IMO.

lib/teslamate/vehicles/vehicle.ex
```diff
@@ -445,7 +445,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
...
```
adriankumpf, line 448: Same here

adriankumpf (changes requested):

adriankumpf:
977f423014d5519cd682f62c3efaabb14e9c8a9f already includes a fix that's compatible with Elixir 1.12 and 1.13 so this is not necessary anymore. 

Thanks for the PR though! 🙏

PR #2281, opened 2022-01-06
Added if the car is preconditioning in the Drive panel

carloscuezva:
Now, you can view if the car is preconditioning the battery during a drive and when

image


niklasfink (changes requested):
That's cool!

Just one thing: "Preaconditioning" -> Preconditioning

carloscuezva:
WTF!!!!! jajajaj @niklasfink Thank you!!

I added the correction

adriankumpf (approved):
Thanks!

Issue #2272, opened 2022-01-03
No consumption data on new installation


walkercam:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Teslamate has been up and running for several days now with multiple drive and charge cycles. All data appears to be showing correctly however I have a complete lack of energy consumption data. I have gone through some of the past issues and it appears there may be others with a similar issue but with no resolution. I've gone through all of the settings and documentation without finding a possible solution. 

Car is a new Chinese assembled M3 SR+.

![image](https://user-images.githubusercontent.com/133103/147896963-25242dc3-30aa-4439-819a-5405ee7da7ce.png)

### Relevant log output

```shell
No obvious issues in the logs (although I'm no expert)
```

### Type of installation

Docker

### Version

1.25.0

BabyYeggie:
Efficiency is not calculated for drives of less than 1km.

DrMichael:
Could well be, that efficiency is empty. What gives `select id, efficiency from cars;`?

That should be filled after a while...

niklasfink:
This is the same problem I'm having and what we already started discussing in #2229. However, since #2229 is probably a slightly different problem (because the issue opener is seeing kWh values, but no efficiencies), I'll continue here.

My cars also do not have an efficiency value, even though they have been charged multiple times for more than 30mins and have lots of drives.

I have found the following in the code base in the `recalculate_efficiency` func:
https://github.com/adriankumpf/teslamate/blob/41f1280e75e8f48db702f18755e1ce0ab1f044bd/lib/teslamate/log.ex#L605

Since there is no documentation on the efficiency feature, I can only guess from that little snippet, that the efficiency is calculated after a charging cycle with very specific properties, such as duration>10, end_battery_level<95 ...

In my case, I mostly charge to 100%, which probably a lot of SR+ LFP battery owners will do.

@adriankumpf what is the reasoning behind the efficiency calculation only being done for charging cycles with end_battery_level<95?

niklasfink:
I finally received an efficiency value for one of the monitored cars after I stopped the charge at 90%. 

Seem like there are quite a few conditions until an efficiency value is calculated. This is currently not explained in the FAQ, therefore a lot of questions are coming up.

![image](https://user-images.githubusercontent.com/1809301/148176911-98055daf-dd49-4e8d-83ee-dc87c1ade828.png)

walkercam:
[...]

Issue #2229, opened 2021-12-19
Consumption will often not shown on drives list, but on drive details


bobo71dream:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Hi. I have in my Drives list often not shown field Consumption. In Drives Details the information is shown. For example see screenshots attached. It is possible to fix it.
PS: this problem was present in all used versions for me. I use teslamate since 12/2020.

Thank you for your good tool and i hope for your support.

### Relevant log output

```shell
no log inserted
```

### Screenshots

![image](https://user-images.githubusercontent.com/10685047/146670892-64274ef7-b32a-414b-9168-ce98559c029c.png)
![image](https://user-images.githubusercontent.com/10685047/146670916-4718d68c-3e39-4711-91f4-d96c541fe647.png)

### Type of installation

Docker

### Version

v1.25.0

DrMichael:
Might be the same as https://github.com/adriankumpf/teslamate/issues/2193

cwanja:
> Might be the same as #2193

Definitely seems possible. But it shows in the drive details. Cannot recall if mine showed in the drive details or not. 

@bobo71dream - can you run the following command from your database / docker installation? 
`select id, efficiency from cars;`

Another alternative can be:
`docker-compose exec database psql teslamate teslamate -c \ "select id, efficiency from cars;"`

**Note:** depending on your installation `database` might need to be replaced with `db`.

cleanev:
Another data point, if it is of any use in identifying why same drive from point A to point B and back, where one of the two is always missing consumption calculation. The drives shown are exactly over same route to and fro.
There seems to be some subtle measure that makes this calculation show or otherwise
![EFC63138-7394-493B-9E1C-F3D816F5165E](https://user-images.githubusercontent.com/45519661/147387969-0369fb02-db0c-4d35-aaff-055ae3ca6ad9.jpeg)

niklasfink:
> > Might be the same as #2193
> 
> Definitely seems possible. But it shows in the drive details. Cannot recall if mine showed in the drive details or not.
> 
> @bobo71dream - can you run the following command from your database / docker installation? `select id, efficiency from cars;`
> 
> Another alternative can be: `docker-compose exec database psql teslamate teslamate -c \ "select id, efficiency from cars;"`
> 
> **Note:** depending on your installation `database` might need to be replaced with `db`.

[...]

PR #2345, opened 2022-01-19
Update faq.md

cwanja:
Clarified how the consumption values are calculated and what triggers the recalculations based on issues #2193, #2229 and #2272


adriankumpf (approved):
Thank you!

PR #2353, opened 2022-01-22
Update timeline.json: add filter for destinations

DrMichael:
Not sure about iteration, uid and version...


DrMichael:
New one coming...

PR #2354, opened 2022-01-22
Add filters for destination

DrMichael:
Not sure about iteration, uid and version...


adriankumpf (approved):
Thanks! 🙌 

Changing the uid should not be a big deal, but for compatibility reasons we should leave it unchanged if possible :)

PR #2368, opened 2022-01-25
Update upgrading.mdx

cwanja:
Added text to clarify that the `pull` command needs to be ran from the directory where the docker YML file is located.


adriankumpf (approved):
Thanks!

PR #2380, opened 2022-01-26
If the recharge cost is null, there is no visible link

carloscuezva:
Set "-" character as default value if is null

Before:
Before

After:
After


adriankumpf (approved):
Good catch! Thank you.

carloscuezva:
Not at all

srivaths:
Related...Is it possible to set the cost for a location/geofence (such as home)?  This above fix seems to address the cost for just one instance of charging.  And even for this single instance, it seems to revert to Total cost (even if I specify the rate /kWh).

carloscuezva:
@srivaths, this option already exists now, you can click in any location and set the cost

image

srivaths:
@carloscuezva Ah! Thanks.  It's on the geofence itself. I got it now.

However I made a mistake while entering the value (cost/kWh).  Instead of entering 0.1 I entered 0.01.  Now I have corrected it in the geofence.  However the charges dashboard still is using the old value (0.01)

srivaths:
@carloscuezva @adriankumpf 
As noted in my previous response, a change to the cost at a geofence is not getting reflected in the **Charges** dashboard.

This is easy to test by changing the cost at a geofence to 0/kWh.  The **Charges** dashboard should then show 0 in the cost column.  It does not.

carloscuezva:
Price changes should never be retroactive by default

srivaths:
@carloscuezva Very valid point.  However what is the solution when a user makes a mistake (like I did)?

PR #2391, opened 2022-01-27
Add datasource to table and map panels.

andrewjw:
This is a bit of a niche case - I integrate TeslaMate with an existing Grafana installation, which has a different datasource as the default. With the most recent release the table panels just displayed "No Data", and the map panels just displayed the world.

It took a while to debug, but I think it came down to the datasource not being specified correctly, and using the default - if that's not the default TeslaMate connection then you get no data.

Grafana don't specify their panel json format, so it's hard to know what the correct fix is. This change is pretty minimal and fixes the issue for me, so I wanted to submit in case it's useful to others. I've not tested it in a more standard installation, but I presume it will work ok.

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -763,6 +763,7 @@
           ]
         }
       ],
+      "datasource": "TeslaMate",
       "title": "Charges",
       "transformations": [
         {
--- a/grafana/dashboards/drives.json
+++ b/grafana/dashboards/drives.json
@@ -805,6 +805,7 @@
           ]
         }
       ],
+      "datasource": "TeslaMate",
       "title": "Drive",
       "transformations": [
         {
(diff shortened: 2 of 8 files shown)


adriankumpf:
In Grafana 7 (and according to my testing also in version 8) the dashboards use the datasource named "TeslaMate", regardless of whether it is set as default. Have you tried if it works if you add or rename your datasource? 

I suspect it would break many installations if the dashboards always used the default datasource.

cguedel:
I can confirm this fixes my issue (described here https://github.com/adriankumpf/teslamate/discussions/2407) as well!

andrewjw:
Hi,

I'm not sure what you mean by rename the datasource - do you mean rename the TeslaMate datasource? I'm not sure what that would do...

Most panels work fine, it's just the new map and tables panel that didn't work for me. #2407 is definitely the same problem that I was seeing.

Thanks,
Andrew

billimek:
Confirming the same experience here and the fix @andrewjw arrived at fixes the issue.

I don't think it would be obvious that this is an issue if you are running Grafana with the TeslaMate datasource as the default, as @andrewjw says.

TL;DR: please merge this! 😉

PR #2415, opened 2022-02-01
Update package versions

howard0su:
Keep up with the new version of packages.


adriankumpf (approved):

adriankumpf:
Thanks. We have @dependabot enabled for updating dependencies. So PR are usually not needed for that.

Issue #2413, opened 2022-01-31
TeslaApi.Error after updating to the last version


ismcan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

After updating to the last version the application is not able to connect to the API anymore with error 408 as below:
I'm in NA so it's using this API endpoint https://owner-api.teslamotors.com/api

I also regenerated the API token and I'm able to login with it

docker-compose.yml file

> version: "3"
>services:
>  teslamate:
>    image: teslamate/teslamate:latest
>    restart: always
>    environment:
>      - DATABASE_USER=${TM_DB_USER}
>      - DATABASE_PASS=${TM_DB_PASS}
>      - DATABASE_NAME=${TM_DB_NAME}
>      - DATABASE_HOST=database
>      - MQTT_HOST=mosquitto
>\#      - VIRTUAL_HOST=${FQDN_TM}
>\#      - CHECK_ORIGIN=false
>      - TZ=${TM_TZ}
>    ports:
>      - 4000:4000
>\#      - 127.0.0.1:4000:4000
>    volumes:
>      - ./import:/opt/app/import
>    cap_drop:
>      - all

### Expected Behavior

TeslaMate should be able to connect and pull data for the car.

### Relevant log output

```shell
teslamate_1  | 2022-01-31 12:54:49.637 [info] GET /
teslamate_1  | 2022-01-31 12:54:49.641 [info] Sent 200 in 3ms
teslamate_1  | 2022-01-31 12:55:39.160 [info] GET /settings
teslamate_1  | 2022-01-31 12:55:39.173 [info] Sent 200 in 13ms
teslamate_1  | 2022-01-31 12:55:45.689 [info] GET /
teslamate_1  | 2022-01-31 12:55:45.693 [info] Sent 302 in 3ms
teslamate_1  | 2022-01-31 12:55:45.697 [info] GET /sign_in
teslamate_1  | 2022-01-31 12:55:45.703 [info] Sent 200 in 5ms
teslamate_1  | 2022-01-31 12:56:04.915 [error] POST https://auth.tesla.com/oauth2/v3/token -> 400 (371.681 ms)
teslamate_1  | 2022-01-31 12:59:03.062 [error] POST https://auth.tesla.com/oauth2/v3/token -> 400 (208.466 ms)
teslamate_1  | 2022-01-31 13:00:24.663 [info] POST https://auth.tesla.com/oauth2/v3/token -> 200 (370.797 ms)
[...]

PR #2431, opened 2022-02-01
Update vehicle.ex

cwanja:
Would it make sense to update the `408` error response per the conversation in #2413? If not, feel free to reject the PR. Chatted with Tyler over at TeslaScope and he mentioned `408` error is when the vehicle is sleeping
Screen Shot 2022-02-01 at 9 28 11 AM

### DIFF
--- a/lib/tesla_api/vehicle.ex
+++ b/lib/tesla_api/vehicle.ex
@@ -95,7 +95,7 @@ defmodule TeslaApi.Vehicle do
       %Tesla.Env{status: 405, body: %{"error" => "vehicle is currently in service"}} = env ->
         {:error, %Error{reason: :vehicle_in_service, env: env}}
 
-      %Tesla.Env{status: 408, body: %{"error" => "vehicle unavailable:" <> _}} = env ->
+      %Tesla.Env{status: 408, body: %{"error" => "vehicle asleep" <> _}} = env ->
         {:error, %Error{reason: :vehicle_unavailable, env: env}}
 
       %Tesla.Env{status: 504} = env ->


adriankumpf:
408s can also occur  while driving, when the car is awake but temporarily unavailable, hence the `{"error": "timeout"}` message:

```
2022-02-01 10:05:42.327 car_id=1 [info] Driving / Start
2022-02-01 10:17:53.168 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/.../vehicle_data -> 408 (9082.294 ms)
2022-02-01 10:17:53.168 [warning] TeslaApi.Error / %{"error" => "{\"error\": \"timeout\"}", "error_description" => "", "response" => nil}
2022-02-01 10:17:53.168 car_id=1 [error] Error / :unknown
```

In some cases, the car may actually already be asleep. In #2413, however, you can also see that sometimes a few seconds before receiving those timeouts, the API has reported the car to be "online". Since a timeout error is returned in both cases, we cannot reliably distinct the two.

Thanks for helping out and answering issues by the way. Much appreciated! 🙌

cwanja:
Sounds good! Appreciate the response. Feel free to reject 😊

Of course, happy to do it where and when I can. I _actually_ want to jump into the web code and try to offer a variable of either Not a Tesla App or TeslaScope for release notes. I think TeslaScope is finally getting back to including release notes. Totally off topic from this pull request though.

adriankumpf:
>  I actually want to jump into the web code and try to offer a variable of either Not a Tesla App or TeslaScope for release notes. I think TeslaScope is finally getting back to including release notes. Totally off topic from this pull request though.

It could perhaps even be generalized it to the point where you can provide a URL pattern yourself, like `https://www.notateslaapp.com/software-updates/version/$VERSION/release-notes`. And both the Updates dashboard and the logic for the car view would need to be adjusted to be consistent.

Feel free to create a separate discussion for this topic.

PR #2447, opened 2022-02-07
add dataSource where needed

billimek:
In situations where the `TeslaMate` datasource is not the default datasource for grafana, it is necessary to properly denote the `TeslaMate` datasource for panel scopes to include FieldConfig as well as some map panel types.

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -58,6 +58,7 @@
       "type": "row"
     },
     {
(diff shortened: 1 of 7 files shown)


andrewjw:
This PR is pretty much the same as #2391. It's definitely solving the same issue.

billimek:
Oops, I should have checked.  The other one looks more complete!

PR #2448, opened 2022-02-07
Added panel with the cost of charges in SuC

carloscuezva:
The data will depend on whether the location of the recharges is correct or not

Captura de pantalla 2022-02-07 a las 18 10 34


ejseguinte:
Would it be possible  to update this to use address name or geofence name? There is a a super charger station near me that is close to a ChargePoint and that address was used for the supercharger session instead of the supercharger. I updated it in geofences but the query won't pick it up since it uses address name. 

My guess would be the query would look something like:

```
SELECT
	sum(c.cost)
FROM
	charging_processes c
	LEFT JOIN addresses address ON c.address_id = address.id
	LEFT JOIN geofences geofence ON geofence_id = geofence.id
WHERE
        $__timeFilter(end_date) AND
        (geofence.name LIKE '%Supercharger%'OR address.name LIKE '%Supercharger%') AND
	c.cost IS NOT NULL AND
        c.car_id = $car_id;
```

DrMichael:
Looks good, makes a big difference for me...

adriankumpf:
-> https://github.com/adriankumpf/teslamate/discussions/2753

Issue #2454, opened 2022-02-08
Detection of Refresh Model X (2022)


natrlhy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Teslamate Web UI does not detect the make/model of the Model X/Model X Plaid

### Expected Behavior

Display the make/model of the car below the vehicle name

### Steps To Reproduce

1. Go to http://teslamate:4000
2. View car(s) listed
3. No make/model for the 2022 Model X

### Relevant log output

```shell
N/A
```

### Screenshots

![PNG image](https://user-images.githubusercontent.com/19696424/153037885-e2229350-44d2-4c2d-8226-440690377bd2.png)

### Type of installation

Docker

### Version

1.26.1

cwanja:
Had the same issue in my refreshed Model S - https://github.com/adriankumpf/teslamate/issues/2193#issuecomment-993406286

We had to update the detection in TeslaMate reading the owners API. Do you know how to query the owners API and get the returned value?

natrlhy:
Yep..

`"car_type": "tamarind"`

@adriankumpf can this be added as well?

Thanks!

cwanja:
Created a PR for it. #2455

natrlhy:
Thanks!

PR #2455, opened 2022-02-08
Detect 2021.5+ Model X car_type

cwanja:
Added support for the 2021.5+ Model X per #2454


adriankumpf (approved):
Thanks!

cwanja:
You need to wait for it to be included in a released version. It will be mentioned in the release notes and linked to this PR. You _technically_ could pull a `dev` branch, but unless you want potentially some bugs I would not suggest it.

natrlhy:
Yeah I normally wait for that reason 😉
Thank you both!

PR #2461, opened 2022-02-10
Update charging-stats.json

DrMichael:
Set Y-Axis to max 100.


adriankumpf (approved):
Yup, that makes sense 👍

PR #2473, opened 2022-02-14
Render Trip piechart legend

cwanja:
Per discussion #2469, looks like the legend got removed. This adds it back as a list.


adriankumpf (approved):
Thanks!

PR #2481, opened 2022-02-18
Update Charging Stats panel styling

cwanja:
Updates the `Cost per 100 $length` panel to not be small text. PR contains before and after screenshots. Makes for consistent formatting across dashboard.

Before
Screen Shot 2022-02-18 at 3 50 06 PM

After
Screen Shot 2022-02-18 at 3 49 45 PM

### DIFF
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -409,7 +409,7 @@
       "id": 26,
       "options": {
         "colorMode": "value",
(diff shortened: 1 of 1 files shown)


adriankumpf (approved):
Good catch! It does look more consistent that way 👍

cwanja:
There is a few others I know I have seen. I will see if I can identify them and fix them either in a single branch and pull or multiple pull requests.

Issue #2494, opened 2022-02-23
Unexpected car model displayed for Model S Refresh


alsun2001:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

My MS LR is displayed as "Model S 100D". It would be nice to have it display as its marketing name, like was done for M3s in #2030 / [6c0c0ee](https://github.com/adriankumpf/teslamate/commit/6c0c0eeaa97b06bd1cdc4c6913d7d33fef93d5c0).

Not sure but additional logic might be needed as 100D is also a valid older Model S trim name.

### Expected Behavior

Model S LR displayed as either 
- "Model S Long Range" (its name when I bought it and what it shows in the Tesla app) or 
- "Model S Dual Motor AWD" (its latest name if ordered today)

### Steps To Reproduce

1. Go to Teslamate front page
2. Observe vehicle model/trim notation for MSLR

### Relevant log output

```shell
N/A
```

### Type of installation

Docker

### Version

1.26.1

natrlhy:
I suspect the Model X Plaid will be the same. I'm guessing it will show up as Model X P100D. Would like it to display Model X Plaid

cwanja:
So there is a couple of things here:
1. First and foremost, Tesla does not provide a _public_ way to identify a Long Range versus a Plaid, etc. Tesla does not even have a "Long Range" designation any more - it is just Dual Motor and Plaid. They stopped identifying the vehicles by battery size with the 100D I believe.
2. Tesla's app likely uses more internal APIs to give you that "Dual Motor AWD" or "Long Range" as opposed to the Owners API that is not documented and not technically publicly available.
3. On that same note, in the public API, there is no way for TeslaMate (or TeslaFi, TeslaScope) to identify more of the model via the `vehicle_data` response. Here is an output for my refreshed Model S. 
```
{"response":
{"badge_version":0,
"can_accept_navigation_requests":true,
"can_actuate_trunks":true,
"car_special_type":"base",
"car_type":"lychee",
"charge_port_type":"US",
"dashcam_clip_save_supported":true,
"default_charge_to_max":false,
"driver_assist":"TeslaAP3",
"ece_restrictions":false,
"efficiency_package":"EfficiencyPackageDefault",
"eu_vehicle":false,
"exterior_color":"SolidBlack",
"has_air_suspension":true,
"has_ludicrous_mode":false,
"has_seat_cooling":true,
"headlamp_type":"HeadlampTypePremium",
[...]

PR #2496, opened 2022-02-24
Display marketing names for Model S and X

cwanja:
All stemming from #2494 - I am going to make this pull request fully knowing it either will be rejected (totally OK) or require some additional work to get it working.

I am not at all close to knowing where the `car.model` or `car.trim_badging` comes from. But I do know to get the right Model S designations, we need to know the `car.car_type`. As such, we need to add this into the logic for lines 64 through 71 (before changes). Also need to add it to line 61 (I believe) to check for this data. **However**, I have no clue what is going to happen to the Model 3 and Y logic that is there if we do not add the `car_type` matching data to those lines (64 - 71). I also have no clue if my modifications would work either. Or if there would just have to be repetitive checks for the 3 and Y, because `car.model` already equals 3 / Y and `car.car_type` also equals 3 / Y (per this comment https://github.com/adriankumpf/teslamate/issues/2494#issuecomment-1049360750)

But to get the right Model S (And X) names, we need not only the `car.model`, we need the `car.car_type` attribute as well  since they started updating `car_type` with the 2021.5 refresh. Because there is a 100D and P100D Model S and X, now they are reusing the 100D designation but changed the `car_type` for the refreshed Model S and X - hence the other two PRs - #2455 and #2431 

So I took at stab at it.

### DIFF
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -58,7 +58,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
   defp put_friendly_name(nil), do: nil
 
   defp put_friendly_name(%Car{} = car) do
-    %Car{car | friendly_name: friendly_name(car.model, car.trim_badging)}
+    %Car{car | friendly_name: friendly_name(car.model, car.trim_badging, car.car_type)}
   end
 
   defp friendly_name("3", "P74D"), do: "LR AWD Performance"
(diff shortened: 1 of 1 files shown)


lib/teslamate/vehicles/vehicle/summary.ex
```diff
@@ -58,7 +58,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
   defp put_friendly_name(nil), do: nil
 
   defp put_friendly_name(%Car{} = car) do
-    %Car{car | friendly_name: friendly_name(car.model, car.trim_badging)}
+    %Car{car | friendly_name: friendly_name(car.model, car.trim_badging, car.car_type)}
```
cwanja, line 61: This is the line I am really concerned about breaking the logic further down.

adriankumpf:
Instead of adding the new `car_type` field to the car, I added the marketing name directly (see a0e12c3808187dee885efba0bb1ab79a0555fd72). Your code for detecting new model S and X is also included in that commit so this MR can be closed. Nevertheless, thank you for investigating the issue.

PR #2527, opened 2022-03-08
Added Battery heater trigger for battery_heater or battery_heater_on.…

woyteck1:
This will ensure that both S,X and M,Y models will show when battery heater is active during charging.

Also added a way to recognize value "2" on phases as three phases, because it's "2" when the 3-phase connection is active.
Data in the DB stays the same, only Grafana changes the visuals.
image

I have applied these onto my Grafana and I am happy with results.

Woyteck


woyteck1:
I am also inclined to change battery heater value to 10 instead of 1, so it's more visible on the graph, like this:
image

PR #2543, opened 2022-03-15
Update docker.md

SailingDad:
Removed leading space character before comments inviting to insert ENCRYPTION_KEY secret or DATABASE_PASS to prevent connection issues with Grafana.

Cf the following discussion
https://github.com/adriankumpf/teslamate/discussions/2540#discussioncomment-2356612


parkr (approved):

adriankumpf:
This is actually intentional: without the space, the keys/passwords would be set to e.g. `#insert your secure database password!`. The space marks everything starting with `#` a YAML comment.

Issue #1336, opened 2021-02-10
teslamate deletes mqtt retained messages if empty


brianmay:
**Describe the bug**

When teslamate publishes a geofence when the car is not in a geofence, it publishes an empty string "" to mqtt.

But on my mqtt server (mosquitto), publishing an empty string results in the retained object being deleted (arrgh!). I believe this is part of the mqtt spec. Google search seems to confirm.

If a client connects, it will not get the message because it is no longer retained, and cannot determine if the car is outside the geofence or just hasn't received the message yet.

If this is not clear enough, I can provide more references, but getting a bit distracted by external events right now :-)

**Expected behavior**

The geofence position message should always be retained by the mqtt server.

**How to reproduce it (as minimally and precisely as possible)**:

Move car from inside geofence to outside geofence. Note that message gets sent with "geofence" = "". Disconnect client and reconnect. Notice that client doesn't receive the message, because it is no longer retained.

**Notes**

I finally correctly diagnosed this issue because I had the exact same problem with another unrelated application I am responsible. Not sure what the best solution is though. I think we can't publish an empty string and expect it to be retained.

In my case I hard coded a magic value to indicate that there was a special condition. Although this solution makes me feel a bit dirty.

Another possibility is to send a json object for geofence. Not sure if I like this or not. It would be sufficient though.

**Environment**

- TeslaMate version: v1.21.2

adriankumpf:
Good catch. I guess the same problem exists with some other topics, namely those for which values are always published, even if they are `nil`.

https://github.com/adriankumpf/teslamate/blob/188eb036725708b209940162ee02c697f0f09ac8/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex#L47-L49

I also can't think of any workaround other than publishing a magic value. Perhaps a space instead of an empty string?

brianmay:
I have been thinking about this some more.

The main problem I have with the magic value, is that it isn't possible to distinguish the magic string and, say if somebody really does have a location called `NONE` for example.

This isn't going to a problem if the magic string is an invalid value. For example, if the published data is normally an integer, and we publish `NONE` it is pretty obvious that there will be no conflict. Or if we add validation rules to to say that any string starting with a `:` character is now considered an invalid geolocation name.

Some options I have been playing around with:

1. Add quotes before/after location. e.g. `"Home"` or `NONE`. This should work even if user entered quotes for value (why?), e.g. `""Home""`.

[...]

Issue #1912, opened 2021-08-17
Can' receive data of Tesla_shift_state from MQTT


crathyd:
Listen the topic with "teslamate/cars/1/shift_state" but no data response;
but other topics are good like "teslamate/cars/1/vision"
<img width="620" alt="截屏2021-08-17 下午6 07 15" src="https://user-images.githubusercontent.com/13013658/129707508-11cbe60a-178b-4a97-b16a-29716e492c18.png">

cryptelli:
I'm also encountering this issue, where MQTT reports the shift_state when set to something like D, P, R etc but when it stops, the value is wiped and the previous value can no longer be seen.

PR #2564, opened 2022-03-30
If no geofence is found map then use default value

brianmay:
This is required because the default value is "" which tells the mqtt
server to delete the retained value.

This in turn means that applications that care about when the car is in
a unknown location will not get the retained value.

This is a bit different from all of the other values in
`@always_published`. For them "" is only published if the value is
unknown or an error occurred. So it makes sense to delete the retained
value.

In this case however, a value of "" generally means "the car is outside
any known geofence" which is a valid condition it itself, and it should
be retained.

There was speculation this might affect the shift_state, but
unfortunately this was not followed up with any evidence or details. #1912 

There are any number of solutions that could be deployed here. The
advantage of this solution however is it is not going to break anything
unless you specifically define the TESLAMATE_DEFAULT_GEOFENCE variable.

Fixes: #1336


brianmay:
Still in the process of testing this to make sure it works. I think I just noticed a problem already, will fix.

This perhaps isn't the perfect solution. But this solution unlike others I can think of avoids breaking the existing API.

brianmay:
As far as I can tell that test failure is unrelated to anything I did...

brianmay:
Tests have been successful so far. Set `TESLAMATE_DEFAULT_GEOFENCE` to "nowhere" and Tesla reported it was "nowhere" instead of "".

config/runtime.exs
```diff
@@ -80,6 +80,9 @@ defmodule Util do
   end
 end
 
+config :teslamate,
+  default_geofence: System.get_env("TESLAMATE_DEFAULT_GEOFENCE")
```
adriankumpf, line 84: Let's drop the `TESLAMATE_` prefix for the sake of consistency.

adriankumpf (changes requested):
That's pragmatic solution to this long-standing issue 👍

Could you add a brief description of that env variable to `website/docs/configuration/environment_variables.md`?

adriankumpf:
> As far as I can tell that test failure is unrelated to anything I did.
> 
> LATER: I notice other PRs are failing on the same test. So probably a flaky test I think.

Yup, there are a bunch of flaky tests that really need to be fixed

brianmay:
> Could you add a brief description of that env variable to `website/docs/configuration/environment_variables.md`?

Done.

Issue #792, opened 2020-06-14
How to manually add charging sessions to complete the history of the monitored car?


petariga:
I started to use TeslaMate when my car was about a month old. Therefore there have been some charging sessions which had not been logged into TeslaMate, but not too many. I kept the track of the sessions in a notebook. It would be nice to restore them somehow **to keep the charging history of my Tesla in TeslaMate complete from the date zero.** My goal would be to have the Dashboards "Charges", "Charging Stats" and maybe "Efficiency" (Consumption net, gross, Logged Distance) as thorough as possible.

Could you suggest what would be the best way how to achieve it? Is there some sort of import from .csv file (utilize the import from TeslaFi?) or some manual way how to type these sessions in? Is it sufficient to add these sessions only to "charging_processes" table or is it more complex and it can ruin the consistency of the entire system (for example there is no odometer attribute in the "charging_processes", so I assume it is kept elsewhere)? I.e. do you also need to add some drives to "drive" table or anything other to keep the consistency?

ngardiner:
Hi @petariga,

I have added some charging sessions manually to the charging_processes table, and it worked. I have been concerned that perhaps a later update of TM may detect invalid data and remove them, so it's not necessarily the case that this is a foolproof approach but it did work for me in tracking charging consumption and cost. 

When I get some time, I'll go back through my SQL history (if it is still there) and find the insert commands that I used. You'll need to look up a few values, such as the place location ID for the charger (if you haven't visited there since, you'll have a challenge there) and also you may need to adjust the timing to account for timezone.

If you have charged at the same location since, my recommendation is to look at the charging_processes row for that charge and grab the entire row and the column headings. This will help you a lot to construct a charging session manually. 

For some columns (such as distance since last charge, perhaps SOC before and after) you may need to just use 0, which will show a blank in the table in grafana for those values. It hasn't impacted any of the grafana data from my perspective.

petariga:
Hi @ngardiner, 

Thank you for your kind reply. I do have the data for charging_processes table ready including the location, distance since the last charge, SOC before & after. However, if you succeed to dig up the exact SQL command you used to submit or import these additional sessions, it would be definitely helpful.

Have you anyhow reflected the sequential order of these sessions? The sessions to be added happened before the others already presented in the db, however they will have a higher id than the later. Does it matter or is it enough to have a correct date & time for each session? Or shall I export all of them, add the historical sessions, reorder them, clear the entire table and reimport them?

[...]

PR #2589, opened 2022-04-03
Update charges dashboard

qgadrian:
This change will allow adding _offline_ or any historical charge not tracked by Teslamate to the `charging_processes` table.

Without this change, previous charges can be added to Teslamate but they won't be visible as there is no way to fetch the charging details from the Tesla API (or at least I couldn't find an endpoint to get that information).

I have a WIP script to migrate _old charges_ from Tesla (as requested in #792), but we will need either this PR or adding a _fake charge_ to the `charges` table to make them visible... @adriankumpf what would be the preferred option for this project?


reddai75:
Hi ! Is there any update on this ? :)

qgadrian:
@reddai75 This PR has been here for a while, if @adriankumpf wants it to be merged I can happily update it otherwise I won't waste the effort to keep it updated.

Dulanic:
This is an old PR, it had conflicts and there has been changes since submitted. Please resubmit a new PR if you still want to bring up.

PR #2655, opened 2022-05-19
Add FAQ around Docker timestamp logs

cwanja:
This PR will add a FAQ around Docker container timestamp logs and why they differ from your machine if you do not configure the TZ environment variable.


adriankumpf (approved):
Thank you!

PR #2656, opened 2022-05-19
Better Charging Stats -> Charge Delta

nicoladefranceschi:
This change is to better handle multiple charges, one after the other.
By using the "odometer" we can see if the car actually didn't move much between charges (this happens for instance if you arrive at a charging point and only the "slow charger" is available, then the "fast" become available, and you can stop charging, switch and restart with the fast)

Before:

image

Now:

image


adriankumpf (approved):
Clever. Thank you! This is a nice improvement over #2566

PR #2657, opened 2022-05-20
Improve power values for AC charging. Fix Addrees Name that was not s…

jheredianet:
Power value on AC charging always shows an integer.  With this fix we calculate it from current and voltage to show a more accurate float value.
The location addres was not showing because the panel was configured to show numeric fields, now it explicitly show the correct field.


adriankumpf:
Cool idea! The AC detection needs some tweaking though:

### before

Screenshot 2022-06-24 at 13 55 03

### after 

Screenshot 2022-06-24 at 13 54 58

Note how the `charger_voltage` fluctuated between 1 and 2.

adriankumpf (changes requested):
.

jheredianet:
> Cool idea! The AC detection needs some tweaking though:
> 
> ### before
> Screenshot 2022-06-24 at 13 55 03
> 
> ### after
> Screenshot 2022-06-24 at 13 54 58
> 
> Note how the `charger_voltage` fluctuated between 1 and 2.

I could not reproduce in my case. I use a Tesla Model 3. On DC charging, charger_voltage is always 2. 
I made the tweak. Could you try it again please?

jheredianet (commented):
I could not reproduce in my case. I use a Tesla Model 3. On DC charging, charger_voltage is always 2. 
I made the tweak. Could you try it again please?

PR #2689, opened 2022-06-05
HTTP with HTPASSWD for Create traefik_http.md

jeonghoonkang:
It gives simple setup for HTTP with Ht Access in the Traefik service.
Some user who using just in house local machine, needs to setup quick ID/PASSWD enabled TESLAMATE.
This quick and simple setup will help many people to use this service, at a certain secure level.


adriankumpf:
We have to make sure the official examples are secure by default as users could otherwise accidentally expose their TeslaMate instances. A blog post would be better suited for such a guide, for example. Thank you anyway!

PR #2691, opened 2022-06-05
Add cost per kwh and min support for charging stations.

damienbr:
Some EMSP charges per kwh and per minutes for charging stations.
I added the ability to set the price per kwh and per minute when defining the cost.
I'm also adding the cost per kwh per minute in the geofence page configuration.


JakobLichterfeld:
Please update your branch

PR #2704, opened 2022-06-14
update HomeAssistant Integration example yaml

star114:
Regarding #2702, I would like to update the documentation for HomeAssistant Integration.

_Defining manually configured MQTT entities directly under the respective platform keys (e.g., fan, light, sensor, etc.) is deprecated, and support will be removed in Home Assistant Core 2022.9._

_Manually configured MQTT entities should now be defined under the mqtt configuration key in configuration.yaml instead of under the platform key._

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -72,8 +72,10 @@ tesla:
   password: !secret tesla_password
   scan_interval: 3600
 
+mqtt:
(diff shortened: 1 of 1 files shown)


JonasSL:
I can confirm that the changed configs of this commit work with the new 2022.6 release. Thank you @star114!

adriankumpf (approved):
Nice. Thank you!

PR #2727, opened 2022-06-29
Add setting to select tire pressure units

cwanja:
This is one of a few changes that need to be made. Not sure how to then write this setting to the database as a configuration setting. Then the dashboard needs to be updated to reference the variable instead of the dropdown.

Next steps are:
1. Update the website to write the setting to the database configuration
2. Update Grafana to have the tire pressure as a variable
3. Update the `Drive details` to reference the database configuration as a variable


cwanja:
Commit 9dd7776 includes step 1 I believe.

adriankumpf:
Hi @cwanja, thanks for the PR! I have incorporated your changes into this commit: a413e8ce2cabf6c1eb024d88c0c87a957f45c241

PR #2730, opened 2022-06-30
Added dark mode support for teslamate UI

NirKli:
Added dark mode to teslamate UI, the setting is controlled by the OS.
Both of the JS files, are listening to `prefers-color-scheme` event
Tested on PC and iPhone

Thanks to @tobiasehlert for his ideas

PC

[PC](https://user-images.githubusercontent.com/45952578/176739872-6b8f462c-332b-4428-b671-04e5f7f929f4.webm)

iPhone

https://user-images.githubusercontent.com/45952578/176739581-df7986a3-b235-4597-8119-fdc92cd4be28.mp4


tobiasehlert:
Hi @NirKli,

I think it's a good start, but there needs to be some polishing.

First off I think the "invert" function in CSS might not be the best way. It might be better considering choosing a ready theme for this that works with the Bulma CSS framework that is currently used.
One thing I noticed was that the menu at the top still was light, so there are definitely some elements left to go through.

This would maybe be a better choice than going all black?
https://jenil.github.io/bulmaswatch/darkly/
https://jenil.github.io/bulmaswatch/slate/

If you look at Grafana, it's not completely black either.. a dark-grey background would be better for the eyes in my opinion at least.

And I don't know if this might be something to look into as well, but there seems to be some scheme for dark-mode variables that might be helpful for us as well. Check at the blog post from 2019:
https://bulma.io/2019/10/15/light-dark-colors/#scheme-variables-for-dark-mode

Kind regards,
Tobias

tobiasehlert:
@NirKli, rebase the branch instead, so you just get your own specific changes in the PR itself :)

PR #2735, opened 2022-07-04
Update HA integration

dcod3d:
Fix km to mi conversion
Add timestamp class to charge time so HA will show relative time.


adriankumpf (approved):
Thank you!

PR #2840, opened 2022-08-08
sync meta theme-color with header background color

libotony:
The theme-color meta tag suggests the color for the status bar in iOS.
REF: [MDN theme-color](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color)
Before:
before
After:
after


JakobLichterfeld (approved):
lgtm

Issue #447, opened 2020-02-13
PostgreSQL - sslmode = required


scottrobertson:
Is there a way of making this work over SSL? Digital Ocean DB service forces sslmode = required.

When i try and start teslamate, i get:

```
22:39:27.576 [error] Postgrex.Protocol (#PID<0.207.0>) failed to connect: ** (Postgrex.Error) FATAL 28000 (invalid_authorization_specification) no pg_hba.conf entry for host "123.4.5.6", user "doadmin", database "defaultdb", SSL off
```

adriankumpf:
This will be possible with the next release. You'll need to set `DATABASE_SSL=true`.

scottrobertson:
Thank you. Would you be able to trigger a docker image build? Or does that only happen when a release is made?

adriankumpf:
The next release will happen soon. Until then, you could run the master branch :)

scottrobertson:
I don't know if it's related, so posting here, but it keeps getting to the following migration, and then just exiting and crashing the pod.

```
10:47:30.273 [info] == Running 20191008191431 TeslaMate.Repo.Migrations.FixLlToEarth.up/0 forward
10:47:30.277 [info] execute "CREATE OR REPLACE FUNCTION public.ll_to_earth(float8, float8)\nRETURNS public.earth\nLANGUAGE SQL\nIMMUTABLE STRICT\nPARALLEL SAFE\nAS 'SELECT public.cube(public.cube(public.cube(public.earth()*cos(radians($1))*cos(radians($2))),public.earth()*cos(radians($1))*sin(radians($2))),public.earth()*sin(radians($1)))::public.earth';\n"
** (Postgrex.Error) ERROR 42501 (insufficient_privilege) must be owner of function ll_to_earth
    (ecto_sql 3.3.3) lib/ecto/adapters/sql.ex:612: Ecto.Adapters.SQL.raise_sql_call_error/1
    (elixir 1.10.1) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
    (ecto_sql 3.3.3) lib/ecto/adapters/sql.ex:699: Ecto.Adapters.SQL.execute_ddl/4
    (ecto_sql 3.3.3) lib/ecto/migration/runner.ex:343: Ecto.Migration.Runner.log_and_execute_ddl/3
    (ecto_sql 3.3.3) lib/ecto/migration/runner.ex:117: anonymous fn/6 in Ecto.Migration.Runner.flush/0
[...]

PR #2846, opened 2022-08-12
Added option to select sslmode when connecting to DB in Grafana

LeeHanYeong:
In teslamate, you can configure an SSL connection with the DB using the `DATABASE_SSL` environment variable.
https://docs.teslamate.org/docs/configuration/environment_variables

But grafana doesn't have that option.

I found the `sslmode` option of DB connection in Grafana's documentation. This is in the jsonData entry in datasources.yml.
https://grafana.com/docs/grafana/latest/administration/provisioning/#data-sources

**Guide documentation**
I configured it to receive this as an environment variable. I also added information to the traefik and apache guides documentation.
```
  grafana:
    image: teslamate/grafana:latest
    restart: always
    environment:
      - DATABASE_USER=${TM_DB_USER}
      - DATABASE_PASS=${TM_DB_PASS}
      - DATABASE_NAME=${TM_DB_NAME}
      - DATABASE_HOST=database
      - DATABASE_SSLMODE=disable  # this line
```

**Dockerfile**
By adding the ENV to Dockerfile, users who don't have this option in docker-compose can use it without any problems.
```
FROM grafana/grafana:8.5.9

ENV GF_ANALYTICS_REPORTING_ENABLED=FALSE \
    ...
    DATABASE_SSLMODE=disable
```

When I ran grafana using the Dockerfile with this option added, it was confirmed that it was connected to PostgreSQL using SSL.
- Using DockerImage: https://hub.docker.com/repository/docker/azelf/grafana

This is also related to #447 from about 2 years ago. It seems that the user who opened the issue had the same problem.

Thank you for your review. have a good day :)


LeeHanYeong:
This problem has not been resolved for several years. I didn't make it an environment variable option like {TM_DB_...} in docker-compose.yml because I thought it wasn't a frequently used option.
I expect users who need this option to modify docker-compose.yml appropriately.

mikesmitty:
Seconded, this would be a nice and simple quality of life change. Manually updating the datasource every time the grafana container restarts is a little less than ideal.

LeeHanYeong:
@adriankumpf 
It would be great if you could review this PR. I need to create a custom Grafana image when updating the TeslaMate version to apply SSL to the DB connection.
Thank you :)

website/docs/guides/apache.md
```diff
@@ -69,6 +69,7 @@ services:
       - DATABASE_PASS=${TM_DB_PASS}
       - DATABASE_NAME=${TM_DB_NAME}
       - DATABASE_HOST=database
+      - DATABASE_SSLMODE=disable
```
JakobLichterfeld, line 72: Since this is the default value, I vote not to include it in the documentation at this location.

website/docs/guides/traefik.md
```diff
@@ -79,6 +79,7 @@ services:
       - DATABASE_PASS=${TM_DB_PASS}
       - DATABASE_NAME=${TM_DB_NAME}
       - DATABASE_HOST=database
+      - DATABASE_SSLMODE=disable
```
JakobLichterfeld, line 82: Since this is the default value, I vote not to include it in the documentation at this location.

JakobLichterfeld (changes requested):
Thanks for your suggestion!

website/docs/guides/traefik.md
DrMichael: Good practise is to include the default action commented out
`#      - DATABASE_SSLMODE=disable  # defaults to disable
`

JakobLichterfeld:
Closing, as no answer from OP since 6 months

PR #2868, opened 2022-08-28
Added Vehicle Speed and Location to the Overview page

ddshd:
- Shows the vehicle speed based on the last speed from their position data point
- Shows the vehicle's last known position from the position data point

image

##

First PR into this project. Let me know if there are PR requirements somewhere, I couldn't find it.

### DIFF
--- a/grafana/dashboards/overview.json
+++ b/grafana/dashboards/overview.json
@@ -1646,6 +1646,18 @@
               "viz": false
             }
           },
+          "links": [
+            {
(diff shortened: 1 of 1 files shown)


Dulanic:
Going to close as I don't feel that this fits the overview dashboard. The overview dashboard has a lot of data over time, so it would be confusing to users to say MPH. MPH when? Also stale and needs to be rebased,

PR #2898, opened 2022-09-14
update korean translations

zipizigi:
Added missing korean translation.
Changed to Tesla official app translation (sentry mode)


adriankumpf (approved):
Thank you!

PR #2951, opened 2022-10-19
Correction of template error in home_assistant.md

shanbs:
Template tesla_elevation_ft contained an extra space after 3.2808 which caused the value cannot be calculated correctly.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -395,7 +395,7 @@ tesla_location:
       unit_of_measurement: ft
       icon_template: mdi:image-filter-hdr
       value_template: >
-       {{ (states('sensor.tesla_elevation') | float * 3.2808 ) | round(2) }}
+       {{ (states('sensor.tesla_elevation') | float * 3.2808) | round(2) }}
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/integrations/home_assistant.md
```diff
@@ -395,7 +395,7 @@ tesla_location:
       unit_of_measurement: ft
       icon_template: mdi:image-filter-hdr
       value_template: >
-       {{ (states('sensor.tesla_elevation') | float * 3.2808 ) | round(2) }}
+       {{ (states('sensor.tesla_elevation') | float * 3.2808) | round(2) }}
```
jonnylangefeld, line 398: This space doesn't matter in Jinja2. You can try this in your 'Developer Tools > Template'. Just paste these two lines there and you'll see it's the same outcome:
```
{{ ('1.5' | float * 3.2808) | round(2) }}
{{ ('1.5' | float * 3.2808 ) | round(2) }}
```

shanbs: I think you are right, I cannot reproduce it any more now, probably was due to something else.

JakobLichterfeld:
Closed as invalid

PR #3025, opened 2022-12-20
Docker compose command fixes

eden881:
Modern Docker installations come bundled with `docker-compose-plugin` and `Compose V2`. This replaces the old standalone `docker-compose` app and doesn't rely on external dependencies such as Python.

I think that Teslamate's documentation should reflect the updated best practice to use with Compose, which is `Compose V2`. This will also benefit beginners who aren't aware of the changes in Compose and won't understand why their commands aren't working out of the box.


eden881:
> > Modern Docker installations
> 
> Out of curiosity, what does "modern" mean? When was this introduced into Docker?

AFAIK, `Compose V2` was launched in 2021 and became generally available this year. It streamlines both the development and the usage of Compose - it allows the Docker devs to push features faster (since it's written in Go, same as Docker Engine), and it allows the user to use a single executable to control all of Docker's aspects.

More information can be found in [this Docker blog post](https://www.docker.com/blog/announcing-compose-v2-general-availability/).

baylanger:
Does this "Docker Compose V2" require a paid license? Seems it was introduced after the licensing for Docker changed. 

Until v2 gets adapted, I'd suggest we add a separate page/sections w/ the V2 instructions.

BowlesCR:
No change to the licensing -- this is available in the CE version as well.
As an example, the official guide for [installing Docker CE on Fedora](https://docs.docker.com/engine/install/fedora/) references it - note docker-compose-plugin. I've been using this for some months now and have enjoyed the improvements.
There's also a [compose-switch](https://github.com/docker/compose-switch) tool which is useful for making the old commands "just work" while you retrain muscle memory.

General reading: https://docs.docker.com/compose/compose-v2/

adriankumpf (approved):
I have postponed this change because there are many users of an older Docker version. There will be issues created on the topic for sure. But by now most people should be able to use `docker compose`. Let's see.

PR #3030, opened 2022-12-23
feat: Avg cost statistics in statistics dashboard

fmossott:
Adds 2 new columns to the statistics dashboard

Avg cost per kWh
Avg cost per km / mi

image


eden881:
Looks good, might be useful in the Drives dashboard too.

Kosta-Github:
I was about to suggest the exact same changes. 👍

PR #3061, opened 2023-01-14
Display VIN in Overview dashboard

natrlhy:
Display the VIN next to the `car_id` variable `name` so one can see the VIN somewhere in the various dashboards.


natrlhy:
Syntax fixes coming

PR #3062, opened 2023-01-14
Dispaly car VIN number in Overview dashboard

natrlhy:
Display the VIN next to the `car_id` variable `name` so one can see the VIN somewhere in the various dashboards.


Dulanic:
I feel this adds noise to the overview dashboard myself and not very useful. I don't know anyone who knows what VIN their car has without looking it up. If others disagree, bring up in discussions and we can discuss further.

fiveangle:
uh… okay, so it's nowhere displayed in Teslamate ?  So that was a waste of time so I'll guess I'll have to go login to the Tesla website, but seems pretty silly to not have it displayed *somewhere* :shrug: 

It's the only fool-proof method to identify a vehicle from another, and I thought I was crazy that I couldn't find it.

PR #3075, opened 2023-01-18
Update docs to manually remove vehicle

cwanja:
Entry added to document how to remove a vehicle from the database until a native solution is implemented. #3074 
#2719


adriankumpf (approved):
Thank you 🙌

Issue #3078, opened 2023-01-18
should send mqtt message if charger connected but no power


brianmay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

After a while charger cables stop working for me, and Tesla reports No Power. Would like to be able to monitor this condition so I can send alerts to get the problem fixed.

Which this happens, I believe the charge state has a `charging_state` field which is set to `"NoPower"`.

But unfortunately this value is not sent to MQTT - not that I can see anyway.

### Relevant log output

```shell
Nothing relevant. By default the charge_state value is not logged.
```

### Additional data

Not sure if a bug report is really relevant, but support request didn't seem to be relevant either.

I am thinking of attempting to create a pull request that will just publish the `charging_state` field to a new MQTT topic. But wanted to get feedback if people thought that was a good idea or not.

### Type of installation

Docker

### Version

1.27.2

EasternPA:
I ended up throwing something together with Node-RED for a buddy of mine. His discount kicks in at 10 pm, and he drives 40 miles/day with his SR+. I assembled a flow that pulls the geofence, charge_port_door_open, charger_actual_current, and usable_battery_level into flow variables. When he first bought the car, he wouldn't insert the charging handle far enough into the port to make a connection.

So on nights before a work day, the flow checks at 10:05 PM to see:

- if he's home
- if the charge port door is open
- if the charging voltage is 0, and
- if the SoC is below 60% 

If all those conditions are met, send him a text message warning him to check the car since it's not charging (and may be too low to get through the next day). It also emails him (and me for troubleshooting). I eventually had to shut it off, though. TM was late detecting that the charge had started by half an hour or more, resulting in several false alarms. He would send me screenshots of the warning from NR and Tessie showing it was charging just fine.

PR #3087, opened 2023-01-25
Add charging_state to VehicleSummary

brianmay:
Fixes: #3078

This will publish the state to MQTT and allow sensing of the `NoPower` state.

### DIFF
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -6,7 +6,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
   alias TeslaMate.Log.Car
 
   defstruct ~w(
-    car display_name state since healthy latitude longitude heading battery_level usable_battery_level
+    car display_name state since healthy latitude longitude heading battery_level charging_state usable_battery_level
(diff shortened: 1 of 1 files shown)


brianmay:
Notes:

1. Not tested yet. About to do that.
2. Wondering if I should add to `@always_published` which will publish `nil` values. But most likely reason for nil value is because `vehicle.charge_state` is `nil`, which means we don't know this value just yet, which means we shouldn't publish.

Actually thinking we shouldn't publish any `vehicle.charge_state` values if that value is `nil`, because it does mean we don't know what the value is. But fixing this beyond the scope of this simple pull request.

brianmay:
As far as I can tell the CI test failures are unrelated.

brianmay:
Preliminary testing seems to indicate it is working...

JakobLichterfeld:
> 1. Not tested yet. About to do that.

Did you test it in the meantime?

brianmay:
Yes, is tested and working. I see there is a conflict, will try to resolve that.

brianmay:
I rebased this against the latest master.

brianmay:
I think the test cases are not connected to this PR.

JakobLichterfeld (approved):
lgtm, failed CI is handled in #3428

PR #3123, opened 2023-02-14
Rearranged Overview Dashboard and added Trackmap

coreGreenberet:
I've rearranged the overview dashboard and added a trackmap.

I think this is now a better view especially as the state, charge level, and charging details are now the same size and stacked. so it is easier to follow between these items.

grafik


JakobLichterfeld:
This looks good on desktop but not on mobile devices, as the rows got stacked in one column and less needed values show up before more important ones imo.

PR #3126, opened 2023-02-18
Fix Korean translation TYPO

nanishin:
Found some weird&missed Korean translation status.
So it's a patch to fix Korean translation TYPO.


adriankumpf (approved):
Thank you!

PR #3136, opened 2023-02-28
Fix grafana datasource in  dashboards

DapengWei:
When grafana not set the Teslamate as the default datasource, the dashboards imported would not work.
So before import dashboards to an exist grafana, need to replace the datasource name with datasource uid.
I add a function  to fix the datasource uid in the script.


JakobLichterfeld:
I could not see the issue, compare https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards

swiffer:
I think it's fixed for now - Imports working fine for me with just having the DataSource specified.

If it's still an issue we should wait for a response in https://github.com/grafana/grafana/issues/77325 before merging that fixture functions. I can't imagine that's how Grafana approaches automated imports of DataSources & Dashboards.
We could instead also hardcode the UID for example...

Issue #2278, opened 2022-01-05
Database Timeout - drives not terminating


w0rd3r:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

With default DATABASE_TIMEOUT (https://docs.teslamate.org/docs/configuration/environment_variables/), drives are not terminating.
Increasing the DATABASE_TIMEOUT to 60000ms solved the issue for about 6 months.
Today, I doubled DATABASE_TIMEOUT again to be able to terminate drives.

The main issue with this is that I lose the first few minutes of every charge (before it terminates the drive)

### Expected Behavior

After a drive, Teslamate terminates the drive immediately.

### Steps To Reproduce

- drive the car
- put it in park
- It takes multiple minutes to register the end of the drive.

### Relevant log output

```shell
2022-01-05 10:21:55.542 car_id=1 [info] Start / :online
2022-01-05 10:21:56.642 car_id=1 [info] Connecting ...
2022-01-05 10:22:08.549 [info] Start of drive initiated by: %TeslaApi.Stream.Data{elevation: ****}
2022-01-05 10:22:08.550 car_id=1 [info] Driving / Start
2022-01-05 10:25:13.577 [info] GET https://owner-api.teslamotors.com/api/1/vehicles//vehicle_data -> 408 (9091.201 ms)
2022-01-05 10:25:13.577 [warn] TeslaApi.Error / %{"error" => "{\"error\": \"timeout\"}", "error_description" => "", "response" => nil}
2022-01-05 10:25:13.578 car_id=1 [error] Error / :unknown
2022-01-05 10:26:04.745 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/***/vehicle_data -> 408 (9074.469 ms)
2022-01-05 10:26:04.746 [warn] TeslaApi.Error / %{"error" => "{\"error\": \"timeout\"}", "error_description" => "", "response" => nil}
2022-01-05 10:26:04.746 car_id=1 [error] Error / :unknown
2022-01-05 10:26:09.322 car_id=1 [info] Received stream data: :inactive
2022-01-05 10:26:55.205 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/***/vehicle_data -> 408 (9095.879 ms)
2022-01-05 10:26:55.205 [warn] TeslaApi.Error / %{"error" => "{\"error\": \"timeout\"}", "error_description" => "", "response" => nil}
[...]

PR #3186, opened 2023-04-17
Add composite index to position : drive_id + date instead of drive_id

IlyaDiallo:
This is a proposed fix to the #2278 issue. On my setup with 5 million records in the position table, the close_drive function executes in under 7 seconds from 55 seconds before the index change.
An (untested) alternative fix would be to change the queries in the close_drive function like this:
```
    filtered_positions =
      from p in Position,
        where:
          p.drive_id == ^id and
            not is_nil(p.ideal_battery_range_km) and
            not is_nil(p.odometer)

    non_streamed_drive_data =
      from p in subquery(filtered_positions),
        select: %{
          start_ideal_range_km: first_value(p.ideal_battery_range_km) |> over(:w),
          end_ideal_range_km: last_value(p.ideal_battery_range_km) |> over(:w),
          start_rated_range_km: first_value(p.rated_battery_range_km) |> over(:w),
          end_rated_range_km: last_value(p.rated_battery_range_km) |> over(:w)
        },
        windows: [
          w: [
            order_by:
              fragment("? RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING", p.date)
          ]
        ],
        limit: 1
```
The idea here is to pre-filter to work with a subset of the positions table (drive_id = id) instead of the whole table.

### DIFF
--- a/priv/repo/migrations/20230417225712_composite_index_to_position.exs
+++ b/priv/repo/migrations/20230417225712_composite_index_to_position.exs
@@ -1,0 +1,8 @@
+defmodule TeslaMate.Repo.Migrations.AddCompositeIndexToPositions do
+  use Ecto.Migration
+
+  def change do
+    create index(:positions, [:drive_id, :date])
+    drop index(:positions, [:drive_id])
+  end
+end


adriankumpf (approved):
Good catch! Thanks.

PR #3225, opened 2023-06-05
Update drive-details.json missing dataSource for pressure units

OrangeBean:
Update drive-details.json missing dataSource for pressure units.

I have to fix this dashboard after every Grafana restart (upgrade) as it pull the latest of the dashboard and bring unfixed version down.


adriankumpf:
Thanks!

PR #3244, opened 2023-06-14
Updated readme, docs and Spanish translation

jheredianet:
Here are some contributions:
- Updated the readme file with links to missing screen shots
- Added missing Spanish translations in default.po file
- Updated some comments on projects using Teslamate


adriankumpf:
Thank you!

PR #3277, opened 2023-07-08
Fix FreeBSD install instructions

uqs:
- add missing shebang
- add optional mqtt username/password, so people can see how it needs to be spelled
- small markup fixes


DrMichael:
This appears to be obsolete?

Dulanic:
> This appears to be obsolete?

If so, why did it get merged? :) Was that an accident?

DrMichael:
I merged only

- add missing shebang
- add optional mqtt username/password, so people can see how it needs to be spelled
Appeared to make sense.

PR #3289, opened 2023-07-20
Update updates.json

Dulanic:
Fix for div by zero error, fixed before to range, fixed now for update dashboard. https://github.com/adriankumpf/teslamate/issues/1460#issuecomment-1585648601


adriankumpf:
Thank you!

PR #3294, opened 2023-07-25
Update charge-details.json

DrMichael:
Show location name in charge details.
As proposed in https://github.com/adriankumpf/teslamate/issues/3293


adriankumpf:
❤️

PR #3296, opened 2023-07-27
Update upgrading.mdx

jwilson-de:
Fixed a type-o with this.  changed from "docker compose pull" to "docker-compose pull"

### DIFF
--- a/website/docs/upgrading.mdx
+++ b/website/docs/upgrading.mdx
@@ -27,13 +27,13 @@ values={[
 Within the directory where the YML file is located, pull the new images:
 
 ```bash
-docker compose pull
+docker-compose pull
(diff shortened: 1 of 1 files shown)


andycui66:
Typo also exists in instruction page. https://github.com/adriankumpf/teslamate/blob/56a6aa3e7f3ca42f41e6e4d80ef61fcf05397b34/website/docs/installation/docker.md?plain=1#L87

IngmarStein:
This is not a typo, but Compose V2 https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2 says "the recommended command-line syntax is `docker compose`".

PR #3299, opened 2023-07-29
Support server_name_indication for ssl

simonsmh:
Tested on neon serverless postgress.
[Reference](https://neon.tech/docs/guides/elixir-ecto#usage-notes)


grafana/Dockerfile
```diff
@@ -1,28 +1,11 @@
-FROM grafana/grafana:9.5.3
+FROM grafana/grafana:latest
```
JakobLichterfeld, line 1: Currently we need to use specific grafana version, and as for all external dependencies pinning the version is best practice. So this is not gonna be merged.

brianmay: In general I think this is a bad idea, because you loose control over when the grafana updates happen. i.e. you make a minor innocent change, and could end up new version of grafana too, that has never been tested before.

Also: if the build system already has an old cached version of grafana/grafana:latest, this will be used instead. And that version might be an old unsupported version.

Which doesn't mean we shouldn't strive for the latest version, but might be better if this is a manual process.

JakobLichterfeld: Thanks for the more detailed explanation, why are we not gonna merge something like this

simonsmh: #3497

PR #3307, opened 2023-08-07
Switch to alexandra-trackmap plugin

pesintta:
This PR switches away from the dated pr0ps-trackmap-panel to alexandra-trackmap-panel plugin. The pr0ps-trackmap-panel is written in AngularJS which is deprecated in later versions of grafana and will cease to work in the future. Alexandra-trackmap-panel has instead a more modern stack (Typescript & React) which is recommended to be used in Grafana.

Teslamate's Grafana Map panels in Drive details, Charge details, Trip and Visited dashboards are converted to alexandra-trackmap with "sensible settings" that look nice and have decent performance.

In this PR alexandra-trackmap is configured with dark raster map background which goes along nicely with the default dark grafana theme. Other graphical features (such as ant-path to show the direction of the drive) are included as well.

Dark TrackMap

n.b. Plugin upstream has not tagged new releases for a while which is why this pull request links to a direct commit hash in the upstream repository.

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -4,7 +4,7 @@ ENV GF_ANALYTICS_REPORTING_ENABLED=FALSE \
     GF_AUTH_ANONYMOUS_ENABLED=false \
     GF_AUTH_BASIC_ENABLED=false \
     GF_PATHS_PLUGINS="/var/lib/grafana-plugins" \
-    GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS=natel-discrete-panel,pr0ps-trackmap-panel,panodata-map-panel,natel-plotly-panel \
+    GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS=natel-discrete-panel,alexandra-trackmap-panel,panodata-map-panel,natel-plotly-panel \
     GF_SECURITY_ADMIN_PASSWORD=admin \
     GF_SECURITY_ADMIN_USER=admin \
     GF_SECURITY_ALLOW_EMBEDDING=true \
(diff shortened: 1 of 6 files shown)


swiffer:
Would prefer using geomap in grafana 10 - that was Dependencies to 3rd Party Plugins is reduced likely resulting in less dependency conflicts in the Future

JakobLichterfeld:
I do see this conflicting with #3439

pesintta:
> I do see this conflicting with #3439

Yes, while I agree with the principle that built-in functionalities of grafana should be used where possible, I do wonder about their feature parity (as of today).

I haven't checked but I suspect that the drive graph in #3439 does not offer the "blue dot" on the map to highlight the position when hovering over other graphs in the dashboard which I find very useful and sought-after feature.
Can someone confirm this either way?

Also updates to the external plugin can probably be released faster than functionalities of built-in grafana visuals. I.e. with plugin there's no need to wait for grafana release cycles if some feature is needed.

That being said, I think the more static map components in dashboards (like charge locations, heatmap etc.) are probably better off with built-in grafana visuals.

JakobLichterfeld:
> I haven't checked but I suspect that the drive graph in #3439 does not offer the "blue dot" on the map to highlight the position when hovering over other graphs in the dashboard which I find very useful and sought-after feature.
> Can someone confirm this either way?

@oivindoh may you comment on this?

oivindoh:
@JakobLichterfeld The map marker is present with the built in visualisation as well.

I really see no reason to use plugins when the builtins offer the functionality we want (and actually offer maintained, timely and supported releases unlike the third party plugins). GeoMap has a rich set of features at this point.

JakobLichterfeld:
Perfect, thanks for feedback, so closing this in favor of #3439

PR #3311, opened 2023-08-13
Add LFP variants of Grafana dashboards

eden881:
For quite some time now, Tesla equips its lower-range vehicles with batteries that use lithium iron phosphate chemistry (also known as LFP). These batteries are [designed](https://driveteslacanada.ca/model-3/tesla-china-recommends-charging-new-lfp-battery-to-100/) to be regularly charged to 100% without the danger of accelerated degradation associated with NMC batteries.

However some dashboards in TeslaMate do not reflect this and display red and yellow indicators when the battery's SOC is high. I have created LFP variants of some dashboards so that owners of LFP-based vehicles can have a slightly better experience when viewing their data.

The added dashboards are:

- `Overview - LFP`: The battery level gauge has red and yellow zones only on the lower end. The maximum value of the charging kW gauge has been lowered to 170, as this is the [maximum](https://ev-database.org/car/1555/Tesla-Model-3#charge-table) DC charging potential for Tesla's LFP batteries.
- `Charging Stats - LFP`: The charge stats table remains green even for entries higher than 80% and 95%.
- `Battery Health - LFP`: The current SOC bar has a red zone only on the lower end.


doronkg:
Looks good @eden881!
Very important, I'd like to see these new features.

zknobody:
Sorry for the beginner question.
How do I check this out ? My docker instance has the latest build but I don't see these dashboards.

Dulanic:
> Sorry for the beginner question. How do I check this out ? My docker instance has the latest build but I don't see these dashboards.

Copy the proposed file from the source and import dashboard. Its in JSON format.

NirKli:
> I'm also torn on this, this may be extra noise. I think a long term option we could possiblly use variables to make the existing dashboards change if it is LFP. We should also know if it is LFP from the API, so it could be automatic in that way also. I'm not sure just adding more dashboards is the "cleanest" route .

I think in the same way, I'll try to work on a solution.

JakobLichterfeld:
Would you like to see this merged and open a seperate PR for a variable based solution?

eden881:
> Would you like to see this merged and open a seperate PR for a variable based solution?

I think so. I also think that a variable based solution is the correct way to go, but I believe it's better to have those dashboards merged in now as they solve the issue, and maybe remove them when a variable is integrated into the main dashboards.

JakobLichterfeld:
I added #3438 to track the variable handling idea

PR #3328, opened 2023-09-01
Support listening on unix domain socket

PhracturedBlue:
This patch allows listening on  a unix-domain-socket instead of via TCP.  This is useful when using rootless containers without slirp4netns by exposing the unix socket through the host rather than mapping unix ports.  In my case, it allows me to run a reverse-proxy with --network=none as it only speaks to unix sockets, and listens on a systemd socket-activated socket.

To use, specify:
 * HTTP_BINDING_ADDRESS=<absolute path to binding socket>
 * ~PORT=0~

PR #3331, opened 2023-09-02
Improved support for listening on unix domain socket

PhracturedBlue:
This is an alternative to PR #3328
It automatically sets port=0 when unix domain sockets are enabled.  It will also automatically remove the socket before (re)creating it, and provides the ability to override the permissions on the socket.

Those last 2 capabilities require using Ranch >= 2.1, which is supported by cowboy, but potentially introduces additional risk, and so I'm providing both options as far as the PR goes.

To enable unix domain sockets, set (for example):
 * HTTP_BINDING_ADDRESS=/socket/sock
 * SOCKET_PERM=666


JakobLichterfeld:
#3328 was merged

PR #3338, opened 2023-09-08
Fix divide by zero in updates dashboard

tacotran:
Same issue as in https://github.com/adriankumpf/teslamate/issues/3092 on another dashboard. Divide by zero occurs if car has ever reached 0% SOC.


tacotran:
Already a PR for this.

PR #3342, opened 2023-09-12
Show all vehicles on drive stats dashboard

yoyostile:
Only one Tesla is shown on the "Drive stats" dashboard. This is a simple fix to show each and every tesla on this dashboard. 
_Edit: same for efficiency and projected range dashboards_

Before:
image

After:
image


adriankumpf:
Thank you!

PR #3348, opened 2023-09-19
Update docker.md

yanggs07:
docker no compose direction?


tobiasehlert:
@yanggs07, it's supposed to be `docker compose` and not `docker-compose` with newer versions of docker.
You might consider updating your binaries :)

PR #3355, opened 2023-10-02
Update faq.md

sebmatton:
Added "Tesla API Token Generator" Python script. Local tool guaranteeing privacy of data, as I don't like using third party which could steal my secrets.


Dulanic:
Closing as no control over the linked repo. If wish to submit a PR /w the python code, that would be different.

PR #3356, opened 2023-10-03
Upgrade esbuild devDependency as it clashes with a dependency version

jnsaff:
Asset builds failed for me as a peer dependency was referring to version `^0.19.4`. Upgrading to this fixed the issue.


JakobLichterfeld:
Closed, as #3477 upgraded node to 20

PR #3368, opened 2023-10-17
Use preferred tile.openstreetmap.org URL

danieldegroot2:
I would like to suggest using the now preferred https://tile.openstreetmap.org URL instead of the current one, see

https://github.com/openstreetmap/operations/issues/737

cc: @adriankumpf


adriankumpf:
Thank you!

PR #3376, opened 2023-10-23
Cleanup Grafana Docker Container

swiffer:
This removed several Plugins based on AngularJS as alternatives are available (geomap (in core) and nline-plotlyjs-panel)

This requires migrating some Dashboards over to the newer Panels. Happy to contribute if there is interest in catching up with latest grafana development


swiffer:
closed in favor of https://github.com/adriankumpf/teslamate/pull/3382

PR #3380, opened 2023-10-26
Fix: broken links when orgId is not 1

jlestel:
The orgId=1 is hardcoded in some links. URL fails when my user's organization is > 1.


adriankumpf:
Thank you!

jlestel:
thank you especially to you!

PR #3382, opened 2023-10-27
Migrate dashboards to Geomap, XYChart visualizations, remove need for plugins

oivindoh:
- Use grafana's built-in Geomap visualization to plot routes, charge points and charging heat map
- Plotly plugin replaced with XYChart
- Update Grafana to 10.2.0
- Clean up plugin references

### DIFF
--- a/grafana/dashboards/internal/charge-details.json
+++ b/grafana/dashboards/internal/charge-details.json
@@ -826,127 +826,11 @@
       },
       "id": 11,
       "links": [],
-      "pconfig": {
-        "fixScale": "",
-        "layout": {
-          "dragmode": "pan",
-          "font": {
-            "family": "\"Open Sans\", Helvetica, Arial, sans-serif"
-          },
-          "hovermode": "closest",
-          "legend": {
-            "orientation": "h"
-          },
-          "showlegend": false,
-          "xaxis": {
-            "rangemode": "normal",
-            "showgrid": true,
-            "title": "SOC [%]",
-            "type": "linear",
-            "zeroline": false
-          },
-          "yaxis": {
-            "rangemode": "normal",
-            "showgrid": true,
-            "title": "kW",
-            "type": "linear",
-            "zeroline": false
-          },
-          "zaxis": {
-            "rangemode": "normal",
-            "showgrid": true,
-            "type": "linear",
-            "zeroline": false
-          }
-        },
-        "loadFromCDN": false,
-        "settings": {
-          "displayModeBar": false,
-          "type": "scatter"
-        },
-        "showAnnotations": true,
-        "traces": [
-          {
-            "mapping": {
-              "color": "Power [kW]",
-              "x": "SOC [%]",
-              "y": "Power [kW]"
-            },
-            "name": "Power",
-            "settings": {
-              "color_option": "solid",
-              "line": {
-                "color": "#005f81",
-                "dash": "solid",
-                "shape": "linear",
-                "width": 6
-              },
-              "marker": {
-                "color": "#8AB8FF",
-                "colorscale": "YlOrRd",
-                "line": {
-                  "color": "#DDD",
-                  "width": 0
-                },
-                "showscale": false,
-                "size": 5,
-                "sizemin": 3,
-                "sizemode": "diameter",
-                "sizeref": 0.2,
-                "symbol": "circle"
-              },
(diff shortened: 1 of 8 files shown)


swiffer:
nice work! did that in my instance as well - great to see it getting upstreamed. changes are looking great for me!

**charging-stats** is currently using markers instead of heatmap visualization - i couldn't get heatmap visualization to zoom in correctly. was it the same for you? for the dc charging curve average i went with an override and style "both" to be closer on the old visualization

grafik
grafik

swiffer:
**charge-details** is no longer working when opened directly (works just fine when opened via **charges**

grafana/dashboards/internal/charge-details.json
```diff
@@ -651,31 +651,131 @@
...
         {
           "alias": "",
           "datasource": "TeslaMate",
+          "editorMode": "code",
           "format": "time_series",
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT\n\t$__time(date),\n\tunnest(ARRAY[latitude, latitude]) AS latitude\nFROM\n\tcharging_processes c\n\tJOIN positions p ON c.position_id = p.id\nWHERE\n\t$__timeFilter(date)\n\tAND c.car_id = $car_id;",
+          "rawSql": "SELECT\n\t$__time(date),\n\tunnest(ARRAY[latitude, latitude]) AS latitude,\n\tunnest(ARRAY[longitude, longitude]) AS longitude\nFROM\n\tcharging_processes c\n\tJOIN positions p ON c.position_id = p.id\nWHERE\n\t$__timeFilter(date)\n\tAND c.car_id = $car_id;",
```
swiffer, line 778: why is an array being build here and then unnested (generates multiple data rows?)

grafik

oivindoh: I really wasn't sure about this query - I just wanted to combine A/B into one as I couldn't understand why we did one for lat and one for long

oivindoh:
> nice work! did that in my instance as well - great to see it getting upstreamed. changes are looking great for me!
> 
> **charging-stats** is currently using markers instead of heatmap visualization - i couldn't get heatmap visualization to zoom in correctly. was it the same for you? for the dc charging curve average i went with an per series override and style "both" to be closer on the old visualization
> 
> grafik grafik

Did not have much luck getting heatmap to work

> **charge-details** is no longer working when opened directly (works just fine when opened via **charges** - before it had shown the latest charge on dashboard load

Are you sure you don't have a modified or older version of that dash than what's on master? If I use the master version of the dash it's full of N/A on load.

swiffer:
we might need to downgrade to 10.1.2 if we would like to continue supporting armv7 for now.

https://grafana.com/blog/2023/09/29/grafana-and-grafana-enterprise-updates-for-armv6-and-armv7-will-be-temporarily-paused/

DrMichael:
I cannot really see the latest version of Grafana with armv7?

swiffer:
When Picking different Versions in there Version Picker you'll notice that when selecting 10.1.2 an armv7 Version is shown

DrMichael:
Right. But we're on 9.5.10 still anyway...

swiffer:
This PR is upgrading to 10.2.0

Dulanic:
I agree with the idea that we need to move to Geomap or something as the other plugin is in the process of being deprecated. However, I do have some concerns. This didn't look very good on my test, The N/A fields, lack of title bars etc... And from the queries, it looks like they are coming from a specific charge, but how it that being passed? And what is the charging curve supposed to indicate? Should we split up some of this to different PRs? 

Any PR's like this need to also show the changes made and what they should look like. I may have screwed something up when i just imported the dashboard, but I have no idea as I don't know what it should look like.

image

oivindoh:
You're using charge-details dash without any info about which charge to show - you need to call it from the charges dashboard link or supply the charge details manually (charging process id, timestamps, car id)

Dulanic:
> You're using charge-details dash without any info about which charge to show - you need to call it from the charges dashboard link or supply the charge details manually (charging process id, timestamps, car id)

Wow how did I screw that up! LOL It had been a bt since I worked on some of this.

oivindoh:
Closing this in favour of https://github.com/teslamate-org/teslamate/pull/3439

PR #3373, opened 2023-10-20
Add endpoints parameter

jsight:

PR #3398, opened 2023-11-11
Endpoints Query Param, Docker 60s delay fixed, Typo fixed, Update several Dependencies

swiffer:
endpoints fix taken from #3373 but modified for better readability / code style
also fix 60s build time penality (caused by using a deprecated install script)
-> https://github.com/nodesource/distributions?tab=readme-ov-file#new-update-%EF%B8%8F
and fixed typo in html view...


lib/teslamate_web/live/charge_live/cost.html.heex
```diff
@@ -1,7 +1,7 @@
 <nav class="breadcrumb" aria-label="breadcrumbs">
   <ul>
-    <li><%= live_redirect gettext("Home"), to: "/" %></li>
-    <li class="is-active"><%= live_redirect gettext("Charge Cost"), to: "" %></li>
+    <li><%= live_redirect(gettext("Home"), to: "/") %></li>
+    <li class="is-active"><%= live_redirect(gettext("Charge Cost"), to: "") %></li>
```
parkr, line 4: These changes (adding parens) seems like a personal preference so I'd revert them since they don't add functionality and it looks like Adrian has consistently chosen to omit them.

parkr (commented):
I'd recommend reverting all non-functional changes (i.e. revert code style changes) so the PR is easier to review.

swiffer:
yes, experimenting a bit - Work in Progress... I will split Changes into multiple PRs tomorrow

lib/teslamate_web/live/charge_live/cost.html.heex
swiffer: skipped the Phoenix update for now - not experienced enough in Elixir to do that...

adriankumpf:
Thank you!

djpalmis:
Can you quickly explain me what is this "Docker 60s delay"?
I ask this because after upgrading to 1.27.4 (or maybe it was already in this way?) I see a delay between app and tesmate. Is this something we wanted?

swiffer:
The way Node JS was installed within the container is deprecated.

https://github.com/nodesource/distributions?tab=readme-ov-file#new-update-%EF%B8%8F

Nodesource added a 60 Seconds sleep duration to make users aware of the deprecation unneccesarrily increasing built time duration. This has no effect on application behvaior.

djpalmis:
oh.. :) Sorry! Thank You so much!

Issue #3158, opened 2023-03-24
TeslaFi data import failure


sirismile:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

not able to import the data after following the steps in the documentation 

### Expected Behavior

not able to import the data after follow the steps in the documentation 
the log says data incomplete, but dont know where to debug. 

apologies, im new to linux and ubuntu. 

my export file from teslafi does indicate a lot of empty spaces. but dont know if its intentionally done like that?! 
i looked one of the previous issue that uploaded file, and mine looks much simpler. 

### Relevant log output

```shell
teslamate_1  | 2023-03-24 01:15:35.695 [info] Importing 1 file(s) ...
teslamate_1  | 2023-03-24 01:15:44.540 [error] GenStateMachine TeslaMate.Import terminating
teslamate_1  | ** (RuntimeError) vehicle data is incomplete
teslamate_1  |     (teslamate 1.27.2) lib/teslamate/import.ex:274: TeslaMate.Import.create_car/1
teslamate_1  |     (teslamate 1.27.2) lib/teslamate/import.ex:128: TeslaMate.Import.handle_event/4
teslamate_1  |     (stdlib 3.17.2.1) gen_statem.erl:1205: :gen_statem.loop_state_callback/11
teslamate_1  |     (stdlib 3.17.2.1) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```

### Screenshots

not able to import the data after follow the steps in the documentation 
the log says data incomplete, but dont know where to debug. 

apologies, im new to linux and ubuntu. 

my export file from teslafi does indicate a lot of empty spaces. but dont know if its intentionally done like that?! 

### Type of installation

Docker

### Version

3

MagoKimbra:
Same problem, I tried several times today, import I follow all the guide, but nothing imported data...

Zjaaspoer:
Agree, tried several exports & imports from TeslaFi into TeslaMate, same error every time. Running the latest docker of TeslaMate

Kjehlje:
+1, same problem here. 
Same log output.

Etus12:
+1, same problem here.
Same log output.

augustcove:
+1, same problem and log output.

vzhivkov:
+1 same problem happened to me yesterday.

ironfistperez:
Same issue here. Any idea if this will ever get fixed?  Got months of data in Teslafi that would be nice to move over.
[...]

PR #3431, opened 2023-11-14
Update TeslaFi importer to accept newer export formats

ithinuel:
I am not sure if using "1" as the default id but that's what's suggested in https://github.com/teslamate-org/teslamate/issues/3158#issuecomment-1721240983 and has worked so far.


ithinuel:
Is the ci failure related to the changes in this PR?

JakobLichterfeld:
> Is the ci failure related to the changes in this PR?

As far as I can see: no.

lib/teslamate/import/line_parser.ex
```diff
@@ -76,6 +76,9 @@ defmodule TeslaMate.Import.LineParser do
       {"id", _val} ->
         Map.put(acc, "id", :rand.uniform(65536))
 
+      {"vehicle_id", ""} ->
+        Map.put(acc, "vehicle_id", "1")
```
JakobLichterfeld, line 80: hard coding vehicle id will not be merged

JakobLichterfeld (changes requested):

lib/teslamate/import/line_parser.ex
ithinuel: Is there any acceptable alternative? This is used in lieu of a default value.

As mentioned in the description, this what is recommended/suggested in the linked comment.

JakobLichterfeld: Proper format input file with correct vehicle id, as suggested in the linked comment. Hardcoding vehicle id to import to 1 will result in overwriting old data if that vehicle id exists before.

ithinuel: > Proper format input file with correct vehicle id, as suggested in the linked comment. 

Isn’t a "proper format input file" for this TeslaFi Importer what is directly exported from TeslaFi without manual modifications?

> Hardcoding vehicle id to import to 1 will result in overwriting old data if that vehicle id exists before.

Doesn’t the current code ignores any input data if there is already some dated before or on the day of the current entry/event?
At worse, using 1 as the default value would add data to the wrong vehicle, but should not overwrite anything.

ithinuel: @JakobLichterfeld should this use `:rand.uniform(65536)` like the `id` field?

JakobLichterfeld: No, as the vehicle id is given by tesla

ithinuel: I don’t know about tesla’s API, but the TeslaFi exports do not report it.

ithinuel:
It is not stale.

To summarize the pending thread:

- @JakobLichterfeld does not want to see a hard codded value for the `vehicle_id`
- This `vehicle_id` is not exported in `TeslaFi`’s data exports.
- In the context of the TeslaFi importer, the input data come from TeslaFi so whether the `vehicle_id` is present or not in Tesla’s own api is irrelevant
- As far as I understand, Teslamate relies on the JSON API’s id, not the `vehicle_id`.

So, is this field essential for the purpose of TeslaMate? Could it be made optionnal and left empty by the importer?

ithinuel:
Still waiting far input from reviewer.

lib/teslamate/import/line_parser.ex
jlestel: @ithinuel Having imported many TeslaFi's, I confirm that it is tedious to fill in the vehicle_id column in each file.

This vehicle_id corresponds to the car_id of the Teslamate "cars" table.

@JakobLichterfeld Is it OK to use a new environment variable to specify the car ID for which TeslaFi data will be imported?

I should ends with something like that: 
```
      {"vehicle_id", ""} ->
        Map.put(acc, "vehicle_id", System.get_env("TESLAFI_IMPORT_VEHICLE_ID", "1"))
```

ithinuel: @jlestel  I (finally) applied your suggestion, and rebased on master.

It looks like CI is happy so the PR is ready for review again :)

jlestel (approved):
LGTM, it's much faster than adding the value in x large csv files by hand...

It might be useful to mention the `TESLAFI_IMPORT_VEHICLE_ID` in the documentation to show that we can use this env variable to choose the vehicle id on which to import the data (may be in the Requirements section of this file `website/docs/import/teslafi.md`)

JakobLichterfeld (changes requested):
Thanks. Lgtm. Please also add a description about the env file in the doc, under Environment Variables and additional mention it on the doc page for tesla fi import

website/docs/configuration/environment_variables.md
```diff
@@ -37,3 +37,4 @@ TeslaMate accepts the following environment variables for runtime configuration:
...
 | **TZ**                            | Used to establish the local time zone, e.g. to use the local time in logs. See [List of tz database time zones](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).                                                                                                                                   |                               |
 | **DEFAULT_GEOFENCE**              | The default GEOFENCE to send via GEOFENCE if car not in geofence. Overrides the default of "" which will delete any retained value.                                                                                                                                                                              | "" (no quotes)                |
+| **TESLAFI_IMPORT_VEHICLE_ID**     | The default Vehicle ID used when importing from TeslaFi. If not defined and the import source is missing that datum, 1 will be used.
```
JakobLichterfeld, line 40: datum -> If it is not defined and the import source does not contain the data, 1 is used.

JakobLichterfeld (changes requested):

JakobLichterfeld (approved):

JakobLichterfeld:
Thank you @ithinuel!

PR #3432, opened 2023-11-15
[CI] Build PR images on GHCR

jlestel:
I modified the CI so that the PR images are build then pushed into the GitHub packages (GHCR).

This does not interfere with how current CI works with Docker Hub.

Functioning :
- When a PR is created or modified: the image is built on the GHCR from fork repo.
- When a branch is pushed: the image is built on the GHCR from main repo.
- When a PR is completed: the image is deleted in the fork repo. Additionally, untagged images are also deleted on this fork repo.

Once this PR is merged, each new / updated PR will trigger CI on the fork repo.
- Users will then see all these images on their fork. Eg: [https://github.com/jlestel/teslamate/pkgs/container/teslamate](https://github.com/jlestel/teslamate/pkgs/container/teslamate)

If necessary, anyone can then modify their docker-compose to use a PR image (eg: `ghcr.io/jlestel/teslamate:master`) for testing this PR instead of `teslamate/teslamate:latest`.


jlestel:
> Thank you for this PR. Can you elaborate on why you chose to modify the Elixir CI rather than create your own workflow?

No reason, I just separated them.

Now I trigger the GHCR build on a pull_request or push master in a separate workflow.

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,57 @@
+name: Build GHCR images
+
+on:
+  push:
+    branches: 
+      - "master"
+  pull_request:
+    branches: [ "master" ]
+
```
JakobLichterfeld, line 9: I would add a `workflow_dispatch:`as well

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,57 @@
...
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
```
JakobLichterfeld, line 19: For maintenance reasons, I would extract the buildx to a separate action since we now use it in different places.

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,57 @@
...
+            runs_on: "ubuntu-latest"
+            cache_id: amd64
+          - platform: "linux/arm/v7"
+            runs_on: "buildjet-2vcpu-ubuntu-2204-arm"
+            cache_id: arm
+          - platform: "linux/arm64"
+            runs_on: "buildjet-2vcpu-ubuntu-2204-arm"
+            cache_id: arm64
+    runs-on: ${{ matrix.runs_on }}
+    steps:
```
JakobLichterfeld, line 35: I would add a timeout to prevent stuck jobs

JakobLichterfeld (changes requested):
lgtm, my findings see inline comments

.github/workflows/ghcr_build.yml
jlestel: Added

.github/workflows/ghcr_build.yml
jlestel: Done

DrMichael:
> If necessary, anyone can then modify their docker-compose to use a PR image (eg: `ghcr.io/teslamate-org/teslamate:pr-123`) for testing instead of `teslamate/ teslamate:latest`

Would that also be the case for grafana?

.github/workflows/ghcr_build.yml
jlestel: I tried to refactor things. Let me know what you think about.

And I believe the Docker Hub user has changed since my initial commit.

jlestel:
> > If necessary, anyone can then modify their docker-compose to use a PR image (eg: `ghcr.io/teslamate-org/teslamate:pr-123`) for testing instead of `teslamate/ teslamate:latest`
> 
> Would that also be the case for grafana?

yes I added grafana also

adriankumpf:
Could this be used by a malicious PR to leak build secrets? Or does building the images require maintainer approval?

brianmay:
I don't like to be the pessimist, but wondering if this will work. "With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository." https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

My experience, not even jobs triggered by dependabot created pull requests have access to secrets. I had to update all of my github workflows as a result.

But nothing wrong with trying it. :-)

Also see: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

JakobLichterfeld:
> Could this be used by a malicious PR to leak build secrets? Or does building the images require maintainer approval?

Good point you raise. As far as my brief research shows, you can extract secrets using malicious github workflows and actions. Building a docker image with a verified pipeline should not be an issue, but PRs related to .github should be excluded from auto-run. But this is nothing introduced by this PR, but more of a general issue in my opinion.

.github/actions/build/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          mkdir -p /tmp/teslamate/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "/tmp/teslamate/digests/${digest#sha256:}"
+          echo ${{ env.DOCKER_METADATA_OUTPUT_VERSION }} > /tmp/teslamate/version
+      - name: Upload digest
+        uses: actions/upload-artifact@v3
+        with:
+          name: digests
+          path: /tmp/teslamate/*
```
JakobLichterfeld, line 62: It should be /tmp/teslamate/digests/*

.github/actions/grafana/action.yml
```diff
@@ -0,0 +1,37 @@
...
+    description: 'Tags'
+    required: false
+    default: ''
+  labels:
+    description: 'Labels'
+    required: true
+  image:
+    description: 'Image target'
+    required: true
+    default: 'myteslamate/grafana'
```
JakobLichterfeld, line 14: I would remove the myteslamate :-)

.github/actions/grafana/action.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ inputs.image }}
+        tags: ${{ inputs.tags }}
+        labels: ${{ inputs.labels }}
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+    - name: Build and push
```
JakobLichterfeld, line 29: Can we really do without it?

```
 - name: Login to DockerHub
        uses: docker/login-action@v3.0.0
        with:
          username: teslamate
          password: ${{ secrets.DOCKER_PASSWORD }}
```

.github/actions/merge/action.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Merge'
+description: 'Merge images'
+runs:
+  using: "composite"
+  steps:
+    - name: Download digests
+      uses: actions/download-artifact@v3
+      with:
+        name: digests
+        path: /tmp/teslamate
```
JakobLichterfeld, line 10: Shouldn't it be /tmp/teslamate/digests?

.github/actions/merge/action.yml
```diff
@@ -0,0 +1,12 @@
...
+    - name: Set up Docker Buildx
```
JakobLichterfeld, line 11: Separation of concerns: Is setup buildx a merge action?

JakobLichterfeld (changes requested):
hank you for your dedicated work! I really like the process of refactoring into separate actions. I added my findings as inline comments.

JakobLichterfeld (commented):
lgtm

.github/actions/build/action.yaml
jlestel: I finished to refactoring. We finally stay on /tmp/digest.

.github/actions/grafana/action.yml
jlestel: It seems to work (I do it in the workflow before this action)

.github/actions/merge/action.yml
jlestel: Reverted

jlestel:
> Could this be used by a malicious PR to leak build secrets? Or does building the images require maintainer approval?

Yes there was a flaw there.

Now, I prevent workflows from being launched when there are changes to the .github directory: only a maintainer can run it manually.

jlestel:
> I don't like to be the pessimist, but wondering if this will work. "With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository." https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow
> 
> My experience, not even jobs triggered by dependabot created pull requests have access to secrets. I had to update all of my github workflows as a result.
> 
> But nothing wrong with trying it. :-)
> 
> Also see: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

You're right. In any case, secrets.GITHUB_TOKEN and secrets.DOCKER_PASSWORD will not be accessible to PRs from external repos... That seems normal to me too: it is rather work branches or release candidates whose content has been qualified by maintainers that could be proposed.

This involves modifying the gitflow to no longer use only a master branch and be able to create GH images on these new branches.

Maybe you already have ideas on a gitflow that could make sense?

jlestel:
> The good news: it should work with secrets.GITHUB_TOKEN and this is the only one we should use.

As you can see [in this action](https://github.com/teslamate-org/teslamate/actions/runs/6891889827/job/18747960155), the GITHUB_TOKEN is in read only.

This seems linked to the [Maximum access for pull requests from public forked repositories](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) which limit write access even if permissive mode is on.

So, i'll try to replace pull_request by pull_request_target to only run an action in the repo of the requester of a PR.

jlestel:
> So, i'll try to replace pull_request by pull_request_target to only run an action in the repo of the requester of a PR.

Voilà @adriankumpf @JakobLichterfeld : the GHCR build workflow is now based on the PR source user. The CI is played on the fork repo  instead of the teslamate-org repo (and the generated images too).

Eg for this PR, you'll find image(s) on my fork:
- TeslaMate image: https://github.com/jlestel/teslamate/pkgs/container/teslamate
- Grafana image: https://github.com/jlestel/teslamate/pkgs/container/teslamate

It build only amd64 as others need a thrid party not everyone has install.

Let me know what you think about that.

.github/workflows/ghcr_build.yml
```diff
@@ -79,6 +79,13 @@ jobs:
         shell: bash
```
JakobLichterfeld: Please remove your debug code. It may look suspicious to a casual observer.

.github/actions/merge/action.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: docker/setup-buildx-action@v3
+
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ inputs.image }}
+        tags: ${{ inputs.tags }}
+
```
JakobLichterfeld, line 30: What about a Login to the docker hub?
```      - name: Login to Docker Hub
        uses: docker/login-action@v3.0.0
```

.github/workflows/buildx.yml
```diff
@@ -31,118 +31,54 @@ jobs:
...
     needs:
       - teslamate_build
-    steps:
-      - name: Download digests
-        uses: actions/download-artifact@v3
+    steps:    
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Login to Docker Hub
```
JakobLichterfeld, line 55: I see this as a requirement inside the merge action

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,85 @@
...
+  teslamate_build:
+    name: Build images
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: "linux/amd64"
+            runs_on: "ubuntu-latest"
+            cache_id: amd64
+#          - platform: "linux/arm/v7"
```
JakobLichterfeld, line 26: please enable all platforms

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,85 @@
...
+#            runs_on: "buildjet-2vcpu-ubuntu-2204-arm"
+#            cache_id: arm
+#          - platform: "linux/arm64"
```
JakobLichterfeld, line 29: please enable all platforms

.github/workflows/buildx.yml
JakobLichterfeld: Forget it, I've got it now. Is outsourced to separate ghcr and docker hub push.

.github/actions/merge/action.yml
JakobLichterfeld: Forget it, I've got it now. Is outsourced to separate ghcr and docker hub

JakobLichterfeld (changes requested):
Thanks again for refactoring and your dedicated work. 

I added once again my small findings.

Additionally, two general points still prevent me from merging:
- I would vote to include ```- '!.github/**'  # Important: Exclude PRs related to .github from auto-run``` in all workflows
- Please lint the files, as there are some trailing whitespaces for example

.github/workflows/ghcr_build.yml
jlestel: Outdated: was in WIP at this moment

.github/workflows/ghcr_build.yml
```diff
@@ -0,0 +1,94 @@
+name: Build GHCR images
+
+on:
+  workflow_dispatch:
+  push:  
+    branches: 
+      - "master"      
+  pull_request_target:
+    branches: [ "master" ]
+#    paths:
```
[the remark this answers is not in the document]
jlestel: When I was testing. I re-added it.

.github/workflows/ghcr_build.yml
jlestel: done. forkers who want to build will have to use the buildjet service

JakobLichterfeld (approved):
lgtm

JakobLichterfeld:
Merging on mobile in the middle of the night without testing in our repo was not the best idea. I suggest opening a CI branch, including the branch in the workflow triggers, and testing it in our repo, as local testing of github workflows is limited in terms of secrets and github usernames.

jlestel:
> @jlestel Sorry, merging on mobile in the middle of the night without testing in our repo was not the best idea. I suggest opening a CI branch, including the branch in the workflow triggers, and testing it in our repo, as local testing of GitHub workflows is limited in terms of secrets and GitHub usernames. The issue lies within the login to pull buildx.

No problem. Yes it will be better to test it with a dedicated branch as well as on our repositories. Keep me informed.

PR #3439, opened 2023-11-16
Use built-in grafana visuals instead of plugins

oivindoh:
(Supersedes https://github.com/teslamate-org/teslamate/pull/3382, this is the dashboard update only to make it more palatable)

The plugins currently used by TeslaMate dashboards are stale and will probably not work at all in later Grafana releases.

Replace them with built-in options that are now available:

- Use grafana's built-in Geomap visualization to plot routes, charge points and charging heat map
- Plotly plugin replaced with XYChart

Tested with Grafana 10.2, but should work with the current grafana docker image from this repo as well (though routes in geomap only graduate to beta in 10.1x)

### Trip

Screenshot 2023-11-17 at 18 26 05

### Visited

Screenshot 2023-11-17 at 18 23 41

### Drive details

Screenshot 2023-11-17 at 18 17 39

### Charge details

Screenshot 2023-11-17 at 18 53 42

### Battery health

Screenshot 2023-11-17 at 15 08 51

### DIFF
--- a/grafana/dashboards/battery-health-lfp.json
+++ b/grafana/dashboards/battery-health-lfp.json
@@ -21,12 +21,6 @@
     },
     {
       "type": "panel",
-      "id": "natel-plotly-panel",
-      "name": "Plotly",
-      "version": "0.0.7"
-    },
-    {
-      "type": "panel",
       "id": "piechart",
       "name": "Pie chart",
       "version": ""
@@ -1196,9 +1190,70 @@
       "type": "bargauge"
     },
     {
-      "datasource": {
-        "type": "postgres",
-        "uid": "TeslaMate"
+      "datasource": "TeslaMate",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisBorderShow": false,
+            "axisCenteredZero": false,
+            "axisColorMode": "text",
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "pointSize": {
+              "fixed": 6
+            },
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "show": "points"
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          }
+        },
+        "overrides": [
+          {
+            "matcher": {
+              "id": "byFrameRefID",
+              "options": "Median"
+            },
+            "properties": [
+              {
+                "id": "custom.lineWidth",
+                "value": 2
+              },
+              {
+                "id": "custom.show",
+                "value": "lines"
+              },
+              {
+                "id": "custom.lineStyle",
+                "value": {
+                  "fill": "solid"
+                }
+              }
+            ]
+          }
+        ]
       },
       "gridPos": {
         "h": 8,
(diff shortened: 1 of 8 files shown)


oivindoh:
Setting WIP because I forgot to include the XY changeset in this PR

DrMichael:
I imported the dasboard into the current Grafana installation. Some remarks regarding the maps:

- Trip: the route is hard to see with OpenStreetMap (routs are also red in Germany), size of 5 makes it better. Or XYZ Tile Layer.
- Visited: took ages for four years on RPi3, but probably a matter of the query, looks good
- Charge Details: the dot in the map is hard to see. Why are there two layers, when you hover the dot?
- Drive Details: the route is hard to see, perhaps better with size 10? (I'd like red indeed better, then size 5 would be fine :-)

All appears to work with the current Grafana version. Great Work!

oivindoh:
@DrMichael I agree on the route colour. I just switched it to blue to try to satisfy demand for a "blue dot", but I think a red dot is fine along with a red route.

The default layer makes the route much more obvious in any case, but I fear it's a bit too devoid of info to use.

I'll finish this up with your suggestions taken into account and add updated screenshots later today.

DrMichael:
> The default layer makes the route much more obvious in any case, but I fear it's a bit too devoid of info to use.

I agree, the default layer is not good in this case.

> I'll finish this up with your suggestions taken into account and add updated screenshots later today.

Too bad, you cannot see the routes in Germany...

oivindoh:
Updated the dashboards with red + wider route lines (3) now.

@DrMichael @JakobLichterfeld what do you feel about OSM layer vs Default?

Trip dash with default for comparison:

Screenshot 2023-11-17 at 18 27 37

DrMichael:
> @DrMichael @JakobLichterfeld what do you feel about OSM layer vs Default? Default feels a lot easier on the eyes, but you also lose quite a bit of information.

Ah, default does not look good. What about the XYZ Tile Layer. It not so visually opulent as OSM, but still has details.

DrMichael:
Dashboards work and look good.
I wonder, whether the map in the charging stats could not be a heatmap?

oivindoh:
Updated the DC charging curve to look like this, with each point linking to the actual session:

Screenshot 2023-11-18 at 16 54 44

oivindoh:
@DrMichael how would you rather have the map in charging stats? There's already a list below. My main issue in the map is home charging is so huge compared to the rest that the map looks kind of silly. Maybe two separate maps for DC/AC would alleviate that - until people who mainly DC charge in one location run into the same problem.

DrMichael:
> @DrMichael how would you rather have the map in charging stats? There's already a list below. My main issue in the map is home charging is so huge compared to the rest that the map looks kind of silly. Maybe two separate maps for DC/AC would alleviate that - until people who mainly DC charge in one location run into the same problem

Ah, we can live with the current map...

swiffer:
Some of the Panels have a broken Query Editor when opening after Import (in Grafana 10.2.1). Anyone else experiencing that? Or is it an issue in recent Grafana versions. The JSON Model looks broken.

grafik

grafik

oivindoh:
@swiffer that's not what the JSON in this PR looks like, but I do see this happens to the dash JSON after importing and attempting to edit the map panel in visited... Not entirely sure why...

swiffer:
When exporting Dashboards in recent Grafana Versions Datasources are structured differently. maybe Migration / Interpretation on import is Not working correctly? Should we Transition to the New Schema with the Grafana Update? Otherwise contributing to the Dashboards is Not easy Afterwards.

oivindoh:
@swiffer I find grafana dash json to be poorly documented - do you have any good links to docs I might have missed?

I fixed the visited dash now in [cb09b1b](https://github.com/teslamate-org/teslamate/pull/3439/commits/cb09b1b6f8f8fcc2404c195b8372b03cee11cea5) - were there any problems with editors in other dashboards?

swiffer:
in the past there had been https://github.com/grafana/dashboard-spec but that seems to be deprecated.

some basic dashboard json schema info is available here - but it's not including any details for panels...
https://grafana.com/docs/grafana/latest/dashboards/build-dashboards/view-dashboard-json-model/#dashboard-json

they seem to have switched to kinds and cue for documenting the available options...
https://grafana.com/docs/grafana/latest/developers/kinds/

maybe this one hels (4 months old)
https://registry.terraform.io/providers/grafana/schemas/latest/docs/data-sources/core_dashboard

While searching for solutions I've found a Plugin for VS Code that allows to edit Dashboards from a repo directly without copying json back and forth - could be a nice time saver (currently not working for me on Windows)

https://github.com/grafana/grafana-vs-code-extension

swiffer:
> @swiffer I find grafana dash json to be poorly documented - do you have any good links to docs I might have missed?
> 
> I fixed the visited dash now in [cb09b1b](https://github.com/teslamate-org/teslamate/pull/3439/commits/cb09b1b6f8f8fcc2404c195b8372b03cee11cea5) - were there any problems with editors in other dashboards?

that works fine here!

I've tried different things two figure out what is causing it by applying changes to the Timeline Dashboard. When modifying the dashboard and setting `"schemaVersion": 34` to `"schemaVersion": 36` imports work fine. We might currently are using invalid json according to the schema we use or there is an issue importing this particular version of the schema in recent grafana versions.

Experiencing the same in "Vampire Drain", "Updates", "Timeline", "Statistics", "States", "Locations", "Drives".

As said - fixed by changing schemaVersion to 36 (from 34).

In addition I've migrated Projected Range to TimeSeries Panel (previously used deprecated Graph (old) Panel).

swiffer:
Opened PR https://github.com/teslamate-org/teslamate/pull/3454 - this fixes the Query Builder. I would propose to merge that one first, and this one afterwards.

swiffer:
@oivindoh - how are you suggest going on with the Grafana Update to 10.1.2 ? You've excluded it from the current PR, right?

oivindoh:
Yes, I didn't want to be very opinionated on grafana version. I hope that go/armv7 issue is worked out soon, but not really holding my breath. I thankfully don't have to worry about armv7 compatibility after I threw out my raspberries and replaced them with some very old mac minis (with their own breed of problems...)

Merging https://github.com/teslamate-org/teslamate/pull/3455 seems the way to go

oivindoh:
OK @JakobLichterfeld I think we're in a state where merging this, https://github.com/teslamate-org/teslamate/pull/3455 and https://github.com/teslamate-org/teslamate/pull/3454 together makes sense now if you feel like it 👍 

We can hash out some further improvements down the line regarding ease of maintaining these dashboards. I'll definately give the extension @swiffer mentioned a whirl, and I think we probably ned to consider making the datasource we provision in grafana for the post gres database have a static uid.

JakobLichterfeld:
Thank you both for your dedicated work!
Waiting with merging until #3464 is solved.

JakobLichterfeld:
So I solved the CI issues, will review your PRs by tomorrow or the day after.

JakobLichterfeld (approved):
Thanks!

JakobLichterfeld:
Failed CI is fixed with #3468 this PR does not change Elixir code therefore good to merge

micves:
Should install documentation also be updated? 
https://docs.teslamate.org/docs/installation/debian

And on a side note, who owns and updates teslamate.org?

swiffer:
Good catch - do we want to continue supporting manual install? Definitly needs some rework

micves:
> Good catch - do we want to continue supporting manual install? Definitly needs some rework

I use it once in a while when I forget the commands and procedures. 
But might also be because its needed before you can proceed with https://docs.teslamate.org/docs/development


> It should auto update, perhaps there is an permission error with repo owner change

I'm guessing it should also only auto update on releases. So maybe it is up-to-date on 1.27.4

brianmay:
We still definitely need the manual install. For users who are suspicious of pre-built binaries. For people who don't trust docker hub. And of course for developers who want to test the code without building a Docker image every time.

oivindoh:
People who are suspicious of prebuilt binaries or have trust issues with docker hub can build using the dockerfile instead, or manually follow the steps laid out by it after reading through all the elixir code they likely don’t understand (like the rest of us) anyway.

No reason to keep the overhead of documenting a non-docker path. Some unrelated PR is probably not the best place to discuss this, however.

JakobLichterfeld:
> And of course for developers who want to test the code without building a Docker image every time.

We do have new CI which build GHCR image on each PR  for that :-)

jensb:
Please continue supporting manual installs. My Raspberry Pi already runs Grafana and Postgres and node, and I'd hate to have Docker duplicating all that. Installing Erlang & Elixir and checking out via Git took just ten minutes, and updating is also working quite well.

JakobLichterfeld:
> Please continue supporting manual installs. My Raspberry Pi already runs Grafana and Postgres and node, and I'd hate to have Docker duplicating all that. Installing Erlang & Elixir and checking out via Git took just ten minutes, and updating is also working quite well.

Can you please open a new Issue for this, hard to follow on merged pr :-)

PR #3443, opened 2023-11-17
style: Dashboard links

jlestel:


JakobLichterfeld (approved):
Thanks for fixing the linting findings.

PR #3445, opened 2023-11-17
[CI] Build PR images on GHCR

jlestel:
I modified the CI so that the PR images are build then pushed into the GitHub packages (GHCR).

This does not interfere with how current CI works with Docker Hub.

Functioning :
- When a PR is created or modified: the image is built on the GHCR from fork repo.
- When a branch is pushed: the image is built on the GHCR from main repo.
- When a PR is completed: the image is deleted in the fork repo. Additionally, untagged images are also deleted on this fork repo.

Once this PR is merged, each new / updated PR will trigger CI on the fork repo.
- Users will then see all these images on their fork. Eg: [https://github.com/jlestel/teslamate/pkgs/container/teslamate](https://github.com/jlestel/teslamate/pkgs/container/teslamate)

If necessary, anyone can then modify their docker-compose to use a PR image (eg: `ghcr.io/jlestel/teslamate:master`) for testing this PR instead of `teslamate/teslamate:latest`.

After the first run, packages are created in the user repo but images can't be pushed to this user repo.

Each user which want to build his images needs to : 
1. Activate Actions in their repo
2. In the packages user page (eg: https://github.com/users/jlestel/packages/`), go to `teslamate` and `teslamate/grafana` settings and add the `user/teslamate` repository with `read/write` permissions.


JakobLichterfeld (approved):
lgtm, thanks again for the refactoring and the new workflow!

DrMichael:
This is awesome!
Perhaps we should have a section for contributors in the documentation?

JakobLichterfeld:
> Perhaps we should have a section for contributors in the documentation?

Yeah, I thought the same thing, one step at a time.

PR #3451, opened 2023-11-18
Display headers on receiving 429 response

brianmay:
Basic code to try to see what the headers are when we get a 429 (rate limited) response from Tesla.

Not tested;  I am not getting 429 responses at present.

I imagine that for a working version we will need to convert the "Retry-After" value from a string into an integer.


brianmay:
Seems I am getting Rust syntax confused with Elixir syntax :-(. Now fixed.

JakobLichterfeld:
Do you think we still need this? I could not re-run the CI on this.

brianmay:
Good question. Maybe not yet. Fingers crossed.

JakobLichterfeld:
solved by #3943

PR #3452, opened 2023-11-18
Add credo lint support

brianmay:
I think having credo support for style linting is desirable, bit surprised we don't already have it.

"mix credo" shows a number of warnings, but this PR is about adding credo checks, not fixing the warnings.

The config file is just the basic created with `mix credo gen.config`.


JakobLichterfeld:
Thanks, had this in my mind as well.

DrMichael:
So, can we merge that then?

(Can someone fix that error "test/teslamate/vehicles/vehicle/charging_test.exs:94: (test)" with the checks?)

JakobLichterfeld:
Will look into it by tomorrow

mix.lock
```diff
@@ -1,4 +1,5 @@
 %{
+  "bunt": {:hex, :bunt, "0.2.1", "e2d4792f7bc0ced7583ab54922808919518d0e57ee162901a16a1b6664ef3b14", [:mix], [], "hexpm", "a330bfb4245239787b15005e66ae6845c9cd524a288f0d141c148b02603777a5"},
```
JakobLichterfeld, line 2: Can you explain why Bunt is needed? Looks like an artifact to me.

JakobLichterfeld (changes requested):
lgtm, one small finding

mix.lock
brianmay: credo depends on bunt, so it got included.

https://github.com/rrrene/credo/blob/master/mix.exs#L151

JakobLichterfeld (approved):

PR #3454, opened 2023-11-19
Ensure Dashboards can be modified after Import - Migrate Projected Range to TimeSeries

swiffer:
As described in https://github.com/teslamate-org/teslamate/pull/3439#issuecomment-1817578606 several Dashboards currently have a broken Panel Query Editor after Import. This PR fixes the issue by updating the schemaVersion.

In addition Projected Range is migrated to the TimeSeries Panel (used the deprecated Graph(old) panel).

**Projected Range (before)**
grafik

**Projected Range (after)**
grafik

**Broken Query Editor**
284014190-5570af86-ff8c-4c00-9d35-21639c073c1e

PR #3455, opened 2023-11-19
Grafana update to 10.1.2

swiffer:
- requires the following PRs to be merged first!
  - #3439
  - #3454

Updates Grafana and removes old Plugins - also disables Angular Support in Grafana, that way we can easily see if we're still using it somewhere


JakobLichterfeld (approved):
Perfect, ty!

JakobLichterfeld:
Failed CI is fixed with #3468 this PR does not change Elixir code therefore good to merge

PR #3469, opened 2023-11-20
Update documentation for Docker Compose v2 notes

cwanja:
Included note about Docker Compose v2 command formatting.

**Note** - not going to patch my own changes to master. Will await a review. It is just documentation, but still good measure 🤷🏻‍♂️

### DIFF
--- a/website/docs/maintenance/backup_restore.md
+++ b/website/docs/maintenance/backup_restore.md
@@ -1,6 +1,9 @@
 ---
 title: Backup and Restore
 ---
+:::note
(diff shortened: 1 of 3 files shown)


DrMichael (commented):
Thanks. Good to note that!

JakobLichterfeld:
Yeah, good practice to not merge your own code without review.

I vote against this change, as docker-compose is deprecated since years and only makes Docu more complex without adding value

cwanja:
Should the note read that instead and advise individuals to update? 

Fully admit - I am still on v1.

JakobLichterfeld:
Yeah, as V1 has end of support on July 2023

cwanja:
On it.

Issue #3464, opened 2023-11-20
ci: failed test since pr 3087


JakobLichterfeld:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Elixir Mix test failed since #3087 

@brianmaand @DrMichael please either revert or fix the test

I thought it was a caching issue at first, but after deleting all caches the problem persists.

### Expected Behavior

CI Test success

### Steps To Reproduce

see runner output

### Relevant log output

First fail on master is this PR, see this runner output: https://github.com/teslamate-org/teslamate/actions/runs/6892784984
Last success was this runner https://github.com/teslamate-org/teslamate/actions/runs/6887784700 for PR #2840

### Screenshots

n.a.

### Additional data

n.a.

### Type of installation

Docker

### Version

v1.24.4

JakobLichterfeld:
cc @jlestel

JakobLichterfeld:
Hm, the behavior is strange, because the runner reported success before, even if it failed: https://github.com/teslamate-org/teslamate/actions/runs/6887784700/job/18735559815#step:11:10 but that was not in the test, but in the coverage step, which does not make sense here.

jlestel:
I've seen the coverage error: `(ExCoveralls.ReportUploadError) Failed to upload the report to 'https://coveralls.io' (reason: status_code = 422, body = {"message":"Couldn't find a repository matching this job.","error":true}).`

It seems that the repository no longer exists in coveralls.io and that's why it fails every time he wants to update it.

Could it be related to organizational change?

JakobLichterfeld:
Yeah oberserved that as well, but is not related to failed elixir test cases

jlestel:
> cc @jlestel

Il me semble que c'est plutôt le repository qui a été supprimé sur coveralls ?

jlestel:
> Yeah oberserved that as well, but is not related to failed elixir test cases

I think that if, at the end of the Elixir check, the coverage is updated.

Until CI can't update it, it returns exit 1.

jlestel:
To simply test, you can deactivate coverage.

JakobLichterfeld:
As `continue-on-error: true`for the coverage the tests with failed coverage still report as success. 
It is strange sometimes the failed elixir test cases are only shown in coverage, so will test it.

jlestel:
[...]

PR #3471, opened 2023-11-20
Add charging_state to VehicleSummary

brianmay:
Fixes: #3078

Was reverted in #3464, sorry, not sure what went wrong.

If there is a test case that needs updating, can somebody please point me to it?

### DIFF
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -6,7 +6,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
   alias TeslaMate.Log.Car
 
   defstruct ~w(
-    car display_name state since healthy latitude longitude heading battery_level usable_battery_level
+    car display_name state since healthy latitude longitude heading battery_level charging_state usable_battery_level
     ideal_battery_range_km est_battery_range_km rated_battery_range_km charge_energy_added
     speed outside_temp inside_temp is_climate_on is_preconditioning locked sentry_mode
     plugged_in scheduled_charging_start_time charge_limit_soc charger_power windows_open doors_open
@@ -84,6 +84,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
 
       # Charge State
       battery_level: charge(vehicle, :battery_level),
+      charging_state: charge(vehicle, :charging_state),
       charge_current_request: charge(vehicle, :charge_current_request),
       charge_current_request_max: charge(vehicle, :charge_current_request_max),
       charge_energy_added: charge(vehicle, :charge_energy_added),
--- a/test/teslamate/vehicles/vehicle/charging_test.exs
+++ b/test/teslamate/vehicles/vehicle/charging_test.exs
@@ -91,6 +91,7 @@ defmodule TeslaMate.Vehicles.Vehicle.ChargingTest do
 
     assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :online, since: ^s2}}}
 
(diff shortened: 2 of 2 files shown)


brianmay:
Never mind, I think I see where the problem is. Looks like the tests failing because %Summary has an extra field.

brianmay:
The problem appears to be because we receive a extra pubsub vehicle summary message that we weren't expecting.

But my code changes the code that receives the pubsub message, it shouldn't mean that there are additional pubsub messages being set.

Still puzzled.

brianmay:
Does appear to be some intermittent test failures still.

Don't think that explains why this is failing though.

(currently this is succeeding because I commented out out of the lines).

brianmay:
Cool, I think I broke lots of tests :-)

It is acting like there is some code somewhere that means the pubsub message will not get sent if it was the same value as the last message.

When I set the field to a constant, this means all tests pass (not counting intermittent faults).

But when I set the field to a random value, I get failures everywhere, because the summary keeps changing every time.

The only problem is I cannot find any code that dos this comparison.

I think for tests we are using our `PubSubMock` which just converts the `PubSub` message into a process message.

I don't see anything anywhere for dropping duplicate messages though.

brianmay:
Sorry, I am an idiot, code is in `test/support/mocks/pubsub.ex`:

```elixir
  def handle_call({:broadcast, _, _, _} = event, _from, %State{last_event: event} = state) do
    {:reply, :ok, state}
  end

  def handle_call({:broadcast, _, _, _} = event, _from, %State{pid: pid} = state) do
    send(pid, {:pubsub, event})
    {:reply, :ok, %State{state | last_event: event}}
  end
```

When we receive and event we save it to last_event. Then if the next event matches, we ignore it. This only happens during tests.

Question is how do we fix this? If I just added extra `assert_receive` lines in the tests in `test/teslamate/vehicles/vehicle/charging_test.exs` - sure that would fix the problem. But it seems to be a very fragile solution.

brianmay:
That error I just got looks kind of bad:

```
** (UndefinedFunctionError) function System.convert_time_unit/3 is undefined or private
    (elixir 1.15.7) System.convert_time_unit(-576460734009829459, :native, :millisecond)
```

Huh?

brianmay:
OK, tests pass, but not convinced this is a good solution...

JakobLichterfeld:
> Sorry, I am an idiot, code is in `test/support/mocks/pubsub.ex`:
> 
> ```elixir
>   def handle_call({:broadcast, _, _, _} = event, _from, %State{last_event: event} = state) do
>     {:reply, :ok, state}
>   end
> 
>   def handle_call({:broadcast, _, _, _} = event, _from, %State{pid: pid} = state) do
>     send(pid, {:pubsub, event})
>     {:reply, :ok, %State{state | last_event: event}}
>   end
> ```
> 
> When we receive and event we save it to last_event. Then if the next event matches, we ignore it. This only happens during tests.
> 
> Question is how do we fix this? If I just added extra `assert_receive` lines in the tests in `test/teslamate/vehicles/vehicle/charging_test.exs` - sure that would fix the problem. But it seems to be a very fragile solution.

Great find, thanks for the in-depth research!

brianmay:
Would this be considered acceptable as?

Not really sure how to make the tests more robust without making them more complicated.

I think they key issue is that these tests only want changes where the `:state` and/or `:since` values are changed. But instead we get all changes.

I guess we could change the pubsub mock, but that seems yuck. We might want to write a test that tests other values in the future.

We could also change the `assert_receive` with a function that loops over the results until it gets a change that it is looking for. That seems overly complicated; tests need to be simple and easy to understand.

JakobLichterfeld:
I completely agree.

JakobLichterfeld (approved):
ty.

PR #3477, opened 2023-11-21
Update to Nodejs 20

### LINKED ISSUE #3476: NodeJS 16 EoL

### What happened?

Right now we are building the docker image with node 16 which is end of life. https://nodejs.org/en/about/previous-releases

This means there are no more updates and we will be vulnerable to any security flaws or issues with the engine here shortly. Right now the current version that will go into LTS is nodejs20.

If we update the project to work with node 20 we will be set well into 2026. 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
No log output.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.27.4

PhilThurston:
Right now we are building the docker image with node 16 which is end of life. https://nodejs.org/en/about/previous-releases

This pull request updates the project to node 20 so that we can remain safe. After building there are still 0 vulnerabilities and building succeeds without issues. 

I've updated "esbuild" to "^0.19.4". This is because "esbuild-sass-plugin": "^2.9.0" as listed requires at least 0.19.4 for esbuild. This allowed the the project to be built and doesn't break anything. 

Closes #3476

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "assets",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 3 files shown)


assets/package-lock.json
```diff
@@ -20,22 +20,14 @@
...
-      "license": "MIT"
-    },
-    "../deps/phoenix_html": {
-      "version": "3.3.1"
-    },
-    "../deps/phoenix_live_view": {
-      "version": "0.17.14",
-      "license": "MIT"
-    },
+    "../deps/phoenix": {},
```
JakobLichterfeld, line 28: Why do you omit the version and license tags?

brianmay: I imagine this might be because we have gone from lockfileVersion 2 to lockfileVersion 3.

PR #3478, opened 2023-11-21
[CI] Success when package doesn't exist

jlestel:
If there is no GHCR image the action should finish with success.


JakobLichterfeld (approved):
ty!

PR #3479, opened 2023-11-22
Run CI with PR sources

jlestel:
In fact the pull_request_target event only plays the CI with the destination branch. This does not embed the changes made by the PR.

I modified the action to pull sources from the PR if necessary so that I can offer images per PR in the packages in this repo.

Besides, we could only trigger this pull_request_target when you set certain labels to control what is build.

PS: I added CI branch to the pull_request_target in order to test it on the CI branch.


JakobLichterfeld (approved):
lgtm

PR #3485, opened 2023-11-23
Add nix flake

brianmay:
This will add support for nix flakes. Which really helps with my development on nixos. This has two basic parts:

1. DEVELOPMENT

After this you can use the following commands on computer with nix and a patched version of direnv:

direnv allow
devenv up

And it will automatically start and instance of postgresql and mosquitto for development.

And the environment variables are already configured for teslamate.

So then can start teslamate with:

npm install --prefix ./assets && npm run deploy --prefix ./assets iex -S mix phx.server

This part is all working as designed. And pretty simple and straight forward.

2. NIX PACKAGE OF TESLAMATE

This is 99% complete, but some pain points I haven't addressed yet. Because I don't yet know of good solutions.

* There are two checksums in flake.nix that needs to be updated whenever mix.lock or package-lock.json changes.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake . --impure
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,12 @@ config/*.env
 
 /priv/plts/*.plt
 /priv/plts/*.plt.hash
+
+# Nix files
+/.direnv
+/.devenv
+/tzdata/
+/result
+
+# Used during tests
+/tmp/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - ci: bump actions/delete-package-versions from 4 to 5 (#3713)
 - Replace tortoise library with tortoise311 (#3697 - @brianmay)
 - build(deps): bump follow-redirects from 1.15.4 to 1.15.6 in /website (#3743)
+- feat: experimental nix flake (#3485 - @brianmay)
 
 #### Dashboards
 
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -25,9 +25,17 @@
         "sass": "^1.63.3"
       }
     },
-    "../deps/phoenix": {},
-    "../deps/phoenix_html": {},
-    "../deps/phoenix_live_view": {},
+    "../deps/phoenix": {
+      "version": "1.6.16",
+      "license": "MIT"
+    },
+    "../deps/phoenix_html": {
+      "version": "3.3.3"
+    },
+    "../deps/phoenix_live_view": {
+      "version": "0.17.14",
+      "license": "MIT"
+    },
     "node_modules/@creativebulma/bulma-divider": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/@creativebulma/bulma-divider/-/bulma-divider-1.1.0.tgz",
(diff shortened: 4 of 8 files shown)


brianmay:
Not sure how people feel about nix support. Just adds files, no change to existing functionality.

Can be simplified if required by leaving off the NIX packaging, which isn't currently working anyway.

JakobLichterfeld:
Do you run your TeslaMate production on NixOS as well? As docker container or as flake?

brianmay:
At present I am currently doing deploys using Docker.

Nix would be better, but haven't got to that point just yet. And would be better if we could avoid the need for the manual update steps in the package build process.

At present time for my other Elixir projects (which I do deploy with Nix) when dependabot updates stuff, I have to manual update things to ensure the nix package build works, which is kind of yuck.

Just setting up a development environment is easy though. This PR currently attempts to do both. Wondering if maybe I should limit it to the development environment as the first step.

brianmay:
Simplified the node dependencies.

At present the package build is attempting to download locale files at build time, which is a big no-no as far as nix is concerned for reproducible reasons.

brianmay:
It builds now. Wow. Amazing.

No idea if it actually works. That can come later.

And the ability to override the LOCALES directory by environment variable probably needs more work. Ideally it should be possible to build it without that environment variable set. Should be easy to fix.

This is why the CI failed.

brianmay:
Mostly works.

Currently getting errors that the token is expired and cannot be refreshed. That is not surprising with this old database.

Also getting errors from tzdata cannot write to read only file system. In theory this should be easy, just change the data directory:

```
config :tzdata, :data_dir, "/writable/directory"
```

But this did not work last I tried (I can't remember why). You can also disable auto updates entirely:

```
config :tzdata, :autoupdate, :disabled
```

Not sure I like this.

I ended up switching to the `tz` library for my other projects. Which appeared to help, but in this library auto updates are disabled by default, and I forgot to enable them. Regardless, this library seems to be better supported.

Relevant docs:

https://github.com/lau/tzdata
https://github.com/mathieuprog/tz

brianmay:
Have fixed the tzdata error.

Have mixed feeling here, this will ensure tzdata is the latest version. And I have never had an update break anything.

But maybe if we just disable auto-updates:

```
config :tzdata, :autoupdate, :disabled
```

This would make stuff more reproducible. And looking at the problems that our users have trying to set TZ correctly, this might be a good thing.

brianmay:
Some reason the "Merge GHCR images" is failing. Not exactly sure what it is trying to do...

JakobLichterfeld:
> Some reason the "Merge GHCR images" is failing. Not exactly sure what it is trying to do...

Sorry about the failed CI, see: #3689

brianmay:
No problem. When this is fixed in master I will update this pull request.

JakobLichterfeld:
> No problem. When this is fixed in master I will update this pull request.

Fixed in the master, thanks for your patience

brianmay:
* Rebased against latest master.
* Added entry to `CHANGELOG.md`.
* Added `teslamate_psql` and `teslamate_sub` wrapper scripts around `psql` and `mosquitto_sub` in devenv for convenience.
* Changed dev postgresql port to 7000 and teslmate port to 7001 so they don't conflict with already running services (if any).

brianmay:
For reasons I don't understand, trying to use the default value for tzdata's data_dir does not work with the Docker container:

```
Mar 19 15:44:33 iot2 teslamate[48492]: ** (File.Error) could not write to file "/opt/app/lib/tzdata-1.1.1/priv/latest_remote_poll.txt": permission denied
```

So I changed the default to `/tmp` instead.

JakobLichterfeld (approved):
lgtm, thanks for your efforts!

brianmay:
Yes, I think I am good to merge now.

PR #3490, opened 2023-11-25
Fix tpms bar sensors name in template for psi

virtualm2000:
In documentation, the template sensors for psi values have the wrong name.


JakobLichterfeld (approved):
nice find, ty!

PR #3498, opened 2023-11-26
grafana: support DATABASE_SSL_MODE

simonsmh:
Make sslmode for grafana changeable


grafana/datasource.yml
```diff
@@ -6,15 +6,14 @@ datasources:
   url: $DATABASE_HOST:$DATABASE_PORT
   user: $DATABASE_USER
   database: $DATABASE_NAME
-  password: ""
```
JakobLichterfeld, line 9: Is it safe to obmit the password field?

JakobLichterfeld (changes requested):

grafana/datasource.yml
simonsmh: Yes. secureJsonData.password should be enough.

JakobLichterfeld (approved):
lgtm, ty!

Issue #3486, opened 2023-11-24
Battery Health Issue with Model S appears to use Ideal Efficiency and not Rated Efficiency, causing far too high an 'old' capacity


jensk22:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

(I don't know how to escalate a Discusiion to an Issue, so here we go) The issue is described in Discussion:
https://github.com/teslamate-org/teslamate/discussions/3470
)

My Tesla Model S 70D Y2015, which has approx 68,8/67,2kWh Usable from new, is reported with 'original' capacity in excess of 83 kWh

![image](https://github.com/teslamate-org/teslamate/assets/92438498/fc03416c-5529-4c74-aff7-40d3575c946d)

Assuming the 
  ('$aux'::json -> 'RatedEfficiency')::text::float
is in effect the Ideal Efficiency of 190Wh/km instead of the Rated Efficiency of 150Wh, I changed the Graph Query to:
and that produces proper 'old' capacity numbers (which for me is around 66.1kWh because I started using Teslamate as late as at odeometer=70kKm). So the :

Below is not a proper fix, the right solution is to update the  ('$aux'::json -> 'RatedEfficiency')::text::float to be correct!

SELECT convert_km(AVG(p.odometer)::numeric,'$length_unit') AS odometer,
AVG(c.ideal_battery_range_km * ('$aux'::json -> 'RatedEfficiency')::text::float / c.usable_battery_level) AS "kWh",
MAX(cp.id) AS id,
to_char(cp.end_date, 'YYYY-MM-dd') AS Title
FROM charging_processes cp
JOIN (SELECT charging_process_id, MAX(date) as date FROM charges WHERE usable_battery_level > 0 GROUP BY charging_process_id) AS last_charges ON cp.id = last_charges.charging_process_id
INNER JOIN charges c
ON c.charging_process_id = cp.id AND c.date = last_charges.date
INNER JOIN positions p ON p.id = cp.position_id
WHERE cp.car_id = $car_id
AND cp.end_date IS NOT NULL
AND cp.charge_energy_added >= ('$aux'::json -> 'RatedEfficiency')::text::float
GROUP BY 4

New graph:
![image](https://github.com/teslamate-org/teslamate/assets/92438498/7c832c3d-45f9-4668-a97a-b1d3e809f590)

### Expected Behavior

[...]

Issue #3438, opened 2023-11-16
feat: use variables in grafana dashboards for battery type


JakobLichterfeld:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

There are different types of batteries, e.g. LFP is normally charged to 100%. Dashboards should reflect this depending on the battery type: see #3311 

### Expected Behavior

Battery type handled with variables to avoid multiple dashboards for each variant

### Relevant log output

```shell
n.a.
```

### Type of installation

Docker

### Version

v1.27.4

swiffer:
Battery type can be derived from VIN - https://teslawissen.ch/tesla-vin-decoder-so-kannst-du-die-vin-entschluesseln/#ziffer-7-ladetyp-2012-2013-akkutyp-seit-2014

happy to open a PR once current grafana PR are merged.

swiffer:
This is what we need for it

https://grafana.com/docs/grafana/next/panels-visualizations/query-transform-data/transform-data/#config-from-query-results

swiffer:
@eden881 - do you have a list of Changes done to the LFP Variants of the Dashboards?

eden881:
> @eden881 - do you have a list of Changes done to the LFP Variants of the Dashboards?
> 
> Is that all that has been Changed?
> 
> Overview - LFP: The battery level gauge has red and yellow zones only on the lower end. The maximum value of the charging kW gauge has been lowered to 170, as this is the maximum DC charging potential for Tesla's LFP batteries. Charging Stats - LFP: The charge stats table remains green even for entries higher than 80% and 95%. Battery Health - LFP: The current SOC bar has a red zone only on the lower end

That is correct.

swiffer:
Tried editing the Overview Dashboard, works fine for "Charging kw" - however for modifying the Thresholds we do need support for setting multiple values via config I guess

https://github.com/grafana/grafana/issues/77491
https://github.com/grafana/grafana/issues/45803

swiffer:
It can be derived from vin, will modify Dashboards and remove LFP ones as soon as we can Set multiple thresholds via Config from query transformation

NirKli:
I've reviewed the potential solution for retrieving information from VIN, but it appears to be inaccurate. 
I've begun developing the solution, implementing a flag on the settings page. However, it seems that Grafana does not support conditional thresholds.

swiffer:
[...]

PR #3531, opened 2023-12-01
Improve Battery Health estimations based on derivate efficiency

jheredianet:
This is an improve on Battery Health estimations, based on derivate efficiency discussed here https://github.com/teslamate-org/teslamate/issues/3486#issuecomment-1835727943

In the other hand, I've removed the same dashboard for LFP batteries since it is not worth having 2 dashboards available as explained here https://github.com/teslamate-org/teslamate/discussions/3513#discussioncomment-7694883 at least until we have this feature (https://github.com/teslamate-org/teslamate/issues/3438) implemented in Teslamate.


JakobLichterfeld:
> In the other hand, I've removed the same dashboard for LFP batteries since it is not worth having 2 dashboards available as explained here [#3513 (reply in thread)](https://github.com/teslamate-org/teslamate/discussions/3513#discussioncomment-7694883) at least until we have this feature (#3438) implemented in Teslamate.

Braking backwards compatibility is not an option.

jheredianet:
> However, breaking backwards compatibility is not an option.
Totally agree, but this is not the case. There is no breaking backwards compatibility. Please check it further.

jheredianet:
> > Totally agree, but this is not the case. There is no breaking backwards compatibility. Please check it further.
> 
> For sure, you remove Dashboards, without including a solution for #3438

Ok, I've made an update. No removed Battery Health LFP dashboard. 
As the title says, this is a PR for improving estimations, based on derivate efficiency.

PR #3534, opened 2023-12-02
Fix Odometer_mi charger dashboard

### LINKED ISSUE #3533: Charges dashboard shows km instead of miles

### What happened?

Charges dashboard shows as kilometer in last column while everything else is set miles

### Expected Behavior

While settings for the car and all other dashboards shows miles, ‘Charges’ dashboard should show miles instead of kilometers in last column

### Steps To Reproduce

_No response_

### Relevant log output

```shell
Nothing relevant
```

### Additional data

_No response_

### Type of installation

Docker

### Version

1.28.2

Edit:
Updated snapshot

mechoriet:
Odometer_mi override was set to km instead of mi in charges dashboard

should fix #3533 

(edited this is in codespaces so dunno if this would break updating it)

also added a version change to the dashboard don't know what is required to be updated or if a single change in the json breaks them

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -1354,6 +1354,6 @@
   "timezone": "",
   "title": "Charges",
   "uid": "TSmNYvRRk",
-  "version": 3,
+  "version": 4,
   "weekStart": ""
 }
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -770,7 +770,7 @@
               },
               {
                 "id": "unit",
-                "value": "km"
+                "value": "mi"
               },
               {
                 "id": "decimals"


cwanja:
@DrMichael will this pull request respect the setting that the user has set within TeslaMate? I know other Grafana dashboards use the user driven setting to show km or mi. This seems like it is hard coded to be one or the other.

mechoriet:
> @DrMichael will this pull request respect the setting that the user has set within TeslaMate? I know other Grafana dashboards use the user driven setting to show km or mi. This seems like it is hard coded to be one or the other.

I only changed the used unit for that override it's not like i'm changing how they are selected just the unit it is displayed as

but I bet the api returns odometer_km or odomoter_mi which ever the case is for that set car

cwanja:
> > @DrMichael will this pull request respect the setting that the user has set within TeslaMate? I know other Grafana dashboards use the user driven setting to show km or mi. This seems like it is hard coded to be one or the other.
> 
> I only changed the used unit for that override it's not like i'm changing how they are selected just the unit it is displayed as
> 
> but I bet the api returns odometer_km or odomoter_mi which ever the case is for that set car

Agree with that entirely. Just wanted a second validation. Digging further into the code, it looks like there was a copy paste with a lack of replace when adding the odometer.

Thanks for the PR 😄

DrMichael:
Right, it was just a naming error on my side... Depending on the setting, either odometer_km or odometer_mi will  be set.

Issue #3535, opened 2023-12-02
Dog Mode not longer recognized


Shadesh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

When Dog Mode is active, Teslamate shows nothing and is going to sleep after a while. So for me it looks like the Dog Mode is not longer recognized. There are no relevant logs, only after a while:

car_id=1 [info] Suspending logging

### Expected Behavior

Detect Dog Mode and show A/C symbol and dont go to sleep on the Teslamate Interface.

### Steps To Reproduce

Enable Dog Mode for more then 15 Minutes and look at the Teslamate Interface

### Relevant log output

```shell
car_id=1 [info] Suspending logging
```

### Type of installation

Docker

### Version

1.28.2

NirKli:
@Shadesh 
Could you clarify if this issue with Dog Mode not being recognized by Teslamate started occurring specifically after updating to version 1.28.2? Or have you been experiencing this problem since you first started using the program?

Shadesh:
Afaik i had the A/C symbol and Teslamate wasn't going to sleep while Dog mode was active. This seems not longer the case. Don't know exactly if it changed with 1.28.2 or earlier. I saw it today the first time because the car was in Dog mode for around 30 minutes.

JakobLichterfeld:
TeslaApi didn't report dedicated dog mode in the past, as they change their API we need to adopt

JakobLichterfeld:
closed by #3538

PR #3538, opened 2023-12-02
Fix an issue when car is could enter sleep mode while dog mode is enabled

NirKli:
Fix an issue when car is could enter sleep mode while dog mode is enabled and also added dog icon that indicates if dog mode is activate.

First

when clicking on try sleep button

second

the current state when dog mode is active (1.28.1)

thrid

#3535


JakobLichterfeld (approved):
Nice feature enhancement, ty

brianmay (approved):
Was hoping to test this, but having [problems](https://github.com/teslamate-org/teslamate/discussions/3554) getting a token right now. Assume it has been tested and works.

It looks good to me.

PR #3544, opened 2023-12-04
Update charges.json: reenable cost per kWh

DrMichael:
Calculation was removed by error.


cwanja (approved):
Looks good! Thank you~

PR #3556, opened 2023-12-10
Refactoring

sakul1234:


JakobLichterfeld:
@jlestel Why was this not prevented from running?

JakobLichterfeld:
stop targeting

jlestel:
> @jlestel Why was this not prevented from running?

Hi Jakob, I'm looking in more detail to find the right answer.

jlestel:
Hum you're right, as mentioned in the GH workflow docs: [You should not use path filtering to skip workflow runs if the workflow has been configured to be required...](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-including-paths)

I'm thinking of using this module to really condition the execution of the workflow: https://github.com/dorny/paths-filter

Let me know what you think before I implement it.

JakobLichterfeld:
Lets track it in #3559

Issue #3524, opened 2023-11-30
Battery health KO 


nehox:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

<img width="1777" alt="image" src="https://github.com/teslamate-org/teslamate/assets/3989362/bc9016e2-ef99-4a04-ad93-3947475a1ca9">

### Expected Behavior

in error I have this message : `Status: 500. Message: db query error: pq: invalid input syntax for type json`

SELECT  
  CASE WHEN $custom_kwh_new > 0 THEN $custom_kwh_new ELSE ('$aux'::json -> 'MaxCapacity')::text::float END as "Usable (new)", 
  ('$aux'::json -> 'CurrentCapacity')::text::float as "Usable (now)",
  ('$aux'::json -> 'CurrentCapacity')::text::float - CASE WHEN $custom_kwh_new > 0 THEN $custom_kwh_new ELSE ('$aux'::json -> 'MaxCapacity')::text::float END as "Difference"

in query $aux is empty. Anyone have the solution ?

### Steps To Reproduce

it's a fresh install, no data import. I want to try this project.

### Relevant log output

```shell
2023-11-30 13:03:59.177 UTC [48] ERROR:  invalid input syntax for type json at character 41
2023-11-30 13:03:59.177 UTC [48] DETAIL:  The input string ended unexpectedly.
2023-11-30 13:03:59.177 UTC [48] CONTEXT:  JSON data, line 1: 
2023-11-30 13:03:59.177 UTC [48] STATEMENT:  SELECT  
	  CASE WHEN 0 > 0 THEN 0 ELSE (''::json -> 'MaxRange')::text::float END as "maxrange_km",
	  (''::json -> 'CurrentRange')::text::float  as "currentrange_km",
	  CASE WHEN 0 > 0 THEN 0 ELSE (''::json -> 'MaxRange')::text::float END - (''::json -> 'CurrentRange')::text::float as "range_lost_km"
```

### Type of installation

Docker

### Version

v1.28.2

nehox:
This request have no result :

`WITH
aux	AS
(
	SELECT cp.charge_energy_added,
		cp.car_id, (SELECT efficiency FROM cars WHERE id = $car_id) * 100.0 AS rated_efficiency,
		(cp.end_rated_range_km - cp.start_rated_range_km) AS added_range_km
	FROM charging_processes cp
		JOIN (SELECT charging_process_id, MAX(date) as date
		FROM charges
[...]

PR #3582, opened 2023-12-22
Fix for "No Data" Error at Battery Health dashboard

inigoml:
Fix error when car battery level is around 0% for a time. This provokes a divide by zero error.
Related with this issue.
https://github.com/teslamate-org/teslamate/issues/3524


jheredianet:
Hi @inigoml thanks for pointing this possible error on my own repository at https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards/issues/62

I've accepted your PR, although I had to make some other adjustments, because the possible divided by zero error was not only on the query you pointed but in other query as well.

I've also taken the opportunity to add some other features and improvements that are included in this PR https://github.com/teslamate-org/teslamate/pull/3593

JakobLichterfeld:
Closed with #3493

PR #3600, opened 2024-01-03
Update default.po

ycjcl868:
「登陆」 means disembark, 「登录」 is correct.


JakobLichterfeld:
Thanks for the change.

PR #3623, opened 2024-01-19
docs/integration: Node-RED required modules should be saved under /data

ToniA:
Node-RED required modules need to be installed into a volume which is preserved when the container is recreated


JakobLichterfeld (approved):
lgtm

PR #3625, opened 2024-01-20
charging-stats: map - add charges count and move label

fmossott:
Couple of small changes on the Charging Stats map:

- move label more on the right to reduce overlap with circle
- added number of charges at the location in the tooltip

image


JakobLichterfeld (approved):
Thanks for the change. lgtm

Issue #3483, opened 2023-11-22
[Feature Request] Add local time when done charging calculated from remaining charging time


adjego:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

If possible i would like to be able to see on the home page of Teslamate, when the car is charging at which time the car will be finished charging. Now it only shows how long it will take to charge, and need to calculate this manually.

### Relevant log output

```shell
Not Relevant
```

### Screenshots

![image](https://github.com/teslamate-org/teslamate/assets/5793288/23d48961-7a15-4a6f-8775-adb5f8d254b3)

### Type of installation

Docker

### Version

v1.27.4

djpalmis:
As per: https://github.com/teslamate-org/teslamate/discussions/2750

NirKli:
I think it will be nice to see when the car will finish charging, I'll look in to it.

djpalmis:
Thanks @NirKli !

NirKli:
I have made same progress.

I'll format it to not include the mills on the dateTime.

what's your thoughts?
@djpalmis @adjego  

<img width="678" alt="image" src="https://github.com/teslamate-org/teslamate/assets/45952578/f1a921c2-bf9a-4006-903e-240fdea22ee0">

djpalmis:
I just LOVE IT!
Honestly I really ignore range (that can be shown hoovering on SoC).
Does anyone really care about that?
Can we do a pool to remove that useless info to make space to your new data? 

On contrary I really care about projected range (I hate to hoover every time) and i would love your new "expected finish..".

Thank you by heart!

NirKli:
Another solution, what do you think, which one is better?

<img width="671" alt="image" src="https://github.com/teslamate-org/teslamate/assets/45952578/f5b5f2e8-e2b3-4e88-9ba1-a71218a21962">

djpalmis:
Wonderful!!

cwanja:
> Another solution, what do you think, which one is better?
> 
[...]

PR #3646, opened 2024-02-01
Added expected finish charging localtime to teslamate site

NirKli:
I've added the requested feature for displaying the expected finishing time of charging on the TeslaMate site, as per the discussion in [issue #3483](https://github.com/teslamate-org/teslamate/issues/3483).

Here's an image showcasing the update:

image

### DIFF
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -148,16 +148,26 @@
               end
             )%></td>
           </tr>
-          <%= if @summary.state == :charging and not is_nil(@summary.time_to_full_charge) do %>
-          <tr>
-            <td class="has-text-weight-medium"><%= gettext "Remaining Time" %></td>
-            <td><%=
+            <%= if @summary.state == :charging and not is_nil(@summary.time_to_full_charge) do %>
+            <%
+              current_time = DateTime.utc_now()
+              finish_time = DateTime.add(current_time, round(@summary.time_to_full_charge * 60 * 60))
+              formatted_finish_time = Timex.format!(finish_time, "%Y-%m-%d %H:%M:%S", :strftime)
+
+              remaining_time_text =
                 round(@summary.time_to_full_charge * 60 * 60)
                 |> Convert.sec_to_str()
                 |> Enum.reject(&String.ends_with?(&1, "s"))
-                |> Enum.join(", ")
-            %></td>
-          </tr>
+                |> Enum.join(", ")
+            %>
+            <tr>
+              <td class="has-text-weight-medium"><%= gettext "Remaining Time" %></td>
+              <td><%= remaining_time_text %></td>
+            </tr>
+            <tr>
+              <td class="has-text-weight-medium"><%= gettext "Expected Finish Time" %></td>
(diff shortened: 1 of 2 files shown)


adgoudsmit:
Very cool this new feature, but it seems it's off by 1 hour for me. Is there a setting somewhere?

image

NirKli:
> Very cool this new feature, but it seems it's off by 1 hour for me. Is there a setting somewhere?
> 
> image

Have you changed the default timezone via TZ env var in docker compose file?
https://docs.teslamate.org/docs/configuration/environment_variables

adgoudsmit:
Thanks for the quick response! Resolved!

adjego:
Resolved! Thanks for the quick response and my feature request!

NirKli:
@JakobLichterfeld,
Perhaps we should include a note in the release notes advising users to adjust the TZ environment variable if the expected time is incorrect.

JakobLichterfeld:
Not sure about this. Grafana can use local browser time. What about Elixir?

Wanted to include your change in yesterday's release and did not fully check the consequences to be honest.
I always set TZ.

brianmay:
It is possible to use local browser time in Elixir live views, but considering the limited scope of the web API, wonder if it is worth it. Might be easier just to set TZ (assuming this works correctly - not tested yet myself).

For details on how to do this see https://alex-min.fr/live-view-browser-timezone/ (top search result; no idea if this is the best or not)

cwanja:
Is there a way to not have it default to 24 hours @NirKli ? Is that defined in the TZ environment variable?

DrMichael:
Perhaps that could be handled via an environment variable. Similar to GF_DATE_FORMATS_FULL_DATE...

cwanja:
I would say either an environment variable or a setting variable (via database) similar to temperature and tire pressure.

PoppyS33d:
I’m getting the wrong time zone too, and I have the TZ line set correctly. Is there something else to change to get the correct time zone?

JakobLichterfeld:
> I’m getting the wrong time zone too, and I have the TZ line set correctly. Is there something else to change to get the correct time zone for estimated charge completion?

Please double-check your value is valid (see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). If the error still persists afterward, please open an issue and share your TZ settings.

PoppyS33d:
> > I’m getting the wrong time zone too, and I have the TZ line set correctly. Is there something else to change to get the correct time zone for estimated charge completion?
> 
> Please double-check your value is valid (see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). If the error still persists afterward, please open an issue and share your TZ settings.

I believe I double checked and it seems to be correct, and even after restarting Docker/whole Mac, I can't seem to get this to work. 

`version: '3'

services:
teslamate:
image: teslamate/teslamate:latest
restart: always
environment:
- ENCRYPTION_KEY=lxxx
- DATABASE_USER=teslamate
- DATABASE_PASS=secret
- DATABASE_NAME=teslamate
- DATABASE_HOST=database
- MQTT_HOST=mosquitto
- TZ=America/Los_Angeles`

JakobLichterfeld:
seems to be right, I opened a seperate issue for you, #3665

PR #3650, opened 2024-02-03
Add details how to configuring the TZ environment variable

NirKli:
Add instructions for configuring the TZ environment variable in Docker setups, ensuring TeslaMate accurately displays local times.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,21 @@
 - doc: Fix TPMS sensor names to match binary_sensor.yaml (#3613 - @kenni)
 - update docker-compose.yml for Docker Compose version v2.24.1 (#3626 - @realJustinLee)
 
+### Configure Timezone Setting for Accurate Local Time Display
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Why choose you to explain things which are already in doc? If needed, we should just link to https://docs.teslamate.org/docs/configuration/environment_variables

JakobLichterfeld (changes requested):
Why choose you to explain things which are already in doc? If needed, we should just link to https://docs.teslamate.org/docs/configuration/environment_variables

NirKli:
> Why choose you to explain things which are already in doc? If needed, we should just link to https://docs.teslamate.org/docs/configuration/environment_variables

My intention was to enhance the visibility of the TZ environment variable configuration, especially considering its impact on user experience regarding local time display in teslamate. However, I agree that maintaining a single source of truth in documentation is crucial to avoid confusion.
I updated the release notes to directly reference the existing documentation for setting environment variables,

PR #3657, opened 2024-02-06
Add navigation active route drive state data

longzheng:
Addresses https://github.com/teslamate-org/teslamate/discussions/3250 

There are additional fields from the `drive_state` object that contain destination name and metadata if it is set in the vehicle routing.

```json
        "drive_state": {
            "active_route_destination": "Home",
            "active_route_energy_at_arrival": 73,
            "active_route_latitude": -1.123456,
            "active_route_longitude": 1.123456,
            "active_route_miles_to_arrival": 6.485299,
            "active_route_minutes_to_arrival": 23.466667,
            "active_route_traffic_minutes_delay": 0.0,
            "gps_as_of": 1686894025,
            "heading": 254,
            "latitude": -1.123456,
            "longitude": 1.123456,
```

Unfortunately I don't have the environment set up so I didn't test this but this matches the existing logic. I'm also hoping this automatically flows through the MQTT broker since the entire drive state is already sent.

### DIFF
--- a/lib/tesla_api/vehicle/state.ex
+++ b/lib/tesla_api/vehicle/state.ex
@@ -167,6 +167,13 @@ defmodule TeslaApi.Vehicle.State do
 
   defmodule Drive do
     defstruct [
+      :active_route_destination,
+      :active_route_energy_at_arrival,
+      :active_route_latitude,
+      :active_route_longitude,
+      :active_route_miles_to_arrival,
+      :active_route_minutes_to_arrival,
+      :active_route_traffic_minutes_delay,
       :gps_as_of,
       :heading,
       :latitude,
@@ -183,6 +190,13 @@ defmodule TeslaApi.Vehicle.State do
 
     def result(drive) when is_map(drive) do
       %__MODULE__{
+        active_route_destination: drive["active_route_destination"],
+        active_route_energy_at_arrival: drive["active_route_energy_at_arrival"],
+        active_route_latitude: drive["active_route_latitude"],
+        active_route_longitude: drive["active_route_longitude"],
+        active_route_miles_to_arrival: drive["active_route_miles_to_arrival"],
+        active_route_minutes_to_arrival: drive["active_route_minutes_to_arrival"],
+        active_route_traffic_minutes_delay: drive["active_route_traffic_minutes_delay"],
         gps_as_of: drive["gps_as_of"],
         heading: drive["heading"],
         latitude: drive["latitude"],


JakobLichterfeld:
Thanks for your suggestion. Please also add a test for this.

longzheng:
> Thanks for your suggestion. Please also add a test for this and update existing once, so CI pass.

@JakobLichterfeld I've set up a local environment and a MQTT broker and verified the new values now show up in MQTT.

I don't have any experience with Elixir so I'm a bit stuck with adding a new test. I'm not really sure where it should go and what we would be testing. Could I get some suggestions/help?

JakobLichterfeld (approved):
thanks for fixing the test case as well and adding the docu

brianmay:
I am kind of uneasy having these sent as separate MQTT topics. For example, think of a program receiving MQTT events, it will receive destination, and then latitude, and then longitude. Although order is unpredictable. There will be a brief time when it has a mismatch of old values and new values. Outputting all of this in one json blob might be better.

Yes, I think this is already a problem with the latitude/longitude output.

Apart from that looks good.

longzheng:
> I am kind of uneasy having these sent as separate MQTT topics. For example, think of a program receiving MQTT events, it will receive destination, and then latitude, and then longitude. Although order is unpredictable. There will be a brief time when it has a mismatch of old values and new values. Outputting all of this in one json blob might be better.
> 
> Yes, I think this is already a problem with the latitude/longitude output.
> 
> Apart from that looks good.

Agreed, I didn't want to break the existing pattern.

I also didn't bother adding in the other fields to the MQTT output but happy to do so. The only one that needs some transforming is `active_route_miles_to_arrival` which I think needs to be converted to km?

WesSec:
> I am kind of uneasy having these sent as separate MQTT topics. For example, think of a program receiving MQTT events, it will receive destination, and then latitude, and then longitude. Although order is unpredictable. There will be a brief time when it has a mismatch of old values and new values. Outputting all of this in one json blob might be better.
> 
> Yes, I think this is already a problem with the latitude/longitude output.
> 
> Apart from that looks good.

I fully agree, for latlong, for the other travel ETA's not so much, these update throughout the journey (where latlong does not), which will require to update the complete blob. Also adding more stuff to a big json requires the processing client (often HA) to parse this json, where normally an mqtt topic can be read easily. 

I think both ways can be done and advocated for, zigbee2mqtt is more json based too, but i'd suggest to not wait for a possible complete revamp of the mqtt structure for this awesome change already.

JakobLichterfeld:
Please shift follow up discussion in #3660

PR #3667, opened 2024-02-11
Add stat panel for selected duration in drive details dashboard

oivindoh:
I sometimes want to compare actual driving time for two different routes between specific points on a drive, so I've added a "selected duration" stat panel that simply shows the number of minutes between the chosen from/to in grafana, e.g. in this screen I've selected a particular subsection of the drive to see how long it took.

Also removed some ancient pinned pluginVersions for panels. We're on at least 10.1 now, so it doesn't make sense to use those.

Screenshot 2024-02-11 at 14 39 44

### DIFF
--- a/grafana/dashboards/internal/drive-details.json
+++ b/grafana/dashboards/internal/drive-details.json
@@ -334,7 +334,6 @@
           "sort": "none"
         }
       },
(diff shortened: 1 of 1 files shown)


JakobLichterfeld (approved):
great idea. Ty!

PR #3670, opened 2024-02-12
Change tracking line color on Maps

jheredianet:
This is a proposal to change the color of the tracking lines in all the Grafana dashboards with Maps, from **red** to **blue**, since the color red often overlaps with the color of the map's own highways and becomes difficult to distinguish.
At the same time, I took the opportunity to update the plugins on the corresponding Grafana dashboards. 

Here's a sample.

Before:
image

After:
image

Most of navigators like Google Maps or the same navigation system of Tesla, the tracking line is "Blue".


JakobLichterfeld:
Thanks for the suggestion. Please change the UID, as we get a strange spelling error with it: https://github.com/teslamate-org/teslamate/actions/runs/7874774381/job/21485158253?pr=3670#step:11:45

JakobLichterfeld:
As I remember correctly, we had the discussion before to change the color of the lines in maps. In the end, blue lines will be a problem in more regions, correct @swiffer and @oivindoh?

oivindoh:
There was some talk at least in https://github.com/teslamate-org/teslamate/pull/3439#issuecomment-1816531294 - i originally switched to blue in that PR, but reverted it.

Personally I buy the comment about blue having become the standard for routes, but the default OSM map makes blues without a border look a bit washed out, maybe.

PR #3672, opened 2024-02-13
Fix agressive switch to offline when stream gets offline message

micves:
Initiated by #3658 

When stream got a message that the vehicle was offline, it switched directly to Start / :offline
This is to aggressive and does not reflect the true state of the vehicle

Now its changed to just fetch data right way and let the result handle the true state of the vehicle.

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -528,7 +528,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
   def handle_event(:info, {:stream, msg}, _state, data)
       when msg in [:vehicle_offline] do
-    Logger.info("Stream reports vehicle as offline, fetching vehicle state ...",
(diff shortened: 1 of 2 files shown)


JakobLichterfeld (approved):
lgtm

jheredianet:
I was testing this in varios scenarios and it does not seem to work well. When the car goes to asleep state it directly becomes offline as commented here https://github.com/teslamate-org/teslamate/discussions/3658#discussioncomment-8857819. 
Maybe for newer updates on Tesla state asleep is offline? I'm not sure.

micves:
This fix just gets the state directly from the car (Tesla api server) so it would not look different.
I'm guessing you are on Tesla version 2024.8.4? Which in generel seems to report asleep as offline.

jheredianet:
> This fix just gets the state directly from the car (Tesla api server) so it would not look different. I'm guessing you are on Tesla version 2024.8.4? Which in general seems to report asleep as offline.

Yes, I'm on version 2024.8.4.
Thanks.

PR #3693, opened 2024-02-27
feat: add missing translation for Korean

zipizigi:
add new message translation for korean

### DIFF
--- a/priv/gettext/ko/LC_MESSAGES/default.po
+++ b/priv/gettext/ko/LC_MESSAGES/default.po
@@ -648,7 +648,7 @@ msgstr "애견 모드"
 #: lib/teslamate_web/live/car_live/summary.ex:139
 #, elixir-autogen, elixir-format
 msgid "Dog mode is enabled"
-msgstr "애견 모드가 켜져있습니다.
+msgstr "애견 모드가 켜져있습니다."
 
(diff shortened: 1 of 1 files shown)


priv/gettext/ko/LC_MESSAGES/default.po
```diff
@@ -638,19 +638,19 @@ msgstr "공기압"
...
 #, elixir-autogen, elixir-format
 msgid "Dog Mode"
-msgstr ""
+msgstr "애견 모드"
 
 #: lib/teslamate_web/live/car_live/summary.ex:139
 #, elixir-autogen, elixir-format
 msgid "Dog mode is enabled"
-msgstr ""
+msgstr "애견 모드가 켜져있습니다.
```
JakobLichterfeld, line 651: this needs to be in `"` atz the end as well

JakobLichterfeld (changes requested):
Thanks for your change. One small finding, see inline comment

priv/gettext/ko/LC_MESSAGES/default.po
zipizigi: Oops, I made a mistake.
problem has been fixed.

JakobLichterfeld (approved):
lgtm

PR #3694, opened 2024-02-27
Add missing German translation, fix mistakes

accolon:


JakobLichterfeld (approved):
Thanks for your change, sieht gut aus :-)

PR #3697, opened 2024-02-27
Replace tortoise library with tortoise311

brianmay:
tortoise311 is better maintained (last release 6 months ago as opposed to 3 years ago).

Not tested yet...


brianmay:
Database values are stored in decimal, since 45d94d52b381f9183bdc6de1995f2c94c481b794. Supposedly this makes things more efficient.

Values from the TeslaApi are retrieved at floats.

I think I am will put all the Decimal types back and have this new code deal with both cases.

JakobLichterfeld (approved):

PR #3699, opened 2024-02-28
Update default.po

occultsound:
Traditional Chinese  updated


JakobLichterfeld (approved):
thanks for your change, lgtm

PR #3709, opened 2024-02-29
Update default.po （Traditional Chinese details updated）

occultsound:
more Traditional Chinese details updated !


tobiasehlert (approved):
thanks for your change 👍

PR #3718, opened 2024-03-03
Fix name in lovelace example to match sensor name

js94x:
this PR is related to #3613 and fixes the sensor name in lovelace example too


tobiasehlert (commented):
@js94x, is there a `sensor.tesla_tpms_pressure_fl` for bar then?
Can't see that in the _sensor.yaml_ file, but I don't run HA myself either so can't check.

tobiasehlert:
Thanks for clarification @js94x 👍

tobiasehlert (approved):
LGTM

PR #3726, opened 2024-03-07
Grafana 10.4.0 & Angular Deprecation

swiffer:


swiffer:
@JakobLichterfeld - looks good for me (amd64) - @oivindoh, if you have some minutes to try out the docker image as well would be great.

oivindoh:
@swiffer I'm running the official 10.4.0 image which doesn't do anything but import dashboards from github master (e.g. not running any of the teslamate env vars except TESLAMATE_DB_PASSWORD, with no plugins, and as far as I can tell everything works as expected.

Not sure the TM specific grafana image gives much value at this point

parkr:
> Not sure the TM specific grafana image gives much value at this point

Not sure exactly what you're proposing here but I value the bundled experience of Teslamate server + graphs as easy as a docker compose file. What does your setup look like using just the stock image? Setting authentication with env vars and automatically including the dashboards is what I care most about -- could we do that with less maintenance burden on the teslamate project?

JakobLichterfeld:
> could we do that with less maintenance burden on the teslamate project?

Yes, if we shift it to the users and rely on GitHub server availability imo. So no real solution imo. 
As new grafana images are available for Armv7 again and swiffer and oivindoh ported all the plugins we are future proof with the current approach. But I'm open to new solutions as well.

swiffer:
Agree that with removal of plugins there is not much value added anymore. I would prefer staying with a Teslamate specific Grafana Image. having Dashboards pre bundled and data source setup still is easier. the additional maintenance burden is super low and it's still more convenient for users to setup (only docker needed) + we have a controlled / recommended environment.

@JakobLichterfeld are you going to merge?

swiffer:
Possible downside: security updates in Grafana require us to bump Grafana and Teslamate Tags if we want to stay on latest version possible.

JakobLichterfeld:
> @JakobLichterfeld are you going to merge?

Yes, will merge by tomorrow and consider to release a new TeslaMate version in the coming days.

PR #3729, opened 2024-03-11
Add location topic

### LINKED ISSUE #3660: feat: bundle MQTT data in one json blob

### What happened?

> I am kind of uneasy having these sent as separate MQTT topics. For example, think of a program receiving MQTT events, it will receive destination, and then latitude, and then longitude. Although order is unpredictable. There will be a brief time when it has a mismatch of old values and new values. Outputting all of this in one json blob might be better.
> 
> Yes, I think this is already a problem with the latitude/longitude output.

(see https://github.com/teslamate-org/teslamate/pull/3657#issuecomment-1933195471)

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
n.a.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.28.3

brianmay:
Fixes #3660.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -74,6 +74,36 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
         nil
     end)
 
+    if state.last_summary == nil or
+         state.last_summary.latitude != summary.latitude or
+         state.last_summary.longitude != summary.longitude do
+      lat_lng =
+        case {summary.latitude, summary.longitude} do
+          {nil, _} -> nil
+          {_, nil} -> nil
+          {%Decimal{} = lat, %Decimal{} = lon} -> {Decimal.to_float(lat), Decimal.to_float(lon)}
+          {lat, lon} -> {lat, lon}
+        end
+
+      case lat_lng do
+        nil ->
+          nil
+
+        {lat, lon} ->
+          location =
+            %{
+              latitude: lat,
+              longitude: lon
+            }
+            |> Jason.encode!()
+
+          case publish({"location", location}, state) do
+            :ok -> nil
+            {:error, reason} -> Logger.warning("Failed to publish location: #{inspect(reason)}")
+          end
+      end
+    end
+
     {:noreply, %State{state | last_summary: summary}}
   end
 
--- a/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
+++ b/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
@@ -95,6 +95,14 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriberTest do
       assert_receive {MqttPublisherMock, {:publish, ^topic, "", [retain: true, qos: 1]}}
     end
 
+    assert_receive {MqttPublisherMock,
+                    {:publish, "teslamate/cars/0/location", data, [retain: true, qos: 1]}}
+
+    assert Jason.decode!(data) == %{
+             "latitude" => 37.889602,
+             "longitude" => 41.129182
+           }
+
     refute_receive _
   end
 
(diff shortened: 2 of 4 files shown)


brianmay:
Not tested yet :-)

brianmay:
Not sure I understand why this test is failing:

```
 1) test publishes vehicle data (TeslaMate.Mqtt.PubSub.VehicleSubscriberTest)
Error:      test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs:27
     Unexpectedly received message {VehiclesMock, {:subscribe_to_summary, 0}} (which matched _)
     code: refute_receive _
     stacktrace:
       test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs:103: (test)
```

The test already has at the top:

```elixir
    assert_receive {VehiclesMock, {:subscribe_to_summary, 0}}
```

brianmay:
Test now works, but I don't understand why. Seems dodgy.

brianmay:
Not sure how the tests were even passing before. All fixed now.

brianmay:
Problem with this is that there are times when we have the latitude, longitude, but not the elevation. So as a result nothing gets published. Arrgh.

Think I will remove elevation. I think the Tesla API treats elevation is very different from latitude and longitude anyway.

brianmay:
CI tests are sometimes randomly passing when they should be failures. e.g. I missed updating a test, and somehow it was passing.

brianmay:
Seems to be working so far.

latitude/longitude are being encoded as string values, not floats. Which is a bit unexpected.

I might try to get them as floats.

brianmay:
teslamate uses Decimal type everywhere for latitude/longitude.

Suspect this is a misunderstanding here. Decimal is required for applications that keep track of numbers - like $ and cents, where it is vital that we represent values like 0.01 exactly.

But this is not one of those applications. It doesn't matter really if a latitude of 0.01 gets converted to 0.0099999997 (https://hexdocs.pm/elixir/1.13.4/Float.html). In fact it is likely the values came out of floating point numbers in Tesla's code anyway.

But will keep it as is for now.

Values are now coming out as floats. And tests appear to be passing.

Early on one of the tests passed, which I hadn't even updated. That was weird. Fixed now.

JakobLichterfeld:
Thanks for your deditcation!

brianmay:
Database values are stored in decimal, since 45d94d52b381f9183bdc6de1995f2c94c481b794. Supposedly this makes things more efficient.

Values from the TeslaApi are retrieved at floats.

I think I am will put all the Decimal types back and have this new code deal with both cases.

brianmay:
We seem to have an unrelated intermittent test failure. https://github.com/teslamate-org/teslamate/actions/runs/8240221994/job/22535198738?pr=3729

brianmay:
Sorry, didn't mean to close, pushed the wrong thing.

brianmay:
Testing again. But next drive not scheduled for this afternoon.

brianmay:
This appears to be working correctly now. If people are happy with the decisions I made, it can be merged.

brianmay:
Just noticed we always publish this every time. We should be checking if the value has changed since the last time. Will fix.

brianmay:
See #3736 concerning the test failures,

All the tests passed on my box :-)

brianmay:
This seems to be working correctly now.

JakobLichterfeld:
Test issue should be resolved by #3733. If you don't mind, upate from master

brianmay:
Rebased against latest master.

Not 100% comfortable adding 30 extra lines to the `handle_info` function call, but I think to early to refactor.

Refactoring probably should happen if we are going to output other stuff as json blobs however.

brianmay:
Had to retry CI tests three times to get a pass :-(

I think they work OK from my box however.

JakobLichterfeld:
Will merge as soon as I came up with a good changelog for this, as it will break some third party integrations

brianmay:
Shouldn't cause any breakage, the current topics remain unchanged.

Unless I missed something here?

JakobLichterfeld:
> Shouldn't cause any breakage, the current topics remain unchanged.
> 
> Unless I missed something here?

You are absolutely right, and I did not look close enough. Will double-check and merge tomorrow.

Thanks for your dedication!

JakobLichterfeld (approved):
lgtm, I added the doc entry for the new topic

vbarrier:
@tobiasehlert I think this new topic, generate a lot of warning on teslamateapi

tobiasehlert:
> @tobiasehlert I think this new topic, generate a lot of warning on teslamateapi

@vbarrier, noticed some errors as well at home. Going to add a fix very soon 👌

Issue #3727, opened 2024-03-10
Different values for Logged Distance


RamonMir20:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

The "Logged Distance" value is different in the "Efficiency" panels compared to "Battery Health", what is the correct value? thank you.

### Expected Behavior

The value should be unique across all panels.

### Relevant log output

```shell
*
```

### Type of installation

Docker

### Version

v1.28.3

cwanja:
One comes from `positions` (Efficiency) and the other comes from `drives` (Battery Health). 

Personally think drives is more accurate. Thoughts @JakobLichterfeld ? @jheredianet any reason you used `positions`?

jheredianet:
> One comes from `positions` (Battery Health) and the other comes from `drives` (Efficiency).
> 
> Personally think drives is more accurate. Thoughts @JakobLichterfeld ? @jheredianet any reason you used `positions`? It also takes longer to pull from `positions`.

Yes, it is as you described. Except that it takes longer to pull, because it's just a query pulling from: **(max(odometer) - min(odometer)**. It takes milliseconds no matter the thousands record you have. 

The reason why I've decided to do it so, it's because I think it's better to know the distance the car has taken since Teslamate began to be used, than how many kilometers or miles have been saved in the database. The difference between one value and another is what could not be saved, either due to disconnection from the car, areas without coverage or because your Teslamate has been out of service.

This question makes me think that, to avoid confusion, it would be good for both values to be in this Dashboard, so that the user has at hand the information about what could not be collected in the database.
I will make the necessary changes so that it can be seen in a future version.

jheredianet:
This is the PR https://github.com/teslamate-org/teslamate/pull/3741

cwanja:
Will be reflected in a future release @RamonMir20. Details in the pull request.

PR #3741, opened 2024-03-16
Added Distance (Logged - Mileage) stats

jheredianet:
This is an improvement to Distance stats to avoid confusing on Mileage and logged distance in TeslaMate as discussed here https://github.com/teslamate-org/teslamate/issues/3727

Now, we can see them separately and its corresponding hint.

image

### DIFF
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -70,7 +70,6 @@
   "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1689940704136,
   "links": [
     {
       "icon": "dashboard",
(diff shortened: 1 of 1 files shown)


cwanja:
Amazing @jheredianet. Thanks so much. I am about to dive more into your dashboards and add them separately.

manikoutai:
Just to mention that I can see the feature inside of "Battery Health" of version 1.28.4 but it missing from "Battery Health - LFP".  Thanks, it is a good addition.

JakobLichterfeld:
> Just to mention that I can see the feature inside of "Battery Health" of version 1.28.4 but it missing from "Battery Health - LFP".  Thanks, it is a good addition.

Yeah, that's expected but not ideal.

PR #3744, opened 2024-03-17
Update timeline.json: column Action larger width

DrMichael:
The Action column is too small (at least for me). Perhaps due to the new Grafana version? So, I bump the width up

### DIFF
--- a/grafana/dashboards/timeline.json
+++ b/grafana/dashboards/timeline.json
@@ -182,7 +182,7 @@
             "properties": [
               {
                 "id": "custom.width",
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
On 1.28.3 column width looks good. Are you on edge?

DrMichael:
> On 1.28.3 column width looks good. Are you on edge?

Right. With latest:
image

With egde:
image

JakobLichterfeld:
Thanks for clarification

PR #3757, opened 2024-03-21
Update overview.json: add time range to states link

DrMichael:
When following the States link, the time range should be preserved...


JakobLichterfeld (approved):
lgtm, ty!

PR #3758, opened 2024-03-21
Update overview.json: add time range to charges links

DrMichael:
When following the charges links, the time range should be preserved...


JakobLichterfeld (approved):
lgtm, ty!

PR #3771, opened 2024-03-24
Add URL_PATH environment variable for reverse proxy deployments

b1c1l1:
For example, this can be used to run TeslaMate behind a reverse proxy, using a URL path such as "/teslamate".


JakobLichterfeld (changes requested):
Thanks for your changes!
Lgtm.

Please update the pot file with `mix gettext.extract --merge`

JakobLichterfeld:
Imo path config is done on reverse proxy side. So I do not see a need for this additional config on TeslaMate side.

b1c1l1:
Thanks for the review!  I added a commit to update the gettext files.

For context, here is an example HAProxy configuration that hosts TeslaMate at path `/teslamate`: [haproxy.cfg.txt](https://github.com/teslamate-org/teslamate/files/14737487/haproxy.cfg.txt).  It can be run using `haproxy -f haproxy.cfg`.

The issue I had was that links generated by the backend would not include the prefix.  For example, the app would redirect to `/sign_in` instead of `/teslamate/sign_in`, and fail to fetch JavaScript at `/assets/app.js` instead of `/teslamate/assets/app.js`.

If there is another way to accomplish the same thing, please let me know!  Currently I am running with this patch locally using HAProxy.

brianmay (approved):
It looks like these change allow hosting teslamate at any path, not just `/`.

I agree in concept to these changes. We do need to be careful that it works as requested and hasn't broken anything.

brianmay:
mix format is complaining...

JakobLichterfeld (approved):

JakobLichterfeld:
pot file still out of date

b1c1l1:
Sorry about that, mix format updated the template, which looks like it requires updating the gettext files again.  Added another commit.

JakobLichterfeld (approved):

JakobLichterfeld:
> hello ! how to configure teslamate for using behind a reverse proxy ? i tried using the new URL_PATH without success, always get 404 I added to my systemd service the following line: Environment="URL_PATH=/teslamate"
> 
> in the log i get 2024-05-15 11:12:21.003 [info] GET /teslamate 2024-05-15 11:12:21.004 [info] Sent 404 in 112µs
> 
> what i'm doing wrong ?

You can either follow: https://docs.teslamate.org/docs/guides/traefik or if you want to change the url_path you can use the environment variable to change it, see https://docs.teslamate.org/docs/configuration/environment_variables

Please open a separate issue, if you need more assistance

PR #3772, opened 2024-03-24
fix: permissions denied error when specifying user

### LINKED ISSUE #3769: v1.28.4 fails to start with Permission denied error

### What happened?

After upgrading to v1.28.4 container fails to start with error:

`
/entrypoint.sh: 14: bin/teslamate: Permission denied
`

My docker-compose file for teslamate is below:

```
  teslamate:
    image: teslamate/teslamate:1.28.4
    restart: always
    container_name: teslamate
    environment:
      - ENCRYPTION_KEY=XXXXX
      - DATABASE_USER=teslamate
      - DATABASE_PASS=XXXXX
      - DATABASE_NAME=teslamate
      - DATABASE_HOST=database
      - MQTT_HOST=mosquitto
    ports:
      - 4000:4000
    user: "1003:1007"
    volumes:
      - /opt/docker/volumes/teslamate/import:/opt/app/import
    cap_drop:
      - all
```

Downgrading to v1.28.3 brought teslamate back to life again. I'm assuming it might be something to do with the fact I run it in its own user/group `user: "1003:1007"`?

### Expected Behavior

I'd expect teslamate to start :joy:

### Steps To Reproduce

- Upgrade image to v1.28.4
- Observe

### Relevant log output

```shell
/entrypoint.sh: 14: bin/teslamate: Permission denied
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.28.4

tobiasehlert:
Fix `Permission denied` error when launching container with `user` variables reported in #3769.

Changes permissions so that content copied by _Dockerfile_ into `/opt/app` gets both readable and executable for owner, group and everyone.

fix #3769
rel #3740

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,8 +70,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     chown -R nonroot:nonroot .
 
 USER nonroot:nonroot
-COPY --chown=nonroot:nonroot --chmod=544 entrypoint.sh /
-COPY --from=builder --chown=nonroot:nonroot --chmod=544 /opt/built .
+COPY --chown=nonroot:nonroot --chmod=555 entrypoint.sh /
+COPY --from=builder --chown=nonroot:nonroot --chmod=555 /opt/built .
 RUN mkdir $SRTM_CACHE
 
 EXPOSE 4000


brianmay:
This line copies a lot of files, I think only `/entrypoint.sh`, `/opt/app/bin/teslamate`, `/opt/app/releases/1.28.4/elixir` need to be executable.  Probably `/opt/app/releases/1.28.4/iex` also. Directories need to be executable also. Everything else though should not be execute permission.

Not sure I like using `--chmod` here. But unfortunately all the solutions are crap. The most obvious solution - using something like  `RUN chmod` is also bad, because, from memory, the docker container format can't represent permission changes in an efficient manner.  IIRC this results in every file being included in full again in the new Docker image layer.

You can also chmod in the entrypoint on starting the container. Have done this, but hate it. Slows down every start operation.

Maybe the best solution seems to be along the lines of "make sure the permissions in the source files are correct before the copy". But in practise, this can vary depending on values such as the users umask is when building the package.

So maybe the best compromise is to make everything executable. Still seems yuck though...

tobiasehlert:
@brianmay, you got a valid point!
I considered the _--exclude_ part to remove files that should have another permission level and also played around with specifying lot of files, but we only end up with lots of additional commands that can get messy.

Therefore I concluded that the `555` level would be the easiest way forward for now, even if it could be a bit more tight for sure permission-vise.. at the end we don't want to have write-permissions assigned to executables.

JakobLichterfeld (approved):

PR #3789, opened 2024-03-28
Publish extra active_route fields to mqtt output

### LINKED ISSUE #3748: MQTT values for active_route_* are not clearing

### What happened?

The active route MQTT values are not resetting after arriving to destination. I navigated to the store - it showed the values correctly. But then when I drove home without navigation - the values remained from the store.

### Expected Behavior

It should clear out your values once you're parked.

### Steps To Reproduce

_No response_

### Relevant log output

```shell
None
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

1.28.4

brianmay:
This refactors the existing code to make it more maintainable.

Nil values are published as "nil" string. This ensures that they will get published, and ensures that MQTT doesn't drop the retained data.

This is WIP because it hasn't had any testing yet, and is likely to fail even the CI tests :-)

Fixes #3748

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -6,10 +6,9 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
 
   alias TeslaMate.Mqtt.Publisher
   alias TeslaMate.Vehicles.Vehicle.Summary
-  alias TeslaMate.Locations.GeoFence
   alias TeslaMate.Vehicles
 
-  defstruct [:car_id, :last_summary, :deps, :namespace]
+  defstruct [:car_id, :last_values, :deps, :namespace]
   alias __MODULE__, as: State
 
   def child_spec(arg) do
@@ -38,28 +37,33 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
     {:ok, %State{car_id: car_id, namespace: namespace, deps: deps}}
   end
 
-  @impl true
-  def handle_info(summary, %State{last_summary: summary} = state) do
-    {:noreply, state}
-  end
-
   @always_published ~w(charge_energy_added charger_actual_current charger_phases
                        charger_power charger_voltage scheduled_charging_start_time
                        time_to_full_charge shift_state geofence trim_badging)a
 
+  @impl true
   def handle_info(%Summary{} = summary, state) do
-    summary
-    |> Map.from_struct()
-    |> Map.drop([:car])
+    values =
+      %{}
+      |> add_simple_values(summary)
+      |> add_car_latitude_longitude(summary)
+      |> add_geofence(summary)
+      |> add_active_route(summary)
+
+    publish_values(values, state)
+    {:noreply, %State{state | last_values: values}}
+  end
+
+  defp publish_values(values, %State{last_values: values}) do
+    nil
+  end
+
+  defp publish_values(values, state) do
+    values
     |> Stream.reject(&match?({_key, :unknown}, &1))
     |> Stream.filter(fn {key, value} ->
       (key in @always_published or value != nil) and
-        (state.last_summary == nil or Map.get(state.last_summary, key) != value)
-    end)
-    |> Stream.map(fn
-      {key = :geofence, %GeoFence{name: name}} -> {key, name}
-      {key = :geofence, nil} -> {key, Application.get_env(:teslamate, :default_geofence)}
-      {key, val} -> {key, val}
+        (state.last_values == nil or Map.get(state.last_values, key) != value)
     end)
     |> Task.async_stream(&publish(&1, state),
       max_concurrency: 10,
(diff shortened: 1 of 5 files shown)


brianmay:
On my box I am seeing 5 test failures. I imagine I broke something, but don't have time to check right now.

brianmay:
Looks like there is an error in the documentation also.

brianmay:
Looks like CI tests agree with my tests, there are 5 tests failing. Looks like the same tests are failing also.

brianmay:
Tests now passing. Seems to be a problem with building the docs.

brianmay:
Made some changes.

* There is now an `active_route` topic which publishes all active route information in one json blob. I consider all of these connected and should update atomically.

* When publishing the json blobs, if there is an error, we will output that as a json value too. Meaning value should always be json, should make it easier to parse.

Wondering if we still need the `active_route_*` values now, or if `active_route` by itself is sufficient. Not fussed, but this could be simplified a little bit if I deleted the `active_route_*` values and everything would still be published.

longzheng:
**RE: changes to reset active_route_ topics when not actually routing**
I'm not familiar with Elixer at all but this look reasonable to check `active_route_destination: nil` to reset all the `active_route_` topics to nil.

**RE: adding JSON topics**
It's my understanding that you're adding two JSON-blob topics `active_route` and `active_route_location`. The existing separate topics are still there. This also looks/sounds reasonable to me. I'm not sure if the goal is to eventually deprecate the separate topics, if so we should update the MQTT and Home Assistant (although I just realised I forgot to update this for the active route sensors) documentations.

brianmay:
I just removed the new non-json topics in favour of the json topics. And marked the old topics as deprecated in the docs.

brianmay:
I haven't tested this yet - apart from the CI tests, I imagine that will be the next step :-)

brianmay:
Dump of all published values on start, hopefully nothing is missing :-)

```
teslamate/cars/1/wheel_type AeroTurbine19
teslamate/cars/1/model S
teslamate/cars/1/geofence Home
teslamate/cars/1/active_route {"error":"No active route available"}
teslamate/cars/1/tpms_soft_warning_rr false
teslamate/cars/1/charging_state Stopped
teslamate/cars/1/tpms_soft_warning_fr false
teslamate/cars/1/active_route_destination nil
teslamate/cars/1/longitude xxx
teslamate/cars/1/charger_power 0
teslamate/cars/1/doors_open false
teslamate/cars/1/since 2024-03-31T23:11:18.410000Z
teslamate/cars/1/charge_current_request 32
teslamate/cars/1/exterior_color Red
teslamate/cars/1/locked true
teslamate/cars/1/power 0
teslamate/cars/1/version 2024.2.7
teslamate/cars/1/shift_state (null)
teslamate/cars/1/plugged_in true
teslamate/cars/1/tpms_pressure_fr 3.375
teslamate/cars/1/active_route_longitude nil
teslamate/cars/1/charger_actual_current 0
teslamate/cars/1/windows_open false
teslamate/cars/1/scheduled_charging_start_time (null)
teslamate/cars/1/ideal_battery_range_km 333.6
teslamate/cars/1/healthy true
teslamate/cars/1/charge_limit_soc 80
teslamate/cars/1/spoiler_type None
teslamate/cars/1/is_preconditioning false
teslamate/cars/1/tpms_soft_warning_rl false
teslamate/cars/1/est_battery_range_km 249.42
teslamate/cars/1/frunk_open false
teslamate/cars/1/trim_badging 90D
teslamate/cars/1/tpms_pressure_fl 3.375
teslamate/cars/1/trunk_open false
teslamate/cars/1/is_user_present false
teslamate/cars/1/usable_battery_level 80
teslamate/cars/1/charge_energy_added 0.0
teslamate/cars/1/charger_phases (null)
teslamate/cars/1/tpms_pressure_rl 3.425
teslamate/cars/1/odometer 158835.27
teslamate/cars/1/location {"latitude":xxx,"longitude":xxx}
teslamate/cars/1/charge_port_door_open true
teslamate/cars/1/latitude xxx
teslamate/cars/1/tpms_pressure_rr 3.4
teslamate/cars/1/charge_current_request_max 32
teslamate/cars/1/time_to_full_charge 0.08
teslamate/cars/1/update_available false
teslamate/cars/1/rated_battery_range_km 415.89
teslamate/cars/1/inside_temp 23.9
teslamate/cars/1/heading 13
teslamate/cars/1/state online
teslamate/cars/1/tpms_soft_warning_fl false
teslamate/cars/1/outside_temp 21.0
teslamate/cars/1/charger_voltage 0
teslamate/cars/1/climate_keeper_mode off
teslamate/cars/1/active_route_latitude nil
teslamate/cars/1/is_climate_on false
teslamate/cars/1/battery_level 80
teslamate/cars/1/update_version (null)
```

I censored latitude, longitude; I don't particularly want to publish the location of my home.

Next step will be to dump the values when navigiating.

brianmay:
Finally got around to looking at the new topic when car is navigating. Everything looks good to me.

If no objections, will merge tomorrow.

PR #3794, opened 2024-03-30
Fix broken link from Docker page to upgrading page

brianmay:
I think this was broken in dbf23faa31d84f222c99dba9f5de52c69dd43afb.

Also remove link from header, seems redundant.


JakobLichterfeld:
Thanks, good catch! Netifly checks do not run on direct commits, I will avoid direct commits for docs in the future.

PR #3813, opened 2024-04-03
publish center_display_state to mqtt

Atrox:
I would love to have the ability to track the `center_display_state` from a published topic in mqtt.

BTW: Does something speak against publishing a `vehicle_state` topic with the full tesla json response? I would love to have something like that.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -88,7 +88,7 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
     charger_actual_current charger_voltage version update_available update_version is_user_present
     model trim_badging exterior_color wheel_type spoiler_type trunk_open frunk_open elevation power
     charge_current_request charge_current_request_max tpms_pressure_fl tpms_pressure_fr tpms_pressure_rl tpms_pressure_rr
-    tpms_soft_warning_fl tpms_soft_warning_fr tpms_soft_warning_rl tpms_soft_warning_rr climate_keeper_mode
+    tpms_soft_warning_fl tpms_soft_warning_fr tpms_soft_warning_rl tpms_soft_warning_rr climate_keeper_mode center_display_state
   )a
 
   defp add_simple_values(map, %Summary{} = summary) do
--- a/lib/teslamate/vehicles/vehicle/summary.ex
+++ b/lib/teslamate/vehicles/vehicle/summary.ex
@@ -17,6 +17,7 @@ defmodule TeslaMate.Vehicles.Vehicle.Summary do
     tpms_soft_warning_fl tpms_soft_warning_fr tpms_soft_warning_rl tpms_soft_warning_rr climate_keeper_mode
     active_route_destination active_route_latitude active_route_longitude active_route_energy_at_arrival
     active_route_miles_to_arrival active_route_minutes_to_arrival active_route_traffic_minutes_delay
+    center_display_state
   )a
 
   def into(nil, %{state: :start, healthy?: healthy?, car: car}) do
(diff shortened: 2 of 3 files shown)


brianmay:
Might be best to merge #3789 first...

JakobLichterfeld:
please update your branch

Atrox:
done :)

kolaCZek:
Yes!!! I want this too! I just wanted to create PR. :D Thank you @Atrox !

kolaCZek:
I tried to run the build from your branch (image ghcr.io/teslamate-org/teslamate:pr-3813) and it doesn't seem to work properly. The center_display_state is not published to mqtt...

Atrox:
I'm running the branch as well and it works for me. Please note that TeslaMate only refetches the vehicle data every x seconds (as far as I can tell, if the vehicle is parked and sentry mode its on, its about every 60 seconds?) if your car is streaming the other info.

To test this, keep sentry mode open on your phone and wait until the topic changes. My `center_display_state` changes accordingly.

brianmay:
Possible a dumb question, but exactly what is `center_display_state`?

kolaCZek:
> Possible a dumb question, but exactly what is `center_display_state`?

https://tesla-api.timdorr.com/vehicle/state/vehiclestate ;)

kolaCZek:
> I'm running the branch as well and it works for me. Please note that TeslaMate only refetches the vehicle data every x seconds (as far as I can tell, if the vehicle is parked and sentry mode is on, its about every 60 seconds?) if your car is streaming the other info.
> 
> To test this, keep sentry mode open on your phone and wait until the topic changes. My `center_display_state` changes accordingly.

I tried letting it run for 10 minutes and still nothing. After another wait with the sentry open, I got a 'daily data limit reached' message on my cell phone. :D I'll try to debug it somehow. Obviously it should work.

Atrox:
> I tried letting it run for 10 minutes and still nothing. After another wait with the sentry open, I got a 'daily data limit reached' message on my cell phone. :D I'll try to debug it somehow. Obviously it should work.

Oh okay, 10 minutes is way too long. That is odd. Maybe someone more familiar with teslamate can check my PR and see if I forgot an edge case where the topic doesn't get published?


> Possible a dumb question, but exactly what is `center_display_state`?

I use it to get notified when sentry mode is triggered (which should be a tesla native feature...). Sadly because of the teslamate refetch delay it is not really accurate and I had to build myself a small companion app that fetches the vehicle data more often and handles the sentry notifications.

cwanja:
Does this update TeslaMate's homepage view as well?

kolaCZek:
> I use it to get notified when sentry mode is triggered (which should be a tesla native feature...). Sadly because of the teslamate refetch delay it is not really accurate and I had to build myself a small companion app that fetches the vehicle data more often and handles the sentry notifications.

This is exactly what I wanted to achieve. But it is true that with such a slow refresh rate of TeslaMate it is impossible. I'll also have to write some lightweight application that will only check the center_display_state more often.

lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
```diff
@@ -39,7 +39,8 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
 
   @always_published ~w(charge_energy_added charger_actual_current charger_phases
                        charger_power charger_voltage scheduled_charging_start_time
-                       time_to_full_charge shift_state geofence trim_badging)a
+                       time_to_full_charge shift_state geofence trim_badging
+                       center_display_state)a
```
brianmay, line 43: You probably don't want or need it in the `always_published` list, this just means the value is published even the value is nil. We probably should have a comment to indicate what these list is used for.

The value does need to get included in the `simple_values` list though, below.

JakobLichterfeld (approved):

PR #3834, opened 2024-04-11
Update Thresholds colors for ºF temperature

jheredianet:
A minor fix for missing colored temperature text column (as shown with ºC) in Charges dasboard.

With ºF is shown like this:
image

With ºC is shown like this:
image

With this update, thresholds colors added for ºF temperature as well.


JakobLichterfeld (approved):
Nice catch, thank you for your change!

Issue #3835, opened 2024-04-11
feat request: velocity histogram in drive dashboard


JakobLichterfeld:
In the Drive Dashboard a velocity histogram would be nice.

Screenshots see: https://tff-forum.de/t/einsteiger-teslamate-oder-teslalogger/140361/413?u=ghostreader

### Type of installation

Docker

### Version

v1.28.5

swiffer:
https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards?tab=readme-ov-file#speed-rates

similar to this, right?

JakobLichterfeld:
> similar to this, right?

Yeah, this even looks better imo

jheredianet:
> @swiffer, @jheredianet, @tobiasehlert What do you think about a velocity histogram in drive dashboard?

Hi, sure. Sometime ago, I found myself in the need to know the maximum and minimum speed of a trip in which I was doing some consumption tests and that is why I was inspired and developed this dashboard https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards?tab=readme-ov-file#speed-rates which you just mentioned @swiffer 

So, I guess you're are suggesting to incluide it in TM as I did with the Battery Health, right? ;-)

cwanja:
PS this should have been a discussion and not an issue 😜

jheredianet:
> PS this should have been a discussion and not an issue 😜

I second that ;-)

cwanja:
Not to hijack this thread entirely, but is there a chance to discuss merging parts of your visuals @jheredianet into our native dashboards? Think our dashboards use too many line charts and there are better visuals out there.

JakobLichterfeld:
> So, I guess you're are suggesting to incluide it in TM as I did with the Battery Health, right? ;-)
> 
> Or are you just asking for something like this? (only speeds of one a trip)

The original idea from the german TFF (tesla "drivers" and "firends" forum) was to have the histogram directly in the drive dashboard. 
Your seperate dashboard would be a second step imo.

jheredianet:
> > So, I guess you're are suggesting to incluide it in TM as I did with the Battery Health, right? ;-)
> > Or are you just asking for something like this? (only speeds of one a trip)
> 
> The original idea from the German TFF (tesla “drivers” and “friends” forum) was to have the histogram directly in the drive dashboard. Your separate dashboard would be a second step, imo.

Sure, this is it https://github.com/teslamate-org/teslamate/pull/3836

[...]

Issue #3763, opened 2024-03-22
Unexpected Relationship between Range (rated) and Range (est.) in Internal Drives Dashboard, Drive panel


tjhart:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

When viewing the 'Drive' panel of the 'Internal Drives' dashboard, the Range (rated) and Range (est.) graph lines relationship to each other is unexpected.

### Expected Behavior

Given that Range (rated) and Range (est.) are measured in the same unit of measurement, I expect values that are close to each other to be graphed close to each other, and values that are relatively far apart to be visually far apart.

### Steps To Reproduce

1. Install Teslamate 1.28.4 (docker implementation)
2. Provide car details for a 2022 Model 3 Tesla (details are probably irrelevant for this issue, but I can provide a partial record on request)
3. import data from `drives.csv` and `drive-2.csv`
4. import data from `positions.csv` and `positions-2.csv`
5. in Grafana, view `drives` dashboard
6. Select drive with time range 2024-03-10 15:39:27.893 UTC to 2024-03-10 17:50:53.855 UTC (`drive-1` ) or time range 2024-03-21 16:30:02 UTC to 2024-03-21 16:40:52 (`drive-2`)
7. Observe unusual relationship between Range (rated) and Range (est.) lines in `Drive` panel (see screen shots)

### Relevant log output

```shell
teslamate-1  | 2024-03-22 14:10:49.587 [info] Migrations already up
teslamate-1  | 2024-03-22 14:10:51.031 [info] No tzdata release files found in custom data dir. Copying release file from tzdata priv dir.
teslamate-1  | 2024-03-22 14:10:51.074 [info] System Info: Erlang/OTP 26 (jit)
teslamate-1  | 2024-03-22 14:10:51.074 [info] Version: 1.28.4
teslamate-1  | 2024-03-22 14:10:52.272 [info] POST https://auth.tesla.com/oauth2/v3/token -> 200 (1189.841 ms)
teslamate-1  | 2024-03-22 14:10:52.272 [info] Refreshed api tokens
teslamate-1  | 2024-03-22 14:10:52.273 [info] Scheduling token refresh in 6 h
teslamate-1  | 2024-03-22 14:10:52.276 [info] Running TeslaMateWeb.Endpoint with cowboy 2.10.0 at :::4000 (http)
teslamate-1  | 2024-03-22 14:10:52.277 [info] Access TeslaMateWeb.Endpoint at http://localhost
[...]

PR #3836, opened 2024-04-12
Velocity/Speed histogram in drive details dashboard

jheredianet:
Here is the adaptation to the change that has been suggested here https://github.com/teslamate-org/teslamate/issues/3835

Please, I would like, before doing a merge, for someone to manually import this dashboard into their test environment to see if any information or improvements are missing.

image


JakobLichterfeld:
Test with

```
grafana:
       # image: teslamate/grafana:latest
      image: ghcr.io/teslamate-org/teslamate/grafana:pr-3836
```

JakobLichterfeld:
Thank you!

Looks nearly perfect!

Minor changes:
- I would prefer if the Speed Histogram goes from 0 from the left to max on the right
- I would remove the unit on every x axis point
- I would add percentage above the bars

jheredianet:
> Minor changes:
> 
> * I would prefer if the Speed Histogram goes from 0 from the left to max on the right
> * I would remove the unit on every x axis point
> * I would add percentage above the bars (or both, Minutes and percentage)

Ok, I will do the updates suggested. Need a little extra time.

jheredianet:
> Minor changes:
> 
> * I would prefer if the Speed Histogram goes from 0 from the left to max on the right
> * I would remove the unit on every x axis point
> * I would add percentage above the bars (or both, Minutes and percentage)

@JakobLichterfeld  What about now?

image

JakobLichterfeld:
Nearly perfect.
image

If there are lots of different speeds, the x-axis got a bit messy, therefore
> * I would remove the unit on every x axis point

image

jheredianet:
> Remove the unit on every x axis point

What about now? I hope this one to be the perfect one ;-)

image

sdwalker:
Missing units for the speed (an existing issue) but looks good otherwise
mi
image
km
image

Standardize the letter casing for all of the text?
Battery heater -> mixed casing
front left -> lower casing
Outside Temperature -> upper casing

jheredianet:
> Standardize the letter casing

HI, I took the opportunity to consider your suggestions as well using upper casing

JakobLichterfeld:
Awesome, ty!
image

tjhart:
@jheredianet 

As I looked at this PR with regard to #3763, I noticed that the `speed` scale in the Drive panel now extends into the negative range. Is that realistic? Would that data set ever extend below zero? If not, why are we limiting that data to half of the vertical space in the graph? Are there other data sets in that graph that are now constrained to the top half of the graph with this change?

cwanja:
@JakobLichterfeld (or anyone for that matter) what is the syntax to update your YML to pull a specific PR? I forget.

Is it:`image: teslamate/teslamate:pr-####`? This would be `image: teslamate/teslamate:pr-3836`.

jheredianet:
> @jheredianet
> 
> As I looked at this PR with regard to #3763, I noticed that the `speed` scale in the Drive panel now extends into the negative range. Is that realistic? Would that data set ever extend below zero? If not, why are we limiting that data to half of the vertical space in the graph? Are there other data sets in that graph that are now constrained to the top half of the graph with this change?

Yes, as I've explained before, it's not only the speed that is represented in that axis. There are many metrics there. So, the challenge is if we adapt the axis for a specific metric, it would not be well represented for other one, like the Power which is the one that has positive and negative values (and even elevation perhaps). Maybe the Solomonic solution could be separating the familiar metrics in another graph (just as an idea)

JakobLichterfeld:
> what is the syntax to update your YML to pull a specific PR? I forget.

```
grafana:
       # image: teslamate/grafana:latest
      image: ghcr.io/teslamate-org/teslamate/grafana:pr-3836
```

similar for TeslaMate image

JakobLichterfeld:
> will update the contributing doc about this

done, see #3856

cwanja:
Hate to say it, but with this PR the drive graph is completely useless for range in my opinion. The lines are all on top of each other.  Beyond speed and power, the lines mean nothing. I think we should roll this back @JakobLichterfeld @jheredianet.

This is the exact same drive from the screenshots in this comment https://github.com/teslamate-org/teslamate/issues/3763#issuecomment-2015943703
Screenshot 2024-04-17 at 8 27 58 PM

jheredianet:
> Hate to say it, but with this PR the drive graph is completely useless for range in my opinion. The lines are all on top of each other. Beyond speed and power, the lines mean nothing. I think we should roll this back @JakobLichterfeld @jheredianet.

We return to the same thing that has already been mentioned before (at https://github.com/teslamate-org/teslamate/issues/3763#issuecomment-2058403045), in this panel the Y axis shows different units (speed per hour, power, range) and that is why it cannot fit for everything. From my point of view, I don't see a major problem about the ranges, beyond the fact that they are seen above the speed, the important thing for me is that both range lines show that there are no significant changes between them.

I don't think it's necessary to revert the PR, since there are other changes outside of this panel, on the contrary, I think we can try to improve it together. 

Maybe you could try to change is the "center cero" axis in this panel and then modify the Overwrite "Add Standard options > Min 0" you suggested to see if you get what you want. If so, a PR with those changes will be enough.

JakobLichterfeld:
> Hate to say it, but with this PR the drive graph is completely useless for range in my opinion.

A reduce in range of 10 miles is so little, the graph is ok from my point of view.

DrMichael:
Actually, I agree with @cwanja. But I think, the ranges and battery heater are not really relevant for a drive. Spee, SOC and Usable SOC might be the essential informations...
With ranges, batter heater:
image
Without (and centred_zero off):
image

jheredianet:
> .... I think, the ranges and battery heater are not really relevant for a drive. Spee, SOC and Usable SOC might be the essential informations... 

I agree with @DrMichael for a drive is only important Speed, SOC, Power (and even elevation as a right Y axis) as I did in this dashboard https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards?tab=readme-ov-file#tracking-drives

image

That's why in a previous comment I've suggested as an idea, that maybe the Solomonic solution could be separating the familiar metrics in another graph (meaning Ranges and other metrics that do not interfere with units)

cwanja:
> > Hate to say it, but with this PR the drive graph is completely useless for range in my opinion.
> 
> 
> 
> A reduce in range of 10 miles is so little, the graph is ok from my point of view.

@JakobLichterfeld I would agree that 10 miles is little range loss, however when comparing old and new graphs for the exact same drive it got worse from a readability standpoint. No more or less data is present on the new graph, it's just less readable and usable in my opinion.

cwanja:
> Actually, turning off only centered zero makes it also a bit more comprehensible?
> 

I'd align with @DrMichael's third graph here. Would love to test it against the same drive.

jheredianet:
> @jheredianet What is the rationale to have centered zero set to on?

It's because of the "Power" since it could be posistive and negative

DrMichael:
But you don't need `centered zero` for that...

jheredianet:
> But you don't need `centered zero` for that...

You're right, but having in mind that in this case I'm showing the elevation

With no "centered cero", the elevation seems that I'm traveling across a mountain (when in reality there is only a small variation around 100m)
image


With "centered cero" it looks better and elevation seems to be almost as an horizontal line (more accurate in reality and it is in tune with the power):
image

PR #3841, opened 2024-04-14
Add average cost per kWh for AC and DC as separate values

woyteck1:
I wanted to know what is my average cost per kWh for my AC charging so I created two more tiles, one for DC and one for AC average cost per kWh.

I think it's useful.


woyteck1:
I was thinking to also add a pie chart to show how much of the total cost is DC vs AC. 
Let me know if that would be of interest and I'll add it.

JakobLichterfeld:
Thanks for your suggestion.

We plan to integrate the Dashboards from @jheredianet 
Regarding Charging costs: https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards?tab=readme-ov-file#charging-costs-stats
image

woyteck1:
This one looks great but has a little bit different information which is total on DC and total on AC. It's also good, but my changes give the cost per kWh (DC vs AC).
I.e. in my cas my average cost on DC is £0.36 and on AC £0.13 - I expected it to be lower, just didn't know the ratio.

jheredianet:
> This one looks great but has a little bit different information which is total on DC and total on AC. It's also good, but my changes give the cost per kWh (DC vs AC). I.e. in my cas my average cost on DC is £0.36 and on AC £0.13 - I expected it to be lower, just didn't know the ratio.

It's a good idea. I could add that info in that dashboard, or maybe we should think to join them together in only one, but not sure if it's going to be confusing or a lot of information to see in only one place. 

Maybe it's a good idea to have one dashboard pro charge stats (all energy related) and the other for Cost stats (all related to prices).

What I'm sure that should be clear, is that if we are planning to have the same dashboards for also for LFP batteries, IMHO, It is not necessary because they do not provide anything different. If it were up to me, I would directly delete the LFP dashboards and focus more on enriching the information in the native panels we already have.

woyteck1:
Hi, I think it's a great idea to have a cost dashboard as separate and also to not have the LFP dashboard for charging stats (just leave one for battery degradation), or have the battery_type variable in settings and then NMC or LFP listed there. Then change the colours accordingly.

Additionally, I wanted to point out that to find the Supercharger locations and associated charging, it would perhaps be easier to check whether charger manufacturer is Tesla and if it's a Fast charger. (I know there are now OEM Tesla chargers but they technically are superchargers as well.) fast_charger variable is true, when it's DC charging (or at least that's my observation).

jheredianet:
> Hi, I think it's a great idea to have a cost dashboard as separate and also to not have the LFP dashboard for charging stats (just leave one for battery degradation), or have the battery_type variable in settings and then NMC or LFP listed there. Then change the colours accordingly.

@JakobLichterfeld I think you could accept this PR as it is. I've checked the changes and tested; they look OK for me. 
 A question to know what you think, may we take te opportunity to ask @woyteck1 to remove the "charging-stats-lfp.json" file? so, we just have to maintain only one dashboard, because the only difference are colors, nothing related to calculations.

> Additionally, I wanted to point out that to find the Supercharger locations and associated charging, it would perhaps be easier to check whether charger manufacturer is Tesla and if it's a Fast charger. (I know there are now OEM Tesla chargers but they technically are superchargers as well.) fast_charger variable is true, when it's DC charging (or at least that's my observation).
> 
> Happy for you to use the query in my change on the cost page.

@woyteck1 about detecting a "Supercharger" the way to do is by this:
````
 .... (addr.name ILIKE '%supercharger%' OR geo.name ILIKE '%supercharger%')
````
I've tried once doing as you suggest using "Fast charger" field by I've realized that some older Model S that field sometimes in some locations it reports null, strange but happens (look here https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards/discussions/67). So, I'm not sure it's a good idea to do it as you suggest. I think is better to let the user themselves be in charge of maintaining their respective geofences or putting the correct names in their databases.

jheredianet:
> Hi, I think it's a great idea to have a cost dashboard as separate and also to not have the LFP dashboard for charging stats (just leave one for battery degradation), or have the battery_type variable in settings and then NMC or LFP listed there. Then change the colours accordingly.

@JakobLichterfeld I think you could accept this PR as it is. I've checked the changes and tested; they look OK for me. 
 A question to know what you think, may we take te opportunity to ask @woyteck1 to remove the "charging-stats-lfp.json" file? so we just maintain only one dashboard, because the only difference are colors, nothing related to calculations.

> Additionally, I wanted to point out that to find the Supercharger locations and associated charging, it would perhaps be easier to check whether charger manufacturer is Tesla and if it's a Fast charger. (I know there are now OEM Tesla chargers but they technically are superchargers as well.) fast_charger variable is true, when it's DC charging (or at least that's my observation).
> 
> Happy for you to use the query in my change on the cost page.

@woyteck1 about detecting a "Supercharger" the way to do is by this:
````
 .... (addr.name ILIKE '%supercharger%' OR geo.name ILIKE '%supercharger%')
````
I've tried once doing as you suggest using "Fast charger" field by I've realized that some older Model S reports that field sometimes as "null" in some locations, it's strange but happens (look here https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards/discussions/67). So, I'm not sure it's a good idea to do it as you suggest. I think is better to let the user themselves be in charge of maintaining their respective geofences or putting the correct names in their databases.

JakobLichterfeld (approved):

PR #3844, opened 2024-04-15
Change car id selector by row on Charge Details dashboard

jheredianet:
Minor change of car_id selector on favor of row title, since this dashboard in order to work must be called by drives dashboard, and this selector doen't make sense here:

Before:
image

Now:
image


JakobLichterfeld (approved):
ty

Issue #3846, opened 2024-04-16
Timeline kWh is incorrect for miles


sdwalker:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Timeline kWh is incorrect for miles due to the missing 1.60934 multiplication for driving and parking

### Relevant log output

```shell
N/A
```

### Screenshots

Most recent drive
![image](https://github.com/teslamate-org/teslamate/assets/904020/62f3b5a9-2d69-4e03-96d0-d7a0ca11a33a)

Without multiplication
![image](https://github.com/teslamate-org/teslamate/assets/904020/2e2fd2c5-816a-47eb-a563-dfcddfa48fc4)

With multiplication
![image](https://github.com/teslamate-org/teslamate/assets/904020/a2607be8-ff32-4981-80dc-6755598abd6b)

### Type of installation

Manual

### Version

v1.28.6-dev

JakobLichterfeld:
related PR #3847

PR #3847, opened 2024-04-16
Timeline: fix kWh values and column layout for miles (#3846)

sdwalker:
Add the kWh multiplication needed for driving and parking when using miles
Use the same index value for the Range Diff column so the layout is the same with miles or kilometers


DrMichael:
May I propose to use convert_km instead of CASE in the SQL statement? Just to be consistent.

sdwalker:
> May I propose to use convert_km instead of CASE in the SQL statement? Just to be consistent.

convert_km divides by 1.60934 so the proper fix for charging looks to be to not use convert_km at all

From
convert_km((end_[[preferred_range]]_range_km - start_[[preferred_range]]_range_km)::NUMERIC, '$length_unit') * efficiency AS "kWh",
To
(end_[[preferred_range]]_range_km - start_[[preferred_range]]_range_km)::NUMERIC * efficiency AS "kWh",

Still looking at how to fix the parking query
convert_km(((LEAD(d.start_[[preferred_range]]_range_km) over w + (LEAD(d.start_km) over w - d.end_km)) - d.end_[[preferred_range]]_range_km)::NUMERIC, '$length_unit') * efficiency AS "kWh",

JakobLichterfeld:
> Still looking at how to fix the parking query

Are you planning a separate PR or include in this one?

sdwalker:
Can we assume the parking query is correct for metric users?
If so, a metric user would be able to remove the convert_km from the parking query, switch their units and confirm the unit conversion matches

swiffer:
efficiency is stored as kwh/km, therefore convert_km seems not to be needed at all in these calculations, agree.

@dyxyl - do you want to update your PR based on latest master or should I open a follow up PR ?

swiffer (changes requested):
efficiency is stored as kwh/km, therefore convert_km seems not to be needed at all in these calculations, agree.

@dyxyl - do you want to update your PR based on latest master or should I open a follow up PR ?

dyxyl:
I don't have a PR for this -- it was @sdwalker who originated it.

swiffer:
😆 - sorry. @sdwalker, would you like to update the PR ?

swiffer:
with #4367 merged this one can be closed @JakobLichterfeld

PR #3850, opened 2024-04-17
Fix missing units when miles is used

sdwalker:
Charges
image
Drive Details
image
Dutch Tax
image

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -638,14 +638,18 @@
               },
               {
                 "id": "unit",
-                "value": "none"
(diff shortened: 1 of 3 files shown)


cwanja:
Assuming this respects the variable the user sets in the `settings` table?

sdwalker:
> Assuming this respects the variable the user sets in the `settings` table?

Correct - switching to the metric units and comparing the dashboards in a duplicated tab is how the imperial discrepancies are being found

cwanja (approved):
Verified and confirmed. Looks good. Thanks @sdwalker.

PR #3854, opened 2024-04-17
Statistics: fix temperature colors and column layout for Fahrenheit

sdwalker:
Add the missing temperature color thresholds when using Fahrenheit
Use the same index value for the temperature column so the layout is the same with Fahrenheit or Celsius

image


cwanja:
My only concern with this is the 'colors' vary by region. Red at 70° in South Texas is wrong. Should be green

PR #3863, opened 2024-04-20
Add-more-ways-to-recognize-Supercharger-locations

woyteck1:
Add more ways to recognize Supercharger locations by treating charge as "at Superchager" when fast charger brand is "Tesla".

I have tested it on my recent trip, and the total cost for Supercharger charges is the same vs what I calculated manually.

### DIFF
--- a/grafana/dashboards/charging-stats-lfp.json
+++ b/grafana/dashboards/charging-stats-lfp.json
@@ -276,7 +276,8 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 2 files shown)


woyteck1:
For all of my data, it added over £200 worth of charges to the count, in line witch my charging payments on my Tesla account.

woyteck1:
Hi, @JakobLichterfeld, do you think can be used?

JakobLichterfeld:
> Hi, @JakobLichterfeld, do you think this can be used?

Generally speaking: yes. Perhaps @jheredianet can take a look into it.

woyteck1:
If the old vehicles do not recognize the "Tesla" brand, then the output visible will just be like the current one and they will still need to do a Supercharger Geo-fences. I think.

jheredianet:
> If the old vehicles do not recognize the "Tesla" brand, then the output visible will just be like the current one and they will still need to do a Supercharger Geo-fences. I think.

Yes, you are right. I've tested and worked ok. 

@JakobLichterfeld lgtm

JakobLichterfeld:
Perfect, thank you both

PR #3866, opened 2024-04-24
Dynamic endpoints and token to use official Tesla API (self-hosted or from third party provider)

jlestel:
I introduced dynamic endpoints and a token to maxime compatibility for future Tesla API/streaming change:
- dynamic endpoints based on ENV vars allows us to connect self-hosted Tesla Fleet API or homemade streaming/testing API
- an optional `token` param to use official Tesla API and Telemetry via by a third party:
  - It is because the `access_token` and `refresh_token` are refreshed by these third parties that this token parameter is mandatory. Thus TeslaMate does not lose access and the API keys are managed on the third party's site.
  - Here are the 2 actors who operate according to this principle: [MyTeslamate (free)](https://app.myteslamate.com/) and [Teslemetry (24$/year)](https://teslemetry.com/)
  - When a `token` is provided, no inputs are needed when the user sign up

A compatible streaming server is also provided by a third party, the stream only sends events with the minimum interval of 1 minute. The streaming interval is the main regression in this operation with official Tesla Telemetry.

However, the new API also contains new data that can make it possible to create interesting new features.


brianmay:
I have been thinking about doing this myself :-)

I note you have some `TESLA_CN_*` variables. Wondering if maybe these should be just `TESLA_*`.

I think it is unlikely that people will need a mixture of China vs non-China variables, and this makes it more complicated trying to figure out which variables need to be set.

lib/tesla_api/auth/refresh.ex
```diff
@@ -15,7 +15,11 @@ defmodule TeslaApi.Auth.Refresh do
       refresh_token: auth.refresh_token
     }
 
-    case post("#{issuer_url}/token", data) do
+    case post(
```
brianmay, line 18: I think I might be inclined to use:

```elixir
endpoint_url = System.get_env("TESLA_AUTH_URL", issuer_url) <> "/token?token=" <> System.get_env("TOKEN", "")
case post(endpoint_url) do
```

I think it is more readable that way.

Just curious, why did you need to add the `token=` parameter here?

jlestel:
> I have been thinking about doing this myself :-)
> 
> I note you have some `TESLA_CN_*` variables. Wondering if maybe these should be just `TESLA_*`.
> 
> I think it is unlikely that people will need a mixture of China vs non-China variables, and this makes it more complicated trying to figure out which variables need to be set.

I made the same observation. 

But since we only know a user's region after login, the same settings will work for all users. This is why I did not factor in a single environment variable.

JakobLichterfeld (approved):
ty, lgtm

JakobLichterfeld:
@jlestel 

Not working on my end without any env variables

> teslamate            | 2024-05-20 14:03:49.632 [notice] Application teslamate exited: TeslaMate.Application.start(:normal, []) returned an error: shutdown: failed to start child: TeslaMate.Api
> teslamate            |     ** (EXIT) an exception was raised:
> teslamate            |         ** (MatchError) no match of right hand side value: "https://auth.tesla.com/oauth2/v3"
> teslamate            |             (teslamate 1.29.0) lib/tesla_api/auth/refresh.ex:12: TeslaApi.Auth.Refresh.refresh/1
> teslamate            |             (teslamate 1.29.0) lib/teslamate/api.ex:220: TeslaMate.Api.refresh_tokens/1
> teslamate            |             (teslamate 1.29.0) lib/teslamate/api.ex:116: TeslaMate.Api.init/1
> teslamate            |             (stdlib 5.2) gen_server.erl:980: :gen_server.init_it/2
> teslamate            |             (stdlib 5.2) gen_server.erl:935: :gen_server.init_it/6
> teslamate            |             (stdlib 5.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3

JakobLichterfeld:
same in #3897 3897

jlestel:
> @jlestel
> 
> Not working on my end without any env variables. Will revert as it breaks backwards compatibility
> 
> > teslamate            | 2024-05-20 14:03:49.632 [notice] Application teslamate exited: TeslaMate.Application.start(:normal, []) returned an error: shutdown: failed to start child: TeslaMate.Api
> > teslamate            |     ** (EXIT) an exception was raised:
> > teslamate            |         ** (MatchError) no match of right hand side value: "https://auth.tesla.com/oauth2/v3"
> > teslamate            |             (teslamate 1.29.0) lib/tesla_api/auth/refresh.ex:12: TeslaApi.Auth.Refresh.refresh/1
> > teslamate            |             (teslamate 1.29.0) lib/teslamate/api.ex:220: TeslaMate.Api.refresh_tokens/1
> > teslamate            |             (teslamate 1.29.0) lib/teslamate/api.ex:116: TeslaMate.Api.init/1
> > teslamate            |             (stdlib 5.2) gen_server.erl:980: :gen_server.init_it/2
> > teslamate            |             (stdlib 5.2) gen_server.erl:935: :gen_server.init_it/6
> > teslamate            |             (stdlib 5.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3

Sorry I thought I checked it. I fix it ASAP to submit a new PR.

florianfish:
Hi @jlestel my teslamate dont work since this night. Can i test your PR ? I used docker version

jlestel:
> Hi @jlestel my teslamate dont work since this night. Can i test your PR ? I used docker version

Obviously, you can setup with instructions from [MyTeslaMate](https://app.myteslamate.com/fleet).

PR #3867, opened 2024-04-24
Update drives.json: show drives with duration=0 but distance>0

DrMichael:
For discussion.
I would like to show also drives, where a duration of 0 is stored, but the distance is not 0. Kind of infinite speed... :-)

I just added  `OR (duration_min = 0 AND distance  >= 0)` to the WHERE clause.


jlestel:
Do these "0 trips" not risk distorting other stats based on the number of trips?

If so, documentation for cleaning up these trips would be sufficient for me.

DrMichael:
> Do these "0 trips" not risk distorting other stats based on the number of trips?
They are not completely zero, they have a distance. I ran into it working on the Statistics dashboard, which did show them, but Drives did not. Either we do this change or I adapt the Statistics dashboard.

> If so, documentation for cleaning up these trips would be usefull for me.
I also realised some invalid drives and some charging processes (no end_date or no significant data). Maybe removing those could be added to the documentation...

jlestel:
> > Do these "0 trips" not risk distorting other stats based on the number of trips?
> 
> They are not completely zero, they have a distance. I ran into it working on the Statistics dashboard, which did show them, but Drives did not. Either we do this change or I adapt the Statistics dashboard.
> 
> > If so, documentation for cleaning up these trips would be usefull for me.
> 
> I also realised some invalid drives and some charging processes (no end_date or no significant data). Maybe removing those could be added to the documentation...

Thank you for the clarification. I already fix incomplete with https://docs.teslamate.org/docs/maintenance/manually_fixing_data/ but maybe it’s another command to fix your invalid drives / charges

DrMichael:
Ah, distance should be greater than zero...

DrMichael:
@JakobLichterfeld I still think, drives with distance > 0 should be shown. Any objections?

JakobLichterfeld:
> @JakobLichterfeld I still think, drives with distance > 0 should be shown. Any objections?

I think so too. Was wondering if this is finished or not as your comment about > 0 wasn`t followed by a commit

JakobLichterfeld (approved):

PR #3868, opened 2024-04-25
fix(geocoder): resolve state for Australian territories

mattew124:
The address returned by the following link returns a territory in place of state so I have added state aliases to include this information:

https://nominatim.openstreetmap.org/reverse?format=xml&lat=-35.1604&lon=149.1049&zoom=19&addressdetails=1

### DIFF
--- a/lib/teslamate/locations/geocoder.ex
+++ b/lib/teslamate/locations/geocoder.ex
@@ -136,6 +136,13 @@ defmodule TeslaMate.Locations.Geocoder do
     "department"
   ]
 
+  @state_aliases [
+    "state",
+    "province",
+    "territory",
+    "state_code"
+  ]
+
   defp into_address(%{"error" => "Unable to geocode"} = raw) do
     unknown_address = %{
       display_name: "Unknown",
@@ -169,7 +176,7 @@ defmodule TeslaMate.Locations.Geocoder do
       city: raw["address"] |> get_first(@city_aliases),
       county: raw["address"] |> get_first(@county_aliases),
       postcode: get_in(raw, ["address", "postcode"]),
-      state: raw["address"] |> get_first(["state", "province", "state_code"]),
+      state: raw["address"] |> get_first(@state_aliases),
       state_district: get_in(raw, ["address", "state_district"]),
       country: raw["address"] |> get_first(["country", "country_name"]),
       raw: raw
--- a/test/teslamate/locations/geocoder_test.exs
+++ b/test/teslamate/locations/geocoder_test.exs
@@ -126,6 +126,109 @@ defmodule TeslaMate.Locations.GeocoderTest do
     end
   end
 
+  test "resolves the state for an Australian territory" do
+    with_mocks [
+      geocoder_mock(-35.1604, 149.1049, %{
+        "address" => %{
+          "road" => "Burrumarra Avenue",
+          "suburb" => "Ngunnawal",
+          "town" => "District of Gungahlin",
+          "territory" => "Australian Capital Territory",
+          "ISO3166-2-lvl4" => "AU-ACT",
+          "postcode" => "2913",
+          "country" => "Australia",
+          "country_code" => "au"
+        },
+        "addresstype" => "road",
+        "boundingbox" => ["-35.1605549", "-35.1598870", "149.1032096", "149.1053282"],
+        "category" => "highway",
+        "display_name" =>
+          "Burrumarra Avenue, Ngunnawal, District of Gungahlin, Australian Capital Territory, 2913, Australia",
+        "extratags" => %{
+          "lit" => "yes",
+          "lanes" => "2",
+          "surface" => "paved",
+          "cycleway" => "lane",
+          "maxspeed" => "50",
+          "lanes:forward" => "1",
+          "maxspeed:type" => "sign",
+          "lanes:backward" => "1"
+        },
+        "importance" => 0.05340591515983822,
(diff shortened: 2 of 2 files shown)


mattew124:
Once this change is applied is it possible to update existing data?

JakobLichterfeld:
Thanks for your suggestion!

> Once this change is applied is it possible to update existing data?

You can code a migration step which can add it for old data, not sure about the performance in this case tbh.

DrMichael:
@JakobLichterfeld I'd say, this PR makes sense....

JakobLichterfeld:
> @JakobLichterfeld I'd say, this PR makes sense....

Sure, but it is neither complete, nor backward compatible, nor proven to work worldwide.

JakobLichterfeld:
> nor proven to work worldwide

No Territory for adresses in germany. For example: https://nominatim.openstreetmap.org/reverse?format=xml&lat=48.95662066509727&lon=9.245291956117336&zoom=19&addressdetails=1

JakobLichterfeld:
This needed to be updated as well: https://github.com/teslamate-org/teslamate/blob/336ba301a9c83703d342741f32eebbb1496dc10f/test/support/mocks/geocoder.ex

JakobLichterfeld:
Thanks for your patience here — and sorry this sat for so long.

I re-checked your report against live Nominatim today and it's still valid: your coordinate returns `"territory": "Australian Capital Territory"` with no `state` key at all, so TeslaMate stores `state: nil` for the entire ACT. Same for the Northern Territory (Darwin → `"territory": "Northern Territory"`).

I also checked whether `territory` could ever shadow an existing value, since that was my main hesitation back then. Yukon (CA), Puerto Rico (US), Chile, Italy, Kenya and Thailand all use `state` or nothing — only the Australian territories use `territory`, and no address carries both. Because `get_first/2` returns the first non-nil hit and `territory` sits behind `state`/`province`, the change is purely additive: no address that resolves a state today changes its value. So my "not backward compatible / not proven worldwide" concern is settled — your fix is correct.

> Once this change is applied is it possible to update existing data?

Better answer than the one I gave in 2024: no migration needed. `TeslaMate.Locations.refresh_addresses/1` already re-geocodes every stored address, and it runs automatically whenever the language is changed in Settings. Switching the language away and back re-resolves all addresses through the new alias list. A DB migration couldn't do this anyway, since it needs network round trips to Nominatim (rate limited to ~1 request/s).

One thing is still missing before this can be merged:

- a regression test in `test/teslamate/locations/geocoder_test.exs` with an ACT payload — the existing tests mock the Tesla adapter with a full raw Nominatim response, so it's essentially a copy-paste of the JSON above

(Ignore my old comment about `test/support/mocks/geocoder.ex` — I looked again, that stub returns already-mapped address maps and never goes through `into_address/1`, so it doesn't need touching here.)

Since the branch is ~2 years behind `main` and the CLA is still unsigned: rather than asking you to rebase, would you be okay with me rebuilding this on current `main` and keeping your authorship via a `Co-authored-by:` trailer? The idea and the analysis are yours; I'd only add the test and the changelog entry. If you'd rather finish it yourself, that's just as welcome — say the word and I'll leave this open and review it.

JakobLichterfeld (approved):
Perfect, thanks for the quick turnaround — rebase and test both look good.

I checked the fixture against the live Nominatim response and it matches field for field, so it's a real regression guard rather than a synthetic payload. The expectations are right too: `city` resolves via `town`, `neighbourhood` via `suburb`, `county` stays `nil`, and `state` now picks up the territory. CI is green on `ff44340`.

I'll add the CHANGELOG entry myself and merge — no need to touch the branch again. Thanks for sticking with this one for two years.

mattew124:
For anyone who already has addresses in their database from before this fix — geocoded back when Nominatim returned `territory` instead of `state`/`province`/`state_code` (Australian territories like ACT/NT are the common case) — those existing rows are stuck with `state = NULL` permanently. TeslaMate caches resolved addresses by `osm_id`/`osm_type` and never re-queries them, so upgrading alone won't backfill historical data.

Since the raw geocoder response is already stored in `addresses.raw`, you can backfill it directly with the same alias priority this PR adds (`state`, `province`, `territory`, `state_code`) — no need to re-query Nominatim:

```sql
-- Preview affected rows first
SELECT count(*) FROM addresses
WHERE state IS NULL
  AND COALESCE(raw->'address'->>'state', raw->'address'->>'province', raw->'address'->>'territory', raw->'address'->>'state_code') IS NOT NULL;

-- Apply the backfill
UPDATE addresses
SET state = COALESCE(raw->'address'->>'state', raw->'address'->>'province', raw->'address'->>'territory', raw->'address'->>'state_code')
WHERE state IS NULL
  AND COALESCE(raw->'address'->>'state', raw->'address'->>'province', raw->'address'->>'territory', raw->'address'->>'state_code') IS NOT NULL;
```

For a Docker Compose setup:
```bash
docker compose exec -T database psql -U teslamate -d teslamate -c "UPDATE addresses SET state = COALESCE(raw->'address'->>'state', raw->'address'->>'province', raw->'address'->>'territory', raw->'address'->>'state_code') WHERE state IS NULL AND COALESCE(raw->'address'->>'state', raw->'address'->>'province', raw->'address'->>'territory', raw->'address'->>'state_code') IS NOT NULL;"
```

Only touches rows where `state IS NULL`, so it can't overwrite an existing value. Tested against a live instance — backfilled 310 rows with no issues.

PR #3877, opened 2024-05-02
docs: Update manually_fixing_data.mdx

cwanja:
Updated docs to include script to close multiple charges and drives in a single file. Credit @dmachadol via #2364


JakobLichterfeld (changes requested):
Thanks for the suggestion.
I'm against this PR, because:
- The docu is not idemponent
- the script use sudo docker ...
- the bash style is not system neutral
- mentioning a single person for credit

mews-se:
This should be revisited.

PR #3884, opened 2024-05-06
Upgrade PostgreSQL to v16

enoch85:
Just tested, and seems to work!


enoch85:
> pg_uprade

Yeah that could be added to the documented steps of upgrading PostgreSQL as well. Want me to fix? Not 100% sure about the command, but will check.

JFTR, I checked all my statistics from 2022 --> now, and they seem to work, everything is there afaik. Will do a `pg_upgrade`now just in case though, good advice! :+1:

enoch85:
OK, so it wasn't as easy as I thought. Will have to revert now, and do it properly with `pg_dumpall` instead.

Maybe the standard command recomended in documentation should be `pg_dumpall` instead of just `pg_dump`?

enoch85:
All good from me now anyway! :+1:

Dulanic:
Now you understand the pain 🤣 I personally run a separate postgres server at I have many containers connect so I just transition when I feel like it. pg_upgrade doesn't work too well with container postgres.

enoch85:
At least this PR could now be merged as it does what proposed - switch to PGSQL 16 and upgrade properly.

parkr:
What are the benefits of 16 over 15? As a user with several years of data, I'd be hesitant to upgrade unless there were benefits.

enoch85:
> What are the benefits of 16 over 15? As a user with several years of data, I'd be hesitant to upgrade unless there were benefits.

Well, newer = better. :) I haven't gone into details, but updating is the best protection against old and unsecure code.

Since it's "easy" to upgrade, why not? I also think that the default for new installs should be the latest version.

enoch85:
@JakobLichterfeld What do you say, can I get your approval on this?

website/docs/maintenance/backup_restore.md
```diff
@@ -10,7 +10,7 @@ If you are using `docker-compose`, you are using Docker Compose v1, which has be
 Create backup file `teslamate.bck`:
 
 ```bash
-docker compose exec -T database pg_dump -U teslamate teslamate > /backuplocation/teslamate.bck
+docker compose exec -T database pg_dumpall -U teslamate > /backuplocation/teslamate.bck
```
JakobLichterfeld, line 13: This will work for upgrading to postgres 16, but not if someone comes from postgres 12 or older I beliefe.

JakobLichterfeld (changes requested):

JakobLichterfeld:
Thanks for your suggestion!
I knew, the time will come to update our db container. I'm with Dulanic, this need additional testing (our CI does not cover backup and restore for example), given the fact some users are still on very old Postgres versions (as we can see in the amount users who post their docker-compose.yml in issues).

enoch85:
> Thanks for your suggestion! I knew, the time will come to update our db container. I'm with Dulanic, this need additional testing (our CI does not cover backup and restore for example), given the fact some users are still on very old Postgres versions (as we can see in the amount users who post their docker-compose.yml in issues).

Hmm, I don't see the issue here actually. You set a tag in your docker-compose.yaml, and as long as that's set, nothing will change - the version will stay the same.

If you on the other hand put a specific version (in this case 16) in the docker-compose.yaml file, it will install with 16, and stay as 16 until you change it.

This PR merly suggests that all new users should get 16 directly, instead of 15 as it is today. It won't update current users.

I run PostgreSQL with your Docker setup, I don't have any external DB, and I managed to upgrade according to your documented instructions. Personally, I went from 14 to 16 I'm one go. 🙂

JakobLichterfeld:
Obviously, you get me wrong. Current state is: some users didn't upgrade Postgres every, they stick with old Postgres version. When they post their docker-compose, we often redirect them to our upgrade guide to be future-proof on that end, even if unrelated to the issue they report.

If we change official installation instructions, we must ensure the manual upgrade process runs smooth, in-depended on the version they come from, this can be Postgres 11 for example.

`pg_dump` and `pg_dumpall` are not the same. May you elaborate why changing makes sense? What are the consequences, what are the differences in our migration?

I would love to upgrade to Postgres 16, but we need to ensure all is working as expected, as, as said above, our CI does not cover these cases.

brianmay:
I am guessing that `pg_dumpall` will save some details (e.g. database user) that may not be recorded by `pg_dump`, but you probably do want to keep. Not tested though.

enoch85:
(Migrated) PostgreSQL 16 is still running fine. ;)

BowlesCR:
Fwiw, I've been running on Postgres 16-alpine since mid-February with no issues beyond the requisite backup/restore to migrate.

At risk of starting a tangential discussion, debates to be had on using the alpine variant vs the default bookworm... Without digging into the numbers the size reduction of alpine (only ~200MB) might be negated by no longer sharing the bookworm-slim base layer between the app and DB. That said, our Grafana image is based on alpine as well, so 🤷

enoch85:
> Seems useful to have new users start on the latest version. I'd recommend splitting this PR into (1) the version change and (2) the backup/restore documentation change.
> 
> FWIW I did the upgrade to 16 (from 13) following the existing instructions (no dumpall) and that worked well for me.

Yeah, worked flawlessly for me as well, so I don't see any reason to use `dumpall` really, but I got insecure as mentioned in previous posts.

website/docs/maintenance/backup_restore.md
JakobLichterfeld, line 13: Please add test results for this, comparing db size, db entries and so on, or revert to pg_dump

JakobLichterfeld (changes requested):

JakobLichterfeld:
In addition to my review findings: Please also adapt all other mentions in the docu.

website/docs/maintenance/backup_restore.md
```diff
@@ -10,7 +10,7 @@ If you are using `docker-compose`, you are using Docker Compose v1, which has be
 Create backup file `teslamate.bck`:
 
 ```bash
-docker compose exec -T database pg_dumpall -U teslamate > /backuplocation/teslamate.bck
+docker compose exec -T database pg_dump -U teslamate teslamate > /backuplocation/teslamate.bck
```
JakobLichterfeld, line 13: the `-U teslamate teslamate` has a duplicate

JakobLichterfeld (changes requested):

website/docs/maintenance/backup_restore.md
DrMichael: Hmmm, that is the database name:
```
dbname
Specifies the name of the database to be dumped. If this is not specified, the environment variable PGDATABASE is used. If that is not set, the user name specified for the connection is used.
```
So, yes, if the user name is same as database name, it could be left out.

JakobLichterfeld: You are right, in the first commits it was removed which confused me

parkr (approved):

enoch85:
> In addition to my review findings: Please also adapt all other mentions in the docu.

Done! :+1:

JakobLichterfeld (approved):
ty!

JakobLichterfeld:
CI is now updated as well, see #3916

enoch85:
Great! 🎊

JakobLichterfeld:
Good point, will add it.

greggitter:
Updated from version 12 to 16 today.  For those using the teslamate manual install on debian (and maybe the same for Docker), this is the procedure I followed.  I haven't seen any problems, if anyone knows why I shouldn't do it this way, please let me know.  Thx.

```
sudo apt install postgresql-16 postgresql-client-16 -y
dpkg -l | grep postgresql (verify install)
sudo service postgresql stop
sudo pg_dropcluster 16 main --stop (drops the new empty 16 cluster)
sudo pg_upgradecluster 12 main
sudo service postgresql start
pg_lsclusters (verify 16 is active)
sudo pg_dropcluster 12 main (permanently deletes v12 cluster - VERIFY before doing this)
sudo apt remove postgresql-12 postgresql-client-12
```

Credit: https://www.youtube.com/watch?v=JvbbI_zxubQ

JakobLichterfeld:
if anyone knows why I shouldn't do it this way, please let me know.  

No migration steps done

greggitter:
OK, thanks...I actually looked and couldn't find the documentation for upgrading.  All I found was this page https://docs.teslamate.org/docs/maintenance/upgrading_postgres but I'm not using docker.

cwanja:
> OK, thanks...I actually looked and couldn't find the documentation for upgrading. All I found was this page https://docs.teslamate.org/docs/maintenance/upgrading_postgres but I'm not using docker.

You won't find _TeslaMate_ documentation on how to upgrade PostgreSQL **manually** because that is not our tool to own.

greggitter:
Apologies, I guess that's not what I meant.  I mean, I am using the MANUAL install of teslamate, so it obviously differs from the DOCKER install...yes?  Maybe there are instructions for the "migration steps" somewhere but I didn't see them after searching for quite some time...including github issues.

JakobLichterfeld:
> Maybe there are instructions for the "migration steps" somewhere but I didn't see them after searching for quite some time...including github issues.

See the restore step in above docu or direct here: 
https://docs.teslamate.org/docs/maintenance/backup_restore

cwanja:
> so it obviously differs from the DOCKER install...yes?

Yes

> Maybe there are instructions for the "migration steps" somewhere but I didn't see them after searching for quite some time...including github issues.

Not sure what you mean by "migration steps"? And for what type of install? Docker is straight forwrad:
1. Backup
2. Shut down TeslaMate
3. Remove PostgreSQL volume
4. Update compose
5. Restart and restore

greggitter:
Thanks but I don't see how this applies since I have neither Docker nor Docker Compose installed.  I have a [native install](https://docs.teslamate.org/docs/installation/debian) in a proxmox lxc.  There is no documentation for upgrading postgresql versions or elixir versions.  Grafana has been updating itself via APT without issue.  I've been running this for nearly a year.  I appreciate the help.

bmerciergallay:
Is upgrading to pgsql 16 mandatory or not ? it's not as clear as it should

enoch85:
> Is upgrading to pgsql 16 mandatory for 1.29.2 or not ? it's not as clear as it should

It's not mandatory, but possible. IMHO you should upgrade.

JakobLichterfeld:
> Is upgrading to pgsql 16 mandatory for 1.29.2 or not ? it's not as clear as it should

If you do not upgrade and open an issue, first answer would be share your docker compose and then upgrade your postgres version. The only supported version is 16 at the time of writing. This means 12 is working fine, but nobody is testing it nor make assumptions about it. 
TeslaMate is not in charge of dependencies, only testing with the versions stated. You can place any parallel here, starting from a car and an owner asking if it is allowed to use his tires without any profile or a cyclist asking if it allowed to run on a flat tire.

bmerciergallay:
I strongly dispute the claim that having a postgresql > 12 and < 16 is like cycling with a flat tire ! 🤣 
got it though...
I'll think about upgrading to 16 when the API landscape will be cleared and certain

PR #3895, opened 2024-05-19
Grafana 11.0.0

swiffer:
not tested yet - angular support is disabled by default in 11 - shouldn't be an issue as it was disabled by env var before anyhow


swiffer:
tested, for me works fine! LGTM

PR #3903, opened 2024-05-21
Dynamic endpoints and token to use official Tesla API (self-hosted or from third party provider)

jlestel:
I introduced dynamic endpoints and a token to maxime compatibility for future Tesla API/streaming change:

dynamic endpoints based on ENV vars allows us to connect self-hosted Tesla Fleet API or homemade streaming/testing API
an optional token param to use official Tesla API and Telemetry via by a third party:
It is because the access_token and refresh_token are refreshed by these third parties that this token parameter is mandatory. Thus TeslaMate does not lose access and the API keys are managed on the third party's site.
Here are the 2 actors who operate according to this principle: [MyTeslamate (free)](https://app.myteslamate.com/) and [Teslemetry (24$/year)](https://teslemetry.com/)
When a token is provided, no inputs are needed when the user sign up
A compatible streaming server is also provided by a third party, the stream only sends events with the minimum interval of 1 minute. The streaming interval is the main regression in this operation with official Tesla Telemetry.

However, the new API also contains new data that can make it possible to create interesting new features.

Initially merged and reverted in [PR 3866](https://github.com/teslamate-org/teslamate/pull/3866).


jlestel:
@JakobLichterfeld I fix the error `no match of right hand side value: "https://auth.tesla.com/oauth2/v3"` on file `lib/tesla_api/auth/refresh.ex:12`.

It was linked to my lack of knowledge about Elixir. I corrected the assignment of the `issuer_url` variable

jlestel:
I tried the `ghcr.io/teslamate-org/teslamate:pr-3903`, seems fine to me. If we have testers, they can try it. I think that was the only typo.

JakobLichterfeld:
> @JakobLichterfeld I fix the error `no match of right hand side value: "https://auth.tesla.com/oauth2/v3"` on file `lib/tesla_api/auth/refresh.ex:12`.
> 
> It was linked to my lack of knowledge about Elixir. I corrected the assignment of the `issuer_url` variable

No worries, neither me nor Brian found it in our review. And I did not test it on my own before merging.

JakobLichterfeld:
@jlestel please update with latest translations from main branch to make merging easier

JakobLichterfeld (approved):

adampetrovic:
Is there any documentation associated with this change?

JakobLichterfeld:
> Is there any documentation associated with this change?

#3959

pkuehnel:
Is there any documentation on how the Fleet API Provider needs serve the Data? Also what is the protocol and how do the interfaces look so TeslaMate can consume Fleet Telemetry data?

JakobLichterfeld:
> Is there any documentation on how the Fleet API Provider needs serve the Data? Also what is the protocol and how do the interfaces look so TeslaMate can consume Fleet Telemetry data?

https://docs.teslamate.org/docs/guides/api

pkuehnel:
What is the long-term goal for this? Will additional fields that are available with Fleet Telemetry be integrated in the future? Will it forever be with the 'faked' streaming API, or will there be another solution in the future? I am asking because, with [TeslaSolarCharger](https://github.com/pkuehnel/TeslaSolarCharger) I have highly Tesla Mate-connected software. Currently, I am hosting a fleet command proxy for all TSC users. I want to host a fleet telemetry server as well. As all the TSC users also use Tesla Mate, it would be cool if I could create a solution that works for Tesla Mate as well. But imitating the unofficial streaming API seems not to be the cleanest way to handle that issue in the long term.

pkuehnel:
Is there anywhere a documentation how the old Streaming API transfers the data? Does TeslaMate work with getting Fleet Telemetry data only? Which data is required for TeslaMate to work?

JakobLichterfeld:
https://docs.teslamate.org/docs/guides/api

JakobLichterfeld:
> > Is there any documentation on how the Fleet API Provider needs serve the Data? Also what is the protocol and how do the interfaces look so TeslaMate can consume Fleet Telemetry data?
> 
> https://docs.teslamate.org/docs/guides/api

pkuehnel:
@JakobLichterfeld Thanks, I already read that, my question is not answered there: How needs streaming Data to Look like so TeslaMate can consume it. I already have a Server getting data from fleet Telemetry but I do not know how the Websocket stream to TeslaMate needs to look like, to fake the old streaming API.

jlestel:
I didn't find any doc, I reengineered to do this: https://github.com/MyTeslaMate/websocket

PR #3906, opened 2024-05-22
Update default.po

tomzt:
Update Thai in default.po. 
Correct some words in default.po.


JakobLichterfeld (approved):
Thanks for your change! lgtm

PR #3908, opened 2024-05-23
doc: update backup path location to current directory to work around no write access to placeholder directory issue

JakobLichterfeld:


DrMichael:
Well, that assumes write access in the current working directory. But, right, people should have some idea of rights...

JakobLichterfeld:
And they needed write access in that directory in the first place to create `docker-compose.yml`

DrMichael:
It does not say, you should issue that command in the directory, where the docker-compose.yml is stored...

PR #3919, opened 2024-05-26
Update MQTT examples of location/active_route

longzheng:
Update the MQTT documentation to give a better example of the `location` and `active_route` payload after it was changed in https://github.com/teslamate-org/teslamate/pull/3789/

image

image

(I also plan on updating the Home Assistant docs to take advantage of the newer MQTT topics in a separate PR)


brianmay (approved):

JakobLichterfeld (approved):
ty for the change!

PR #3923, opened 2024-05-27
Update HomeAssistant docs to use new MQTT values

longzheng:
Related to the MQTT changes in https://github.com/teslamate-org/teslamate/pull/3789 and https://github.com/teslamate-org/teslamate/pull/3919

We can now configure Home Assistant to use the new JSON `location` topic as a `device_tracker` `json_attributes_topic` directly, without having to manually update a device tracker based on `latitude` and `longitude` changes. 

This makes the tracker much more smooth and contain less jagged lines because both the latitude and longitude update together when the location changes (instead of one or the other).

image

Also add another `device_tracker` for "active route location" and sensors for all the other active route metadata extracted from the `active_route` topic.
- Active route destination
- Active route energy at arrival
- Active route distance to arrival (mi)
- Active route distance to arrival (km) - converted using a `sensor.yaml` template sensor
- Active route minutes to arrival
- Active route traffic minutes delay

The active route sensors have an availability based on the `active_route` `error` property since the driver may not always be routing.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -88,7 +88,7 @@ Don't forget to replace `<teslamate url>` and `<your tesla model>` with correct 
     unique_id: teslamate_1_active_route_location
     availability: &teslamate_active_route_availability
       - topic: "teslamate/cars/1/active_route"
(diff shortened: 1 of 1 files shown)


JakobLichterfeld (approved):
ty for the change!

Issue #3910, opened 2024-05-23
A lot of empty space on the drive graph


woyteck1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Looks liek after new update, the bottom of the graph is symmetrical to the top, thus leaving a large, unused space on the graph.
Let's make bottom just a bit smaller than maximum regen.
Also, while we're at it - can we have second scale for 100% or miles.
Also looks like Rated range is not a cenered zero, but all other graphs are.
I think turning off Centered zero looks better.
![toomuchempty](https://github.com/teslamate-org/teslamate/assets/25311523/e051a064-12e7-4b38-bcd9-74dd18b4650d)

See graph with added percentages on the left, and not auto-centered.
![that's better](https://github.com/teslamate-org/teslamate/assets/25311523/dd1aaa5d-01fb-4469-8db9-05b130cb5e92)

### Relevant log output

```shell
This is a visual isue.
```

### Type of installation

Docker

### Version

v1.29.1

woyteck1:
A fix to these would be to overrride min to be 0
![image](https://github.com/teslamate-org/teslamate/assets/25311523/4a0209f2-1036-4ace-b65f-0805481a0cea)

JakobLichterfeld:
Thanks for reporting. @jheredianet or @swiffer  can you take a look into this?

jheredianet:
@woyteck1 great catch. Thank you!

Here is the PR https://github.com/teslamate-org/teslamate/pull/3927

PR #3927, opened 2024-05-29
Remove A lot of empty space on the drive graph

jheredianet:
This is a fix discussed here https://github.com/teslamate-org/teslamate/issues/3910


JakobLichterfeld:
TY!

Issue #3928, opened 2024-05-29
v1.29.0 Tesla Fleet API via MyTeslaMate stopped working - unsure if local TeslaMate or MyTeslaMate proxy issue


lits101:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Yesterday I configured a new local docker-compose instance of TeslaMate 1.29.0, using the FleetAPI available via the free tier of MyTeslaMate. This worked perfectly for a few hours, then while I was out driving it seemed to just stop. I'm getting some non-descript errors relating to the API: unknown, offline, Vehicle_1_api_error, etc...

I've tried refeshing my tokens via MyTeslaMate, reconfigured my local TeslaMate with the new token code, to no avail. 

Initially upon failure, my local TeslaMate instance was just showing a failed health check for the car and not reporting its position or any telemetry. After leaving it running for a few hours, it seems to be reporting the cars location at the time of API failure yesterday (this is incorrect, obviously since the car is home now) and it's occasionally grabbing its asleep state and time correctly. All other telemetry is stuck at the point of failure yesterday (SoC, tyre pressure, heading, etc). 

I'm a bit lost as to what to do. I've tried creating a new local docker instance of TeslaMate on another host, and I'm having the exact same issue. So maybe it's not anything I've done?

### Relevant log output

```shell
2024-05-30 08:46:38.853 [notice]     :alarm_handler: {:clear, TeslaMate.Vehicles.Vehicle_1_api_error}
2024-05-30 08:40:28.887 [info] GET https://api.myteslamate.com/api/1/vehicles/xxxx/vehicle_data?token=xxxx -> 429 (6293.800 ms)
2024-05-30 08:40:28.887 [warning] TeslaApi.Error / ""
2024-05-30 08:40:28.888 car_id=1 [error] Error / :unknown
2024-05-30 08:41:11.593 car_id=1 [info] Start / :offline
2024-05-30 08:43:30.386 [info] GET /
2024-05-30 08:43:30.404 [info] Sent 200 in 17ms
2024-05-30 08:43:39.858 [info] GET /settings
2024-05-30 08:43:39.885 [info] Sent 200 in 27ms
2024-05-30 08:46:38.853 [notice]     :alarm_handler: {:clear, TeslaMate.Vehicles.Vehicle_1_api_error}
```

### Screenshots

[...]

PR #3940, opened 2024-06-05
Customize intervals between vehicle_data calls

jlestel:
As discussed [here](https://github.com/teslamate-org/teslamate/issues/3928), the `vehicule_data` endpoint is limited for Tesla Fleet API users (300 hits per vehicle per day).

With the current hard-coded polling values, this limit is quickly exceeded. 

I want to let the Tesla Telemetry send the real time events via web streaming API and be able to customize polling intervals to try to not exceed this limit on `vehicle_data`.


jlestel:
@brianmay @JakobLichterfeld let me know what you think about this PR.

The current intervals are kept by default. I added some environment variables to be able to modify the intervals.

lib/teslamate/vehicles/vehicle.ex
```diff
@@ -401,10 +418,10 @@ defmodule TeslaMate.Vehicles.Vehicle do
...
         interval =
           case state do
-            {:driving, _, _} -> 10
-            {:charging, _} -> 15
-            :online -> 20
-            _ -> 30
+            {:driving, _, _} -> round(default_interval() * 2 / 3)
+            {:charging, _} -> default_interval()
+            :online -> round(default_interval() * 4 / 3)
+            _ -> default_interval() * 2
```
brianmay, line 424: My eyes immediately get drawn to these hardcoded factors - what do they do?

brianmay (commented):
In general looks good.

Have mixed feelings about this. We don't want people to think they have to configure these values and end up with incorrect values that are loading up Tesla's servers.

Suggestion - add somewhere a note that it is strongly recommended that you do not set that values unless you really know what you are doing.

lib/teslamate/vehicles/vehicle.ex
jlestel: Initially, I preferred to keep this factor rather than the current numerical values. Current values ​​are 10, 15, 20, 30.

It's only in the case of unhandled errors that these values ​​are used, so I hesitated to leave the current values ​​hard…

As only the 200 response code are important for the Tesla WAF, I think now it’s better to revert to current hardcoded values.

jlestel:
> In general looks good.
> 
> Have mixed feelings about this. We don't want people to think they have to configure these values and end up with incorrect values that are loading up Tesla's servers.
> 
> Suggestion - add somewhere a note that it is strongly recommended that you do not set that values unless you really know what you are doing.



> Suggestion - add somewhere a note that it is strongly recommended that you do not set that values unless you really know what you are doing.

I will add a bold note about this in the env vars description.

lib/teslamate/vehicles/vehicle.ex
jlestel: Reverted to default values.

lib/teslamate/vehicles/vehicle.ex
```diff
@@ -28,11 +28,24 @@ defmodule TeslaMate.Vehicles.Vehicle do
...
   # Static
+  def interval(env_var, default) do
+    System.get_env(env_var)
+    |> case do
+      nil -> default
+      interval -> String.to_integer(interval)
+    end
+  end
+
+  def asleep_interval, do: interval("POLLING_ASLEEP_INTERVAL", @asleep_interval)
```
JakobLichterfeld, line 43: My vote would be to not let the user set intervals via env variables shorter than our defaults.

JakobLichterfeld (changes requested):

lib/teslamate/vehicles/vehicle.ex
jlestel: Good idea, it acts as a safeguard for those who would like to play with what they do not master. I'll implement it.

jlestel:
> will Teslamate notice this in real time via streaming API, regardless of this interval? Or will we run the risk of losing the first few minutes of each drive

Yes, the data retrieved from the streaming API is independent of these intervals and Tesla continues to get it in real time.

Obviously with long intervals, the data is much less precise... but streaming helps limit losses.

Note that for the "future": the new Telemetry only sends this information every minute while the "owner" streaming API sends every second during drives.

lib/teslamate/vehicles/vehicle.ex
brianmay: It also means we can increase the defaults later on, if deemed necessarily, and people will get the increased limits even if they have smaller values.

Maybe the limits should be displayed in the website. So the user knows what they got (otherwise I can just imagine bug reports - I changed the interval to 1 second but it doesn't seem to have changed anything!). But may be not worry about that just, that is starting to sound like feature creep for  what should be a simple pull request. Instead, create a new issue if this is merged.

brianmay:
Not that the streaming parameters supported is very restricted. Anything not supported in streaming will be affected, e.g. state of doors being opened/closed. State of vehicle being plugged in, etc.

JakobLichterfeld (approved):

jlestel:
Thank you !

PR #3941, opened 2024-06-05
ci: add yeec and leex compilers to mix.exs

JakobLichterfeld:
To fix the linter warnings, see for example here: https://github.com/teslamate-org/teslamate/actions/runs/9380988328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - ci: Upgrade PostgreSQL to v16 in elixir workflow (#3916 - @JakobLichterfeld)
 - build(deps): bump docker/login-action from 3.1.0 to 3.2.0 (#3933)
 - build(deps): bump elixir from 1.16.1-otp-26 to 1.16.2-otp-26 (#3935 - @JakobLichterfeld)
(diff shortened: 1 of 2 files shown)


brianmay:
I suspect the problem is in compiling one/more dependencias. Can't fix it here :-(

In general I think the Elxir community aren't particularly proactive about removing compiler warnings - have a look at the compiler logs.

Although this could also be because we are using old versions of packages (I suspect probably).

It could also be sign that the packages we rely on are no longer being maintained.

JakobLichterfeld:
At least the linter warnings dissapear with these changes.

From time to time I remember your idea of rewriting in rust.

brianmay:
Are you sure that they disappeared?

I still see them. e.g. https://github.com/teslamate-org/teslamate/actions/runs/9382022104?pr=3941

JakobLichterfeld:
> Are you sure that they disappeared?
> 
> I still see them. e.g. https://github.com/teslamate-org/teslamate/actions/runs/9382022104?pr=3941

You are right, damn. I looked in the single job :-(

PR #3943, opened 2024-06-06
Respect Retry-After header on 429 response code

jlestel:
As seen [in this ticket](https://github.com/teslamate-org/teslamate/issues/3928) and in the @brianmay [related PR](https://github.com/teslamate-org/teslamate/pull/3451), the API response code 429 contains a `Retry-After` header which must be respected to avoid unnecessary calls.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ config/*.env
 .eclipse/
 .elixir_ls/
 /.project
+.vscode/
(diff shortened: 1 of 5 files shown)


brianmay (approved):

JakobLichterfeld (approved):
lgtm, ty!

PR #3950, opened 2024-06-09
Remove A lot of empty space on the drive graph

jheredianet:
This is a quick fix on the min Y-axes of the drive graph for a better visualization (when there is a big difference on Y values)


JakobLichterfeld (approved):

JakobLichterfeld:
TY!

PR #3951, opened 2024-06-09
Update and fix timeline dashboard

jheredianet:
This is an improve on "Timeline" dashboard:

- Use of ILIKE instead of LIKE to avoid case sensitive text filter
- Update scheme version
- Remove fixed preferences
- Remove fixed urls
- Remove fixed Units


JakobLichterfeld (approved):
ty!

PR #3954, opened 2024-06-12
Remove LFP dashboards

jheredianet:
This is a proposal to remove the "LFP" dashboards that have been in the project for a long time, but they only cause confusion to users and I see that they do not really contribute anything new. What's more, they are a clone from the core dashboards that have already been updated several times and have become obsoletes.

It has been commented several times in different posts (which I refer to at the end) of the intention to remove them and have the code modified so that it tries to detect the type of battery that the car has automatically and that the panels do I calculate them using that data. But even so, I don't see the need to make special calculations for the type of battery since the car itself is already responsible for providing that data specifically and we don't need to categorize it again.

The only thing I have seen is that Teslamate, to consider the efficiency of the charges, only considers charges that last longer than 10 minutes and are less than 95% SOC (because we already know that after that the car is no longer efficient at charging) which is correct, and since those with LFP batteries usually charge at 100%, that's where the controversy comes from that the formulas have to change.

As a developer of the "Battery Health" dashboard I receive many questions regarding the LFP dashboards and why they show different values. In my humble opinion, the LFP dashboards should not have been approved from the beginning. I run a Spanish group of Teslamate enthusiast (more than 500 users) and everyone uses the core dashboards directly instead of the LFP ones because the data is more accurate. 

If you consider that the LFP dashboards should not be removed, I understand and respect that. If so, please feel free to reject this PR. Thank you

These are other posts talking about this:

- https://github.com/teslamate-org/teslamate/discussions/3513
- https://github.com/teslamate-org/teslamate/pull/3311 
- https://github.com/teslamate-org/teslamate/issues/3438
- https://github.com/teslamate-org/teslamate/discussions/2595
- https://github.com/teslamate-org/teslamate/discussions/3752


KILYBMW:
Good idea.
Those dashboards don't add value because the main ones (NO-LFP) have updated data and take everything necessary into account, regardless of the vehicle's battery type.

RafaG:
Great Idea! I really don't understand why we have separate dashboards for LFP batteries

JakobLichterfeld:
I tend to agree

JakobLichterfeld (approved):

swiffer:
Supporting the idea of removing them. The changes that have been made back then are listed here:

https://github.com/teslamate-org/teslamate/issues/3438#issuecomment-1823585862

@KILYBMW - visualization changes have not been incorporated into the main ones - so there is still value but having them diverged over time is bad.

as stated in https://github.com/teslamate-org/teslamate/issues/3438 we should come back to one dashboard and utilize config / queries instead of adding additional dashboards.

Grafana 11 supports setting multiple thresholds by config query results.

So for example to remove upper caps from overview dashboard battery level gauge is possible via query - the only thing needed is somehow storing battery type per car and using that in the queries going forward.

grafik

grafik

swiffer:
UPDATE - checking the Overview Dashboard I see that changed based on VIN have been partially added already. I thought VIN is no longer trustworthy - happy to add the threshold changes using the same logic on other places later today.

```sql
SELECT
  CASE
    WHEN substring(vin, 7, 1) = 'F'
    THEN 170
    ELSE 250
  END as max_charging_kw
from cars 
```

JakobLichterfeld:
solved by #3966

PR #3953, opened 2024-06-10
Update projects.md with new MMM

top-gun:
MMM-Teslamate has a new maintainer, updated the link to the new repo

PR #3973, opened 2024-06-18
doc: fix MMM-Teslamate and ha-addon-teslamate project links

olexs:
Fixes an incorrect change made in #3953


JakobLichterfeld (approved):
Thank you so much!

PR #3978, opened 2024-06-19
Documentation for new endpoint changes

jlestel:
Add documentation the new endpoints as asked in #3959

### DIFF
--- a/website/docs/guides/api.md
+++ b/website/docs/guides/api.md
@@ -1,40 +1,46 @@
 ---
-title: Using the new Tesla Fleet API and Telemetry streaming
+title: Using the Official Tesla Fleet API and Telemetry Streaming
 ---
 
-## Why Tesla Fleet and Telemetry should be needed?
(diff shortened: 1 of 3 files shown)


zlymeda:
If using self-hosted HTTP proxy, doesn't the token refresh also need `client_id`? https://developer.tesla.com/docs/fleet-api#third-party-token

jlestel:
> If using self-hosted HTTP proxy, doesn't the token refresh also need `client_id`? https://developer.tesla.com/docs/fleet-api#third-party-token

You're right, I tried it and it works now with this PR #3984.

Check updates on doc on [the direct Tesla API part](https://deploy-preview-3978--teslamate.netlify.app/docs/guides/api#tesla-fleet-api).

JakobLichterfeld (approved):
Ty!
I hesitated at first because the reference to external projects for Fleet API use is not necessarily complete.

jlestel:
Thank you

PR #3980, opened 2024-06-19
SSL: use ENV var to verify/dont verify HTTP certs

zlymeda:


jlestel:
My suggestions: 
1. add [a line in the doc](https://docs.teslamate.org/docs/configuration/environment_variables) below `TESLA_API_HOST`
1. rename to `TESLA_API_TLS_ACCEPT_INVALID_CERTS` to be consistent in naming

JakobLichterfeld:
In Addition please remove debugging code

PR #3984, opened 2024-06-20
Allow Fleet API users to change client_id

jlestel:
As seen by @zlymeda [on the doc PR](https://github.com/teslamate-org/teslamate/pull/3978#issuecomment-2180005542), `client_id` need to be changed when direct use Tesla Fleet API.


lib/tesla_api/auth.ex
```diff
@@ -6,7 +6,7 @@ defmodule TeslaApi.Auth do
   @web_client_id "ownerapi"
   @redirect_uri "https://auth.tesla.com/void/callback"
 
-  def web_client_id, do: @web_client_id
+  def web_client_id, do: System.get_env("TESLA_AUTH_CLIENT_ID") || @web_client_id
```
JakobLichterfeld, line 9: shouldn't we do at least a format check before applying it directly to the variable?

JakobLichterfeld (changes requested):

lib/tesla_api/auth.ex
jlestel: I prefer to add a mention of the current format of Tesla client_ids in the env vars doc rather than add a dependency on an identifier managed by Tesla.

Adding a check in the code is not really useful since if the client_id is not correct the API indicates it.

JakobLichterfeld: fair take, ty

JakobLichterfeld (approved):

PR #3989, opened 2024-06-23
Update vehicle.ex with marketing name for Model Y Long Range RWD

stefanlange:
Add marketing name for Model Y LR RWD. This solves an issue where the new Model Y is currently named "Model Y 74" in TeslaMate's main screen.


JakobLichterfeld (approved):
ty for your change

PR #3993, opened 2024-06-23
feat: Add ability to enable and disable data collection for cars

mark3-dev:
This feature will allow individual cars to be enabled and disabled from the settings page.

There is one new column added to the car_settings table for storing the enabled flag. It is not nullallable and has a default value of true.

When a change to the enabled value of any car it will call the existing restart method in the vehicles file.

NOTE: This will not affect any grafana dashboards other than no new data for a car will be collected. 

I found two discussions that relate to this.
https://github.com/teslamate-org/teslamate/discussions/3918
https://github.com/teslamate-org/teslamate/discussions/1049

### DIFF
--- a/lib/teslamate/import.ex
+++ b/lib/teslamate/import.ex
@@ -291,7 +291,8 @@ defmodule TeslaMate.Import do
         settings = %CarSettings{
           suspend_min: 0,
           suspend_after_idle_min: 99999,
-          use_streaming_api: false
+          use_streaming_api: false,
+          enabled: true
         }
 
         %Car{car | settings: settings}
--- a/lib/teslamate/settings.ex
+++ b/lib/teslamate/settings.ex
@@ -7,8 +7,9 @@ defmodule TeslaMate.Settings do
   alias TeslaMate.Repo
 
   alias __MODULE__.{GlobalSettings, CarSettings}
-  alias TeslaMate.{Log, Locations}
+  alias TeslaMate.{Log, Locations, Vehicles}
   alias TeslaMate.Log.Car
+  import Core.Dependency, only: [call: 2]
 
   def get_global_settings! do
     case Repo.all(GlobalSettings) do
@@ -46,6 +47,7 @@ defmodule TeslaMate.Settings do
   def update_car_settings(%CarSettings{car: %Car{}} = pre, attrs) do
     Repo.transaction(fn ->
       with {:ok, post} <- pre |> CarSettings.changeset(attrs) |> Repo.update(),
+           :ok <- on_enabled_change(pre, post),
            :ok <- broadcast(pre.car, post) do
         post
       else
@@ -82,6 +84,12 @@ defmodule TeslaMate.Settings do
 
   defp on_language_change(%GlobalSettings{}, %GlobalSettings{language: lang}) do
     Locations.refresh_addresses(lang)
+  end
+
+  def on_enabled_change(%CarSettings{enabled: preEnabled}, %CarSettings{enabled: postEnabled}) do
+    if preEnabled != postEnabled do
+      call(Vehicles, :restart)
+    end
   end
 
   defp broadcast(car, settings) do
--- a/lib/teslamate/settings.ex
+++ b/lib/teslamate/settings.ex
@@ -90,6 +90,7 @@ defmodule TeslaMate.Settings do
     if preEnabled != postEnabled do
       call(Vehicles, :restart)
     end
+
     :ok
   end
 
--- a/lib/teslamate/settings.ex
+++ b/lib/teslamate/settings.ex
@@ -90,6 +90,7 @@ defmodule TeslaMate.Settings do
     if preEnabled != postEnabled do
       call(Vehicles, :restart)
     end
+    :ok
   end
 
   defp broadcast(car, settings) do
--- a/lib/teslamate/settings/car_settings.ex
+++ b/lib/teslamate/settings/car_settings.ex
@@ -10,7 +10,7 @@ defmodule TeslaMate.Settings.CarSettings do
     field :req_not_unlocked, :boolean, default: false
     field :free_supercharging, :boolean, default: false
     field :use_streaming_api, :boolean, default: true
-    field :enabled, :boolean
+    field :enabled, :boolean, default: true
 
     has_one :car, Car, foreign_key: :settings_id
   end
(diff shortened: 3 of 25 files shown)


JakobLichterfeld:
Thanks for your suggestion. This is obviously not as hard as https://docs.teslamate.org/docs/maintenance/manually_fixing_data#remove-a-vehicle-from-the-database

JakobLichterfeld (approved):

mark3-dev:
> Thank you for your suggestion. This is obviously not as permanent as removing the vehicle from the database: https://docs.teslamate.org/docs/maintenance/manually_fixing_data#remove-a-vehicle-from-the-database

Thanks JakobLichterfeld, I agree this is not as permanent as removing the vehicle from the database. This is more for stopping data collection on active cars associated with your Tesla account. I have a car on my account that belongs to a family member and I need to prevent data collection on that car without removing driver access.

JakobLichterfeld (approved):

JakobLichterfeld (approved):

JakobLichterfeld:
@mark3-dev  Why the feature is not yet merged:
I'm thinking about how we can document this feature for the user. The most user-friendly way would certainly be an explanation directly in the settings.

mark3-dev:
> @mark3-dev Why the feature is not yet merged: I'm thinking about how we can document this feature for the user. The most user-friendly way would certainly be an explanation directly in the settings.

Is there anything else I can do? 

My original entry in the settings page was "Enabled" as the section title and "Enable data collection" for the text beside the slider. But after I realized that there were many different languages that the interface was translated to I changed it to Enabled for both since that was already translated in all the languages.

cwanja:
I'd suggest put a 'Learn more' link and add a page under the 'Maintenance' section with more details. Including a link back to the removing a vehicle from the database details as well. 

Could also seeing merging that section and this PR under a 'Vehicle Management' page under 'Maintenance'.

JakobLichterfeld:
> My original entry in the settings page was "Enabled" as the section title and "Enable data collection" for the text beside the slider. But after I realized that there were many different languages that the interface was translated to I changed it to Enabled for both since that was already translated in all the languages.

For now I would go with your initial suggestion. Necessary translation should not prevent us from being precise in our wording for the sake of UX.

mark3-dev:
> > My original entry in the settings page was "Enabled" as the section title and "Enable data collection" for the text beside the slider. But after I realized that there were many different languages that the interface was translated to I changed it to Enabled for both since that was already translated in all the languages.
> 
> For now I would go with your initial suggestion. Necessary translation should not prevent us from being precise in our wording for the sake of UX.

I'll try and get that added later today.

JakobLichterfeld:
Not saying you need to translate the message, we start with English

mark3-dev:
Screenshot from 2024-07-04 16-26-44

mark3-dev:
I think I like this one better
Screenshot from 2024-07-04 16-35-45

cwanja:
> I think I like this one better
> 
> Screenshot from 2024-07-04 16-35-45
> 
> 

Agreed.

JakobLichterfeld (approved):
perfect, ty!

git-anish:
Hey folks, @mark3-dev @JakobLichterfeld 


Using this feature when the car is disabled it still appears and auto populates in Grafana (understandably so).

- If I try deleting the car using: DELETE FROM cars WHERE id = X; then the teslamate.com/settings page stops functioning
- If I then use DELETE FROM car_settings WHERE id = 10; then the car reappears during the next teslamate reboot or API key refresh starting the cycle all over again

Any tips on how to disable the car then prevent it from auto populating in the grafana dashboards while maintaining access to the settings page? Does this need to be a net new feature that accounts for the disabled car & deletes the data from Grafana?

Thanks!

JakobLichterfeld:
> * If I try deleting the car using: DELETE FROM cars WHERE id = X; then the teslamate.com/settings page stops functioning
> * If I then use DELETE FROM car_settings WHERE id = 10; then the car reappears during the next teslamate reboot or API key refresh starting the cycle all over again
> 
> Any tips on how to disable the car then prevent it from auto populating in the grafana dashboards while maintaining access to the settings page? Does this need to be a net new feature that accounts for the disabled car & deletes the data from Grafana?
> 
> Thanks!

See: https://docs.teslamate.org/docs/maintenance/manually_fixing_data#remove-a-vehicle-from-the-database

git-anish:
Thanks for getting back to me, @JakobLichterfeld 

Just to clarify — I’ve already followed the exact steps outlined in the documentation you linked for removing a vehicle from the database. Here’s what’s happening in my case:

1. When I run the full set of deletion commands:

```
DELETE FROM cars WHERE id = 12;
DELETE FROM car_settings WHERE id = 12;
DELETE FROM charging_processes WHERE car_id = 12;
DELETE FROM charges WHERE charging_process_id IN (SELECT id FROM charging_processes WHERE car_id = 12);
DELETE FROM drives WHERE car_id = 12;
DELETE FROM positions WHERE car_id = 12;
DELETE FROM states WHERE car_id = 12;
DELETE FROM updates WHERE car_id = 12;
```

— the car is successfully removed.
— but during the next TeslaMate reboot or API key refresh, it automatically reappears as a new car record (e.g. `car_id = 13`), restarting the cycle.

2. If I skip the `DELETE FROM car_settings WHERE id = 12;` command to avoid the re-creation of a new car record & data collection enabled by default, the Settings page (teslamate.com/settings) stops functioning.

So, I end up in a circular situation:

deleting the car_settings entry breaks the Settings page,

but keeping it causes the car to auto-recreate.

My goal is **to keep the car disabled for data collection without it reappearing in Grafana or causing settings to break**.

Do you think this requires an additional feature to persist the “disabled” state across API refreshes, or is there a workaround I might be missing?

JakobLichterfeld:
> — but during the next TeslaMate reboot or API key refresh, it automatically reappears as a new car record (e.g. `car_id = 13`), restarting the cycle.

Ah, that's for sure intended. All vehicles in your account will be added to TeslaMate. If you sell a vehicle and do not want to have the data, then you can follow the linked description.
Otherwise, you can use the setting to disable data collection for this vehicle.

JakobLichterfeld:
Discussion for feature enhancement can be found here: https://github.com/teslamate-org/teslamate/discussions/2736

PR #3998, opened 2024-06-25
feat: Create NixOS module

scottbot95:
### Overview

Adds a semi-opinionated NixOS module to ease deploying TeslaMate on NixOS.

### Trade-offs

- Does not use docker/podman for service. This was done as it is typical for NixOS services to run natively since Nix itself + systemd provides much the same isolation/reproducibility that docker does.
- Uses built-in NixOS grafana/postgres modules. This makes development easier/more-consistent however it does mean we are polluting the "global" grafana/postgres instances. I felt this was worth the trade-off but am open to discussion.
- Provides native nix bindings for most teslamate configuration but not all. I chose the minimal set of configurations option that would be required to run Teslamate+Grafana behind a reverse proxy. Additionally, any missing configuration can still be provided by adding to `systemd.services.teslamate.environment`
- Does not setup MQTT server. I chose not to include this since setting up a MQTT server in NixOS is quite simple and TeslaMate is not opinionated on its configuration.
- `RELEASE_COOKIE` must be passed in via the `secretFile`. Values used in the NixOS config are globally readable so secrets should be read from disk at runtime. We could use either a hard-coded RELEASE_COOKIE value or generate something based on hashes, but given it is technically a secret, I chose to just require the end user to provide it.

### Testing
`nix flake check --impure` (`--impure` require due to devenv) will run a simple telsamate configuration in a VM and assert that a server starts up on port 4000


flake.nix
```diff
@@ -153,11 +156,39 @@
...
+        moduleTest = (nixpkgs.lib.nixos.runTest {
+          hostPkgs = pkgs;
+          defaults.documentation.enable = false;
+          imports = [{
+            name = "teslamate";
+            nodes.server = {
+              imports = [ self.nixosModules.default ];
+              services.teslamate = {
+                enable = true;
+                secrestFile = builtins.toFile "teslamate.env" ''
```
brianmay, line 169: Possible typo in `secretsFile` here.

module.nix
```diff
@@ -0,0 +1,295 @@
...
+            {
+              PORT = toString cfg.port;
+              DATABASE_USER = cfg.postgres.user;
+              DATABASE_NAME = cfg.postgres.database;
+              DATABASE_HOST = cfg.postgres.host;
+              DATABASE_PORT = toString cfg.postgres.port;
+              VIRTUAL_HOST = cfg.virtualHost;
+              URL_PATH = cfg.urlPath;
+              HTTP_BINDING_ADDRESS = mkIf (cfg.listenAddress != null) cfg.listenAddress;
+              DISABLE_MQTT = mkIf (!cfg.mqtt.enable) "true";
```
brianmay, line 178: I am a bit uneasy that `postgres.enable` enables/disables the postgres server, but `mqtt.enable` enables functionality on the client. This could be a bit confusing.

Maybe `postgresl.enable_server` might be better here? As we still configure the postgres server in the client even if it is disabled.

brianmay (commented):
In general, looks good to me.

flake.nix
scottbot95: Oops yeah. Those force pushes earlier were fixing that, I guess I missed this one lol

module.nix
scottbot95: Yeah that's a fair critique. I felt very awkward about that as well. Perhaps `postgresql.create_server`? Or `postgresql.setup_server`? Or would it be better to take the opposite approach and rename `mqtt.enable` to `mqtt.disable` to align it with the actual env var that will get set? Thoughts?

JakobLichterfeld:
I love the work on the NixOS side. What I wonder is how to do backup and restore with this implementation? This is the only reason why I have not yet migrated my TeslaMate instance to my NixOS server.

brianmay:
Backup/restore is just:

* Backup the database as usual.
* Backup the nixos configuration for the system that lets you recreate the system. Personally I have a nix flake in a central git that lets me configure all my systems.

I have not used postgresql on nixos, so I am not really familiar with upgrading databases, etc. I think I saw a webpage with instructions somewhere.

The potential gotcha is configuring secrets. You don't want to include secrets in plain text format in the nix repo, or at least this is frowned upon. I personally use agenix for these.

brianmay:
At present I just install teslamate on my nixos system with a docker image, but I think using this flake will be simpler.

module.nix
brianmay: Thinking about this some more, I think I would go for `enable_server`.

Leaving `mqtt.enable` is more in line with my expectations from other projects. And same with `grafana.enable` also.

Not extremely fussed though.

JakobLichterfeld:
> * Backup the database as usual.

Sometimes the most obvious is the hardest to see :-)

> * Backup the nixos configuration for the system that lets you recreate the system. Personally I have a nix flake in a central git that lets me configure all my systems.

Yeah, that's what I do as well


> The potential gotcha is configuring secrets. You don't want to include secrets in plain text format in the nix repo, or at least this is frowned upon. I personally use agenix for these.

Same for me, using agenix for secrets and git-crypt as well

module.nix
scottbot95: Ok that sounds reasonable. I will go with `enable_server`

brianmay (approved):

JakobLichterfeld (approved):
TY!

Issue #4001, opened 2024-06-25
"Current Energy" instead of "Current Capacity"


woyteck1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

It's a cosmetic change, but I do not agree with "Current Capacity" name in this graph.
To make sense, it should be "Current Energy" or "Currenty Stored Energy", and the vaule is in units of energy - kWh. 
![image](https://github.com/teslamate-org/teslamate/assets/25311523/2e482001-e408-4e56-9f8c-9464a97ea4d2)

### Steps To Reproduce

Go to Battery Health.

### Relevant log output

```shell
N/A
```

### Type of installation

Docker

### Version

v1.29.2

JakobLichterfeld:
Agree on this. What do you think  @jheredianet ?

jheredianet:
Great suggestion. I'll make the change ASAP.

PR #4005, opened 2024-06-26
Improve visualizations on Battery Health dasboard

jheredianet:
Just a minor improve on Battery Health dashboard on labels and data suggested here https://github.com/teslamate-org/teslamate/issues/4001


JakobLichterfeld (approved):
ty!

PR #4031, opened 2024-07-03
Improve "Charges" dashboard

jheredianet:
This adds more information on the charges dashboard:
- Summary of Period: Costs and average duration
- Graphic bars for efficiency
- Filters for location, costs and duration
- General information if there are incomplete charges (charges without ending date)

image

PR #4033, opened 2024-07-04
Improve "Drives" dashboard

jheredianet:
This improves the "Drives Dashboard"
 - Update filter labels
 - Add new filters
 - Add Summary rows and new panels
 - Add Incomplete drive table

image


jheredianet:
@cwanja with this PR and this other one https://github.com/teslamate-org/teslamate/pull/4031
You shouldn't have any incomplete Drive or Charge  😉

JakobLichterfeld (approved):
Awesome work, thank you!
I do like the idea of not having a separate Dashboard for incomplete data but include it in drives and charges dashboard.

cwanja:
UX thoughts here @jheredianet - either:
1. Moving the new tiles and data to the top of both dashboards
2. Limiting the number of rows displayed (if that is possible).

To be honest, I do not think most are going to know the additional data is there. I am torn between which one I think is "right". @DrMichael @swiffer any thoughts or preferences on way or another? 

Would apply to #4031 as well.

JakobLichterfeld:
> To be honest, I do not think most are going to know the additional data is there.

This was a comment in a forum just minutes after the release:



I personally do like the idea of not having a separate Dashboard for it. If someone is complaining about not completed drives we now have a standard way via UI.

jheredianet:
> To be honest, I do not think most are going to know the additional data is there. I am torn between which one I think is "right". 

If you prefer to give more visibility to this additional information, I think the most suitable thing would be limiting the number of rows displayed, because the main purpose of these dashboards is to show the drives/charges info. The rest is just an extra info. 

Personally, I would leave it as is, I don't share the idea that no one will find it out. All of us who use TeslaMate are involved in IT in some way, just out of curiosity, whenever something new comes out we look at the screens scrolling from beginning to end, as well as reading the change logs and looking for the new features of a new version.

DrMichael:
I think, the stats row might be on top. Also there I would recommend to use custom units like in the other dashboards, e.g. Total Distance show as "2Mm", which is kind of unusual. Better might be "1532 km". So "Custom unit: km".
And for the duration perhaps rather "clock (s)".

cwanja:
> I personally do like the idea of not having a separate Dashboard for it. If someone is complaining about not completed drives we now have a standard way via UI.

Not wanting to go back to separate dashboards for the data. But re-arranging the data layout of the current dashboards. Or changing the height of the panel listing Drives or Charges.

PR #4045, opened 2024-07-06
Update overview.json: fix Net

DrMichael:
Readd Wh/km and Wh/mi


JakobLichterfeld (approved):

sdwalker:
https://github.com/teslamate-org/teslamate/blob/master/grafana/dashboards/overview.json#L664 needs to be Wh/mi

PR #4038, opened 2024-07-05
feat: add LFP Battery setting for car settings

JakobLichterfeld:
> Sorry, due to merging #3993, may I ask you to update the pot again? And merge the settings_test? Sorry for the inconvenience.

I try to solve it via a separate branch as I do not have write access to your repo.

_Originally posted by @JakobLichterfeld in https://github.com/teslamate-org/teslamate/issues/4007#issuecomment-2211136329_
            

Successor of #4007

Issue #4030, opened 2024-07-03
perf: updates dashboard slow


JakobLichterfeld:
Ok, I think we will go with query 8, will open a PR later. Any other queries to speed up? ,🤓

_Originally posted by @swiffer in https://github.com/teslamate-org/teslamate/issues/3921#issuecomment-2198258530_

---> The updates dashboard is slow @swiffer

swiffer:
Could you run an explain analyze on the slow query and add it here? Guess my data is to small or the compute power to big 😀

JakobLichterfeld:
:-D Yeah for sure. Just need to figure out the single query.

> or the compute power to big
Did not spot it on my Mac Studio M2 Pro, but on my Raspberry Pi 3 B and I know the updates are stored in separate table, therefore the time needed does not make sense to me.

![image](https://github.com/teslamate-org/teslamate/assets/16510328/476ee92c-882c-4249-aa5d-3184c56a5320)

swiffer:
Took a Look already - it Scans positions and charges for 10 years. I guess this can be simplified without big sacrifice in data accuracy. Will come up with an alternative query, let's see how much they differ.

swiffer:
@JakobLichterfeld - can you run this query (once with explain analyze and once without to compare data outcome)

```sql
with u as (

  select *, coalesce(lag(start_date) over(order by start_date desc), now()) as next_start_date 

  from updates

  where car_id = '2' and start_date BETWEEN '2014-07-04T12:37:24.227Z' AND '2024-07-04T12:37:24.227Z'

),

rng as (

  SELECT

	  date_trunc('hour', date) AS date,

	  (sum(rated_battery_range_km) / nullif(sum(usable_battery_level),0) * 100 ) AS "battery_rng",
	  
	  sum(case when action = 'Charge' then 1 else 0 end) as chg_ct

  FROM (

    select coalesce(usable_battery_level, battery_level) as usable_battery_level, start_date as date, start_rated_range_km as rated_battery_range_km, start_ideal_range_km as ideal_battery_range_km, 'Drive' as action

    from drives d
    
    inner join positions p on d.start_position_id = p.id 

    where d.car_id = '2' and d.start_date BETWEEN '2014-07-04T12:37:24.227Z' AND '2024-07-04T12:37:24.227Z'

    union all

[...]

PR #4051, opened 2024-07-07
use lfp_battery setting in dashboards & speed up updates dashboard

swiffer:
Follow Up of #4038 - make use of lfp_battery setting instead of relying on vin detection

Closing #4030 - stop relying on positions / charges and use drives / charging_processes instead when calculating range in updates dashboard as performance is otherwise bad for low compute devices (rpi)


swiffer:
@JakobLichterfeld - lgtm

JakobLichterfeld (approved):
Perfect, thank you!

PR #4056, opened 2024-07-07
Fix hard coded informations (timezone and orgId)

jlestel:
I noticed that the timezone was hard in 2 dashboards

The browser is the default behavior: https://grafana.com/docs/grafana/latest/administration/organization-preferences/#change-the-grafana-default-timezone

On the other hand, by forcing it, overrides (coming from the server, organization, team, or individual user level) are not taken into account.

### DIFF
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -1978,7 +1978,7 @@
   "timepicker": {
     "hidden": true
   },
-  "timezone": "browser",
+  "timezone": "",
   "title": "Battery Health",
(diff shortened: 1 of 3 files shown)


jlestel:
@DrMichael perhaps it is intended to leave these "timezones" hard?

DrMichael:
No, don't think so. I am fine with that change, we will see.

JakobLichterfeld:
Good spot!

With #4064 we enable browser based time settings.

JakobLichterfeld (approved):

PR #4069, opened 2024-07-09
feat: extend version with more build info

### LINKED ISSUE #3682: Extend version with more build info
It could be useful to have some more info than just the version number when testing out PR's

It could be just a log output when starting teslamate, but might also be handy on settings on web.

I guess we would need something like this to be able to backtrace it completely:
- PR #
- PR origin: fork and revision 
- build date and time

_Originally posted by @micves in https://github.com/teslamate-org/teslamate/discussions/3658#discussioncomment-8507266_

JakobLichterfeld:
closes: #3682


JakobLichterfeld:
image

Not yet working correctly

brianmay:
> Not yet working correctly in our ghcr image, as git command is not available in our docker file

I would imagine the nix flake build will have the same issue. The git repo stuff not available from the directory.

Think we will need to use environment variables, as per https://github.com/teslamate-org/teslamate/issues/3682#issuecomment-1951462478.

For Docker these need to be passed to the docker build command. `--build-arg=arg=value`

Nix is easier, with something like this would create the environment variables (this needs to be used in appropriate place):

```
 build_env = {
          BUILD_DATE =
            with flockenzeit.lib.splitSecondsSinceEpoch { } self.lastModified;
            "${F}T${T}${Z}";
          VCS_REF = "${self.rev or "dirty"}";
        };
```

Although it is unfortunate you need the flockenzeit just to format the date time. This doesn't use the VERSION file, combining the two shouldn't be hard.

brianmay:
Forgot to say:

* Version number should be set at build time. Oh wait, I think we already do this. Good.
* Suggest using environment variables as first preference. If they are not defined, then fall back to something else.

JakobLichterfeld:
> I would imagine the nix flake build will have the same issue. The git repo stuff not available in the build.

Yeah, assume this as well. As we use a two staged docker build, we could add git in the builder step to have the command available with no real downside imo. Will test this. In the end build env are the best choice imo as well.

PR #4019, opened 2024-06-30
feat: Grafana 11.1.0 & Dashboard improvements

swiffer:
Closes #4018
Closes #3921

PR #4071, opened 2024-07-09
downgrade grafana until maps issue in upstream grafana is resolved

swiffer:
before releasing this potentially as 1.30.1 please check if maps are working fine again

https://github.com/teslamate-org/teslamate/pull/4019#issuecomment-2218427564

sorry rushing the grafana update. always try to stay on latest. (at least 1.30.1 would bring some grafana localization improvements as well ;))

once grafana 11.1.1 is released we can upgrade again as it will include this fix: https://github.com/grafana/grafana/pull/89248

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -1,6 +1,6 @@
 # Ensure selecting a tag that is available for arm/v7, arm64, and amd64
 # https://hub.docker.com/r/grafana/grafana/tags
-FROM grafana/grafana:11.1.0
+FROM grafana/grafana:11.0.1
 
 ENV GF_ANALYTICS_REPORTING_ENABLED=false \
     GF_AUTH_ANONYMOUS_ENABLED=false \


swiffer:
I think I've edited two dashboards after upgrading to 11.1.0 - are these still working fine after downgrade? Can test tomorrow and chance dashboard yank if needed...

brianmay:
We should put a link to the upstream grafana issue. Assuming there is one. If there isn't maybe we need to create one.

swiffer:
looks good to me - dashboards still working

cwanja:
Guess I am confused. I thought we were **not** downgrading based upon discussions in #4019

JakobLichterfeld:
> Guess I am confused.

Now I'm confused as well. Was on mobile yesterday evening and later agreed to downgrade by today.

cwanja:
Oh well. Release is out there. No harm, less issues and discussions. 

We can close the open / pinned issue when the new Grafana release happens.

JakobLichterfeld:
Yeah, overall I think this is the best User Experience and Developer Experience :-)

swiffer:
https://grafana.com/blog/2024/07/09/upgrade-with-confidence-strategies-for-updating-your-self-hosted-grafana-instance/

Grafana 11.1.1 will be released in approx. 2 weeks. So think it's good to have that release out for that period.

Issue #4081, opened 2024-07-11
Unauthorized using Tesla Fleet API directly


XxTopKillerzZ:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I am trying to make the change to Tesla Fleet API like the docs say. Everything used to work with the old API.
I already have a developer Tesla account with client id and secret. Hosted public key. Proxy working for HomeAssistant Tesla Integration and everything works fine.

My interpretation of the docs at https://docs.teslamate.org/docs/guides/api#tesla-fleet-api is that I only need to change TESLA_AUTH_CLIENT_ID which I did. 
It says to "Add the following environment variable with your own domain" but then does not say which one.

I used the following configuration:

  - TESLA_API_HOST=https://fleet-api.prd.eu.vn.cloud.tesla.com
  - TESLA_AUTH_HOST=https://auth.tesla.com
  - TESLA_AUTH_PATH=/oauth2/v3
  - TESLA_AUTH_CLIENT_ID=my-client-id-from-tesla

When I initialize I get the logs that i'm putting here.

### Steps To Reproduce

1. Working instalation of teslamate using old API
2. Changed to FleetAPI like the docs say
3. Error 401 unauthorized

### Relevant log output

```shell
Attaching to teslamate-database-1, teslamate-teslamate-1
teslamate-database-1   | 
teslamate-database-1   | PostgreSQL Database directory appears to contain a database; Skipping initialization
teslamate-database-1   | 
teslamate-database-1   | 2024-07-11 23:04:15.807 UTC [1] LOG:  starting PostgreSQL 14.12 (Debian 14.12-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
teslamate-database-1   | 2024-07-11 23:04:15.808 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
teslamate-database-1   | 2024-07-11 23:04:15.808 UTC [1] LOG:  listening on IPv6 address "::", port 5432
teslamate-database-1   | 2024-07-11 23:04:15.812 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
teslamate-database-1   | 2024-07-11 23:04:15.821 UTC [27] LOG:  database system was shut down at 2024-07-11 23:02:26 UTC
[...]

PR #4095, opened 2024-07-22
Fix 401 on direct Fleet API calls

jlestel:
Fixed client_id passed in env var for direct calls to the Fleet API.

Due to a syntax error, the client_id was still "owner_api".

This should resolve the issue #4081


JakobLichterfeld (approved):

JakobLichterfeld:
Multiple users tested it successfull

PR #4104, opened 2024-07-27
Correct TPMS Pressure naming

helmo:
#3613 and #3718 modified the TPMS Pressure sensor naming, but part of is was still inconsistent.

this should match the names of the mqtt sensors with the calculated sensors (for psi)


JakobLichterfeld:
Thank you for your PR. The names in Home Assistant do not need to match the names of the MQTT topic name, innit?

helmo:
Correct mqtt name can be different, but the config to map that to an entity should have the same name as the ui dashboard config.

JakobLichterfeld:
Thanks, now I see the mismatch. Good catch!

PR #4124, opened 2024-08-05
small visual distinguish between AC & DC charging in charges dashboard and unification of the DC coloring in all dashboards

stauffenberg2020:
This commit just adds a different colour to the DC charges so that it can be easily visualized

image

### DIFF
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -896,7 +896,7 @@
                         "index": 0
                       },
                       "DC": {
-                        "color": "orange",
+                        "color": "light-orange",
                         "index": 1
                       }
                     },
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -876,6 +876,35 @@
                 "value": 62
               }
             ]
+          },
+	      {
+            "matcher": {
+              "id": "byName",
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
Thanks for your suggestion!

May I ask to align the color to Chargin Stats?

JakobLichterfeld:
> So should I change the charging stats dashboard to orange so that DC is represented by orange color in all these 3 dashboards?

That would be awesome!

grafana/dashboards/charges.json
```diff
@@ -876,6 +876,35 @@
...
+                "id": "mappings",
+                "value": [
+                  {
+                    "options": {
+                      "AC": {
+                        "color": "green",
+                        "index": 0
+                      },
+                      "DC": {
+                        "color": "orange",
```
JakobLichterfeld, line 899: this should be `light-orange` as well

JakobLichterfeld (changes requested):
Thanks for your change

JakobLichterfeld (approved):
now it looks perfect

sdwalker:
The battery health and charge pie chart greens are different as well

I agree with getting a consistent shade of colors used throughout all of the dashboards

PR #4144, opened 2024-08-19
Odometer in charges

jheredianet:
This is an improvement suggested in https://github.com/teslamate-org/teslamate/discussions/4109#discussioncomment-10252388

I've taken the opportunity to rearrange the stats as suggested in other posts. 

image


JakobLichterfeld (approved):
Thanks for your effort!

PR #3971, opened 2024-06-18
Export all Dashboards via Grafana 11.0.0

swiffer:
@JakobLichterfeld - as discussed yesterday this is all remaining dashboards exported once straight out of Grafana 11 (using the latest master branch as a base)

PR #4153, opened 2024-08-22
Fix the issue of failing to restore  efficiency dashboard.

ghostiee:
The __inputs field in efficiency.json will cause the efficiency dashboard restore to fail.

The error message is: 
{"message":"dashboard import failed: missing dashboard input variable DS_TESLAMATE","traceID":""}

A sample command to reproduce this behavior is: 
```bash
curl   --user "admin:admin" --show-error   -X POST -H "Content-Type: application/json"   -d "{\"dashboard\":$(cat ./grafana/dashboards/efficiency.json), \
      \"overwrite\":true, \
      \"folderId\": null, \
      \"inputs\":[{\"name\":\"DS_CLOUDWATCH\", \
      \"type\":\"datasource\", \
      \"pluginId\":\"cloudwatch\", \
      \"value\":\"TeslaMate\"}]}"   "localhost:3000/api/dashboards/import"
```

Due to the absence of error messages within dashboards.sh and the uninterrupted nature of the restore process, I've added the --fail parameter to the curl command. This modification will cause the restore process to terminate upon encountering an restore failure for the dashboard, allowing for timely detection and resolution of the issue.

### DIFF
--- a/grafana/dashboards.sh
+++ b/grafana/dashboards.sh
@@ -62,7 +62,7 @@ restore() {
       while IFS= read -r -d '' dashboard_path; do
           folder_id=$(get_folder_id "$(basename "$dashboard_path" .json)")
           curl \
-            --silent --show-error --output /dev/null \
+            --silent --fail --show-error --output /dev/null \
             --user "$LOGIN" \
             -X POST -H "Content-Type: application/json" \
             -d "{\"dashboard\":$(cat "$dashboard_path"), \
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
@swiffer May I ask you to look into the `__inputs` field in [efficency.json](https://github.com/teslamate-org/teslamate/blob/0c154ae11f6e19efea80eaa7cb125d224a19145e/grafana/dashboards/efficiency.json#L2) , as it was introduced by you with #3971.

swiffer:
Ups - sorry, nice find. yes change looks valid and approved from my end!

Issue #4169, opened 2024-08-29
After recent update, charges with negative cost are not visible by default


woyteck1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Until the redesign of charges page, all charges were visible, even the ones with negative cost. These happen due to negative electricity tarrifs, where electricity supplier pays us to use electricity, therefore we use it.
Now, a field appeared on the Charges page, and it is a required one, and by default it does NOT show any charges with negative cost.
![image](https://github.com/user-attachments/assets/95dd4746-88c8-4adf-844d-1367346da301)
Can this please be made an optional field, or Zero changed to all numbers - and this being the default ?
I thought I was going mad because my charges started disappearing from the list!!

### Expected Behavior

Expected behavior is that when I do not select any filters, I expect to see all of the charges on the list.
Instead I have a lot of them hidden from the view.

### Steps To Reproduce

1. Goto a charge and change the cost to negative.
2. Go to charges pages and observe how this charge is hidden by the default filtering.

### Relevant log output

```shell
No logs. This is purely UI design issue.
```

### Type of installation

Docker

### Version

v1.30.1

woyteck1:
So, this valuse could be changed to Custom, with All being default and any other value being an option, like this: 
![image](https://github.com/user-attachments/assets/3acb7812-4778-4f3b-a262-5eab20aed0ae)

Also this query has a flaw, because if someone changes the default '0' to something else, then entries which did not have the cost set up yet will not be visible.
```
    AND CASE WHEN $cost = 0 THEN (cost IS NULL OR cost >= 0) ELSE cost >= $cost END
```

JakobLichterfeld:
Thanks for reporting and nice find. Now that you've said it, it's obvious.

@jheredianet May I ask you to look into this?

PR #4179, opened 2024-09-03
feat: Improve cost filter on Charges dashboard to show charges with negative cost as well

jheredianet:
This is an improvement on the cost filter on Charges dashboard suggested here: https://github.com/teslamate-org/teslamate/issues/4169#issuecomment-2320326037


JakobLichterfeld (approved):
Awesome, ty!

PR #4186, opened 2024-09-07
improve charge level dashboard

### LINKED ISSUE #4185: Charge Level maxing out Grafana Row Limit of 1000000 for larger periods

### What happened?

When selecting larger periods in Charge Level Grafana Row Limit of 1000000 becomes effective as Query is selecting from positions table.

grafik

### Expected Behavior

_No response_

### Steps To Reproduce

1. Go to Charge Level Dashboard
2. Select Period of 6 month / a year

### Relevant log output

```shell
-
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.30.1

swiffer:
fixes #4185

by making use of $__interval_ms and time bucketing long periods can be shown without hitting the row limit of 1000000.

$__interval_ms is divided by 1000 to ensure avg() calcuations to not smooth out the charge level or still show min / max values reflecting reality but significantly reducing number of data points returned by the query (70.000 instead of 2.000.000) for 6 months in my case.

would be great if someone with long lasting history or slow performing hardware could test to see performance impact (as we are time bucketing now this could mean longer query load time). to make it a fair comparsion going like for like (6 months vs 6 months) would be fair i guess.

After the change - see linked issue for a screenshot before introducing the change

grafik

### DIFF
--- a/grafana/dashboards/charge-level.json
+++ b/grafana/dashboards/charge-level.json
@@ -171,12 +171,13 @@
             "type": "grafana-postgresql-datasource",
             "uid": "TeslaMate"
           },
+          "editorMode": "code",
           "format": "time_series",
           "group": [],
           "hide": false,
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT\n\t$__time(date),\n\tbattery_level AS \"Battery Level\",\n\tusable_battery_level AS \"Usable Battery Level\"\nfrom positions\n\tWHERE $__timeFilter(date) AND car_id = $car_id\n\tORDER BY Time ASC\n;",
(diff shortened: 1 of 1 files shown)


sdwalker:
Before 
image

After
image

New version loaded in ~1/3 of the time the previous version took for 6 months

swiffer:
will most likely close this one in favour of #4187 if we can performance to an acceptable level (or introduce a toggle on the dashboard to show / hide the results of this query). thanks for testing @sdwalker - if you have some time please try out the latest query from #4187

swiffer:
closed in favor of #4200

PR #4191, opened 2024-09-09
increase max battery charge gauge threshold to 101 in case of LFP

### LINKED ISSUE #4184: battery level gauge shows red value if battery is 100% and LFP is selected

### What happened?

In case of LFP Tesla, that is, lfp option is flagged in teslamate config, the battery level gauge correctly shows different alert thresholds, however, the edge case where the battery is actually 100% charged still shows as red.
It's probably enough to change the limit config query to use "101" as max level in case of LFP. See screenshot for proposed solution.

### Expected Behavior

See screenshot

### Steps To Reproduce

_No response_

### Relevant log output

```shell
See screenshot
```

### Screenshots

issue with 100 percent

proposed change

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.30.1

neothematrix:
fixes #4184: visualization issue of gauge turning "red" in case of 100% battery charged on lfp teslas

### DIFF
--- a/grafana/dashboards/overview.json
+++ b/grafana/dashboards/overview.json
@@ -203,7 +203,7 @@
           "format": "table",
           "hide": false,
           "rawQuery": true,
-          "rawSql": "SELECT\r\n  0 as lowest,\r\n  10 as low,\r\n  20 as mid,\r\n  CASE WHEN lfp_battery THEN 100 ELSE 81 END as high,\r\n  CASE WHEN lfp_battery THEN 100 ELSE 91 END as highest\r\nfrom cars inner join car_settings on cars.settings_id = car_settings.id\r\nwhere cars.id = $car_id",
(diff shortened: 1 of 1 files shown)


grafana/dashboards/overview.json
```diff
@@ -2169,4 +2169,4 @@
   "uid": "kOuP_Fggz",
   "version": 9,
```
JakobLichterfeld, line 2170: version should be updated

JakobLichterfeld (changes requested):
Thanks for your suggestion.

grafana/dashboards/overview.json
neothematrix: @JakobLichterfeld  thanks, I missed it, new commit is ready

JakobLichterfeld (approved):

swiffer:
looks good!

PR #4192, opened 2024-09-11
Update manually_fixing_data.mdx to use simple quotes to fix data

florianfish:


JakobLichterfeld:
Thanks for your suggestion. The current documentation works for all systems, single quotes do not

PR #4198, opened 2024-09-14
feat: display vehicle VIN as a fallback for vehicle name on grafana dashboards

arcastro:
### Context

* In cases of unnamed vehicles, 
    * The grafana dashboards currently display "null".
    * This pull-requests adds fallback logic to display the vehicle vin instead.
* for cases of named vehicles
    * the behavior remains unchanged and the vehicle name is still displayed.

### Links
* https://github.com/teslamate-org/teslamate/discussions/4180


JakobLichterfeld (approved):
Nice find and thanks for your changes!

Issue #4156, opened 2024-08-22
Auto update hashes in nix


brianmay:
Purpose of this bug report is to document my current findings on this issue in a public place.

I would like a script that can automatically update the hashes in the flakes.

https://github.com/teslamate-org/teslamate/blob/9b2de5ffa0d0943c9bd89825af681fea14af91df/flake.nix#L27
https://github.com/teslamate-org/teslamate/blob/9b2de5ffa0d0943c9bd89825af681fea14af91df/flake.nix#L35

Here is a solution for npm: https://sitr.us/2024/03/08/nix-npm-and-dependabot.html/

Mix is the problem, I don't know obtain the new hash without building with fakeHash, and then replacing the hash with the hash in the error.

Google search showed me this project which attempted to solve this problem https://github.com/argoproj/argo-workflows/pull/12520 with the following shell script:

```
# this is reliant on the vendorHash being on line 195
sed -i '195s/vendorHash = \"\([^\"]*\)\"/vendorHash = ""/g' ./dev/nix/flake.nix
NIX_HASH=$(nix --log-format raw build ./dev/nix 2>&1 | grep  "got: " | awk '{ split($0,a,"got:    "); print a[2] }')
sed -i '195s|vendorHash = \"\([^\"]*\)\"|vendorHash = "'$NIX_HASH'"|g' ./dev/nix/flake.nix
echo "Changed Nix hash to : $NIX_HASH"
rm -rf result
```

Unfortunately this is broken, it will try to build the project with the old hash, and if the old hash exists, it will be the wrong version.

brianmay:
I asked about this on the matrix channel https://matrix.to/#/!RRerllqmbATpmbJgCn:nixos.org/$HxXrDb80i36hBG3uY4kE8PaX_G3NsM4bKFInux3Q-Xg?via=nixos.org&via=matrix.org&via=tchncs.de

![image](https://github.com/user-attachments/assets/8891d8b2-77b7-446a-9899-04c0e8831071)

Probably need to find a better forum where the maintainers of the elxir and npm code can see it.

brianmay:
Haven't looked into detail,  but this might help (?) https://docs.determinate.systems/guides/automatically-fix-hashes-in-github-actions/

brianmay:
Trying this out on my project. https://github.com/brianmay/phone_db/

Some points:

* Instructions are not 100% correct, had to make some minor changes.

* The Action fails on successfully updating hash, which was confusing at first.

[...]

PR #4207, opened 2024-09-18
ci: enable dependabot for mix and npm

JakobLichterfeld:
reenable Dependabot for

- [x] mix
- [x] npm in `/assets`
- [x] nom in `/website`

it was disabled 2 years ago in https://github.com/teslamate-org/teslamate/commit/6e876d76852c34b48b3ca664ac0cd5df1d97ff56

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,10 @@ updates:
     schedule:
       interval: "monthly"
 
-  # - package-ecosystem: "npm"
-  #   directory: "/assets"
-  #   schedule:
-  #     interval: "monthly"
+  - package-ecosystem: "npm"
+    directory: "/assets"
+    schedule:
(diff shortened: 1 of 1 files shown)


brianmay:
On thing to note is that any change to the dependencies will break the nix build. See #4156 

I think we should apply this regardless.

But prehaps before we make a release or something we should make sure the hashes in `flake.nix` are up-to-date.

brianmay (approved):

JakobLichterfeld:
> On thing to note is that any change to the dependencies will break the nix build. See #4156

Yeah, thanks for pointing it out. Will try to wirte some sort of how-to-release doc in the future until we are able to automate it via ci.

> I think we should apply this regardless.

I debated with myself about doing the first update by hand, but I think this is the more transparent way.

PR #4217, opened 2024-09-18
timezone fixes and moving window on beginning of period

swiffer:
Hi @JakobLichterfeld

as promised:

- some refinements for beginning of period
- and ensuring joins considering time zone correctly where needed


JakobLichterfeld (approved):
Awesome, ty!

PR #4219, opened 2024-09-18
ci: add treefmt as code formatting multiplexer, refactor CI to avoid duplication, reorg CI into DevOps workflow

### LINKED ISSUE #4218: ci: add treefmt as code formatting multiplexer
As we do have a multitude of programming languages, running each linter separate is not efficient.

[Treefmt](https://github.com/numtide/treefmt) solves this and comes with speed improvements as well. See the [documentation](https://treefmt.com/) for more information.

Using the VS Code extension it is very convenient to run it on each save, for example.

JakobLichterfeld:
As we do have a multitude of programming languages, running each linter separate is not efficient.

[Treefmt](https://github.com/numtide/treefmt) solves this and comes with speed improvements as well. See the [documentation](https://treefmt.com/) for more information.

- [x] treefmt.toml config
- [x] lint whole project accordingly
  - [x] exclude Grafana Dashboards
- [x] CI
  - [x] refactor CI to avoid duplication
  - [x] reorg CI into DevOps workflow
  - [x] allow ghcr_build parallel to elixir test
  - [x] GitHub Action for treefmt ~or Pre-Commit Hook~ using [treefmt-nix](https://github.com/numtide/treefmt-nix)

closes #4218


swiffer:
👍, but would prefer formatting as created by grafana export for dashboards

JakobLichterfeld:
> 👍, but would prefer formatting as created by grafana export for dashboards

Yes, I stumbled over that too. Will exclude the dashboard config from the prettier linter.

JakobLichterfeld:
> Will exclude the dashboard config from the prettier linter.

Done

brianmay:
Those instructions seem to assume the use of flake-root, which is not something I am familiar with. Seems weird. But maybe a natural reaction from me for anything I am not familiar with ;-) - doesn't mean it is bad.

Maybe these instructions might be better? https://github.com/numtide/treefmt-nix?tab=readme-ov-file#flakes

But I guess the problem here is that you have duplicated the treefmt in nix format and toml format. I can see why they did it that way - we need to be able to specify the full paths to the nix programs, but it isn't particularly great. It would be better if we could read the toml file, and override the paths.

Did I understand your concern correctly?

JakobLichterfeld:
Yeah exactly. The instructions in the treefmt-nox readme differ from the detailed walkthrough mentioned and linked in the same readme.

Was thinking about reading the toml config to avoid duplication as well. 

1. In the first step I wanted to have it run without nix, which is working.
2. The next step would be to have it run in nix, which I tried but is not yet working.
3. The CI checks linting with running nix flake check which invokes treefmt.
4. Then a refactoring to use the same config

I'm stucked with  No. 2, and before throwing more time on it decided to ask for your help 😊

.github/workflows/ensure_linting.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@ba0dd844c9180cbf77aa72a116d6fbc515d0e87b #v27
+        with:
+          nix_path: nixpkgs=channel:nixos-unstable
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: run treefmt in CI mode
+        run: nix flake check --impure
```
scottbot95, line 22: Just want to call out that `nix flake check` will run *all* checks (which is something we really probably should be doing in CI). IMO we should just rename this job to something like "run nix checks" (and adjust any comments as needed) and call it a day.

If you do specifically want to just run the formatter here, unfortunately there's no built-in way to run a specific check so you'd have to do something like `nix build .#checks.x86_64-linux.formatting` (or just invoke treefmt outside of nix entirely)

scottbot95 (changes requested):
It seems like there's not really an easy way to have treefmt read the `toml` file from nix (without using something like the [parseTOML](https://github.com/mozilla/nixpkgs-mozilla/blob/master/lib/parseTOML.nix) helper from mozilla and I'm not really a huge fan of adding another flake input just for this), so I think if we have a hard requirement on being able to run treefmt both as a flake check and outside of nix entirely, we simply have to duplicate the config... :(

.github/workflows/ensure_linting.yml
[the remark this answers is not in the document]
brianmay: I imagine `--impure` is a left over from before we defined the `checks` output. Without the `checks` it seems to want top check `devShell`, and that does require `--impure`. So maybe we don't need it here.

Supposedly this PR was to prevent `devShell` requiring `--impure` also,  https://github.com/cachix/devenv/pull/1018 - - I think that requires adding a new input to our flake though (if I understand this correctly).

flake.nix
```diff
@@ -97,98 +117,179 @@
...
-              enable = true;
-              package = pkgs.postgresql_16;
-              listen_addresses = "127.0.0.1";
-              port = postgres_port;
-              initialDatabases = [{ name = "teslamate"; }];
-              initialScript = ''
-                CREATE USER teslamate with encrypted password 'your_secure_password_here';
-                GRANT ALL PRIVILEGES ON DATABASE teslamate TO teslamate;
-                ALTER USER teslamate WITH SUPERUSER;
+          modules = with pkgs; [
```
brianmay, line 120: This is probably my fault. But using `with` seems to be very much discouraged, the way it works is potentially confusing. Particularly on a large scope such as used here.

A number of people have written about the issues, and there are different options for replacing it. e.g. https://www.tweag.io/blog/2023-01-24-nix-with-with-nickel/

flake.nix
```diff
@@ -97,98 +117,179 @@
...
-        nixosModules.default = import ./module.nix { inherit self; };
-      };
+
+        # for `nix fmt`
+        formatter = eachSystem (pkgs: treefmtEval.${pkgs.system}.config.build.wrapper);
+        # for `nix flake check`
+        checks = eachSystem (pkgs: {
+          default = moduleTest;
+          formatting = treefmtEval.${pkgs.system}.config.build.check self;
+        });
```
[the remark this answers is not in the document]
JakobLichterfeld: > `flake-utils`.

`flake-utils` was entirely new for me

flake.nix
```diff
@@ -2,13 +2,27 @@
...
 
   inputs = {
-    nixpkgs = { url = "github:NixOS/nixpkgs/nixos-unstable"; };
-    flake-utils = { url = "github:numtide/flake-utils"; };
+    nixpkgs = {
+      url = "github:NixOS/nixpkgs/nixos-unstable";
+    };
+    flake-utils = {
+      url = "github:numtide/flake-utils";
+    };
```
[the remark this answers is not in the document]
JakobLichterfeld: > IMO I prefer the latter.

Me also, just did not want to change your code :-)

.github/workflows/ensure_linting.yml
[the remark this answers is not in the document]
JakobLichterfeld: Yeah it was a result of the devShell.

JakobLichterfeld: > Just want to call out that `nix flake check` will run _all_ checks (which is something we really probably should be doing in CI). IMO we should just rename this job to something like "run nix checks" (and adjust any comments as needed) and call it a day.

Yeah, will add this to CI as well, which enhance our elixir test cases with a real test of db connection for example.


> If you do specifically want to just run the formatter here, unfortunately there's no built-in way to run a specific check so you'd have to do something like `nix build .#checks.x86_64-linux.formatting` (or just invoke treefmt outside of nix entirely)

JakobLichterfeld:
> If you do specifically want to just run the formatter here, unfortunately there's no built-in way to run a specific check so you'd have to do something like `nix build .#checks.x86_64-linux.formatting` (or just invoke treefmt outside of nix entirely)

I wanted to add treefmt via nix, as I realized how many github actions I need to add to setup all the formatters in ci.
I now think about adding all dependencies to nix devShell and get rid of many github actions, which will result in improved sec and improved performance.

flake.nix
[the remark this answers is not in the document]
brianmay: flake-utils is somewhat controversial, there is a school of thought that you should not use it, but rather publish for a list of known good architectures. I am not really convinced this is a good idea though when publishing architecture independant code.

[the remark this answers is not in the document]
JakobLichterfeld: Thanks for the background information. Even though I'm ill, I've started reading more about nix in the last few days and will continue to do so.

JakobLichterfeld:
I tend to use treefmt via flakeModule like [here](https://github.com/brianmcgee/astronvim.nix/blob/main/flake.nix)

flake.nix
```diff
@@ -97,98 +117,179 @@
...
+
+            programs.nixpkgs-fmt.enable = true;
+          }
+        );
+      in
+      {
+        imports = [
+          inputs.treefmt-nix.flakeModule
+          inputs.flake-root.flakeModule
+        ];
```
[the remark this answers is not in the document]
scottbot95: Haha I know the feeling.

.github/workflows/ensure_linting.yml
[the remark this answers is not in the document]
scottbot95: Ah bummer. Well if we need it, we need it. I'll just cry every time I see this ;)

scottbot95:
> I tend to use treefmt via flakeModule like [here](https://github.com/brianmcgee/astronvim.nix/blob/main/flake.nix)

Yeah that's probably better. Maybe we just also migrate this repo to flake-parts while we at it 😉(jk this PR is already doing a lot, we shouldn't pile on the refactors)

.github/workflows/ensure_linting.yml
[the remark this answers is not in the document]
brianmay: I believe it shouldn't be required anymore, because the tests don't construct the devenv anymore.

JakobLichterfeld:
@brianmay I'm not able get the devenv running, even with latest master branch I get:

```bash
direnv: loading ~/development/repos/teslamate/.envrc                                                                                                                           
direnv: using flake . --impure
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'devenv-shell'
         whose name attribute is located at /nix/store/3qsyy1d9wlxy0zaa5l27nwis7cjf93qs-source/pkgs/stdenv/generic/make-derivation.nix:352:7

       … while evaluating attribute 'DEVENV_PROFILE' of derivation 'devenv-shell'

         at «none»:0: (source not available)

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: A definition for option `packages."[definition 4-entry 3]"' is not of type `package'. Definition values:
       - In `<unknown-file>': null
 ```

**Comment by brianmay** [comment]
Very odd. "Works for me" :-)

What happens if you try:

```shell
nix develop --impure
```

Should do the same thing, actually, but just checking something isn't wrong with your devenv.

JakobLichterfeld:
> Very odd. "Works for me" :-)

Was just waiting for this 😀


> What happens if you try:
> 
> ```shell
> nix develop --impure
> ```
> 
> Should do the same thing, actually, but just checking something isn't wrong with your devenv.

Same result, so I assume it is not releated to my devenv setup. But will try with a clean git clone

JakobLichterfeld:
> But will try with a clean `git pull`

Same result. Deleted whole folder, did a clean `git pull`. Running on arm64 here

brianmay:
Hmmm. I tested with x86, I have an arm64 VM, will try that if I remember tomorrow.

brianmay:
I am guessing that one of the packages listed in the `packages = ` might be broken on arm64, although a quick glance it looks OK to me.

JakobLichterfeld:
> I am guessing that one of the packages listed in the `packages = ` might be broken on arm64, although a quick glance it looks OK to me.

Yeah, good point. Will test around the coming days. Just want this pr to be done and therefore wanna test the devenv as well.

brianmay:
Reproduced. Under OSX (arm64-darwin), so guessing you are using OSX also.

The problem is the `glibcLocales` reference. At least if I remove that, it works. Still a bit confused, I can install that manually, although it does pull in a number of other packages...

brianmay:
I think glibcLocales is not used on darwin, so probably this needs to be made conditional on Linux.

JakobLichterfeld:
> I think `glibc` is not used on darwin, so probably this needs to be made conditional on Linux.

Good catch, I can confirm that this solved the problem.

JakobLichterfeld:
> > I tend to use treefmt via flakeModule like [here](https://github.com/brianmcgee/astronvim.nix/blob/main/flake.nix)
> 
> Yeah that's probably better. Maybe we just also migrate this repo to flake-parts while we at it 😉(jk this PR is already doing a lot, we shouldn't pile on the refactors. Maybe as a follow-up PR though....)

Any idea where to put `imports = [  inputs.treefmt-nix.flakeModule ];`? I am lost on this :-(

.github/workflows/ensure_linting.yml
JakobLichterfeld: > Yeah, will add this to CI as well, which enhance our elixir test cases with a real test of db connection for example.

will be tracked with #4234

.github/workflows/elixir_test.yml
```diff
@@ -0,0 +1,55 @@
...
+  CACHE_NAME_COMPILED_TEST: cache-compiled-test-build
+  ELIXIR_ASSERT_TIMEOUT: 1000
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-20.04
```
swiffer, line 17: Bumping to 24.04? Or does it cause problems?

.github/workflows/elixir_test.yml
```diff
@@ -0,0 +1,55 @@
...
+
+    permissions:
+      contents: read
+
+    services:
+      db:
+        image: postgres:16
```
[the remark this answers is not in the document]
JakobLichterfeld: yeah, this will be solved with merge-conflicts

.github/workflows/elixir_test.yml
JakobLichterfeld: I do want to keep changes as small as possible, therefore will go with ubuntu-20.04 in the first step

JakobLichterfeld:
> Any idea where to put `imports = [ inputs.treefmt-nix.flakeModule ];`? I am lost on this :-(

I just refactored to use flake-parts. Looks promising, beside the build folder issue in `postBuild`. Will keep looking, any hints are welcome.

JakobLichterfeld:
Finaly the devenv is running correctly :-)

flake.nix
```diff
@@ -2,193 +2,252 @@
...
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
+      # See ./nix/modules/*.nix for the modules that are imported here.
+      imports = [
+        inputs.devenv.flakeModule
+      ] ++ builtins.map (fn: ./nix/modules/${fn}) (builtins.attrNames (builtins.readDir ./nix/modules));
```
scottbot95, line 31: It is tempting to be clever and automatically discover modules to load, however IMO it generally better to just list them all out explicitly (especially when we only have a small handful of modules). It makes it more clear what files are being imported and gives us more control if we want to do anything beyond a flat directory with nothing but *.nix flake modules.

nix/modules/formatter.nix
scottbot95, line 1: It is convention that "flake modules" (eg modules used by flake-parts) are put in a directory labeled "flake-modules" instead of just "modules" to help differentiate between flake-parts modules and nixos modules.

scottbot95 (changes requested):
I love the move to flake-parts. I think we should probably also make the following refactors, but they probably deserve a separate PR since this PR is already going a lot.

- move `module.nix` to the `nix/nixos` or maybe something like `nix/flake-modules/nixos-modules`
- extract the package creation stuff to its own flake module
- extract the checks to their own flake module
- extract the dev shell to its own flake module (and possible migrate to use [partitions](https://flake.parts/options/flake-parts-partitions))

I don't mind submitting a PR with these changes once this one is completed.

JakobLichterfeld:
> I love the move to flake-parts.

Thanks.

> I think we should probably also make the following refactors, but they probably deserve a separate PR since this PR is already doing a lot.
> 
> * move `module.nix` to the `nix/nixos` or maybe something like `nix/flake-modules/nixos-modules`
> * extract the package creation stuff to its own flake module
> * extract the checks to their own flake module
> * extract the dev shell to its own flake module (and possible migrate to use [partitions](https://flake.parts/options/flake-parts-partitions))

Agree. Already tried to do some parts of it, but I have struggled with getting the build to run in extracted part. Maybe need to inherit the src part, idk.

> I don't mind submitting a PR with these changes once this one is completed.

Would love to see your PR.

flake.nix
JakobLichterfeld: agree on this

nix/modules/formatter.nix
JakobLichterfeld: makes sense

JakobLichterfeld:
glibcLocales seems to be broken on linux. I disabled it, but the runner now gets issues with locales:

```bash
mix-format error:
warning: the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8. Please ensure your locale is set to UTF-8 (which can be verified by running "locale" in your shell) or set the ELIXIR_ERL_OPTIONS="+fnu" environment variable
```

We use https://github.com/elixir-cldr/cldr for locales

JakobLichterfeld:
> but the runner now gets issues with locales:

Solved by setting `ELIXIR_ERL_OPTIONS="+fnu"`

JakobLichterfeld:
Locale seems to be correct

runner has

```bash
LANG=C.UTF-8
LANGUAGE=
LC_CTYPE="C.UTF-8"
LC_NUMERIC="C.UTF-8"
LC_TIME="C.UTF-8"
LC_COLLATE="C.UTF-8"
LC_MONETARY="C.UTF-8"
LC_MESSAGES="C.UTF-8"
LC_PAPER="C.UTF-8"
LC_NAME="C.UTF-8"
LC_ADDRESS="C.UTF-8"
LC_TELEPHONE="C.UTF-8"
LC_MEASUREMENT="C.UTF-8"
LC_IDENTIFICATION="C.UTF-8"
LC_ALL=
```

direnv in runner has
LANG=C.UTF-8
LC_CTYPE="C.UTF-8"
LC_NUMERIC="C.UTF-8"
LC_TIME="C.UTF-8"
LC_COLLATE="C.UTF-8"
LC_MONETARY="C.UTF-8"
LC_MESSAGES="C.UTF-8"
LC_PAPER="C.UTF-8"
LC_NAME="C.UTF-8"
LC_ADDRESS="C.UTF-8"
LC_TELEPHONE="C.UTF-8"
LC_MEASUREMENT="C.UTF-8"
LC_IDENTIFICATION="C.UTF-8"
LC_ALL=
```
```bash

**Comment by JakobLichterfeld** [comment]
Still getting

```bash
Could not find "rebar3", which is needed to build dependency :unicode_util_compat
Shall I install rebar3? (if running non-interactively, use "mix local.rebar --force") [Yn] ** (Mix) Could not find "rebar3" to compile dependency :unicode_util_compat, please ensure "rebar3" is available
```

brianmay:
In what way is glibcLocales broken on Linux? Maybe we should submit a bug report?

Wonder why we are only getting issues with rbar3 now. 

We use the `mixRelease` function, which is defined here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/beam-modules/mix-release.nix

This sets the `MIX_REBAR3` environment variable: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/beam-modules/mix-release.nix#L129C5-L130C1

I am assuming Elxiir is then able to find the library from the environment variable, but for some reason it isn't working.

JakobLichterfeld:
> In what way is glibcLocales broken on Linux? Maybe we should submit a bug report?

devShell does complain it is not a package (what it really isn't). https://github.com/teslamate-org/teslamate/actions/runs/11157566497/job/31012111403

> Wonder why we are only getting issues with rbar3 now.
> 
> We use the `mixRelease` function, which is defined here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/beam-modules/mix-release.nix
> 
> This sets the `MIX_REBAR3` environment variable: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/beam-modules/mix-release.nix#L129C5-L130C1
> 
> I am assuming Elxiir is then able to find the library from the environment variable, but for some reason it isn't working.

Yeah, it is even stranger: doing a 

```bash
mix local.rebar --force
mix local.hex --force
```

when entering the shell (similar to our [DOCKERFILE](https://github.com/teslamate-org/teslamate/blob/dc64687ba8f150b9da664cc411c4f7111f231418/Dockerfile#L18)) does not solve it.

JakobLichterfeld:
I'm about to give up...

When running `devenv up` I do get

```bash
❯ devenv up
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'devenv-up'
         whose name attribute is located at /nix/store/fpivx4sjcp2vk4rp9nhliln5cwcp3kc6-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'text' of derivation 'devenv-up'

         at /nix/store/fpivx4sjcp2vk4rp9nhliln5cwcp3kc6-source/pkgs/build-support/trivial-builders/default.nix:101:16:

          100|       ({
          101|         inherit text executable checkPhase allowSubstitutes preferLocalBuild;
             |                ^
          102|         passAsFile = [ "text" ]

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'unix-socket' missing

       at /nix/store/m0j6n8wnynywa7fqa42h7f5l70ffh0id-source/src/modules/process-managers/process-compose.nix:42:54:

           41|       ${cfg.package}/bin/process-compose --config ${cfg.configFile} \
           42|         --unix-socket ''${PC_SOCKET_PATH:-${toString config.process.process-compose.unix-socket}} \
             |                                                      ^
           43|         --tui=''${PC_TUI_ENABLED:-${lib.boolToString config.process.process-compose.tui}} \
```

On [latest master](https://github.com/teslamate-org/teslamate/commit/dc64687ba8f150b9da664cc411c4f7111f231418) it runs fine (when disabling glibcLocales on darwin). When running `nix flake update` on latest master I get the same error.

brianmay:
> On [latest master](https://github.com/teslamate-org/teslamate/commit/dc64687ba8f150b9da664cc411c4f7111f231418) it runs fine (when disabling glibcLocales on darwin). When running nix flake update on latest master I get the same error. 

Confirmed, I see that too. will investigate.

brianmay:
This appears to be a bug in devenv. I filled a bug report: https://github.com/cachix/devenv/issues/1497

Version e316837bf4d0256e56ef26aa3778a10086f8db1a - the git commit immediately before the problem commit - should be fine also.

Switching to nixos-23.05 might be a good idea also, but it won't solve this problem.

In mean time if you just want to update, e.g. nixpkgs without dev, use:

```
nix flake lock --update-input nixpkgs
```

JakobLichterfeld:
> This appears to be a bug in devenv. I filled a bug report: [cachix/devenv#1497](https://github.com/cachix/devenv/issues/1497)

Ty!

> Version e316837bf4d0256e56ef26aa3778a10086f8db1a - the git commit immediately before the problem commit - should be fine also.

So we can pin devenv to this, innit?

brianmay:
> So we can pin devenv to this, innit?

Yes, that sounds good for now.

JakobLichterfeld:
Pinning devenv solved it, `devenv up` is now working as expected with new flake.

JakobLichterfeld:
> I am assuming Elxiir is then able to find the library from the environment variable, but for some reason it isn't working.

My guess is that the mix installed via `builtins.attrValues config.treefmt.build.programs` collides with elixir installed via beam package. Commenting it out seems to fix it.

JakobLichterfeld:
> My guess is that the mix installed via `builtins.attrValues config.treefmt.build.programs` collides with elixir installed via beam package. Commenting it out seems to fix it.

This turns out not to be true. We need nixpgks-unstable for treefmt 2.0.5. When using unstable we get the rebar error even without `builtins.attrValues config.treefmt.build.programs`

scottbot95:
> Running nix develop . --impure --command treefmt --ci locally (aarch64-darwin) does not result in the rebar3 error, only on GitHub Runner.

And that is exactly why we shouldn't be using `--impure` 😉 

Debugging might easier if you could run locally in the same way as GH runner. In that vein, might be worth trying the trick from https://github.com/cachix/devenv/pull/1018 to see if we can get away with not needing `--impure`. Normally I'm against extra inputs unless needed, but the input in this case is just `/dev/null` so it's fine IMO. @brianmay mentioned issues getting this to work with this project, but it might be worth spending more time to figure it out if we are somewhat blocked on things running differently locally vs in GH actions.

That said, I know essentially nothing about elixir so any suggestion I might give is from a purely nix-oriented perspective and may not actually be useful.

brianmay:
> @brianmay mentioned issues getting this to work with this project, but it might be worth spending more time to figure it out if we are somewhat blocked on things running differently locally vs in GH actions.

I didn't try too hard, I might have done something stupid.

brianmay:
It seems to work now. https://github.com/teslamate-org/teslamate/pull/4245

brianmay:
> This turns out not to be true. We need nixpgks-unstable for treefmt 2.0.5. When using unstable we get the rebar error even without builtins.attrValues config.treefmt.build.programs

Possibly we could add another input and add nixpgks-unstable just for treefmt. Not sure if it is worth it though.

brianmay:
At present one editor (visual code) I use 
wants to reformat the nix files one way, and my other editor (nixvim) wants to reformat the files another way. Arrgh. Hopefully I can try to address this better once this is merged.

JakobLichterfeld:
> And that is exactly why we shouldn't be using `--impure` 😉

Total agree. 

> Debugging might easier if you could run locally in the same way as GH runner.

Yes, this is my motivation to move to nix for our DevOps. There is https://github.com/nektos/act to run GitHub Actions locally, but it has limitations, so local execution is never the same as on GitHub runners. I want to have the most robust DevOps pipeline using nix. And I want to get rid of works on my machine. :-)

JakobLichterfeld:
> Possibly we could add another input and add nixpgks-unstable just for treefmt. Not sure if it is worth it though.

Yeah, tried that already but with no success, seems I'm not fully on the page how to then use it with the `config.treefmt.build.wrapper`.

JakobLichterfeld:
> At present one editor (visual code) I use wants to reformat the nix files one way, and my other editor (nixvim) wants to reformat the files another way. Arrgh. Hopefully I can try to address this better once this is merged.

I know the pain, that's one additional reason for using this PR.

JakobLichterfeld:
> It seems to work now. #4245

Ty. `devenv up` now works without  `--impure`. Flake check and `nix develop .` don't.

JakobLichterfeld:
I'm getting there.

`nix develop . --command treefmt` (without `--impure`) works fine when commenting out: `devenv-up = devShell.config.procfileScript;` and `devShells.default = devShell;`

JakobLichterfeld:
> I'm getting there.
> 
> `nix develop . --command treefmt` (without `--impure`) works fine when commenting out: `devenv-up = devShell.config.procfileScript;` and `devShells.default = devShell;`

It works without `--impure`with devenv enabled:

`nix develop --override-input devenv-root "file+file://"<(printf %s "$PWD") . --command treefmt --ci`

JakobLichterfeld:
Even without `--impure` the Action fails on GitHub runner due to [missing rebar3](https://github.com/teslamate-org/teslamate/actions/runs/11210968994/job/31158903535?pr=4219#step:5:897), [even if it is created](https://github.com/teslamate-org/teslamate/actions/runs/11210968994/job/31158903535?pr=4219#step:5:472)

brianmay:
> Even without --impure the Action fails on GitHub runner due to [missing rebar3](https://github.com/teslamate-org/teslamate/actions/runs/11210968994/job/31158903535?pr=4219#step:5:897), even if [it is created](https://github.com/teslamate-org/teslamate/actions/runs/11210968994/job/31158903535?pr=4219#step:5:472): * creating /home/runner/.mix/elixir/1-16/rebar3. Treefmt use latest elixir treefmt: error: formatting failure: formatter '/nix/store/r2bjg78m7b7s75lwcgn69svk010j3sqb-elixir-1.17.3/bin/mix' with options '[format]' failed to apply: exit status 1.

It kinds of sounds like something depends on something concerning your environment. Even though that isn't suppose to happen.

> Locale is also an issue as told by elixir, even if locale output on runner in nix develop is fine:

Strange.

If I get a chance, will see if I can look tomorrow.

JakobLichterfeld:
> It kinds of sounds like something depends on something concerning your environment. Even though that isn't suppose to happen.

Yeah, direnv does something spooky with $PATH I assume. The `+__impureHostDep` for example.

```bash
direnv: export +AR +AS +CC +CONFIG_SHELL +CXX +DATABASE_HOST +DATABASE_NAME +DATABASE_PASS +DATABASE_PORT +DATABASE_USER +DEVENV_DOTFILE +DEVENV_PROFILE +DEVENV_ROOT +DEVENV_RUNTIME +DEVENV_STATE +ENCRYPTION_KEY +IN_NIX_SHELL +LD +LD_DYLD_PATH +LOCALES +MACOSX_DEPLOYMENT_TARGET +MQTT_HOST +MQTT_PORT +NIX_BINTOOLS +NIX_BINTOOLS_WRAPPER_TARGET_HOST_aarch64_apple_darwin +NIX_BUILD_TOP +NIX_CC +NIX_CC_WRAPPER_TARGET_HOST_aarch64_apple_darwin +NIX_CFLAGS_COMPILE +NIX_DONT_SET_RPATH +NIX_DONT_SET_RPATH_FOR_BUILD +NIX_DONT_SET_RPATH_FOR_TARGET +NIX_ENFORCE_NO_NATIVE +NIX_HARDENING_ENABLE +NIX_IGNORE_LD_THROUGH_GCC +NIX_LDFLAGS +NIX_NO_SELF_RPATH +NIX_PKG_CONFIG_WRAPPER_TARGET_HOST_aarch64_apple_darwin +NIX_STORE +NM +NODE_PATH +OBJCOPY +OBJDUMP +PATH_LOCALE +PGDATA +PGHOST +PGPORT +PKG_CONFIG +PKG_CONFIG_PATH +PORT +RANLIB +RELEASE_COOKIE +SIZE +SOURCE_DATE_EPOCH +STRINGS +STRIP +TEMP +TEMPDIR +TMP +TZDATA_DIR +ZERO_AR_DATE +__darwinAllowLocalNetworking +__impureHostDeps +__propagatedImpureHostDeps +__propagatedSandboxProfile +__sandboxProfile +cmakeFlags +configureFlags +mesonFlags +name +system ~PATH ~TMPDIR ~XDG_DATA_DIRS
```



> If I get a chance, will see if I can look tomorrow.

That would be awesome, as I run out of energy on it.

Witout `--impure` I do not see a reason to use a container for local test but can be wrong.

brianmay:
I can't say I entirely understand... Regardless the following probably is a good change to make.

It fails for me also, with the same error. If I make the following change, it works:

```diff
@@ -185,8 +183,7 @@
                   export MQTT_PORT="${toString mosquitto_port}"
                   export RELEASE_COOKIE="1234567890123456789"
                   export TZDATA_DIR="$PWD/tzdata"
-                  mix local.rebar --force
-                  mix local.hex --force
+                  export MIX_REBAR3="${rebar3}/bin/rebar3";
                   mix deps.get
                   echo "Debug"
                   locale
```

Just to make sure I did a `rm -rf /home/brian/.mix/` and retested.

brianmay:
Curiously, my shell prompt - I use starship - still claims it is an impure shell.

```
☸ <aws context> in teslamate on  treefmt [$!?] via 💧 v1.16.3 (OTP 26) via  v20.17.0 via ❄️  impure (devenv-shell-env) took 3s
at 11:16:05 ❯
```

There probably is a good reason for this, I see that on all my projects even ones without devenv, but not sure I understand.

Also note that `nix develop` has a `—ignore-environment` option that looks independant of the `--impure` flag.

brianmay:
See this discussion here also: https://github.com/starship/starship/issues/6188

brianmay:
Noe that use of `—ignore-environment` unsets the locale, and my terminal stops working probably, which is not surprising. Because we didn't set values for `LANG`, `TERM`, etc.

So using `--ignore-environment` in general probably a bad idea, but good to know it is there for debugging stuff.

We probably do want to set `LANG="en_AU.UTF-8"` (or something like that) somewhere, maybe the CI. Because probably the CI doesn't come with a good default.

JakobLichterfeld:
> I can't say I entirely understand... Regardless the following probably is a good change to make.

Thanks! Now it seems obvious to "just" set the path correct, but didn't tried before.

Ci runs fine as well: https://github.com/teslamate-org/teslamate/actions/runs/11229295521/job/31214608563?pr=4219

> Curiously, my shell prompt - I use starship - still claims it is an impure shell.

Same here (even if I didn't noticed before)


> Noe that use of `—ignore-environment` unsets the locale, and my terminal stops working probably, which is not surprising. Because we didn't set values for `LANG`, `TERM`, etc.

Yeah, that's why I didn't used it.

> We probably do want to set `LANG="en_AU.UTF-8"` (or something like that) somewhere, maybe the CI. Because probably the CI doesn't come with a good default.

GitHub Runners do have proper `C.UTF-8`set, imo best for compiling and CI. In local devenv it is nice UX to have your own lang settings.

JakobLichterfeld:
CI ran successful on push but not on pull_request_target.

Now we ensure the cache name does not include invalid character

JakobLichterfeld:
I will ensure pull_request_target only runs on real forks to avoid invalid cache name and unnecessary duplicate runs.

brianmay:
I think the mix hash needs to be updated also:

```
error: hash mismatch in fixed-output derivation '/nix/store/gly3c0rmyx308dzyl4qssf0irh175wgh-teslamate-mix-deps-1.30.2-dev.drv':
         specified: sha256-ukE7PKrsRaw6TWDSqTJljsPJQedi39mZgGFrMINEJlw=
            got:    sha256-Y+CGgvnSCiiuyhtsQ+j0vayq1IHO5IEPVl+V/wwTd6w=
error: 1 dependencies of derivation '/nix/store/1ladsmz0i5x6zzx15bqqpjv2fywa746g-teslamate-1.30.2-dev.drv' failed to build
```

brianmay (commented):
Looks mostly good to me, just needs the two issues resolved.

.github/workflows/ensure_linting.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Install Nix
+        uses: cachix/install-nix-action@ba0dd844c9180cbf77aa72a116d6fbc515d0e87b #v27
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Nix binary cache
+        uses: DeterminateSystems/magic-nix-cache-action@87b14cf437d03d37989d87f0fa5ce4f5dc1a330b #v8
+
+      - name: Run treefmt in CI mode
+        run: nix develop --override-input devenv-root "file+file://"<(printf %s "$PWD") . --command treefmt --ci
```
scottbot95, line 24: I see we're using `nix develop`  to run treefmt, we could probably just do `nix build .#checks.x86_64-linux.treefmt` instead which should be a more reproducible result since then `treefmt` will run inside the nix sandbox.

nix/flake-modules/formatter.nix
```diff
@@ -0,0 +1,66 @@
...
+{
+  imports = [
+    inputs.treefmt-nix.flakeModule
+  ];
+  perSystem =
+    { config, pkgs, ... }:
+    {
+      # Auto formatters. This also adds a flake check to ensure that the
+      # source tree was auto formatted.
+      treefmt.config = {
```
scottbot95, line 11: Is this actually the right setting? I was reading the [treefmt-nix flake module](https://github.com/numtide/treefmt-nix/blob/4446c7a6fc0775df028c5a3f6727945ba8400e64/flake-module.nix) a bit, and it seems like all the settings are just under `treefmt.*` (eg `treefmt.settings` instead of `treefmt.config.settings`).

scottbot95 (commented):
So I did a little digging because it was bothering me that we have to copy the treefmt config in two places and I discovered a potential workaround. `treefmt.settings` [is just used to generate treefmt.build.configFile](https://github.com/numtide/treefmt-nix/blob/4446c7a6fc0775df028c5a3f6727945ba8400e64/module-options.nix#L278) and `treefmt.build.configFile` is what is [actually used as the config file for the treefmt executable](https://github.com/numtide/treefmt-nix/blob/4446c7a6fc0775df028c5a3f6727945ba8400e64/module-options.nix#L190). So we *could* simply set `treefmt.build.configFile` ourselves with `lib.mkForce` and point it to the `treefmt.toml` in repo. That said, from their code it's unclear to me whether this is an intended extension point or merely an implementation detail so I'm not sure if we really want to rely on it. It's probably fine... What's the worst that could happen 😅

nix/flake-modules/formatter.nix
brianmay: "it must be right otherwise it would be showing an error".

I am somewhat curious now too. I don't see any definitions for `treefmt.config` but if I renamed it to `treefmt.xxx` it clearly showed an error. So `treefmt.config` must be defined somewhere.

brianmay: I see options documented here, no mention of `treefmt.config`: https://flake.parts/options/treefmt-nix

scottbot95: > it must be right otherwise it would be showing an error

Haha I was thinking this too. Well I guess if it's working then it's fine. It is odd though that `treefmt.config` doesn't seem to be defined anywhere... 

Oh maybe it's quirk of the fact that `treefmt` is defined a submodule? Like if you did `treefmt = { ... }: { /* whatever */ }` I'm guessing that might work as well.

JakobLichterfeld:
> That said, from their code it's unclear to me whether this is an intended extension point or merely an implementation detail so I'm not sure if we really want to rely on it. It's probably fine... What's the worst that could happen 😅

Yeah, I found that hole in the fence too, but was too excited when I finally got it working to do another refactoring :-)

nix/flake-modules/formatter.nix
JakobLichterfeld: > Is this actually the right setting?

Yeah, this bothered me a lot, until I found the right setting in the flake-parts dock

> I see options documented here, no mention of `treefmt.config`: https://flake.parts/options/treefmt-nix
Exactly this.

.github/workflows/ensure_linting.yml
JakobLichterfeld: Yes and no. `mix format` does need the dependencies fetched beforehand. Running `nix build .#checks.aarch64-darwin.treefmt` locally results therefore in:

```bash
❯ nix build .#checks.aarch64-darwin.treefmt
warning: Git tree '/Users/jakob/development/repos/teslamate' is dirty
error: builder for '/nix/store/3nj3634djh1s9acq79jd04n17ki0v2w2-treefmt-check.drv' failed with exit code 1;
       last 7 log lines:
       > Leeres Git-Repository in /private/tmp/nix-build-treefmt-check.drv-0/project/.git/ initialisiert
       > treefmt v2.0.5
       > mix-format error:
       > Mix requires the Hex package manager to fetch dependencies
       > Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :castore from TeslaMate.MixProject
       >
       > treefmt: error: formatting failure: formatter '/nix/store/irh1cr1a71355gl7z329qx5fp3kmqjbr-elixir-1.17.3/bin/mix' with options '[format]' failed to apply: exit status 1
       For full logs, run 'nix log /nix/store/3nj3634djh1s9acq79jd04n17ki0v2w2-treefmt-check.drv'.
```

That's why I make use of the devshell to first pull the dependencies.

scottbot95: Oh that's lame... There's probably a way to make it work correctly with the check, but I doubt it's particularly simple.... 

In that case, we should probably disable the check (by setting `treefmt.flakeCheck` to false). IMO having a check that is known to not work is worse than having no check at all.

nix/flake-modules/formatter.nix
brianmay: Not 100% sure I understand where we got to here.

I would recommend replacing `treefmt.config = {` with just `treefmt = {` - as per the documentation - unless there is a good reason not to (e.g. it is breaking something).

> until I found the right setting in the flake-parts dock

What is this?

flake.nix
brianmay: This probably should be fixed after this PR is merged.

.github/workflows/ensure_linting.yml
brianmay: I think I can imagine a possible solution here. But lets get this PR merged first.

We probably should set `treefmt.flakeCheck = false` for now though. Marking as unresolved so this doesn't get missed.

brianmay:
This error was a bit unexpected:

```console
$ nix flake check --override-input devenv-root "file+file://"<(printf %s "$PWD") .
warning: Git tree '/home/brian/tree/3rdparty/teslamate' is dirty
warning: not writing modified lock file of flake 'git+file:///home/brian/tree/3rdparty/teslamate':
• Updated input 'devenv-root':
    'file:///dev/null?narHash=sha256-d6xi4mKdjkX2JFicDIv5niSzpyI0m/Hnm8GGAIU04kY%3D'
  → 'file:///proc/self/fd/11?narHash=sha256-MR36Ywqs/zTxXcyqXW6MESFZ3X7OT91FF2wTpsZPoEg%3D'
error:
       … while checking flake output 'checks'

         at «none»:0: (source not available)

       … while checking the derivation 'checks.x86_64-linux.default'

         at «none»:0: (source not available)

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'nixosModules' missing

       at /nix/store/ngxcbaks38d1vpxflw88983n205bkfic-source/flake.nix:227:33:

          226|                     nixosModules.default = import ./module.nix { inherit self'; };
          227|                     imports = [ self'.nixosModules.default ];
             |                                 ^
          228|                     services.teslamate = {
```

flake.nix
```diff
@@ -2,204 +2,267 @@
...
+
+          moduleTest =
+            (nixpkgs.lib.nixos.runTest {
+              hostPkgs = pkgs;
+              defaults.documentation.enable = false;
+              imports = [
+                {
+                  name = "teslamate";
+                  nodes.server = {
+                    nixosModules.default = import ./module.nix { inherit self'; };
```
scottbot95, line 226: Looks like this line got moved to the wrong spot after the flake-parts change? This `nodes.server` section is a NixOS config so `nixosModules.default` does nothing here. (this actually doesn't even work but I'm guessing we don't see the error since nix lazy-evaluates things and we don't run `nix flake check` in CI and this test only runs on linux systems). Also `inherit self'` won't work since `module.nix` is expecting the normal flake `self`, not `self'` from `perSystem`.

`nixosModules.default` needs to be set as `flake.nixosModules.default` in the top-level attribute set passed to the `mkFlake` call. 

@brianmay this is causing the error you just mentioned were seeing.

scottbot95 (changes requested):

.github/workflows/ensure_linting.yml
JakobLichterfeld: Agree and  I have implemented it.

flake.nix
JakobLichterfeld: Thanks for reminding, stumbled over this as well and forgot about it during illness.

nix/flake-modules/formatter.nix
JakobLichterfeld: > I would recommend replacing `treefmt.config = {` with just `treefmt = {` - as per the documentation - unless there is a good reason not to (e.g. it is breaking something).

Not running in issues with just `treefmt = {}`. I tooked it from: https://github.com/srid/haskell-template/blob/master/nix/modules/formatter.nix

JakobLichterfeld:
Not yet sure what is happening with the CI. Elixir test fails, even when all testcases pass. I assume this is due to postgres17 update and was not found with CI before as the caching of the build was used in the postgres17 pr.

brianmay:
Tests "succeeded" now, after I clicked retry.

Actually if you look at the logs, there was one scary looking error that isn't generating a test failure:

```
** (FunctionClauseError) no function clause matching in DateTime.to_naive/1
    (elixir 1.16.2) lib/calendar/datetime.ex:925: DateTime.to_naive({:ok, ~U[2003-07-01 00:00:00Z], 0})
    (teslamate 1.30.2-dev) lib/teslamate/log.ex:185: TeslaMate.Log.get_positions_without_elevation/2
    (teslamate 1.30.2-dev) lib/teslamate/terrain.ex:77: TeslaMate.Terrain.handle_event/4
    (stdlib 5.2.3.2) gen_statem.erl:1397: :gen_statem.loop_state_callback/11
    (stdlib 5.2.3.2) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
```

relevant code is at `lib/teslamate/log.ex:185` is:

```elixir
    date_earliest =
      cond do
        min_id == 0 ->
          DateTime.add(DateTime.utc_now(), -10, :day)

        true ->
          DateTime.from_iso8601("2003-07-01T00:00:00Z")
      end

    naive_date_earliest = DateTime.to_naive(date_earliest)
```

Problem is `DateTime.from_iso8601` returns a tuple, and could in theory fail. Sigh. Which Elixir had strong type checking...

In actual practice, this probably function I think can't fail, the value is constant.

brianmay:
I think the above is not a regression in this pull request, so opened a new issue. #4260. Probably better to merge it separately so it doesn't get lost.

Think it is safe to merge this PR now...

brianmay (approved):
Looks good to me now. Minor stuff can be cleaned up as we find it after this is merged.

scottbot95 (approved):
Also looks good to me. I'll have a PR before to long to do some further clean-up on the nix code.

JakobLichterfeld:
> Tests "succeeded" now, after I clicked retry.

Yeah, GitHub Runners have been slow and I assume a timing error.

> Actually if you look at the logs, there was one scary looking error that isn't generating a test failure:

 Saw it and was confused as to why there was no failure.

> Wish Elixir had strong type checking...

That explains it.

> I think the above is not a regression in this pull request, so opened a new issue. #4260. Probably better to merge it separately so it doesn't get lost.

Ty!

PR #4227, opened 2024-09-27
fix: add nix module option to specify postgres package

brianmay:
Without this upgrading postgres - which must be done manually - is likely to be painful.

For postgresql upgrade instructions see
https://nixos.org/manual/nixos/stable/#module-services-postgres-upgrading


brianmay:
Also of interest may be this pull request, looks like it needs some work still (based on comments): https://github.com/NixOS/nixpkgs/pull/327966

JakobLichterfeld (approved):

PR #4231, opened 2024-09-28
PostgreSQL 17

swiffer:
PostgreSQL 17 is mentioning a change in how functions are using search_path. "Functions used by expression indexes and materialized views that need to reference non-default schemas must specify a search path during function creation."

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=2af07e2f7

OOTB earthdistance is not specifying a search path causing migration failures as earth_box and ll_to_earth have been used in indexes (they aren't any longer).

Therefore:

- change the migrations executed in the past to allow setup & testsuite to complete when using PostgreSQL 17

And as last indexes involing ll_to_earth / earth_box have been dropped in `20200502140646_drop_unused_indexes.exs` / `20191007105010_add_new_fkey_indexes.exs`:

- Add a new migration that drops and recreates the extensions so we ensure they are not manipulated any longer (`20191008191431_fix_ll_to_earth.exs`)
- rework Backup & Restore, as ll_to_earth is not used in indexes anymore the SQL used there is not needed

Updated Solution is taken from here (Adrian mentioned this issue back in 2019 as well):

- https://github.com/diogob/activerecord-postgres-earthdistance/issues/30#issuecomment-2122829036

Next to upgrading to PostgreSQL 17 I changed / updated some install docs as well to stay in line with currently supported / recommended versions of Grafana & Postgres.

Updated Restore instructions have been tested successfully on my instance.

PostgreSQL issue discussion: https://postgrespro.com/list/thread-id/2377019

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -129,7 +129,7 @@ jobs:
 
     services:
       db:
-        image: postgres:16
+        image: postgres:17
         ports: ["5432:5432"]
         env:
           POSTGRES_PASSWORD: postgres
--- a/priv/repo/migrations/20190925152807_create_geo_extensions.exs
+++ b/priv/repo/migrations/20190925152807_create_geo_extensions.exs
@@ -2,8 +2,15 @@ defmodule TeslaMate.Repo.Migrations.CreateGeoExtensions do
   use Ecto.Migration
 
   def change do
-    execute("CREATE EXTENSION IF NOT EXISTS cube", "DROP EXTENSION cube")
-    execute("CREATE EXTENSION IF NOT EXISTS earthdistance", "DROP EXTENSION earthdistance")
+    execute("CREATE EXTENSION IF NOT EXISTS cube WITH SCHEMA public", "DROP EXTENSION cube")
+
+    execute(
+      "CREATE EXTENSION IF NOT EXISTS earthdistance WITH SCHEMA public",
+      "DROP EXTENSION earthdistance"
+    )
+
+    execute("ALTER FUNCTION ll_to_earth SET search_path = public")
+    execute("ALTER FUNCTION earth_box SET search_path = public")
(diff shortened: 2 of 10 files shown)


JakobLichterfeld:
I have read about it as well. I do like the new incremental backups' functionality.

swiffer:
yes, running on pg17 locally without any problems but see tests failing in the pipeline now 😬

module.nix
```diff
@@ -187,7 +187,7 @@ in {
       (mkIf cfg.postgres.enable_server {
         services.postgresql = {
           enable = true;
-          package = pkgs.postgresql_16;
+          package = pkgs.postgresql_17;
```
brianmay, line 190: I think postgresql isn't available in nixos yet, not even unstable has it. https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=postgresql_17

Also see #4227

swiffer: Yes, was bumping first just to see if the pipeline runs through, which is not the case...

swiffer: changed back to PG 16 for now.

swiffer:
@JakobLichterfeld - ready for review

JakobLichterfeld:
Thanks!

We will handle postgres in nix installs in #4227 

> Next to upgrading to PostgreSQL 17 I changed / updated some install docs as well to stay in line with currently supported / recommended versions of Grafana & Postgres.

Not sure if PostgreSQL 17 is available on freebsd for examle but we can keep it this way.

> Updated Restore instructions have been tested successfully on my instance.

I do love the reduction of commands needed.

JakobLichterfeld (approved):
Thanks for your efforts!

swiffer:
As docs have been published already and users start installing with PostgreSQL 17 (https://tff-forum.de/t/teslamate-erfahrungs-und-datenthread/84179/2824?u=spongyhat) we should aim for a release soon. Guess you want to finish #4219 first?

PR #4250, opened 2024-10-09
docs: add recommended RAM size

JakobLichterfeld:
1GB of Ram is sufficient for installation. I personally have been running TeslaMate on a Raspberry Pi 3B+ with 1 GB Ram for 5 years without any problems. And yes, this does require some background knowledge, such as how to set up ZRAM.

Nevertheless, in this day and age it is advisable to have at least 2 GB Ram.

In version 1.30.1, we have increased the performance of a dashboard by a factor of 100. The next version will include further performance improvements and the dashboards run smoothly on a Raspberry Pi3 B with 1 GB Ram (there is only one value that takes a long time to load in a dashboard, but we will fix that too). In addition, we have improved the background load that occurs, e.g. to add missing elevation data etc. by a factor of 900.


brianmay:
I think how much ram you need is probably going to depend on if you running postgresql on the same host or not.

I imagine the most common setup will be to run it on the same host though.

swiffer:
While we are working to improve performance by ensuring proper index usage and query optimization databases definitely will grow over time. I therefore recommend using a vm / pi with 2GB RAM.

For me it doesn't mean we stop improving performance or declining issues if recommendation of 2GB is not matched. We still should aim for calculations & reports as efficient as possible.

swiffer (approved):

JakobLichterfeld:
Wow, GitHub Runners are really slow today.

PR #4252, opened 2024-10-09
perf: fix skipping streaming data

### LINKED ISSUE #4251: perf: cost per 100 km in charging stats is slow
The query for cost per 100 km in charging stats dashboard is slow on slow host hw.

### Expected Behavior

improved query

### Type of installation

Docker

### Version

v1.30.1

swiffer:
fixes #4251 - positions table grows over time by the amount of data collected by streaming data. to exclude streaming data we are making use of the a new index (part of the upcoming release). when testing please make sure you are having latest migrations of master branch applied.

### DIFF
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -532,7 +532,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


swiffer:
4x query improvement here - should be larger impact on slow hw. @JakobLichterfeld

swiffer:
@JakobLichterfeld - data required for calculation is available in drives

-> so: use drives instead of positions for another 60x runtime improvement (totalling to 240x on my instance) ;)

JakobLichterfeld (approved):

JakobLichterfeld:
Successfully tested. It is now nearly instant, more than 240x on Rpi 3B+ :-)

Thanks so much!

PR #4253, opened 2024-10-09
feat: add max speed & speed histogram to drive stats

js94x:
I have implemented long term statistics about maximum speed and a speed histogram (already known from "Drive Details" dashboard). 

Imho a good place for that kind of statistics is the "Drive Stats" dashboard.

For speed histogram I added a "hack" to hide standing times (speed == 0) at traffic lights or while parking -> see last WHERE condition. Otherwise the bar/spike at the left side of histogram would be even bigger.

```sql
SELECT 
    speed_section_$length_unit AS "Speed",
    SUM(seconds_elapsed) * 100 / MAX(duration) as "Elapsed",
    TO_CHAR((SUM(seconds_elapsed) || ' second')::interval, 'HH24:MI:SS') AS "Time"
FROM (
    SELECT
        ROUND(convert_km(p.speed::numeric, '$length_unit') / 10,0) * 10 AS speed_section_$length_unit,
        EXTRACT(EPOCH FROM (LEAD(p."date") OVER (ORDER BY p."date") - p."date")) AS seconds_elapsed,
        EXTRACT(EPOCH FROM (end_date - start_date)) AS duration
    FROM drives d
    INNER JOIN positions p ON p.drive_id = d.id
    WHERE d.car_id = $car_id AND $__timeFilter(d.start_date)
) AS drivedata
WHERE speed_section_$length_unit > 0
GROUP BY 1
ORDER BY 1
```

See screenshot for more details:

skitch_export

PR #4258, opened 2024-10-12
perf: improvements drive stats

### LINKED ISSUE #4193: km-driven in drive-stats only updates once a day

### What happened?

km-driven in drive-stats only updates once a day at 2am, is this as designed or is this a bug. 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
06/09/2024, 02:00:00 55.8 km
07/09/2024, 02:00:00 74.7 km
08/09/2024, 02:00:00 115 km
09/09/2024, 02:00:00 100.0 km
10/09/2024, 02:00:000 0km
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.30.1

### LINKED ISSUE #4199: perf: drive stat - km driven slow
the query for km driven in drive stats dashboard is slow on slow host hw.

### Expected Behavior

improved query

### Type of installation

Docker

### Version

v1.30.1

swiffer:
This brings several improvements to the Drive Stats Dashboard

- make use of Drives instead of Positions where possible (250ms vs 1ms here)
- reuse queries accross mutliple panels to reduce stress on database (3 queries no longer executed at all)

In addition some oddities have been fixed

- While Panels mentioned "Average" within the queries a "Median" has been calculated, change Panel titles
- Fix visualizations in first dashboard row by gapfilling data for days without drives
- Fix "Exclude Locations" filter (when empty locations containing spaces have been filtered out)

fixes #4199 
fixes #4193

### DIFF
--- a/grafana/dashboards/drive-stats.json
+++ b/grafana/dashboards/drive-stats.json
@@ -125,11 +125,12 @@
             "type": "grafana-postgresql-datasource",
             "uid": "TeslaMate"
           },
+          "editorMode": "code",
           "format": "time_series",
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT\n  date_trunc('day', start_date) as \"time\",\n  count(*)\nFROM drives\nWHERE $__timeFilter(start_date) AND car_id = $car_id\nGROUP BY 1\nORDER BY 1;",
+          "rawSql": "WITH since as (\n\tSELECT min(start_date) at time zone 'UTC' as date FROM drives\n\tWHERE car_id = $car_id\n\tGROUP BY car_id\n),\n\nactual AS (\n\tSELECT\n\t\tdate_trunc('day', start_date at time zone 'UTC') AS date,\n\t\tcount(*) AS distance\n\tFROM drives\n\tWHERE car_id = $car_id and $__timeFilter(start_date) and $__timeFilter(end_date) \n\tGROUP BY 1\n),\n\nbase_line AS (\n\tSELECT date from generate_series(date_trunc('day', (select date from since)), date_trunc('day', timestamp with time zone $__timeTo()), '1 day'::interval) date\n)\n\nSELECT\n  base_line.date as time,\n\tconvert_km(COALESCE(actual.distance, 0)::numeric, '$length_unit') as \"distance_$length_unit\"\nFROM base_line\nLEFT JOIN actual ON actual.date = base_line.date\nWHERE date_trunc('day', timestamp with time zone $__timeFrom()) <= base_line.date\norder by base_line.date\n",
           "refId": "A",
           "select": [
             [
@@ -141,6 +142,23 @@
               }
             ]
           ],
+          "sql": {
+            "columns": [
+              {
+                "parameters": [],
+                "type": "function"
+              }
+            ],
+            "groupBy": [
+              {
+                "property": {
+                  "type": "string"
+                },
+                "type": "groupBy"
+              }
+            ],
+            "limit": 50
+          },
           "table": "addresses",
           "timeColumn": "inserted_at",
           "timeColumnType": "timestamp",
@@ -244,7 +262,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "SELECT \n    DATE(start_date)  AS \"time\",\n    SUM(convert_km((end_km - start_km)::numeric, '$length_unit')) AS \"distance_$length_unit\"\nFROM drives \nWHERE car_id = $car_id\n\tAND $__timeFilter(start_date)\nGROUP BY DATE(start_date)\nORDER BY DATE(start_date);\n\n",
+          "rawSql": "WITH since as (\n\tSELECT min(start_date) at time zone 'UTC' as date FROM drives\n\tWHERE car_id = $car_id\n\tGROUP BY car_id\n),\n\nactual AS (\n\tSELECT\n\t\tdate_trunc('day', start_date at time zone 'UTC') AS date,\n\t\tsum(distance) AS distance\n\tFROM drives\n\tWHERE car_id = $car_id and $__timeFilter(start_date) and $__timeFilter(end_date) \n\tGROUP BY 1\n),\n\nbase_line AS (\n\tSELECT date from generate_series(date_trunc('day', (select date from since)), date_trunc('day', timestamp with time zone $__timeTo()), '1 day'::interval) date)\n\nSELECT\n  base_line.date as time,\n\tconvert_km(COALESCE(actual.distance, 0)::numeric, '$length_unit') as \"distance_$length_unit\"\nFROM base_line\nLEFT JOIN actual ON actual.date = base_line.date\nWHERE date_trunc('day', timestamp with time zone $__timeFrom()) <= base_line.date\norder by base_line.date\n",
           "refId": "A",
           "select": [
             [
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Thanks for your efforts!

On my Raspberry Pi 3B+ this dashboard is now unusable. Loading ~ 46 seconds. Median not showing up at all.

https://github.com/user-attachments/assets/c7c1a184-b8e9-4c50-8de9-8e92682d44fe

swiffer:
ok, that went totally in the wrong direction 🤔 will have a look...

swiffer:
@JakobLichterfeld - can you run an explain analyze on this query please

```sql
explain analyze WITH since as (
	SELECT date at time zone 'UTC' as date FROM positions
	WHERE car_id = 2
	ORDER BY date ASC
	LIMIT 1
),

actual AS (
	SELECT
		date_trunc('day', start_date at time zone 'UTC') AS date,
		count(*) AS distance
	FROM drives
	WHERE car_id = 2
	GROUP BY 1
),

base_line AS (
	SELECT date from generate_series(date_trunc('day', (select date from since)), NOW(), '1 day'::interval) date
)

SELECT
  base_line.date as time,
	convert_km(COALESCE(actual.distance, 0)::numeric, 'km') as "distance_km"
FROM base_line
LEFT JOIN actual ON actual.date = base_line.date
WHERE base_line.date BETWEEN '2023-10-13T08:20:26.94Z' AND '2024-10-13T08:20:26.94Z'
order by base_line.date;
```

swiffer:
never mind, found it!

swiffer:
ready for retest

JakobLichterfeld:
> @JakobLichterfeld - can you run an explain analyze on this query please

```bash
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=208989.57..208989.71 rows=56 width=40) (actual time=36763.214..36769.724 rows=366 loops=1)
   Sort Key: date.date
   Sort Method: quicksort  Memory: 31kB
   InitPlan 1 (returns $0)
     ->  Limit  (cost=208795.94..208795.94 rows=1 width=8) (actual time=35885.387..35885.393 rows=1 loops=1)
           ->  Sort  (cost=208795.94..224757.68 rows=6384697 width=8) (actual time=35874.792..35874.796 rows=1 loops=1)
                 Sort Key: ((positions.date AT TIME ZONE 'UTC'::text))
                 Sort Method: top-N heapsort  Memory: 17kB
                 ->  Seq Scan on positions  (cost=0.00..176872.46 rows=6384697 width=8) (actual time=0.284..31167.745 rows=6383679 loops=1)
                       Filter: (car_id = 1)
   ->  Hash Right Join  (cost=127.22..192.01 rows=56 width=40) (actual time=36760.720..36762.525 rows=366 loops=1)
         Hash Cond: ((date_trunc('day'::text, (drives.start_date AT TIME ZONE 'UTC'::text))) = date.date)
         ->  HashAggregate  (cost=112.15..145.59 rows=2229 width=16) (actual time=80.388..81.257 rows=700 loops=1)
               Group Key: date_trunc('day'::text, (drives.start_date AT TIME ZONE 'UTC'::text))
               Batches: 1  Memory Usage: 145kB
               ->  Seq Scan on drives  (cost=0.00..101.01 rows=2229 width=8) (actual time=28.246..61.703 rows=2107 loops=1)
                     Filter: (car_id = 1)
         ->  Hash  (cost=15.01..15.01 rows=5 width=8) (actual time=36680.199..36680.201 rows=366 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 16kB
               ->  Function Scan on generate_series date  (cost=0.01..15.01 rows=5 width=8) (actual time=36674.824..36675.131 rows=366 loops=1)
                     Filter: ((date >= '2023-10-13 08:20:26.94+00'::timestamp with time zone) AND (date <= '2024-10-13 08:20:26.94+00'::timestamp with time zone))
                     Rows Removed by Filter: 1471
 Planning Time: 24.053 ms
 JIT:
   Functions: 26
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 11.140 ms, Inlining 0.000 ms, Optimization 20.642 ms, Emission 553.109 ms, Total 584.891 ms
 Execution Time: 38848.518 ms
(28 rows)
```

JakobLichterfeld:
> ready for retest

you are too fast for my rpi to take up :-)

JakobLichterfeld:
> never mind, found it!

I don't think so. It still takes ages.

swiffer:
https://github.com/user-attachments/assets/a11bc06e-02f8-4eaa-ae0a-0cb5ca6e8b9b

swiffer:
you are using this one?

https://github.com/teslamate-org/teslamate/pull/4258/commits/f2d13ea55847ed3fc15a50a512c1114bff523a0c

JakobLichterfeld:
> I don't think so. It still takes ages. (45 seconds)

> you are using this one?
> 
> [f2d13ea](https://github.com/teslamate-org/teslamate/commit/f2d13ea55847ed3fc15a50a512c1114bff523a0c)

Nope. I was faster than my Rpi was able to pull the new image, so I thought I would use the latest one.

Now it runs at instant speed. But Median per day is still empty.

JakobLichterfeld:
Do you have a prefered merger order for #4253?

JakobLichterfeld:
> But Median per day is still empty.

Just to confirm: In v1.30.1 The average fields are populated correctly on my setup.

swiffer:
> > But Median per day is still empty.
> 
> Just to confirm: In v1.30.1 The average fields are populated correctly on my setup.

using grafana 11.2.2 locally, will check in grafana 11.0.1

> Do you have a preferred merger order for #4253?

i would prefer merging this first and add the additional panels afterwards.

swiffer:
50% Percentile Calculations aren't available in Grafana 11.0.1 - reverted two panels to using it's own sql query instead of calculating with dataset from another panel. should work now.

swiffer:
> Nope. I was faster than my Rpi was able to pull the new image, so I thought I would use the latest one.
> 
> Now it runs at instant speed (two refreshes in attached video). Even with last 10 years it is instant. But Median per day is still empty.

great to here - the trick is to avoid positions whenever possible / ensure proper index usage / exclude streaming data as described here https://github.com/teslamate-org/teslamate/pull/4253#issuecomment-2408563522 for now...

JakobLichterfeld:
> reverted two panels to using it's own sql query instead of calculating with dataset from another panel. should work now.

With your latest push it works, values seems reasonable for last 7 days but not for last 2.
2 Drives with 20.8 km is correct. The median is a bit off :-)
image

JakobLichterfeld:
For comparison with v1.30.1: It counts 4 drives which is in fact correct. There has been 2 ~800m drives as well.

image

swiffer:
> 22,3703 km driven in 4 drives means 5,592575 average innit? Ah it is average per day. Happy to test more, but not today :-) Happy weekend.

in v1.30.1 it currently is percentile_disc(0,5) per day.... i actually vote for percentile_cont(0,5) per day instead in case of 2 days timeframe...

swiffer:
Ok, ended up now basically beeing a rewrite of the dashboard... it could be that numbers are different from before but they should be better now, give it a try.

Some Notes:

- when selecting "Last 2 days" this is a relative timeframe and basically means: last 48 hours. the dashboard now has been adapted to show data from last 48 hours only.
- when selecting "Last 7 days" and you only had drives in day 1 and 2 of the week the estimated mileage most likely was way to high - this should have been corrected now.
- switched to percentile_cont (this should be noticeable on short timeframes only).

would be interesting to see v1.30.1 vs this PR in your instance again

JakobLichterfeld:
v1.30.1:
image

PR-4258:
image

swiffer:
depends - it looks like you are not driving enough 😆 

median means 50 percentile - so if there are more than 50% of days where you are not driving at all then 0 km is correct.

should we switch to average calculations for all 3 ? (like the titles said).

JakobLichterfeld:
> depends - it looks like you are not driving enough 😆

Yeah you are right, but don't miss the time on the road that much :-)

> median means 50 percentile - so if there are more than 50% of days where you are not driving at all then 0 km is correct.

> should we switch to average calculations for all 3 ? (like the titles said).

In this case I vote yes, as it is more intuitve, innit?

swiffer:
rebased and back to dashboard queries & average calculations. should be ready!

JakobLichterfeld (approved):
perfect, ty!

sdwalker:
Went from 1135 to 705 drives over the same 1 year timespan with average distance driven and average kWh used being wrong

Before:
image

After:
image

swiffer:
@sdwalker thanks for reporting, will have a look! Could you check the actual number of drives for that car and period (drives dashboard, validate drives table data, execute and see number of rows).
Do you have any incomplete drives?

sdwalker:
> @sdwalker thanks for reporting - see #4285 for a fix of drives count.

Drive count is fixed

sdwalker:
> the average values are different but should be "better" now as these are averages per day (incl. drives without days) instead of median (over days with drives).
> 
> so different output expected here.

Count me in the group that preferred the median values

PR #4267, opened 2024-10-15
feat: add max Speed column to drives table

js94x:
In addition to the max Power column, I would also find it helpful to see the max Speed of past drives. 
Screenshot 2024-10-14 at 23 46 24

### DIFF
--- a/grafana/dashboards/drives.json
+++ b/grafana/dashboards/drives.json
@@ -1078,6 +1078,60 @@
           },
           {
             "matcher": {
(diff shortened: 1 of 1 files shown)


swiffer:
approved, added this commit via cherry-pick to #4284 so you can merge both changes via #4284 @JakobLichterfeld

JakobLichterfeld:
Thanks for your suggestion! Merged via #4284

PR #4270, opened 2024-10-16
zstd compress for 10x size reduction of backups

swiffer:
recommend taking backups with zstd compression for 10x size reduction

https://www.cybertec-postgresql.com/en/lz4-zstd-pg_dump-compression-postgresql-16/

### DIFF
--- a/website/docs/maintenance/backup_restore.md
+++ b/website/docs/maintenance/backup_restore.md
@@ -8,10 +8,10 @@ If you are using `docker-compose`, you are using Docker Compose v1, which has be
 
 ## Backup
 
-Create backup file `teslamate.bck`:
+Create backup file `teslamate.bck.zst`:
 
 ```bash
-docker compose exec -T database pg_dump -U teslamate teslamate > ./teslamate.bck
+docker compose exec -T database pg_dump -U teslamate teslamate --format=custom --compress=zstd > ./teslamate.bck.zst
 ```
 
 :::note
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Thanks for the suggestion. As `zstd` is not standard on all operating systems, this is not operating system-independent. So a restore is possible as we use our docker but the backup file itself is not open-able on every system.

I strongly vote against any compression in backup creation. The user can compress backups afterward and store them compressed. I do a compressed upload every night for example. But this is total in user hand.

This also makes the command more error-prone, especially because of the spaces. An old backup cannot be restored with the command, etc.

swiffer:
ok, fine - would you consider a new page in "Guides" on how to backup (eg. daily) with a small backup script, compression options and crontab or is what we have enough?

website/docs/maintenance/backup_restore.md
```diff
@@ -49,7 +49,7 @@ CREATE EXTENSION earthdistance WITH SCHEMA public;
 .
 
 # Restore
-docker compose exec -T database psql -U teslamate -d teslamate < teslamate.bck
+docker compose exec -T database psql -U teslamate -d teslamate < teslamate.bck.zstd
```
JakobLichterfeld, line 52: This will result in

```bash
The input is a PostgreSQL custom-format dump.
Use the pg_restore command-line client to restore this dump to a database.
```

JakobLichterfeld (changes requested):

PR #4284, opened 2024-10-20
fix: for drives not showing if duration < 1 minute and feat: add max speed in drives dashboard

swiffer:
this is a fix for drives not showing in the drives dashboard if the `duration_min` is 0.

fallback to diff of `end_date - start_date` instead to calculate `avg_speed`.

this brings number of drives in line with drive stats dashboard.

in addition cherry pick changes from #4267


JakobLichterfeld (approved):

JakobLichterfeld (approved):
Thanks! And thanks for cherry-picking, makes life way easier. :-)

PR #4285, opened 2024-10-20
fix: number of drives per day in drive stats

swiffer:
idkw - number of drives was using convert_km, must have been a copy / paste issue in #4258

fixes https://github.com/teslamate-org/teslamate/pull/4258#issuecomment-2424194101


JakobLichterfeld (approved):
Perfect, ty!

PR #4287, opened 2024-10-20
perf: exclude streaming data in visited dashboard

swiffer:
exclude streaming data in visited dashboard as well. query has been found to be "slow" by checking pg_stat_statements.

as data is bucketed by minute and non streaming data is gathered every 15 seconds i see no reason not to do. greatly reduces query time.

i just want to mention one discussion around a similar change (index on odometer) back from april #3801 where @DrMichael voted against adding this index.

i still think it's right to add it based on all the improvements making it into this release.

@JakobLichterfeld - last PR from my end. voting to include these and cutting a release afterwards.

### DIFF
--- a/grafana/dashboards/visited.json
+++ b/grafana/dashboards/visited.json
@@ -213,7 +213,7 @@
           "format": "table",
           "hide": false,
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


DrMichael:
> i just want to mention one discussion around a similar change (index on odometer) back from april #3801 where @DrMichael voted against adding this index.

Ah, it wasn't me. I would never dare to judge about indexes... :-)

swiffer:
Yes, right. Sorry it was @Dulanic

JakobLichterfeld (approved):
Thank you very much! Tested successfully on slow HW.

swiffer:
i think it's good to merge - the discussion related to the index is not part of this PR. if @Dulanic has concerns over the index (that is already added) please open an issue for that.

PR #4301, opened 2024-10-23
perf: improvement in charge level

swiffer:
that one slipped through... @JakobLichterfeld, thanks as always for all the work!


JakobLichterfeld (approved):
Thanks for catching that! Always happy to contribute—appreciate your efforts as well!

PR #4304, opened 2024-10-26
perf: improvement and deprecated syntax removal

swiffer:
- from Grafana [docs](https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/#variable-syntax): `[[varname]]` Do not use. Deprecated old syntax, will be removed in a future release.
- in addition found 2 more places where we can make use of new index of positions table


JakobLichterfeld (approved):
awesome, ty!
lgtm.
I do have the feeling Charge Level is slower now, rest tested successfully.

PR #4305, opened 2024-10-27
perf: add ideal_battery_range_km as query condition

### LINKED ISSUE #4295: Drive Stats: Top 10 Destinations query issue with postgres <16

### What happened?

After importing the pulling the latest Grafana dashboards from GitHub, the Top 10 Destinations panel in the Drive stats is no longer loading data. 

The error is very explicit: 
`db query error: pq: subquery in FROM must have an alias`

If the query is run as is in pgAdmin, the same error is reached and can easily be fixed by adding an alias at the end of the subquery: 
eg: 
```
SELECT * FROM (
SELECT
	COALESCE(g.name, COALESCE(a.name, nullif(CONCAT_WS(' ', a.road, a.house_number), ''))) as name,
	count(*) AS visited
FROM drives t
INNER JOIN addresses a ON end_address_id = a.id
LEFT JOIN geofences g ON end_geofence_id = g.id
WHERE t.car_id = $car_id AND $__timeFilter(t.start_date) and $__timeFilter(t.end_date) 
GROUP BY 1
ORDER BY visited DESC) as foo
WHERE name NOT ILIKE ALL ${exclude_formatted_string:raw}
LIMIT 10;
```

### Expected Behavior

The query should run normally

### Steps To Reproduce

Go to Grafana, Drive Stats, Top 10 Destinations should fail to load.

### Relevant log output

```shell
`db query error: pq: subquery in FROM must have an alias`
```

### Screenshots

Screenshot 2024-10-22 at 09-48-39 Drive Stats - TeslaMate - Dashboards - Grafana

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.30.1

swiffer:
after updating the dashboards to v1.31.0 and resetting all gathered stats i've found 2 more queries that are currently rather slow and exec perf. improved by adding the filter for ideal_battery_range_km

in addition, fixes #4295 for postgres < 16

### DIFF
--- a/grafana/dashboards/drive-stats.json
+++ b/grafana/dashboards/drive-stats.json
@@ -1034,7 +1034,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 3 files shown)


JakobLichterfeld (approved):
Ty. lgtm

JakobLichterfeld (approved):

Issue #4307, opened 2024-10-27
beam.smp chewing CPU on v1.31.0 on arm


leewillis77:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I've upgraded to v1.31.0 and after the upgrade I noticed* that there's a beam.smp process chewing up CPU time:

```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 1041 10000     20   0  192432  64284   4732 S 144.9   3.4   6:13.18 beam.smp
```

The system is idle currently (car is asleep, and neither teslamate nor grafana are open on any devices). 

* It's possible this was happening before the latest upgrade - I can't say for sure.

### Expected Behavior

System is idle with no excessive CPU hogging. 

### Relevant log output

```shell
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 1041 10000     20   0  192432  64284   4732 S 144.9   3.4   6:13.18 beam.smp
```

### Type of installation

Docker

### Version

v1.31.0

swiffer:
Seems to be caused by rammitmq (quick google search) which is not utilized by teslamate. Could that be ?

leewillis77:
@swiffer There's nothing else running on the box other than teslamate, and the processes in question look like they're coming from docker:

```
systemd(1)-+-agetty(455)
           |-agetty(456)
           |-alsactl(331)
           |-avahi-daemon(344)---avahi-daemon(356)
           |-colord(1279)-+-{colord}(1280)
           |              `-{colord}(1282)
           |-containerd(454)-+-{containerd}(458)
           |                 |-{containerd}(459)
           |                 |-{containerd}(460)
           |                 |-{containerd}(461)
           |                 |-{containerd}(462)
           |                 |-{containerd}(480)
           |                 |-{containerd}(481)
           |                 |-{containerd}(482)
           |                 |-{containerd}(498)
           |                 |-{containerd}(812)
           |                 |-{containerd}(813)
           |                 `-{containerd}(814)
           |-containerd-shim(828)-+-tini(889)-+-beam.smp(1041)-+-erl_child_setup(1438)---inet_gethost(1486)---inet_gethost(1487)
           |                      |           |                |-{beam.smp}(1435)
           |                      |           |                |-{beam.smp}(1436)
           |                      |           |                |-{beam.smp}(1437)
[...]

PR #4309, opened 2024-10-27
docs: fix ghcr image path in contributing guide

swiffer:
this fixes image path causing image cannot be loaded:

https://github.com/teslamate-org/teslamate/issues/4307#issuecomment-2440140607


JakobLichterfeld (approved):
Nice catch, ty.

PR #4311, opened 2024-10-28
build: add wx and observer as extra applications

JakobLichterfeld:
Beam is a VM. Probably will have problems debugging this with system tools, think you will need beam tools.

One such tool is the erlang observer. https://blog.sequinstream.com/how-we-used-elixirs-observer-to-hunt-down-bottlenecks/

Having said that, it may be fiddly to get this working in a Docker container. Either the observer needs to have X access somehow (e.g. https://github.com/c0b/docker-elixir/issues/19#issuecomment-257637757), or you need to run a local beam that connects to the remote beam somehow (https://stackoverflow.com/questions/42164298/using-the-erlang-observer-app-with-a-remote-elixir-phoenix-server-inside-docker). Either way likely to be fiddly.

Also: It seems that our nix flake doesn't compile in the observer. I tried some things, but so far have been failures. Later: Oh, wait, this fixes it:

```diff
diff --git a/mix.exs b/mix.exs
index d159eefd..afa1c98a 100644
--- a/mix.exs
+++ b/mix.exs
@@ -27,7 +27,7 @@ defmodule TeslaMate.MixProject do
   def application do
     [
       mod: {TeslaMate.Application, []},
-      extra_applications: [:logger, :runtime_tools]
+      extra_applications: [:logger, :runtime_tools, :wx, :observer]
     ]
   end
```

(solution copied from https://stackoverflow.com/a/76760603)

_Originally posted by @brianmay in https://github.com/teslamate-org/teslamate/issues/4307#issuecomment-2440184233_

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -27,7 +27,7 @@ defmodule TeslaMate.MixProject do
   def application do
     [
       mod: {TeslaMate.Application, []},
-      extra_applications: [:logger, :runtime_tools]
+      extra_applications: [:logger, :runtime_tools, :wx, :observer]
     ]
   end
 


brianmay:
According to the report here: https://github.com/teslamate-org/teslamate/issues/4307#issuecomment-2440874346:

It is looking for libglu1-mesa which we haven't provided in the docker container.

How very odd. But probably due to the X UI stuff required by observer. I thought this wouldn't be required though unless you actually started the observer.

Need to test this, I imagine other libraries might be required also. Maybe we only want to add this if there is some debugging flag set, if it becomes too much overhead.

JakobLichterfeld:
#4307 solved via #4319

If we wanna go debug route in the future this ca be taken as a first step with above mentioned additions.

PR #4310, opened 2024-10-28
fix: re-add missing changes from pr 4124

swiffer:
this fixes #4306, in addition in cleanes up the overwrites (lot of them not actually needed)

PR #4313, opened 2024-10-29
fix: correct unit of date column in drives dashboard

swiffer:
fixes https://github.com/teslamate-org/teslamate/pull/4310#issuecomment-2442940153

no mention in changelog needed as regression introduced after latest release


JakobLichterfeld (approved):
Thanks! lgtm

PR #4299, opened 2024-10-22
feat: Update Grafana to 11.4.0

swiffer:
Upgrades Grafana to 11.4.0. Grafana switched to scenes powered dashboards for which the experimental setting date_formats.use_browser_locale is breaking timerange selector & urls containing absolute time ranges.

Therefore date_formats.use_browser_locale has been temporarily disabled (see comment in Dockerfile).

Tracking issue: https://github.com/grafana/grafana/issues/95209

PR #4317, opened 2024-10-29
fix: ensure dutch-tax dashboard is not repeating multiple times per car (row and table)

swiffer:
this brings fixes from grafana 11.3.0 exploration to current bundled grafana (broken and outdated here as well)

-> https://github.com/teslamate-org/teslamate/pull/4299#issuecomment-2430131657


JakobLichterfeld (approved):

JakobLichterfeld (approved):
nice catch. Ty.

PR #4318, opened 2024-10-29
fix: remove convert_km from kwh calculations in timeline dashboard

swiffer:
follow up of #3847

removes convert_km from kwh calc in Driving, Parking and Missing.
efficiency is stored in kwh/km, therefore it's not needed to convert units here.


JakobLichterfeld (approved):
Ty! lgtm

Issue #4297, opened 2024-10-22
ci: nix caches cause the 10GB cache space to be used up quickly, slowing down Github actions


JakobLichterfeld:
![image](https://github.com/user-attachments/assets/c4ff44b5-8b31-40ad-a842-7e394729e93d)

Introducing the cleanup caches job on merge helped, but was not enough: https://github.com/teslamate-org/teslamate/commit/0dd5ef1076d810e838d6695cba057804475425fa

JakobLichterfeld:
I do run a `gh cache delete --all`right now once again as a workaround

JakobLichterfeld:
as of October 21st:
> This extension should be removed and replaced with usage of gh cache directly (see: https://github.com/actions/gh-actions-cache)

- ci(fix): update cleanup_caches.yml to use new cache management commands and fix permissions ([d6793ce](https://github.com/teslamate-org/teslamate/commit/d6793ce5717687b9e984067bf4c208415e15fdac), [b0b694f](https://github.com/teslamate-org/teslamate/commit/b0b694fc8c3c45036aafda45200f3b0d068a2f50), [16bb503](https://github.com/teslamate-org/teslamate/commit/16bb5032c7d81cb86e76cc19662e3332456291a0) - @JakobLichterfeld)
- ci: Add workflow to manually cleanup largest 100 caches ([dad7e3d](https://github.com/teslamate-org/teslamate/commit/dad7e3dea0ae1d799398bf1b31a0d598eff784bf), [523419d](https://github.com/teslamate-org/teslamate/commit/523419d35a610c7b06bbf7e9c2edd105e7d089aa) - @JakobLichterfeld)

JakobLichterfeld:
We switched to cache-nix-action in #4626

PR #4319, opened 2024-10-29
fix: use elixir-1.17.3-otp-26 to avoid beam.smp chewing CPU on arm

JakobLichterfeld:
fixes: #4307


swiffer (approved):

swiffer:
Is it stuck?

JakobLichterfeld:
> Is it stuck?

The cache space cot eaten up, causing to slow down all GitHub Actions: #4297

I deleted the biggest manually, run locally `gh cache delete --all` which will fail after too many requests to API and corrected the CI jobs to automatically delete the caches. Need to delete some more before restarting the jobs

PR #4323, opened 2024-10-29
perf: improve performance in drive stats after #4284

### LINKED ISSUE #4321: perf: max speed and speed histogram slow in drive stats
The max speed and speed histogram introduced by #4253 is slow on slow hw

### Type of installation

Docker

### Version

v1.31.1

swiffer:
fixes #4321

@JakobLichterfeld - max speed should be blazing fast but i guess speed histogram is still not performing best, right?

### DIFF
--- a/grafana/dashboards/drive-stats.json
+++ b/grafana/dashboards/drive-stats.json
@@ -3,6 +3,12 @@
   "__requires": [
     {
       "type": "panel",
+      "id": "barchart",
+      "name": "Bar chart",
+      "version": ""
+    },
+    {
+      "type": "panel",
       "id": "bargauge",
       "name": "Bar gauge",
       "version": ""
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Thanks, will test vy tomorrow

JakobLichterfeld:
> max speed should be blazing fast but i guess speed histogram is still not performing best, right?

Tested and yes, max speed is instant, histogram takes ~5 seconds. Speed histogram in drive details is instant

swiffer:
Yes, let me see later today if we can improve further, within drive details data size is way smaller so longer runtime is expected...

JakobLichterfeld (approved):

swiffer:
closed as included in #4299

Issue #4327, opened 2024-10-30
Home Assistant MQTT sensors unavailable due to healthy but data is available


longzheng:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

I've noticed that my Home Assistant MQTT sensors sometimes report "unavailable" even though data is actually available and logged in Teslamate.

For example there was no "speed" data between 6:51am and 6:56am.

<img width="1089" alt="image" src="https://github.com/user-attachments/assets/a159f56b-edaf-4ca5-b4b6-d51411d94ae1">

But looking at Teslamate's trip, speed was logged for the whole strip including the portion before 6:56am. I suspect this data came from the "Streaming API" which I have enabled.

<img width="1897" alt="image" src="https://github.com/user-attachments/assets/adbdfbf3-e346-4c8c-b39e-7869f82ae810">

Looking at some of the other sensors, I then realised that the sensor state was "unavailable" at the same time

![image](https://github.com/user-attachments/assets/34c59e5e-145c-46d7-84e1-817f3a829176)

Looking at the [Home Assistant MQTT sensors documentation](https://docs.teslamate.org/docs/integrations/home_assistant#mqtt_sensorsyaml-mqtt-section-of-configurationyaml) I then realised that all of the sensors have an `availability` tied to the `teslamate/cars/X/healthy` topic

```yaml
- sensor:
    name: Speed
    object_id: tesla_speed
    unique_id: teslamate_1_speed
    availability: *teslamate_availability
    device: *teslamate_device_info
    state_topic: "teslamate/cars/1/speed"
    unit_of_measurement: "km/h"
    icon: mdi:speedometer
```

```yaml
- sensor:
    name: Display Name
    object_id: tesla_display_name # entity_id
    unique_id: teslamate_1_display_name # internal id, used for device grouping
    availability: &teslamate_availability
      - topic: teslamate/cars/1/healthy
        payload_available: "true"
        payload_not_available: "false"
    device: &teslamate_device_info
      identifiers: [teslamate_car_1]
      configuration_url: https://teslamate.zxxz.io/
      manufacturer: Tesla
      model: Model 3
      name: Tesla Model 3
    state_topic: "teslamate/cars/1/display_name"
    icon: mdi:car
```

From the container logs I can verify that my vehicle became "unhealthy" around the same time due to consequtive `408` errors from the Owners API. 

```
...
[...]

PR #4362, opened 2024-11-07
Remove availability from Home Assistant MQTT sensors

longzheng:
Fixes https://github.com/teslamate-org/teslamate/issues/4327

Remove the `availability` from most Home Assistant MQTT sensors because it can be misleading and prevent sensors from receiving updated values (from streaming API) when the vehicle is "Unhealthy".

Refer to the `binary_sensor.tesla_healthy` sensor to actually check if the vehicle is healthy.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -64,10 +64,6 @@ Don't forget to replace `<teslamate url>`, `<your tesla model>` and `<your tesla
     name: Display Name
     object_id: tesla_display_name # entity_id
     unique_id: teslamate_1_display_name # internal id, used for device grouping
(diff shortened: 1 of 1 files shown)


longzheng:
@JakobLichterfeld one of the jobs failed due to rate limits

```
buildx failed with: ERROR: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

JakobLichterfeld:
> @JakobLichterfeld one of the jobs failed due to rate limits

Yeah sorry for that. For this PR it does not matter as it updates only the docs

PR #4369, opened 2024-11-10
Update vehicle.ex for Model S LR+

### LINKED ISSUE #4368: Model S long range miss labeled as a 100D

### What happened?

Very minor issue, the teslamate dashboard labels by Model S long range plus as a 100D. I know they are basically the same but it just doesn't line up with the Tesla naming from the car info and app.

 
image

image

I'm guessing this might also be the same with other raven Model Ss (long Range, Long Range Plus, Performance) 

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
NA
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

1.31.1

cwanja:
- Update marketing name to recognize Model S LR+

Fixes #4368

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -65,6 +65,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
           with str when is_binary(str) <- type do
             case String.downcase(str) do
               "models" <> _ -> "S"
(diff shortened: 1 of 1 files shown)


cwanja:
Derp. Reversing and publishing from forked repo.

PR #4372, opened 2024-11-10
ci: create PR to udpate flake.lock every saturday

brianmay:
If this works it should create PRs that look a bit like https://github.com/brianmay/robotica-rust/pull/701

Caveats:

* Might have problems because I didn't create a personal token. Hoping this won't be required, but for personal project found it was. Particularly with dependabot. See https://github.com/DeterminateSystems/update-flake-lock?tab=readme-ov-file#with-a-personal-authentication-token
* Might have implications for running out of cache space, see #4297.
* Not sure what timezone this cron job uses.

### DIFF
--- a/.github/workflows/update-flack-lock.yml
+++ b/.github/workflows/update-flack-lock.yml
@@ -1,0 +1,25 @@
+---
+name: update-flake-lock
+on:
+  workflow_dispatch:
+  schedule:
+    # runs weekly on Saturday at 00:00 UTC time
+    - cron: "0 0 * * 6"
+
+jobs:
+  lockfile:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Install Nix
+        uses: cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72 #v30
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Update flake.lock
+        id: update
+        uses: DeterminateSystems/update-flake-lock@a2bbe0274e3a0c4194390a1e445f734c597ebc37 #v24
+        with:
+          pr-title: "Update flake.lock"
+      - name: Print PR number
+        run: echo Pull request number is ${{ steps.update.outputs.pull-request-number }}.


brianmay:
There is a request for native nix flake support in dependabot, but don't hold your breath waiting for it to happen :-)

https://github.com/dependabot/dependabot-core/issues/7340

.github/workflows/update-flack-lock.yml
```diff
@@ -0,0 +1,23 @@
...
+  schedule:
+    # runs weekly on Saturday at 00:00
+    - cron: "0 0 * * 6"
+
+jobs:
+  lockfile:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
JakobLichterfeld, line 14: Please use dep pinning: `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2`

.github/workflows/update-flack-lock.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@main
```
JakobLichterfeld, line 16: What is the advantage of this nix installer? We use the cachix nix installer in our other nix workflow: https://github.com/teslamate-org/teslamate/blob/93cc34437c6466c993fde98bb271faaa9ed8a58d/.github/workflows/ensure_linting.yml#L16

.github/workflows/update-flack-lock.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Update flake.lock
+        id: update
+        uses: DeterminateSystems/update-flake-lock@main
```
JakobLichterfeld, line 19: Please use dep pinning, `DeterminateSystems/update-flake-lock@v24`

.github/workflows/update-flack-lock.yml
```diff
@@ -0,0 +1,23 @@
+---
+name: update-flake-lock
+on:
+  workflow_dispatch:
+  schedule:
+    # runs weekly on Saturday at 00:00
+    - cron: "0 0 * * 6"
```
JakobLichterfeld, line 7: This is UTC

.github/workflows/update-flack-lock.yml
```diff
@@ -0,0 +1,23 @@
+---
+name: update-flake-lock
+on:
+  workflow_dispatch:
+  schedule:
+    # runs weekly on Saturday at 00:00
```
JakobLichterfeld, line 6: This comment does not reflect the cron job timer 😃

JakobLichterfeld (changes requested):
Thanks a lot! 
I do like the approach of having a Dependabot style for nix flake dependencies.

.github/workflows/update-flack-lock.yml
brianmay: Good question. I hadn't noticed that there are multiple competing solutions.

Perhaps I should change it to the one we already use.

I see a comparison at https://discourse.nixos.org/t/which-github-nix-installer-action-is-faster/25878

brianmay: Supposedly https://github.com/nixbuild/nix-quick-install-action is fastest, it is single user only. But I think we would need to benchmark this before switching.

brianmay:
Thanks for the feedback. I agree with all points raised.

brianmay:
Hopefully I addressed all the feedback.

I left the `DeterminateSystems/nix-installer-action@main` open though so you can read what I said :-)

Obviously this has not been tested...

.github/workflows/update-flack-lock.yml
JakobLichterfeld: Yeah, I found the above-mentioned benchmark after deciding for the cachix installer, which still claims to be fastest in most cases.

JakobLichterfeld (approved):
lgtm

JakobLichterfeld:
As feared, there are access problems: https://github.com/teslamate-org/teslamate/actions/runs/11813145746/job/32909622410#step:4:913

JakobLichterfeld:
Ok, we getting there:

https://github.com/teslamate-org/teslamate/actions/runs/11813338621/job/32910192341#step:4:921

JakobLichterfeld:
I changed our org settings: 
image

brianmay:
Great!

PR #4405, opened 2024-11-26
Adding backup for manual install on FreeBSD

ivanovd:


ivanovd:
Will add the restore part later, although it is even easier

JakobLichterfeld:
Thank you for your PR. CI fails here, doc preview cannot be built.

Additionally: As backup and restore is the most linked doc, I do not see adding steps for unsupported installation methods. People who choose manual installation should be able to transfer the steps themselves.

My suggestion: move to the manual installation instructions.

PR #4419, opened 2024-12-02
fix(nix): wait for mosquitto to start before starting teslamate

brianmay:
After reboot, I noticed a number of errors from teslamate, because it started before mosquitto started.


JakobLichterfeld (approved):

JakobLichterfeld:
ty!

PR #4430, opened 2024-12-08
doc: align node req with what is used in CI

swiffer:
bump package version in freebsd / debian install guide so that they align with what is used in CI


JakobLichterfeld (approved):
lgtm

Issue #4445, opened 2024-12-18
Status: 500. Message: db query error: pq: invalid input syntax for type json


dayal18:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Followed the exact steps mentioned here:

https://docs.teslamate.org/docs/installation/docker

Description:    Ubuntu 24.04.1 LTS
Release:        24.04
Codename:       noble

Docker version 27.3.1, build ce12230

  teslamate:
    image: teslamate/teslamate:latest
  database:
    image: postgres:17
  grafana:
    image: teslamate/grafana:latest
  mosquitto:
    image: eclipse-mosquitto:2

### Expected Behavior

The dashboards are supposed to show data

### Steps To Reproduce

1. Creating docker-compose.yml file as mentioned in this document (https://docs.teslamate.org/docs/installation/docker)
2. docker compose up
3. access the webUI using http://ipaddress:4000 (provide the access and refresh tokens)
4. access the Dashboard UI http://ipaddress:3000 (change admin password, and access battery dashboard)

### Relevant log output

```shell
database-1   | 2024-12-18 22:09:35.034 UTC [89] ERROR:  invalid input syntax for type json at character 41
database-1   | 2024-12-18 22:09:35.034 UTC [89] DETAIL:  The input string ended unexpectedly.
database-1   | 2024-12-18 22:09:35.034 UTC [89] CONTEXT:  JSON data, line 1:
database-1   | 2024-12-18 22:09:35.034 UTC [89] STATEMENT:  SELECT
database-1   |    CASE WHEN 0 > 0 THEN 0 ELSE (''::json ->> 'MaxRange')::float END as "maxrange_mi",
database-1   |    (''::json ->> 'CurrentRange')::float  as "currentrange_mi",
database-1   |    CASE WHEN 0 > 0 THEN 0 ELSE (''::json ->> 'MaxRange')::float END - (''::json ->> 'CurrentRange')::float as "range_lost_mi"
grafana-1    | logger=context userId=1 orgId=1 uname=admin t=2024-12-18T22:09:35.03529099Z level=info msg="Request Completed" method=POST path=/api/ds/query status=400 remote_addr=192.168.86.130 time_ms=6 duration=6.782616ms size=577 referer="http://192.168.86.115:3000/d/jchmRiqUfXgTM/battery-health?orgId=1" handler=/api/ds/query status_source=downstream
[...]

PR #4448, opened 2024-12-19
fix for battery health dashboard erroring out if no charge data has been collected so far

swiffer:
fix for #4445 - ensure the json is build correctly even if no charges have been collected so far to avoid panel errors in the battery health dasboard.


JakobLichterfeld (approved):
ty!

PR #4463, opened 2025-01-01
replace cross join with left join

swiffer:
cross join will still lead to empty result set if at least one of the tables is not returning a result


swiffer:
draft for now, will improve first cte of the aux query further tomorrow

swiffer:
lgtm

JakobLichterfeld (approved):
Nice, ty!

PR #3262, opened 2023-06-26
Handle sleep behavior of MCU2 upgraded cars

micves:
There is a lot more info in this issue: https://github.com/adriankumpf/teslamate/issues/3084

The short story is:
Older cars upgraded to MCU2 have a little wakeup every hour to check subsystems.
They report online but if vehicle_data is requested they wake up completely (cars clicks) and is awake for 15 minutes instead of a 2-3 minutes.
The only difference (known at this moment) is stream reports power=nil in subsystem online and reports power as a number when it is a real online.

Fix implemented by staying in state :start when online is detected on non-vehicle_data.
Stream receives data and new variable fake_online is set dependent on power is nil or a number.
The variable fake_online is used to decide whether to get only non-vehicle_data or the full blown wakeup vehicle_data
In suspended only non-vehicle_data is allowed to be better at going to sleep. Usage is detected if power>0 to go back to online state.

Other things:
- Handle stream getting vehicle offline and change to state :offline in vehicle
- Try to suspend in online when getting error on fetch_result
  This will lay of vehicle_data and use non-vehicle_data instead and
  hopefully not keep the car awake when errors suddenly stop again
- I'm not quite into nice coding guidelines for Elixir, so some code-review might be good :)
- I don't have possibility to test on other types of cars than my own 2017 Model S with upgraded MCU2.

PR #4453, opened 2024-12-23
feat: handle sleep behavior of MCU2 upgraded cars

JakobLichterfeld:
changes from @micves from #3262, fixes #3084

to use this just follow https://docs.teslamate.org/docs/development#testing-with-our-ci-which-builds-the-docker-images-automatically-per-pr with `pr-4453`

PR #4467, opened 2025-01-02
feat: improve logging messages

brianmay:
changes from @micves from #3262


brianmay:
I split the #4453 into two. Not sure if I can go any further, but these changes look like they should be relatively straight forward.

JakobLichterfeld (approved):
Ty, lgtm!

PR #4257, opened 2024-10-12
docs: fix Home Assistant MQTT sensor JSON templates warnings

longzheng:
Fixes #4247

More gracefully handles `teslamate/cars/1/active_route` topic JSON when the car is not routing, the `json_attributes_template` would cause a Home Assistant warning since we're trying to access properties on the JSON that aren't there.

MQTT topic `active_route`
```
{"error":"No active route available"}
```

Home Assistant logs
```
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.127 WARNING (MainThread) [homeassistant.components.mqtt.mixins] TypeError: Object of type LoggingUndefined is not JSON serializable rendering template for entity 'device_tracker.tesla_active_route_location', template: '{{ value_json.location | tojson }}' and payload: {"error":"No active route available"}
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.127 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'destination' when rendering '{{ value_json.destination }}'
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.127 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'energy_at_arrival' when rendering '{{ value_json.energy_at_arrival }}'
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.128 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'miles_to_arrival' when rendering '{{ value_json.miles_to_arrival }}'
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.128 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'minutes_to_arrival' when rendering '{{ value_json.minutes_to_arrival }}'
Oct 07 15:10:43 iot2 homeassistant[107792]: 2024-10-07 15:10:43.128 WARNING (MainThread) [homeassistant.helpers.template] Template variable warning: 'dict object' has no attribute 'traffic_minutes_delay' when rendering '{{ value_json.traffic_minutes_delay }}'
```

[...]

PR #4470, opened 2025-01-04
doc: Fixing typo for sensor psi calculation

Phazz:
The psi sensors were not referencing the correct sensor for their calculation.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -684,40 +684,40 @@ Don't forget to replace `<teslamate url>`, `<your tesla model>` and `<your tesla
       friendly_name: Front Left Tire Pressure (psi)
       unit_of_measurement: psi
       icon_template: mdi:car-tire-alert
-      availability_template: "{{ has_value('sensor.tesla_tpms_fl') }}"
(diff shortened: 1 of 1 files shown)


Phazz:
Thanks a lot for this beautiful project btw.

JakobLichterfeld:
@longzheng may I ask you for review, as it was latest edited in #4257 by you?

JakobLichterfeld (approved):

longzheng:
This change looks good.

However this just reminded me perhaps using a template sensor to offer sensors in different units is not the best approach here. Perhaps we can use MQTT sensors with a value template to do the transformations. The benefits would be:
- remove sensor reference by entity ID
- reduce `has_value` logic in value template
- have device association

I'll open a separate PR for that change.

JakobLichterfeld:
> This change looks good.

Perfect, ty.


> However this just reminded me perhaps using a template sensor to offer sensors in different units is not the best approach here. Perhaps we can use MQTT sensors with a value template to do the transformations. The benefits would be:
> - remove sensor reference by entity ID
> - reduce `has_value` logic in value template
> - have device association
> 
> I'll open a separate PR for that change.

That sounds really like the best way forward, thanks in advance!

PR #4472, opened 2025-01-05
doc: Simplify Home Assistant sensors, add device_class to allow changing measurement units

longzheng:
When reviewing https://github.com/teslamate-org/teslamate/pull/4470 I had a thought why are the Home Assistant sensors defined as template sensors rather than MQTT sensors with a transform in `value_template`. Then I went down a rabbit hole, why even have different sensors for different measurements (miles vs. km) at all if Home Assistant can automatically transform the unit of measurement

image

(It's my assumption that Home Assistant automatically picks the correct unit of measurement based on the server's locale/region, but please correct me if I'm wrong)

So I made the following changes
- moved parking brake sensor from `binary_sensor.yaml` to `mqtt_sensors.yaml` with a `value_template` transform
- removed several "mi" template sensors and added `device_class: distance` to the "main" sensor so the unit can be changed
- renamed some sensors like `tesla_est_battery_range_km` to `tesla_est_battery_range` so they are not unit specific (since the unit can be changed)
- kept the separate bar/psi tyre pressure sensors since that isn't so much a region-specific unit of measurement but a preference (Home Assistant still allows changing between bar/psi)
- removed separate `sensor.yaml` and `binary_sensor.yaml` files from the docs
- updated `ui-lovelace.yaml` to reference just the main sensors (and not both km/miles)

I think this approach reduces the complexity of the sensor definitions but I'm not sure if there's any downsides I've missed.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -51,8 +51,6 @@ tesla:
   scan_interval: 3600
 
 mqtt: !include mqtt_sensors.yaml
-sensor: !include sensor.yaml
-binary_sensor: !include binary_sensor.yaml
 ```
 
 ### mqtt_sensors.yaml (mqtt: section of configuration.yaml)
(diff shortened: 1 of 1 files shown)


brianmay:
I would really like to see us implement mqtt auto discovery, so this config can be part of teslamate and not have to be manually updated.

https://www.home-assistant.io/integrations/mqtt/#mqtt-discovery

But I have no idea how feasible this would be.

longzheng:
> I would really like to see us implement mqtt auto discovery, so this config can be part of teslamate and not have to be manually updated.
> 
> https://www.home-assistant.io/integrations/mqtt/#mqtt-discovery
> 
> But I have no idea how feasible this would be.

I agree auto discovery would be handy but I also don't have emough experience with Home Assistant ot Elixer to implement it. 

I think this PR will make it easier to transition to auto discovery as the duplicate measurement unit sensors are removed and it's all just MQTT sensors remaining.

JakobLichterfeld (approved):
Awesome, thanks you so much!

longzheng:
Do you want me to resolve the conflicts due to the other merge?

PR #4541, opened 2025-02-14
Update docs

jheredianet:
- Add script for maintenance: REINDEX
- Rearrange Projects using TeslaMate as DataSource, removing outdated projects or those that don't really use Teslamate as datasource and order by name
- Minor fixes in docs for Website

### DIFF
--- a/.github/workflows/ensure_linting.yml
+++ b/.github/workflows/ensure_linting.yml
@@ -18,8 +18,11 @@ jobs:
           github_access_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Nix binary cache
-        uses: DeterminateSystems/magic-nix-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9
-
+        # Magic Nix Cache has been deprecated due to a change in the underlying GitHub APIs and will stop working on 1 February 2025.
+        # To continue caching Nix builds in GitHub Actions, use FlakeHub Cache instead.
+        #uses: DeterminateSystems/magic-nix-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9  
+        uses: DeterminateSystems/flakehub-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9
+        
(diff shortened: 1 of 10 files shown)


.github/workflows/ensure_linting.yml
```diff
@@ -18,8 +18,11 @@ jobs:
           github_access_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Nix binary cache
-        uses: DeterminateSystems/magic-nix-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9
-
+        # Magic Nix Cache has been deprecated due to a change in the underlying GitHub APIs and will stop working on 1 February 2025.
+        # To continue caching Nix builds in GitHub Actions, use FlakeHub Cache instead.
+        #uses: DeterminateSystems/magic-nix-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9  
+        uses: DeterminateSystems/flakehub-cache-action@6221693898146dc97e38ad0e013488a16477a4c4 #v9
```
brianmay, line 24: I think this change really deserves a pull request all of its own. Something like that can get lost if combined with unrelated changes. And it looks somewhat important.

Where can I find this change documented? I tried looking https://github.com/DeterminateSystems/magic-nix-cache-action but I don't see anything.

Did you forget to update the hash? Seems kind of suspect that the hash is exactly the same.

brianmay (changes requested):
Please split out the nix cache action into separate pull request.

JakobLichterfeld (changes requested):
Thanks for your changes. As indicated, please split up. Commenting out unused github actions is not good practise imo. Changing the nix action is not needed atm.
Please use proper linting.

JakobLichterfeld:
As the PR is not only a doc change tzhis PR will be closed

.github/workflows/ensure_linting.yml
JakobLichterfeld: -> #4544

PR #4549, opened 2025-02-21
Update doc files

jheredianet:
- Add script for maintenance: REINDEX
- Rearrange Projects using TeslaMate as DataSource, removing outdated projects or those that don't really use Teslamate as datasource and order by name
- Minor fixes in docs for Website

### DIFF
--- a/website/docs/maintenance/manually_fixing_data.mdx
+++ b/website/docs/maintenance/manually_fixing_data.mdx
@@ -69,7 +69,7 @@ docker compose exec teslamate bin/teslamate rpc \
 
 If for some reason a drive or charge was recorded incorrectly, you can delete it.
 
-1.  Attach to the **running** database container:
+1. Attach to the **running** database container:
 
     ```bash
     docker compose exec database psql teslamate teslamate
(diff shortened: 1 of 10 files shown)


JakobLichterfeld:
Thanks for your suggestions. Please ensure proper linting before marking as ready for review. https://docs.teslamate.org/docs/development#code-formatting

website/docs/maintenance/manually_fixing_data.mdx
```diff
@@ -124,19 +124,53 @@ If for some reason a drive or charge was recorded incorrectly, you can delete it
...
+    DELETE from charges WHERE charging_process_id in (select id from charging_processes where car_id    = num);
+    DELETE FROM drives WHERE car_id = num;
+    DELETE FROM positions WHERE car_id = num;
+    DELETE FROM states WHERE car_id = num;
+    DELETE FROM updates WHERE car_id = num;
+    ```
+
+## Reindex Database
+
+**NOTE:** Maintaining the efficiency of database operations is imperative for any application relying on relational databases. One of the tools provided by PostgreSQL to help maintain performance is the REINDEX command. It is recommended that the database be reindexed periodically to keep it up to date, at least once a month or more frequently if you use the car intensively.
```
JakobLichterfeld, line 146: Thats technically not true imo. Only if you mess with update and delete operations. So normal use does not need any reindex

JakobLichterfeld (changes requested):
Not correctly linted

reindex not needed

website/docs/maintenance/manually_fixing_data.mdx
jheredianet: You're right. I will change the recommendation and comment when it is needed.

PR #4556, opened 2025-02-25
Update Document Files

jheredianet:
- Add script for maintenance when importing or removing massive data: REINDEX
- Rearrange Projects using TeslaMate as DataSource, removing outdated projects or those that don't really use Teslamate as datasource and order by name
- Minor fixes in docs for Website

PR #4558, opened 2025-02-27
docs: add reindexing instructions for database maintenance to improve performance in case of index bloat due to frequent updates or deletions

JakobLichterfeld:
successor of #4556

- docs: add reindexing instructions for database maintenance to improve performance in case of index bloat due to frequent updates or deletions (by @jheredianet)
- docs: fix links to use markdown format to fix netlify build


brianmay (approved):

website/docs/maintenance/manually_fixing_data.mdx
```diff
@@ -140,3 +140,39 @@ DELETE FROM positions WHERE car_id = num;
...
+   ```sql
+   DO $$
+   DECLARE
+       t text;
+   BEGIN
+       FOR t IN (SELECT table_name FROM information_schema.tables WHERE table_schema = 'public' AND    table_name IS NOT NULL)
+       LOOP
+           EXECUTE 'REINDEX TABLE public.' || t;
+       END LOOP;
+   END $$;
```
swiffer, line 171: can't we simply use `REINDEX DATABASE teslamate` ?

jheredianet: Yes, sure. But I'm not sure if REINDEX DATABASE is available in all PostgreSQL versions.

PR #4572, opened 2025-03-02
Fix Battery Health difference estimation

jheredianet:
This is a minor fix on estimation.

### DIFF
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -107,7 +107,7 @@
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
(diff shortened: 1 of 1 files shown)


swiffer:
I don't really understand the change - isn't it always 0 now with the updated query?

Looks correct for me (latest master) with a custom kwh when new entered:

grafik

swiffer (changes requested):
I don't really understand the change - isn't it always 0 now with the updated query?

Looks correct for me (latest master) with a custom kwh when new entered:

grafik

jheredianet:
> I don't really understand the change - isn't it always 0 now with the updated query?
> 
> Looks correct for me (latest master) with a custom kwh when new entered:

Yes, apparently the calculation formula was correct. A few days ago, when I received my new car, I realized that it gave me these values ​​(without having to put in any custom values). That's when I realized that the calculation was wrong and that it should be the difference between Usable (new) - Usable (now). Before the fix I was getting this:

image

PR #4573, opened 2025-03-02
docs: Update projects using TeslaMate

jheredianet:
Uppdate projects using TeslaMate as DataSource, removing outdated projects or those that don't really use Teslamate as datasource and sorted by name.


JakobLichterfeld (approved):
ty!

PR #4574, opened 2025-03-02
docs: Update REINDEX notes

jheredianet:
This is an update suggested in https://github.com/teslamate-org/teslamate/pull/4558


swiffer:
7.4 was released 2003, hope no one is using that...

website/docs/maintenance/manually_fixing_data.mdx
```diff
@@ -159,6 +159,18 @@ In summary, you don't necessarily need to run REINDEX periodically unless your d
...
+   REINDEX DATABASE teslamate
+   ```
+
+3. Exit the prompt by typing:
+
+   ```bash
+   \q  (or press CTRL + C)
+   ```
+
+**NOTE:** If you are using previous versions of PostgreSQL (surely not), the REINDEX DATABASE command was introduced in PostgreSQL 7.4. So, if you're using a version older than 7.4, this command won't be available. In versions older than 7.4, you would need to reindex individual tables or indexes manually. Here’s an example of how you could perform the same operation table by table:
```
JakobLichterfeld, line 172: We do not support PostgreSQL below v15, so this notes does not add any value imo

JakobLichterfeld (changes requested):
Thanks for simplyfining.

website/docs/maintenance/manually_fixing_data.mdx
jheredianet: You are right. I've removed it.

JakobLichterfeld (approved):

PR #4577, opened 2025-03-06
fix: update mix dependency hash in nix builds.

weiren2:
Fixed the hash mismatch so nix can build.

Tested locally on an x86 machine running NixOS 24.11.


JakobLichterfeld:
> Fixed the hash mismatch so nix can build.

Thanks for your contribution!

It seems like we need to improve on this end, as our CI only updates the flake.lock

Will manually test on my end before merging

JakobLichterfeld (approved):

PR #4629, opened 2025-04-05
feat: make use of car filter, add timefilter in locations dashboard

### LINKED ISSUE #4609: Car filter in Locations is not used

### What happened?

The Car Filter in the Locations Dashboard seems to have no functionality.
The filter is not used in any panel. It always shows data from all cars combined.

### Expected Behavior

Filter all panels to the selected car.

### Steps To Reproduce

Change Car filter from All to any Car

### Relevant log output

```shell
No log needed
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.33.0

swiffer:
closes #4609

### DIFF
--- a/grafana/dashboards/locations.json
+++ b/grafana/dashboards/locations.json
@@ -95,11 +95,12 @@
             "type": "grafana-postgresql-datasource",
             "uid": "TeslaMate"
           },
+          "editorMode": "code",
           "format": "table",
           "group": [],
           "metricColumn": "none",
           "rawQuery": true,
-          "rawSql": "select count(*) from addresses;",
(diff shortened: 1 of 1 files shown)


DrMichael:
@swiffer Hmmm, `charging_processes` also uses `addresses`. Might be worth adding also.

swiffer:
I had that in initially...

But i guess you cant charge without driving to the charger in the first place ;)

Has no difference with my data and left it out.

DrMichael:
> But i guess you cant charge without driving to the charger in the first place ;)

:-) I can
image

swiffer:
okay, fair enough :) PR updated.

I tried to consolidate additional queries but that would have involved adding to many transformations, that's why i kept combining the first row only.

think ready to be merged now.

DrMichael:
> :-) I can, I am abroad sometimes for month...

But admittedly only one address... :-)

PR #4633, opened 2025-04-05
Update Makefile

Firebasky:
123

PR #4637, opened 2025-04-06
ci(sec): remove pull_request_target workflow triggers to improve sec even further, the downside is that test images are now only created for repo's own PRs

JakobLichterfeld:


JakobLichterfeld:
> And #4633 also (this one is still available).

Yeah reported that user as well.

> I think with the increasing action against Tesla, we may accidentally become targets too, so we do need to ensure our security is as tight as possible.

Agree. We now do not use any dangerous workflow according to OpenSSF scorecard ([the metadata there](https://deps.dev/project/github/adriankumpf%2Fteslamate) is outdated, March 24th).
Our continuous effort to pin dependencies and other fuel the same engine to keep sec as tight as possible.

> Is the docker secret the only one we have to worry about?

From what I see yes, this is the only secret beside of token for coveralls report, see https://github.com/teslamate-org/teslamate/settings/secrets/actions

JakobLichterfeld:
> * I do have a solution in mind since Friday, but in the meantime this seems to be the best possible response.

This one: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

JakobLichterfeld:
Additionally, I activated CodeQL automated scanning

brianmay:
The referenced documentation has multi-line JavaScript code to download the artifact. Suspect it might be possible to simplify this with the download-artifact action: https://github.com/actions/download-artifact

JakobLichterfeld:
Yeah, we already using the download-artifact action in our merge action: https://github.com/teslamate-org/teslamate/blob/f5d0631d513bf3c5be6aa079eb4be23f204ea0c4/.github/actions/merge/action.yml#L22

PR #4639, opened 2025-04-07
feat: check Postgres version on startup, require 16.7 / 17.3, update earthdistance extension

### LINKED ISSUE #4638: earthdistance v1.2 in PG16/17 -> migration needed

### What happened?

we are currently installing the bundled postgres extensions cube and earthdistance.

earthdistance has been updated to 1.2 in PG 16.7 / 17.3

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=31daa10fa
https://learn.microsoft.com/en-us/azure/postgresql/extensions/concepts-extensions-versions#earthdistance

upgrading PostgreSQL is not automatically updating bundled extensions.

we can do so via migration but to ensure this is running as planned we should require updating to PG 16.7 / 17.3 beforehand.

The proposal would be to add a Postgres Version check to the Migration and fail if the version is to low asking users to upgrade their Postgres version first.

This also ensures users are updating the Postgres instances and we can rely on recent functions without causing issues.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.33.0

swiffer:
This PR adds a check for on startup that validates the PostgreSQL version used.
Requirements are set to 16.7+ / 17.3+ - if not matched, an exception is raised.

In addition a migration is added that updates the earthdistance extension to 1.2 (bundled since 16.7 / 17.3).

Feedback welcome, fixes #4638

### DIFF
--- a/lib/teslamate/application.ex
+++ b/lib/teslamate/application.ex
@@ -11,6 +11,8 @@ defmodule TeslaMate.Application do
     :ok = :telemetry.detach({Phoenix.Logger, [:phoenix, :socket_connected]})
     :ok = :telemetry.detach({Phoenix.Logger, [:phoenix, :channel_joined]})
 
+    TeslaMate.DatabaseCheck.check_postgres_version()
+
     Supervisor.start_link(children(), strategy: :one_for_one, name: TeslaMate.Supervisor)
   end
 
--- a/lib/teslamate/database_check.ex
+++ b/lib/teslamate/database_check.ex
@@ -1,0 +1,61 @@
+defmodule TeslaMate.DatabaseCheck do
+  alias Ecto.Adapters.SQL
+  alias TeslaMate.Repo
+
+  # Minimum versions for supported major releases
+  @min_version_16 "16.7"
+  @min_version_17 "17.3"
+
+  def check_postgres_version do
+    # Start the Repo manually without running migrations
+    {:ok, _pid} = Repo.start_link()
+
+    # Query the PostgreSQL version
+    {:ok, result} = SQL.query(Repo, "SELECT regexp_replace(version(), 'PostgreSQL ([^ ]+) .*', '\\1') AS version", [])
+    raw_version = result.rows |> List.first() |> List.first()
+
+    # Normalize to SemVer by appending .0 if needed
+    version = normalize_version(raw_version)
+
+    # Split into major and minor parts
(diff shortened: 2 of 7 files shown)


swiffer:
to be clear - this will intentionally break existing installs not running on 16.7+ / 17.3+ and enforces users to upgrade.

from my checks supported versions should be available on all supported platforms.

@JakobLichterfeld - i suggest releasing it as part of v1.34 and mention it in the release notes?

cwanja:
What is this going to do users who are running on installations that do not support PostgreSQL 16 or 17, but are running fine on 15 or below? Are we comfortable locking them at v1.33?

swiffer:
That would be the consequence...

Based on supported installs there shouldn't be any but i guess there are some installs out there...

parkr:
Is there a way users can check their minor version? Migrating is a bit of work. I'm on 16 but not sure which minor version. Is this something we could display in Grafana for example?

swiffer:
Migrating between minors is "no" work, compose pull, down, up

You can see your current version in the database dashboard

parkr:
Yes, easier when you trust postgres minor versions. Is this also documented on the migration page or is that too much detail and users will be confused? The best way is to back up, clear the volume, pull the new version, and restore. It's best because it works for any version upgrade but takes much longer. So if folks can understand what they're doing, the pull/down/up is a great shortcut.

swiffer:
It's not explicitly stated (the shortcut for minors) but the title in the docs is: 

"Upgrading PostgreSQL to a new major version".

All install install code is not pinning to specific minors so upgrades can happen anytime.

Manual installs are upgraded via package managers as well.

See Upgrades in the official docs. https://www.postgresql.org/support/versioning/

Open for extending the docs if others agree that providing more context helps!

JakobLichterfeld:
> Open for extending the docs if others agree that providing more context helps!

Imo the docs are sufficent there: https://docs.teslamate.org/docs/upgrading

It is not in TeslaMate's responsibility to educate the users how to update their docker stacks, nor when to do so. Pulling new minor PostgreSQL, grafana or mosquitto or anything else should be done on regular basis according to own update strategy.

PR #4648, opened 2025-04-11
feat: check Postgres version on startup, require 16.7 / 17.3, update earthdistance extension

### LINKED ISSUE #4638: earthdistance v1.2 in PG16/17 -> migration needed

swiffer:
follow up of #4639 

closes #4638

### DIFF
--- a/.github/workflows/ghcr_build.yml
+++ b/.github/workflows/ghcr_build.yml
@@ -20,10 +20,33 @@ jobs:
   check_paths:
     uses: ./.github/workflows/check_paths.yml
 
-  teslamate_build:
-    name: Build images
+  check_if_pr_from_outside_repo:
+    name: Check if PR from outside repo
+    # as we currently only push ghcr images for PRs from our own repo
     needs: check_paths
     if: needs.check_paths.outputs.githubfolder != 'true'
+    runs-on: ubuntu-24.04
+    steps:
+      - name: check if PR from outside repo
+        id: check_if_pr_from_outside_repo
+        shell: bash
+        run: |
+          echo "::group::Check if PR from outside repo"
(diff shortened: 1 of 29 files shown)


JakobLichterfeld:
ci fixed in main for our repo

swiffer:
@JakobLichterfeld - still failing 🤔

JakobLichterfeld:
> @JakobLichterfeld - still failing 🤔

as expected as it is from outside our repo

swiffer:
I thought it's better for keeping things well organized. If it can't be solved quickly i can change branch names as well.

JakobLichterfeld:
> I thought it's better for keeping things well organized.

Makes total sense.

> If it can't be solved quickly i can change branch names as well.

Fixed on master

JakobLichterfeld (approved):
lgtm, ty!

PR #4659, opened 2025-04-13
docs: allow to add energy added to the Home Assistant's Energy tab to measure how much energy each session uses

alexsapran:
This commit adds the `state_class: total` in the energy sensor that allows to add it in the energy tab to measure how much energy each session consumes. The value is set to `total` as each charge session does not cumulative increase.


JakobLichterfeld:
Failed CI on GHCR build is expected behavior atm. Sorry for this. We needed to shut down the pull_request_target trigger therefore it will fail for PRs from forks and will only work for PR from branches in this repo. We will bring back our dedicated testing images per PR for forks as well, but this needs some rework.

Jobs will be skipped with #4660

JakobLichterfeld (approved):
nice, ty.

PR #4662, opened 2025-04-15
feat: use Grafana 11.6.1 to fix issues when using browser locale for date formats

swiffer:
- Bumps Grafana to 11.6.1, fixes issues when using browser locale for date formats
- Widens Date Columns to fully show date strings formatted in US locale
- Removes obsolete config entries from Dashboard json
- Moves database to jsonData in datasources.yaml according to current spec

fixes #4656, fixes #4657, fixes #4690

PR #4678, opened 2025-04-19
feat: add cost per mileage to trips dashboard

jheredianet:
This is a little improvement to the Trips dashboard. It adds information about the cost per mileage.


swiffer:
The value is misleading if battery level on beginning of trip and end of trip differs, right?

JakobLichterfeld:
Adittionally, please ensure 

> "When making larger changes to Grafana Dashboards please include updated screenshots (1920x1080 @ x2, Grafana running in Kiosk mode) in ./website/static/screenshots. Ensure to blur sensitive information." 

(https://docs.teslamate.org/docs/development#making-changes-to-grafana-dashboards)

jheredianet:
> misleading

I don't understand what you mean.  To estimate the cost per mileage, I'm just dividing the cost / mileage (obtained by the odometer). I'm not using the SOC. So in a Trip it's possible to have the battery level with the same value or more than the initial value when starting a trip, because when you travel you usually charge the battery on the road, one o more times.

jheredianet:
> Adittionally, please ensure
> 
> > "When making larger changes to Grafana Dashboards please include updated screenshots (1920x1080 @ x2, Grafana running in Kiosk mode) in ./website/static/screenshots. Ensure to blur sensitive information."
> 
> (https://docs.teslamate.org/docs/development#making-changes-to-grafana-dashboards)

Done!

swiffer:
👍 can we keep size at 3840x2160 ?

grafana/dashboards/trip.json
```diff
@@ -1234,6 +1234,108 @@
...
+          "select": [
+            [
+              {
+                "params": [
+                  "efficiency"
+                ],
+                "type": "column"
+              }
+            ]
+          ],
```
swiffer, line 1306: can be removed, guess comes from duplicating an existing panel?

grafana/dashboards/trip.json
```diff
@@ -1234,6 +1234,108 @@
...
+          "table": "cars",
+          "timeColumn": "inserted_at",
+          "timeColumnType": "timestamp",
+          "where": [
+            {
+              "name": "$__timeFilter",
+              "params": [],
+              "type": "macro"
+            }
+          ]
```
swiffer, line 1333: can be removed, guess comes from duplicating an existing panel?

grafana/dashboards/trip.json
```diff
@@ -1234,6 +1234,108 @@
...
+      "targets": [
+        {
+          "datasource": {
+            "type": "grafana-postgresql-datasource",
+            "uid": "TeslaMate"
+          },
+          "editorMode": "code",
+          "format": "table",
+          "group": [],
+          "metricColumn": "none",
```
swiffer, line 1293: can be removed, guess comes from duplicating an existing panel?

swiffer (changes requested):

grafana/dashboards/trip.json
```diff
@@ -1234,6 +1234,108 @@
...
+            "mean"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "showPercentChange": false,
+        "textMode": "value_and_name",
+        "wideLayout": true
+      },
+      "pluginVersion": "11.6.0",
```
swiffer, line 1283: please change to 11.6.1 - see #4662

grafana/dashboards/trip.json
jheredianet: Great catch. Yes, It seems to be inherited by an existing panel. Done!

swiffer:
> > misleading
> 
> I don't understand what you mean.  To estimate the cost per mileage, I'm just dividing the cost / mileage (obtained by the odometer). I'm not using the SOC. So in a Trip it's possible to have the battery level with the same value or more than the initial value when starting a trip, because when you travel you usually charge the battery on the road, one o more times. 

Assume you start your trip with a full battery, drive, charge, drive, end your trip.

See calc for consumption which takes this into account.

I would add a note saying cost can vary based on soc at start / end at least. Same for totals).

swiffer:
What about (consumption/1000 * avg price per kwh * km) driven.

Or a hint / (N/A) if soc before first drive differs more than x% when compared with soc after last charge or drive in selected period.

jheredianet:
> Assume you start your trip with a full battery, drive, charge, drive, end your trip.
> 
> See calc for consumption which takes this into account.
> 
> I would add a note saying cost can vary based on soc at start / end at least. Same for totals).

Sorry, but I still don't understand. If I'm going on a trip, the SOC (State Of Charge) doesn't matter. It's simply that if charging at home + charging on the road cost me X € and I've traveled Y kilometers, then X/Y gives me the cost per km, or multiplied by 100, it would be the cost per 100 km. The part I don't understand here is what SOC or Consumption has to do with it.

swiffer:
Sorry, was on mobile - the explanation was a little rough.

Your calculation is completely right - but only if you are charging as much energy as needed for driving.

scenario:

beginning of trip
30% state of charge - you charge to 90% - cost: 4 eur
driving 350km to 10% of charge
charging to 50% - cost: 3 eur
end of trip

total cost: 7 eur, cost per 100km: 2 eur

but - you end up with 20% more energy in your battery then what you have had at the beginning of the charge

the total /per 100km cost would be lower if only charged to 30%

jheredianet:
> Your calculation is completely right - but only if you are charging as much energy as needed for driving.
> 
> scenario:
> 
> beginning of trip 30% state of charge - you charge to 90% - cost: 4 eur driving 350km to 10% of charge charging to 50% - cost: 3 eur end of trip
> 
> total cost: 7 eur, cost per 100km: 2 eur
> 
> but - you end up with 20% more energy in your battery than what you have had at the beginning of the trip.
> 
> the total /per 100km cost would be lower if only charged to 30% at the end of your trip.

Ok, I understand your point. Normally, on a trip, you always leave home with the highest possible SOC and charge only enough to get to your destination.
But, if you need to be more precise, the ideal would be to end the trip at the same SOC you left with. If that's the case, what I've often done is:

Since this dashboard allows you to select the start and end date (which can be hours, days, or weeks), what I've done is to set the start and end dates and times (before and after the trip) so that the SOC matches at both ends. 

For example, I leave the car charging overnight, leave at 6 a.m., make the trip, charge along the way, and arrive at my destination, let's say with 30%. So, what I do is to use this dashboard and set the start date and time when the car was at 30% (to match the end SOC of the whole trip). This way, the price would be more accurate, as you indicate. 

What I don't want to include is the consumption variable in the calculation, because it could happen that I went on a picnic in the countryside and had the car stopped hours with the doors open, the music playing, and the Sentinel active. Therefore, it's energy consumption that wasn't necessarily used to move the car. I don't know if I explained myself well.

In any case, what estimation do you consider appropriate for this? Because what you proposed previously (consumption/1000 * avg price per kwh * km) driven, I don't think is accurate either.

I think the goal of this dashboard is for the user to know how much energy has been spent during a trip from start to finish.

JakobLichterfeld:
> Ok, I understand your point. Normally, on a trip, you always leave home with the highest possible SOC and charge only enough to get to your destination.

This is only valid for some users. For users without a charging point at home, or for users without a charging point at their destination, the approach is completely different.

Personally, the only solution I can see is to manually set the cost of charging and calculate the price per km in the relevant time frame.

jheredianet:
> > Ok, I understand your point. Normally, on a trip, you always leave home with the highest possible SOC and charge only enough to get to your destination.
> 
> This is only valid for some users. For users without a charging point at home, or for users without a charging point at their destination, the approach is completely different. Even with a charger at home leaving with 100% doesn't make sense it many cases.

I believe it doesn't matter where you charge, at home or on the street. In any case, there's always a price to pay. And that's what it's all about: knowing how much your trip cost, not precisely by the meter travelled, but rather a global estimate of the expense.
 
> Personally, the only solution I can see is to manually set the cost of charging and calculate the price per km in the relevant time frame.
In Teslamate, you actually always have to enter the charging costs manually (unless you have a geofence), dnd in this panel, those are the costs it uses to calculate.

Based on your comments, are you suggesting not including this information panel?

JakobLichterfeld:
> What I don't want to include is the consumption variable in the calculation, because it could happen that I went on a picnic in the countryside and had the car stopped hours with the doors open, the music playing, and the Sentinel active. Therefore, it's energy consumption that wasn't necessarily used to move the car.

As a user, this is consumed energy for that specific trip, imo.

> Based on your comments, are you suggesting not including this information panel?

In the suggested way it is not matching users expectations as you can see on the discussion.

jheredianet:
> > What I don't want to include is the consumption variable in the calculation, because it could happen that I went on a picnic in the countryside and had the car stopped hours with the doors open, the music playing, and the Sentinel active. Therefore, it's energy consumption that wasn't necessarily used to move the car.
> 
> As a user, this is consumed energy for that specific trip, imo.
> 

Yes, you are right. The actual estimation considers it (the actual cost per km with be a little higher)

> > Based on your comments, are you suggesting not including this information panel?
> 
> In the suggested way it is not matching users expectations as you can see on the discussion.

Ok, feel free to discard this PR. I will include it on my own [Custom Grafana Dashboards](https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards/) repository for those who finds it useful.

swiffer:
Currently on a longer trip (cologne to barcelona). Will play around with some calculations afterwards.

JakobLichterfeld:
> Currently on a longer trip (cologne to barcelona).

Safe trip!

> Ok, feel free to discard this PR. I will include it on my own [Custom Grafana Dashboards](https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards/) repository for those who finds it useful.

I think the idea is a great addition from a user perspective! I just want to make sure we are on the same page.

> Will play around with some calculations afterwards.

Great idea, so we can all come to a common understanding and solution @swiffer and @jheredianet

jheredianet:
> Currently on a longer trip (cologne to barcelona). Will play around with some calculations afterwards.

Have a nice trip 👍🏼

JakobLichterfeld:
More discussion also here: https://github.com/teslamate-org/teslamate/discussions/2433#discussioncomment-13402292

PR #4684, opened 2025-04-21
fix: set $__timezone explicitly in dashboards to ensure truncation is done with respect to the Grafana timezone

### LINKED ISSUE #4675: Incorrect filter for daily statistics dashboard

### What happened?

I have noticed that when filtering the Statistics dashboard for daily drives, some drives appear to be tagged to the following day. It seems to happen with drives that occur in the afternoon or evening (I'm in Eastern Time). For example, in the attached screenshot, there are two drives listed under the 4/15 dashboard, even though they actually took place on 4/14.

I have confirmed my time zone is correct in the .env file. 

Image

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
See attached screenshots.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

1.33

### PostgreSQL version

_No response_

swiffer:
Grafana is not setting the PostgreSQL session timezone. To ensure truncation is done with respect to the Grafana timezone setting set the [optional time_zone argument](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC) for `DATE_TRUNC()`.

Fixes #4675


JakobLichterfeld (changes requested):
Thanks!
Unfortunately there seems to be some syntax error in Drive Stats
image

swiffer:
on it

JakobLichterfeld:
# of drives is different as well

swiffer:
yes, drives happening over night have been missing so far - current number should be correct. can you compare with drives dashboard?

JakobLichterfeld:
Syntax Error is fixed, ty"!

> yes, drives happening over night have been missing so far - current number should be correct. can you compare with drives dashboard?

Nope, drives dashboard shows 122 rows for last 6 months which seems to be right
image

v1.33.0 drive stats shows same number but #4684 does show 137

swiffer:
Okay, will have a look...

DrMichael:
Let's collect it here...
https://github.com/teslamate-org/teslamate/issues/4675#issuecomment-2817997916
Almost there... :-)
You still need to replace in the Statistics' dashboard the CASE $period's the date by TIMEZONE('$__timezone', date).
Currently, the period shows up correctly in my edge case, but the period is the day before...

Charging Stats for my edge case looks good.

Perhaps @giterdone509 can also test it. He is 4 hours off UTC...

swiffer:
@JakobLichterfeld do you have any incomplete drives? drives dashboard and stat are identical looking at my data 🤔

JakobLichterfeld:
> @JakobLichterfeld do you have any incomplete drives? drives dashboard and stat are identical looking at my data 🤔

Yeah but not on this timeframe

swiffer:
ok, should those be ex or included?

swiffer:
@DrMichael - thanks checking this! fixed all places where making use of to_char!

DrMichael:
> @DrMichael - thanks checking this! fixed all places where making use of to_char!

My pleasure.

JakobLichterfeld:
> ok, should those be ex or included?

A drive is a drive, innit? So in my case it is one drive which was split in 15 due to whatever reason.
So as a user I would expect unclosed drives to be included.

swiffer:
Ready to be merged

JakobLichterfeld:
> > ok, should those be ex or included?
> 
> A drive is a drive, innit? So in my case it is one drive which was split in 15 due to whatever reason.
> So as a user I would expect unclosed drives to be included.
> Or we write number of finished drives

Sorry, during my run in the woods I changed my mind. We do handle unclosed drives and charges as corrupt in other places, aren't we? So they should not be included and you are right.

JakobLichterfeld (approved):
Successfully tested. Ty!
And sorry for the back and forth

Issue #4661, opened 2025-04-15
Grafana 11.2.3, Can't display image on welcome dashboard


cyqtmxk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

the docker runs on Openwrt24.10，the platform is arm/v7, however, on the welcome dashboard, there is no image in the middle:

![Image](https://github.com/user-attachments/assets/a7ccda4a-b357-475a-a758-20e419664603)

### Expected Behavior

I try run on Synology NAS, it works well~

### Relevant log output

```shell
teslamate-1  | 2025-04-15 01:07:45.532 [info] Refreshed api tokens
teslamate-1  | 2025-04-15 01:07:45.582 [info] Scheduling token refresh in 6 h
teslamate-1  | 2025-04-15 01:07:45.737 [info] Running TeslaMateWeb.Endpoint with cowboy 2.13.0 at :::4000 (http)
teslamate-1  | 2025-04-15 01:07:45.757 [info] Access TeslaMateWeb.Endpoint at http://localhost
teslamate-1  | 2025-04-15 01:07:46.342 [info] Starting logger for 'Model Y'
teslamate-1  | 2025-04-15 01:07:46.819 car_id=1 [info] Start / :offline
teslamate-1  | 2025-04-15 01:07:47.669 [info] MQTT connection has been established
teslamate-1  | 2025-04-15 01:09:31.885 [info] GET /
teslamate-1  | 2025-04-15 01:09:32.006 [info] Sent 200 in 119ms
teslamate-1  | 2025-04-15 03:23:13.727 car_id=1 [info] Start / :online
teslamate-1  | 2025-04-15 03:23:13.881 car_id=1 [info] Stream connecting ...
teslamate-1  | 2025-04-15 03:26:16.921 car_id=1 [info] Suspending logging
teslamate-1  | 2025-04-15 03:36:53.341 car_id=1 [info] Stream :inactive in suspended, fetching vehicle state ...
teslamate-1  | 2025-04-15 03:36:54.342 car_id=1 [info] Start / :asleep
teslamate-1  | 2025-04-15 03:36:54.389 car_id=1 [info] Stream disconnecting ...
teslamate-1  | 2025-04-15 07:07:45.584 [info] Refreshing access token ...
```

### Type of installation

Docker

### Version

v1.33.0

### Are you running latest major supported PostgreSQL version?

- [x] I run the latest major supported PostgreSQL version

277601502:
[...]

PR #4702, opened 2025-04-27
fix: new image on home dashboard as old got removed from Tesla mediaserver

JakobLichterfeld:
closes: #4661

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 #### Dashboards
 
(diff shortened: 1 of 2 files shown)


swiffer (approved):

arcastro:
If it ever happens again, maybe internet archive can be used. Here's an image snapshot that should hopefully remain indefinitely.

https://web.archive.org/web/20250510162935if_/https://digitalassets.tesla.com/tesla-contents/image/upload/c_pad,dpr_1.0,f_auto,q_auto/c_pad/Group_76?pgw=1

PR #4705, opened 2025-04-28
Update Teslemetry

Bre77:
Hey Teslamate team! Teslemetry here.

We are [moving away](https://teslemetry.com/blog/road-away-from-polling) from `vehicle_data` polling on July 1st, which means the Teslamate approch of using Teslemetry's vehicle_data endpoint is likely going to get very expensive for users.

I am notifying my customers each month, but I want to make sure your documentation sets the right expectation too. If you would prefer to remove Teslemetry all together I fully understand.

I also tweaked the line about streaming. We do offer streaming, and in fact thats the primary way of consuming data in Teslemetry moving forward, however it is not a legacy compatible WSS stream like MyTeslaMate offers, but rather a SSE stream containing the near-raw fleet telemetry messages.

That said, looking at the code for `MyTeslaMate/websocket` I might consider adding a WSS option to Teslemetry if that helps mitigate the loss of vehicle_data.

### DIFF
--- a/website/docs/guides/api.md
+++ b/website/docs/guides/api.md
@@ -76,6 +76,8 @@ MyTeslaMate also provides streaming by [reproducing the old streaming from the d
 
 #### Teslemetry Fleet API
 
+**_Important: Teslemetry's vehicle data polling is [pay-per-use](https://teslemetry.com/blog/road-away-from-polling)._**
+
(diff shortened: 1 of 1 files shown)


USAFPride:
@brianmay @JakobLichterfeld 

What impacts does this have?

JakobLichterfeld:
Thanks for your heads-up!

At the moment I am leaning towards removing Teslemetry from our docs as it is a paid service.

PR #4715, opened 2025-05-02
feat: preserve the folders when restoring dashboards

IngmarStein:
Use the dashboard UID instead of the dashboard name to get the folder ID. This way, the dashboards get restored to their current location instead of the root folder.

### DIFF
--- a/grafana/dashboards.sh
+++ b/grafana/dashboards.sh
@@ -63,13 +63,9 @@ restore() {
 				--user "$LOGIN" \
 				-X POST -H "Content-Type: application/json" \
 				-d "{\"dashboard\":$(cat "$dashboard_path"), \
-                    \"overwrite\":true, \
-                    \"folderId\":$folder_id, \
-                    \"inputs\":[{\"name\":\"DS_CLOUDWATCH\", \
-                                 \"type\":\"datasource\", \
-                                 \"pluginId\":\"cloudwatch\", \
-                                 \"value\":\"TeslaMate\"}]}" \
-				"$URL/api/dashboards/import"
+					\"overwrite\":true, \
+					\"folderId\":$folder_id}" \
+				"$URL/api/dashboards/db"
 
 			echo "RESTORED $(basename "$dashboard_path")"
 		done
--- a/grafana/dashboards.sh
+++ b/grafana/dashboards.sh
@@ -57,7 +57,7 @@ backup() {
 restore() {
 	find "$DASHBOARDS_DIRECTORY" -type f -name \*.json -print0 |
 		while IFS= read -r -d '' dashboard_path; do
-			folder_id=$(get_folder_id "$(basename "$dashboard_path" .json)")
+			folder_id=$(get_folder_id "$(jq -r '.uid' <"$dashboard_path")")
 			curl \
 				--silent --fail --show-error --output /dev/null \
 				--user "$LOGIN" \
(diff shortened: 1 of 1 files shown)


brianmay (approved):

brianmay:
Will the restore process create the folders if they don't already exist?

IngmarStein:
I don't think so - that is unchanged from before (where the URLs were just broken). It would indeed be nice to call `POST /api/folders` to create them.

IngmarStein:
@brianmay I'm also curious about the `DS_CLOUDWATCH` data source. Is this just a leftover from an example and can be removed?

brianmay:
> @brianmay I'm also curious about the `DS_CLOUDWATCH` data source. Is this just a leftover from an example and can be removed?

No idea sorry.  Git history shows that it has been there for a since the beginning.

IngmarStein:
> No idea sorry. Git history shows that it has been there for a since the beginning.

The `dashboards/import` endpoint is not documented as part of Grafana's official API. People have discovered it by observing the calls made by the Grafana UI and that knowledge was shared in various forums. The CloudWatch data source was used in some examples, so it might just have been carried over from https://github.com/sivakrishna2606/-grafana-aws-cloudwatch-dashboards.

I've added a commit which removes it because it's not needed.

IngmarStein:
This should eventually just use https://grafana.github.io/grafanactl and `grafanactl pull` / `grafanactl push`. The tool is in public preview currently.

swiffer:
grafanactl is looking great but would further complicate the manual install procedure (not bundled yet)

- https://grafana.github.io/grafanactl/installation/#prebuilt-binaries
- https://grafana.github.io/grafanactl/configuration/#defining-contexts

checking the docs I noted that there are offical API endpoints for creating/updating dashboards:

- https://grafana.com/docs/grafana/v11.6/developers/http_api/dashboard/#create--update-dashboard

as well as endpoints for creating / updating folders:

- https://grafana.com/docs/grafana/v11.6/developers/http_api/folder/#create-folder

those should be used imho.

Note that Grafana 12 introduces an entirely new API Structure, however Grafana 12 is not available on all endpoints yet but could mean another rewrite / change down the road.

https://grafana.com/docs/grafana/latest/developers/http_api/apis/

IngmarStein:
Targeting the Kubernetes-style APIs which are available in v12 as the `v1beta1` version makes sense, but I would do that in a follow-up PR. This one is a small change and a strict improvement over the status quo and uses the existing API.

grafana/dashboards.sh
```diff
@@ -57,18 +57,14 @@ backup() {
...
 				-X POST -H "Content-Type: application/json" \
 				-d "{\"dashboard\":$(cat "$dashboard_path"), \
                     \"overwrite\":true, \
-                    \"folderId\":$folder_id, \
-                    \"inputs\":[{\"name\":\"DS_CLOUDWATCH\", \
-                                 \"type\":\"datasource\", \
-                                 \"pluginId\":\"cloudwatch\", \
-                                 \"value\":\"TeslaMate\"}]}" \
+                    \"folderId\":$folder_id}" \
 				"$URL/api/dashboards/import"
```
swiffer, line 68: can we switch to the officially documented endpoint here ?

`POST /api/dashboards/db`

IngmarStein: Sure, done!

swiffer:
thanks - looking good. lgtm.

PR #4723, opened 2025-05-06
fix(nix): update cldr hash

brianmay:
See https://github.com/teslamate-org/teslamate/discussions/4503

My feeling is that there should be a better solution. e.g. we should be able to use the files downloaded by `mixRelease`.

But my current home environment is somewhat distracting right now :-(

So this should do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Improvements and bug fixes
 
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
Ty. On darwin I still get the permission error with da.json.
As darwin is not a server machine I do see this as low priority.

PR #4726, opened 2025-05-06
draft: fix(docs): revise database procedures

brianmay:
The database instructions, while technically correct, are not idiot proof. There are so many different ways you can stuff up and end up losing all your data.

For example, the current docs say:

docker compose exec -T database pg_dump -U teslamate teslamate > ./teslamate.bck

If anything goes wrong with this command, it might overwrite any potentially good backup with a 0 byte file. And some of the users won't notice until after they have deleted the current database and tried to restore. By this stage it is too late to recover.

These instructions attempt to make the process more resilient by not deleting the current data until after it is confirmed that the restore is good.

The instructions also use a unique filename for every backup. So you are not overwriting the same backup file every time.

### DIFF
--- a/website/docs/maintenance/backup_restore.md
+++ b/website/docs/maintenance/backup_restore.md
@@ -8,10 +8,10 @@ If you are using `docker-compose`, you are using Docker Compose v1, which has be
 
 ## Backup
 
-Create backup file `teslamate.bck`:
+Create backup file:
 
 ```bash
-docker compose exec -T database pg_dump -U teslamate teslamate > ./teslamate.bck
+docker compose exec -T database pg_dump -U teslamate teslamate > teslamate_database_$(date +%F_%H-%M-%S).sql
 ```
 
 :::note
@@ -19,7 +19,7 @@ docker compose exec -T database pg_dump -U teslamate teslamate > ./teslamate.bck
 :::
 
 :::note
-Be absolutely certain to move the `teslamate.bck` file to another safe location, as you may lose that backup file if you use a docker-compose GUI to upgrade your teslamate configuration. Some GUIs delete the folder that holds the `docker-compose.yml` when updating.
+Be absolutely certain to move the backup file to another safe location, as you may lose that backup file if you use a docker-compose GUI to upgrade your teslamate configuration. Some GUIs delete the folder that holds the `docker-compose.yml` when updating.
 :::
 
 :::note
@@ -36,21 +36,36 @@ If you changed `TM_DB_USER` in the .env file from one of the advanced guides, ma
 Replace the default `teslamate` value below with the value defined in the .env file if you have one (TM_DB_USER and TM_DB_NAME)
 :::
 
-```bash
-# Stop the teslamate container to avoid write conflicts
-docker compose stop teslamate
+1. Stop the teslamate container to avoid write conflicts
 
-# Drop existing data and reinitialize (Don't forget to replace first teslamate if using different TM_DB_USER)
-docker compose exec -T database psql -U teslamate teslamate << .
-DROP SCHEMA public CASCADE;
-CREATE SCHEMA public;
-CREATE EXTENSION cube WITH SCHEMA public;
-CREATE EXTENSION earthdistance WITH SCHEMA public;
-.
+   ```bash
+   docker compose stop teslamate
+   ```
 
-# Restore
-docker compose exec -T database psql -U teslamate -d teslamate < teslamate.bck
+2. Drop existing data and reinitialize (Don't forget to replace first teslamate if using different TM_DB_USER)
 
-# Restart the teslamate container
-docker compose start teslamate
-```
+   ```bash
+   docker compose exec -T database psql -U teslamate teslamate << .
+   DROP SCHEMA public CASCADE;
+   CREATE SCHEMA public;
+   CREATE EXTENSION cube WITH SCHEMA public;
+   CREATE EXTENSION earthdistance WITH SCHEMA public;
+   .
+   ```
+
+3. Restore
+
+   Use the same filename that was used in the backup step.
+
+   Replace `teslamate_database_2025-05-07_07-33-48.sql` with the actual filename generated during the backup step.
+
+   ```bash
+   DATABASE="teslamate_database_2025-05-07_07-33-48.sql"
+   docker compose exec -T database psql -U teslamate -d teslamate < "$DATABASE"
+   ```
+
+4. Restart the teslamate container
+
+   ```bash
+   docker compose start teslamate
+   ```
--- a/website/docs/maintenance/upgrading_postgres.md
+++ b/website/docs/maintenance/upgrading_postgres.md
@@ -3,28 +3,117 @@ title: Upgrading PostgreSQL to a new major version
 sidebar_label: Upgrading PostgreSQL
 ---
 
-1. Create a [backup](backup_restore.md)
-2. Stop all TeslaMate containers
+In these instructions it is assumed you are upgrading from postgresql 16 to postgresql 17. Please update
+references to 16 to your start version and 17 to your desired version.
+
+These instructions also assume you have the standard `docker-compose.yml` file. If this is not the case
+you might need to make adjustments. For example, older installed might call the container "db" not "database".
+
+:::note
+If you are using `docker-compose`, you are using Docker Compose v1, which has been deprecated. Docker Compose commands refer to Docker Compose v2. Consider upgrading your docker setup, see [Migrate to Compose V2](https://docs.docker.com/compose/migrate/)
+:::
+
+## Upgrade
+
+1. Create a backup:
+
+   ```bash
+   docker compose exec -T database pg_dump -U teslamate teslamate > teslamate_database_$(date +%F_%H-%M-%S).sql
+   ```
+
+2. _Check to make sure there are no obvious errors in this backup file._ There should not be any errors appearing
+   on screen. The file should be more than 0 bytes long. Do not proceed if there is any doubt.
+
+3. Stop all TeslaMate containers
 
    ```bash
    docker compose down
    ```
 
-3. Delete the database volume. **Be careful**, this will delete all your previously recorded data! Make sure that your backup can be restored before you start.
+4. Make a copy of your current `docker-compose.yml`
 
    ```bash
-   docker volume rm "$(basename "$PWD")_teslamate-db"
+   cp docker-compose.yml docker-compose-16.yml
    ```
 
-4. Change the postgres version in docker-compose.yml and start the container
+5. Update the source to point to a new location that doesn't exist already and update the postgres version.
 
-   ```yml {2}
-   database:
-     image: postgres:xx
+   ```yaml
+   services:
+      [...]
+      database:
+        image: postgres:17
+        [...]
+        volumes:
+          - teslamate-db-17:/var/lib/postgresql/data
+   [...]
+   volumes:
+     teslamate-db-17:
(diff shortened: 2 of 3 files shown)


brianmay:
I have marked this as draft because I have not tested these changes in any way shape or form.

website/docs/maintenance/backup_restore.md
```diff
@@ -36,21 +36,35 @@ If you changed `TM_DB_USER` in the .env file from one of the advanced guides, ma
...
+  .
+  ```
+
+3. Restore
+
+  Use the same filename that was used in the backup step.
+  Replace `<backup_filename>` with the actual filename generated during the backup step (e.g., `teslamate_database_2025-05-07_07-33-48.sql`).
+
+  ```bash
+  docker compose exec -T database psql -U teslamate -d teslamate < <backup_filename>
```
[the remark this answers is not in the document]
brianmay: That change is wrong. The `<` in `<backup_filename>` is not a redirection operator.

brianmay:
Also see https://thomasbandt.com/postgres-docker-major-version-upgrade

swiffer:
You may still want to add the command to delete the old volume after verified that the new setup is working?

DrMichael:
Perhaps we should make 2. under Upgrade in bold face?

brianmay:
> You may still want to add the command to delete the old volume after verified that the new setup is working?

Good question. I deliberately left that out because I thought it might be risky. But it could be desirable all the same.

> Perhaps we should make 2. under Upgrade in bold face?

Why? Because it is an important point that needs highlighting maybe?

Also I just noticed most of the step start with upper case letter, but one of the points under Rollback doesn't. Oops.

DrMichael:
> > Perhaps we should make 2. under Upgrade in bold face?
> Why? Because it is an important point that needs highlighting maybe?

Genau...

> Also I just noticed most of the step start with upper case letter, but one of the points under Rollback doesn't. Oops.

Bummer. :-)

JakobLichterfeld (changes requested):
Thank you for your efforts!

I agree that the process has its pitfalls and we can improve it.

From my point of view, we need to find the sweet spot between easy copy-paste and robustness.
Given the number of tickets and other feedback channels, I do not feel that users are left without data.
I also want to remain OS independent.

JakobLichterfeld:
I would love to get your perspective on DB backup and restore, @parkr

website/docs/maintenance/backup_restore.md
```diff
@@ -36,21 +36,34 @@ If you changed `TM_DB_USER` in the .env file from one of the advanced guides, ma
...
+3. Restore
 
-# Restore
-docker compose exec -T database psql -U teslamate -d teslamate < teslamate.bck
+Use the same filename that was used in the backup step.
+Replace `<backup_filename>` with the actual filename generated during the backup step (e.g., `teslamate_database_2025-05-07_07-33-48.sql`).
 
-# Restart the teslamate container
+```bash
+docker compose exec -T database psql -U teslamate -d teslamate < <backup_filename>
```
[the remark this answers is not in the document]
brianmay: I could replace it with a dummy file name, e.g. teslamate_database_2025-05-08_07-00-22

At least this way the error might be more understandable:

```bash
[brian@miacis:~]$ cat < <backup_filename>
bash: syntax error near unexpected token `<'

[brian@miacis:~]$ cat < teslamate_database_2025-05-08_07-00-22.sql
bash: teslamate_database_2025-05-08_07-00-22: No such file or directory
```

Using the first one is more standardized. But the characters can be misinterpreted as redirection.

The second one gives better errors. But do people even read errors these days? :-(

Perhaps another option:

```bash
BACKUP_FILENAME=teslamate_database_2025-05-08_07-00-22.sql
docker compose exec -T database psql -U teslamate -d teslamate < "$BACKUP_FILENAME"
```

website/docs/maintenance/upgrading_postgres.md
```diff
@@ -3,28 +3,109 @@ title: Upgrading PostgreSQL to a new major version
...
    ```bash
    docker compose down
    ```
 
-3. Delete the database volume. **Be careful**, this will delete all your previously recorded data! Make sure that your backup can be restored before you start.
+4. Make a copy of your current `docker-compose.yml`
 
    ```bash
-   docker volume rm "$(basename "$PWD")_teslamate-db"
+   cp docker-compose.yml docker-compose-16.yml
```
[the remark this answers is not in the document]
brianmay: Any suggestions? Maybe just have text "Copy file docker-compose-16.yml over docker-compose.yml?

website/docs/maintenance/backup_restore.md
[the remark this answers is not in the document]
brianmay: Hopefully I have addressed this now :-)

website/docs/maintenance/upgrading_postgres.md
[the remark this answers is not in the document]
brianmay: I think cp should work on all OS we care about though, right?

website/docs/maintenance/backup_restore.md
```diff
@@ -8,18 +8,18 @@ If you are using `docker-compose`, you are using Docker Compose v1, which has be
 
 ## Backup
 
-Create backup file `teslamate.bck`:
+Create backup file:
 
 ```bash
-docker compose exec -T database pg_dump -U teslamate teslamate > ./teslamate.bck
+docker compose exec -T database pg_dump -U teslamate teslamate > teslamate_database_$(date +%F_%H-%M-%S).sql
```
swiffer, line 14: What about using only a date (2025-05-16 here) - that way copy and paste of commands should still work (as users are most likely updating on the same day).

website/docs/maintenance/upgrading_postgres.md
```diff
@@ -3,28 +3,114 @@ title: Upgrading PostgreSQL to a new major version
...
```
swiffer, line 36: Why should users backup the compose file. We could say - in case of error please revert your changes to the compose file to get back to a working version (that should be fine to ask for).

website/docs/maintenance/upgrading_postgres.md
```diff
@@ -3,28 +3,114 @@ title: Upgrading PostgreSQL to a new major version
...
-   database:
-     image: postgres:xx
+   ```yaml
+   services:
+      [...]
+      database:
+        image: postgres:17
+        [...]
+        volumes:
+          - teslamate-db-17:/var/lib/postgresql/data
```
swiffer, line 48: The volume name is listed at the end of the compose file again, isnt it?

brianmay: Yes, you are right. Now fixed.

website/docs/maintenance/backup_restore.md
brianmay: Sorry, not entirely sure what you are asking for?

website/docs/maintenance/upgrading_postgres.md
brianmay: On one hand, I note that there is a comment that some environments may automatically delete any unexpected files, such as this backup. Oops.

On the other hand, not really convinced our users are good at following written instructions... Think spelling it out might be required.

swiffer: those would also delete the postgres backup, right?

as long as the data is still accessible everything is alright. if someone isn't able to follow a "revert changes made to docker-compose.yaml" file they still can open up a discussion / ask in forums for help.

website/docs/maintenance/backup_restore.md
swiffer: going with `$(date +%Y-%m-%d)` and not using %H-%M-%S will allow the restore command to be a copy / paste without users manually searching for the exact filename taken earlier (assuming that most upgrades happen at the same day).

website/docs/maintenance/upgrading_postgres.md
brianmay: These will delete the updated `docker-compose.yml` for postgres. The postgres data itself should remain intact.

At least that is the theory. Possibly some platform might decide that since the data is no longer referenced it can be deleted. Hoping that doesn't actually happen...

website/docs/maintenance/backup_restore.md
brianmay: Good question.

Was trying to avoid the situation here where you might have a good backup, but then you accidentally run the above command again after the database is shutdown/replaced and replace the good backup with a bad backup. And not notice the glaring error message saying it failed.

Yes, would like to think this wouldn't happen, but we did get a report from somebody who had a 0 byte backup file, and this is one way that could occur.

At least with the above, there is some increased chance you might still have the good backup file.

swiffer:
Completely different idea:

What about using https://github.com/pgautoupgrade/docker-pgautoupgrade in one shot mode instead for the upgrade (Or as the postgres image)?

JakobLichterfeld:
I do think about including backup and restore script in our reo, which will be used via docker-compose service or as nix ideomatic overlay.

While setting up my new NixOS server I stumbled about @scottbot95 s idea here: https://github.com/scottbot95/nixos-config/blob/58fa26c76980a983e4835e0e4cb6ded3194a9e51/machines/teslamate/backup.nix (which is outdated as he uses our flake and the script uses podman).
I then thought about how to implement it not only for nix but for docker compose installs as well. As a script in the container cannot stop other containers, the solution seems to be a docker compose service.

brianmay:
> What about using https://github.com/pgautoupgrade/docker-pgautoupgrade in one shot mode instead for the upgrade (Or as the postgres image)?

That does seem like an interesting idea. But can you make a full procedure that uses it and is idiot proof?

> I then thought about how to implement it not only for nix but for docker compose installs as well. As a script in the container cannot stop other containers, the solution seems to be a docker compose service.

I think these two cases probably will end up being totally different. Unfortunately.

For reference, here is the official nix upgrade procedure: https://nixos.org/manual/nixos/stable/#module-services-postgres-upgrading - we could include such a script in our nixos module. Possibly we might need some additional parameters, not sure. I guess we can assume the user wants to upgrade to the latest postgres in nix stable, but we might need details for the current database.

The Docker case may be complicated because we don't supply a docker-compose file, only give a sample file, so everyone is likely to have differences. In fact that is known to be the case already, some people have the database service called "db", where as the docs say it should be "database".

JakobLichterfeld:
> I think these two cases probably will end up being totally different. Unfortunately.

Agree, that's why I only want a backup and restore script, maintained in our repo which will be called via docker-compose service or via nix.


> some people have the database service called "db", where as the docs say it should be "database".

In ancient versions of the doc, it was called "db".

swiffer:
> That does seem like an interesting idea. But can you make a full procedure that uses it and is idiot proof?

it still would be

1. create a backup (as is)
2. docker compose down (as is)
3. upgrade postgres (see below)
4. modify docker-compose image (as is)
5. docker compose up -d

current step 3 (deletion of volume) and 5 (restore) would no longer be neccessary

upgrade postgres would be:

```sh
docker run --name pgauto -it \
	--mount type=volume,src="$(basename "$PWD")_teslamate-db",dst=/var/lib/postgresql/data \
	-e POSTGRES_PASSWORD=password \
	-e PGAUTO_ONESHOT=yes \
	pgautoupgrade/pgautoupgrade:17-bookworm
```

swiffer:
The next major upgrade might be different (requiring changes to docker compose) and in addition will open new possibilties (see docs):

https://github.com/docker-library/docs/pull/2582/files

PR #4727, opened 2025-05-07
Create Catalan default.po

pellix:


JakobLichterfeld:
Ty for your contribution!

> hoping that the translated version of Grafana will arrive some day

Grafana does not support localization, and maintaining all dashboards in all languages is not an option imo.

PR #4737, opened 2025-05-11
docs: add note on accessory power functionality can prevent car sleep

darox:
After I enabled the accessory power functionality in my Model Y Juniper, my car didn't go to sleep anymore even when no accessory was connected. So I thought it might make sense to add a note in the docs.


JakobLichterfeld (changes requested):
Ty for your contribution!

I would place it first in this section. Currently it is in between the explanation for MCU1 cars

JakobLichterfeld:
Additionally regarding failed CI: please follow https://docs.teslamate.org/docs/development#code-formatting

PR #4744, opened 2025-05-17
fix hardcoded id

swiffer:


swiffer:
@JakobLichterfeld - as this is such a minor change i'll go ahead and merge this one. hope it's ok.
changelog entry not needed imho as grafana 12.0.0 has not been released yet.

JakobLichterfeld:
lgtm.

PR without a review is generally speaking bad practice and gets down rates on OSF score card. 
So imo only if sec risk or really nobody available fir hotfix

swiffer:
ok, will tag you next time!

PR #4746, opened 2025-05-20
fix: consumption for drives with positive soc diff

### LINKED ISSUE #3659: Trips with negative gradient do not record correctly

### What happened?

Trips with negative gradient do not record kWh consumption, average consumption or travel efficiency, which means that the total averages do not show the real data. Living in a mountainous area, only half of the trips are recorded.

### Expected Behavior

If the consumptions are 0kWh or negative, they should be displayed correctly and be part of the overall consumption calculation.

### Steps To Reproduce

_No response_

### Relevant log output

```shell
*
```

### Screenshots

image

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.28.3

### LINKED ISSUE #4717: Total Energy consumed from Trip dashboard is wrongly calculated when regenerating more than start or end rated_battery_range_km

### What happened?

The **Total Energy consumed (gross)** stats from the "Trip" dashboard is wrongly calculated in situations when the max or min rated range is not actually the start and end rated range of the trip. 

For example if I have a trip:

1. Start with 300km rated range
2. Drive downhill and regen to 330km rated range
3. Drive and consume until 170km rated range
4. Drive and regen to 200km rated range
5. Stop with the final 200km rated range

The total consumption is calculated as `(max - min) * efficiency`: (330 - 170) * 149 

### Expected Behavior

The total consumption should be calculated as `(start_rated_range - end_rated_range) * efficiency`: (300 - 200) * 149 

### Steps To Reproduce

_No response_

### Relevant log output

```shell
Logs are not relevant
```

### Screenshots

Image

### Additional data

_No response_

### Type of installation

Docker

### Version

v2.0.0

### PostgreSQL version

v17

swiffer:
fixes #3659 

driving downhill and recovering more energy than what is actually consumed leads to wrong data as queries have actively excluded these cases so far, mostly by wrapping the range_diff used for consumption calculation with `greatest(range_diff, 0)`.

multiple dashboards have been touched, see before and after for two examples that show the improvements.

**before - drives**
grafik

**after - drives**
grafik

**before - drive details**
grafik

**after - drive details**
grafik

fixes #4717 

the explanation in #4717 is perfect, this pr is addressing that by entirely changing the way total consumption gross is calculated. It's now reusing the query results of Ø consumption gross.


swiffer:
@micku7zu - please have a look :)

micku7zu:
I've saw in the code that this case was actively excluded by using the `greatest(range_diff, 0)` but I didn't know the reason and didn't have the time to dig all that deep to understand why. 

**What you've fixed here is a great addition, thanks for your work!** 

But my edge case described here https://github.com/teslamate-org/teslamate/issues/4717 it's a little bit different. It's not about the case when a drive has a positive soc difference at the end of the trip. it's about a drive that has a greater soc during the drive than the start of the trip, or it has a lower soc during the drive than the end of the trip.

This pull request does what it's supposed to do based on its description, so I think we can leave it as is and continue the discussion on the issue I reported.

swiffer:
Sorry, will update this PR to cover that case as well!

swiffer:
@micku7zu - updated this PR - performance should be better as well, I'm now reusing the query of the consumtion gross avg panel. could you maybe post a before and after screenshot hopefully showing correct numbers for the latter?

micku7zu:
Thanks! Will do but I'm currently away from laptop for 24 hours.

JakobLichterfeld (approved):

micku7zu:
Before:


After:


Before: 41.35 kWh
After: 40.61 kWh

It's good now 👍 

**Thanks!**

JakobLichterfeld:
I will test on slow HW before merging

swiffer:
One query deleted, rest as is. Should be good!

JakobLichterfeld:
shall we merge this before #4756?

swiffer:
Yes, will rebase #4756 afterwards!

JakobLichterfeld:
will do a test on slow HW right now and then merge

JakobLichterfeld (approved):
Successfully tested on slow host HW

JakobLichterfeld (approved):

PR #4749, opened 2025-05-20
install metric drilldown app, required for query inspector to function

swiffer:
while upgrading to grafana 12.0.0 preinstall of drilldown apps has been disabled.

however, metric drilldown seems to be required for the panel query inspector to function. as it's useful to have while developing, let's ensure the metric drilldown app is installed.


swiffer:
seems not to be the reason - looking further

swiffer:
https://github.com/grafana/grafana/issues/104256

PR #4758, opened 2025-05-23
docs: add Unraid installation instructions (no support)

vypergts:
I've been responding to different reddit posts about setup on Unraid. I've had this working well for quite some time now so I figured it would be good to contribute directly here to the documentation. I haven't ever worked on github before though so please forgive if I didn't do this correctly or let me know how I can improve it for future updates.

### DIFF
--- a/website/docs/installation/unraid.md
+++ b/website/docs/installation/unraid.md
@@ -13,6 +13,8 @@ This document provides the necessary steps for installation of TeslaMate on Unra
 
 ## Instructions
 
+Unlike the Compose installation which sets up the following containers in one go, on Unraid you will need to install and configure each container individually from Community Apps.
+
 **Postgres Docker**
 1. Go to the Apps tab for the Community Apps and search for postgresql17 (current version supported) and click install.
 2. Verify that no other applications are running on port `5432`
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
Thank you for your contribution!

I am not sure about adding this. We do not have the capacity to support all variants, therefore we only support docker and NixOS installation. All other tickets will be closed. Separate backup and restore instructions will delute the documentation even further.

swiffer:
Can't one use the compose add in and use the standard install we provide via docker?

https://docs.ibracorp.io/docker-compose/docker-compose-for-unraid

vypergts:
I can totally understand if you don’t want to support other installs, it’s still docker though, just easier? Updating a docker container in Unraid is one click with how I’ve documented it. It’s a way, way less command line. I suppose people could run the compose plugin too, my method is keeping it closer to the docker-for-dummys approach that Unraid takes.

fiveangle:
On many other projects I've see a "Community Resources" section (with appropriate "not maintained or monitored by this project" disclaimer) in the project documentation that links to community-maintained resources, forums, docs, plugins, etc.  It is a much lower level of management for the project maintainer since even if the links die well before the maintainer "cleans it up" once or twice a year, it doesn't reflect poorly on the project (plus there's always archive.org ;)
This allows rich resources for impassioned communities to be easily found without requiring users to have a minimum blue-belt in google-fu to find it. Something to consider.

JakobLichterfeld:
closed in favor of #4829

Issue #4761, opened 2025-05-26
nix(darwin): cldr download/build fails


JakobLichterfeld:
as discussed in https://github.com/teslamate-org/teslamate/discussions/4503 `nix build --override-input devenv-root "file+file://"<(printf %s "$PWD")` fails on darwin with 

```bash
       > == Compilation error in file lib/teslamate_web/cldr.ex ==
       > ** (File.Error) could not write to file "/nix/store/xrh3fjc2cv9rfi30gp3050p64yz0cc47-source/priv/cldr/locales/da.json": permission denied
       >     (elixir 1.17.3) lib/file.ex:1144: File.write!/3
       >     (ex_cldr 2.42.0) lib/cldr/install.ex:107: Cldr.Install.do_install_locale_name/3
       >     (elixir 1.17.3) lib/enum.ex:987: Enum."-each/2-lists^foreach/1-0-"/2
       >     (ex_cldr 2.42.0) lib/cldr/install.ex:29: Cldr.Install.install_known_locale_names/1
       >     (ex_cldr 2.42.0) lib/cldr.ex:102: Cldr.install_locales/1
       >     (ex_cldr 2.42.0) expanding macro: Cldr.Backend.Compiler.__before_compile__/1
       >     lib/teslamate_web/cldr.ex:1: TeslaMateWeb.Cldr (module)
```

JakobLichterfeld:
Switching to `force_locale_download: false` it seems to work on darwin

JakobLichterfeld:
From the [cldr doc](https://hexdocs.pm/ex_cldr/readme.html#additional-cldr-packages):

> :force_locale_download determines whether to always download locale files during compilation. Locale data is ex_cldr version dependent. When a new version of ex_cldr is installed, no locales are installed and therefore locales are downloaded at compilation time as required. This ensures that the right version of the locale data is always associated with the right version of ex_cldr. However, if locale data is being cached in CI/CD there is some possibility that there can be a version mismatch. Since reproducible builds are important, setting the force_locale_download: true in a backend or in global configuration adds additional certainty. The default setting is false thereby retaining compatibility with existing behaviour. 

But as @brianmay spotted, we skip the download in our docker build, see https://github.com/teslamate-org/teslamate/blob/3ba23da7a7b1c1c9e5a3cecaf7eb029328067797/Dockerfile#L46

So I lean to disable it for your nix build, but I'm not 100% certain about this.

brianmay:
[...]

PR #4763, opened 2025-05-26
fix(nix): skip duplicate cldr download to fix build on darwin and improve reproducibility

JakobLichterfeld:
fixes: #4761

### DIFF
--- a/nix/flake-modules/package.nix
+++ b/nix/flake-modules/package.nix
@@ -65,6 +65,9 @@
           mixFodDeps
           ;
 
+        # set the environment variables for the build
+        mixEnv = "prod"; # build for production
+        SKIP_LOCALE_DOWNLOAD = true; # do not download locales during build as they are already included in the cldr package from github
         LOCALES = "${cldr}/priv/cldr";
 
         postBuild = ''
--- a/nix/flake-modules/package.nix
+++ b/nix/flake-modules/package.nix
@@ -66,8 +66,8 @@
           ;
 
         # set the environment variables for the build
-        mixEnv = "prod"; # build for production
-        SKIP_LOCALE_DOWNLOAD = true; # do not download locales during build as they are already included in the cldr package from github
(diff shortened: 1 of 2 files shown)


nix/flake-modules/package.nix
```diff
@@ -65,6 +65,9 @@
           mixFodDeps
           ;
 
+        # set the environment variables for the build
+        MIX_ENV = "prod"; # build for production
```
JakobLichterfeld, line 69: Not sure if we should explicitly set it, as we use it in our devenv as well, innit?

brianmay:
Am I bit unclear, is this still not working?

Anyway, [docs](https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/beam.section.md#mixrelease---example-mix-release-example) have:

```
  # if using mix2nix you can use the mixNixDeps attribute
  mixFodDeps = packages.fetchMixDeps {
    pname = "mix-deps-${pname}";
    inherit src version;
    # nix will complain and tell you the right value to replace this with
    hash = lib.fakeHash;
    mixEnv = ""; # default is "prod", when empty includes all dependencies, such as "dev", "test".
    # if you have build time environment variables add them here
    MY_ENV_VAR = "my_value";
  };
```

I think this makes in clear that mixEnv="prod" is the default.

And not sure I entirely understand, but it suggests `SKIP_LOCALE_DOWNLOAD` should be set inside the `fetchMixDeps` block. If this was still not working, that might be the solution.

nix/flake-modules/package.nix
brianmay: The devenv environment won't be used during package builds.

JakobLichterfeld:
> The only thing here I am not sure of is setting `MIX_ENV` explicitly to prod, I think `mixEnv` is the required setting, and it already defaults to prod anyway.

I agree, my bad, and I've removed it.

brianmay (approved):

PR #4764, opened 2025-05-26
feat: use the k8s-style API introduced in Grafana v12 if using manual dashboard setup

IngmarStein:
Ensure that folders are created before restoring dashboards.

Change authentication from user/password to the recommended API token.

Note: the on-disk format will change in the next backup/restore roundtrip because it now includes both the dashboard spec as well as metadata. The restore function reads both formats.


grafana/dashboards.sh
```diff
@@ -36,82 +61,346 @@ DASHBOARDS_DIRECTORY: $DASHBOARDS_DIRECTORY
...
+	mkdir -p "$DASHBOARDS_DIRECTORY" # Ensure base directory exists
 
-		if [[ -z $dashboard_json ]]; then
-			echo "ERROR:
-  Couldn't retrieve dashboard $dashboard.
-      "
-			exit 1
+	list_dashboards | while IFS= read -r dashboard_info_json; do
+		dashboard_uid=$(echo "$dashboard_info_json" | jq -r '.uid')
+		dashboard_title=$(echo "$dashboard_info_json" | jq -r '.title')
```
brianmay, line 79: Maybe I am being excessively paranoid. But the lack of double quotes around many of these variables in a shell script makes me nervous.

```sh
dashboard_title="$(echo "$dashboard_info_json" | jq -r '.title')"
```

Otherwise if the title happens to have spaces it could cause problems.

IngmarStein: Looking at `dutch-tax.json`: it has the title "Drives - Dutch tax". This results in:

```
BACKED UP: ./grafana/dashboards/reports/drives---dutch-tax.json (UID: lBIoQIggk)
```

So it's not the same filename as before (that is the case for two dashboards: this one and `database-information.json` instead of `database-info.json`), but it doesn't look like there's a problem with spaces, does it?

brianmay: Yes, might be OK. Still makes me somewhat nervous though.

IngmarStein: No problem, I've added quotes.

IngmarStein:
@brianmay given the staleness notification above, should I keep this PR open or close it?

brianmay:
Sorry, was this waiting for me? I just approved it.

IngmarStein:
Apologies, @brianmay. I tagged you only because you reviewed the change earlier and I wanted to check whether there's something missing. It's probably up to @JakobLichterfeld now?

JakobLichterfeld:
Hi Ingmar,

> should I keep this PR open or close it?

> **It could** be closed if no further activity occurs.

Does mean no automatic closing will happen for PRs. It is more like an indication, if something is missing and the contributor is not responding to questions, findings etc.

So in this case it can be ignored.

I have spent time touching the grass, so I have not yet merged this PR. Currently, I won't merge it, as I cannot foresee the implications of '# Make sure to guide the user to replace this', for example. Could you please elaborate on this a bit? In our Nix installation, we don't use the `dashboard.sh`, so this won't be affected, will it?

IngmarStein:
> I cannot foresee the implications of '# Make sure to guide the user to replace this', for example.  Could you please elaborate on this a bit?

Sure. I've replaced this comment with a link to instructions how to create an API token.
The difference to the status quo is that the token _must_ be specified. The current version "just works" for users who did not change the default credentials ("admin:admin").

> In our Nix installation, we don't use the dashboard.sh, so this won't be affected, will it?

Afaik, that installation method doesn't use this script so won't be affected.

JakobLichterfeld (approved):

PR #4768, opened 2025-05-27
docs: add NixOS installation guide

JakobLichterfeld:


brianmay:
My only concern is we say "If you intend to expose TeslaMate directly to the internet use a reverse proxy like [Caddy](https://nixos.wiki/wiki/Caddy)." - but then the Caddy instructions don't actually say how to secure the endpoint.

brianmay:
Also not sure if we need to mention this will conflict with any existing postgres/grafana servers (shame the nixos modules don't support multiple instances actually).

PR #4781, opened 2025-06-01
Add Google Maps link for car's current location

clhuang:
- Add Google Maps link in car summary view
- Link opens car's current location in new tab

### DIFF
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -23,7 +23,18 @@
       disabled
     />
   </div>
-  <div class="card-content">
+  <div class="card-footer">
+    <div class="card-footer-item">
+      <%= link("Open in Google Maps",
+        to: "https://www.google.com/maps?q=#{@summary.latitude},#{@summary.longitude}",
+        target: "_blank",
+        rel: "noopener noreferrer",
+        class: "button is-small is-link is-outlined"
+      ) %>
+    </div>
+  </div>
+  <div class="card-content"></div>
+
     <div class="media is-flex mb-5">
       <div class="media-content">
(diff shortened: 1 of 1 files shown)


cwanja:
Curious. What is the benefit to this? With the OpenStreet view?

clhuang:
allows easier navigation to the car

clhuang:
OpenStreet view doesn't provide an easy way to get the car's actual location, you can only guess based on visible nearby landmarks

cwanja:
More dumb questions. Why are you not using the native Tesla app's location to get navigation details to your parked cars?

clhuang:
It's less responsive than teslamate, and sometimes I like to look at it on a computer

JakobLichterfeld:
Thanks for your contribution. 
I do not really know if we should include this. I do see the use case for this, but on the other hand, Google Maps is closed source.

JakobLichterfeld:
If we decide to implement it, the test needs to include this new link --> I changed to  draft as the PR is not completed

clhuang:
if you don't want a link to google maps, would an additional text field with the coordinates be acceptable?

JakobLichterfeld (approved):

JakobLichterfeld:
I fixed your linting issues, please update the pot files: https://docs.teslamate.org/docs/development#update-pot-files-extract-messages-for-translation

sdwalker:
image

Was there a reason for going with a button compared to a pin icon next to the other status icons?

JakobLichterfeld:
> Was there a reason for going with a button compared to a pin icon next to the other status icons?

Thank you for your comment and for pointing me in this direction. The UX for this is really poor (additionally it shrinks the map). Need to be improved or reverted.

JakobLichterfeld:
-> #4837

PR #4787, opened 2025-06-08
fix: align usage of time-range filters to fix consumption gross issues in some edge cases

swiffer:
aligns usage of filters used in consumption gross queries with other panels to avoid some edge cases.

as all other queries are using date filters on start_date of a drive let's ensure we are doing the same for queries relying on positions / advanced logic.

before this change - selecting a period with a start date at the time of driving (actual drive start out of selected period) consumption gross was off.

grafik

after this change - numbers to align correctly

grafik


JakobLichterfeld (approved):
lgtm. Thanks so much!

PR #4795, opened 2025-06-10
feat(db-check): allow beta / rc PostgreSQL versions for testing

swiffer:
fix parsing of prelease versions to allow testing them

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - feat: use Grafana 12.0.2 (#4805 - @swiffer)
 - feat(mqtt): always publish healthy status and disable retain to prevent stale healthy status via mqtt (#4817 - @allivshits)
 - feat: use the k8s-style API introduced in Grafana v12 if using manual dashboard setup (#4764- @IngmarStein)
+- feat(db-check): allow beta / rc PostgreSQL versions for testing (#4795 - @swiffer)
 
 #### Build, CI, internal
 
--- a/lib/teslamate/database_check.ex
+++ b/lib/teslamate/database_check.ex
@@ -2,70 +2,71 @@ defmodule TeslaMate.DatabaseCheck do
   alias Ecto.Adapters.SQL
   alias TeslaMate.Repo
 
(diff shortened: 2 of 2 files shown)


swiffer:
should be good now 🤔

swiffer:
@JakobLichterfeld - could you have a look?

JakobLichterfeld (approved):

JakobLichterfeld (approved):

JakobLichterfeld:
The use of Regex.run with direct pattern matching on the result list (e.g., `[_, major, minor, patch, pre]`) is concise, but it introduces a few potential concerns:
- If any of the optional capture groups (minor, patch, or pre) are not matched, their positions in the list are filled with nil. This means the pattern match only works reliably when all elements (including nils) are present in the expected order.
- This introduces an implicit dependency on the exact number and order of matches returned by the regex, which can become fragile if the pattern is ever modified.

That’s why I was considering something like this:

```elixir
defp normalize_version(version) do
  case Regex.run(~r/^(\d+)(?:\.(\d+))?(?:\.(\d+))?((?:beta|rc)\d+)?$/, version) do
    [_, major, minor, patch, pre] ->
      "#{major}.#{minor || "0"}.#{patch || "0"}" <> if pre, do: "-#{pre}", else: ""
    nil ->
      raise "Invalid PostgreSQL version format: #{version}"
  end
end
```

I also wanted to introduce a test case first to verify this.
Unfortunately, since the function is declared as `defp`, it’s not directly accessible in the test suite.

We can either make it `def` temporarily (just for testing), or extracting it to a small utility module.

swiffer:
@JakobLichterfeld - rewrite making use of server_version_num

- https://www.postgresql.org/docs/current/runtime-config-preset.html#GUC-SERVER-VERSION-NUM
- https://pgpedia.info/s/server_version_num.html
- https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION

JakobLichterfeld (approved):
Perfect, ty!

PR #4798, opened 2025-06-13
feat: add total ascent & descent to the drive table, introduce slope-adjusted efficiency metric in drives dashboard

swiffer:
this PR adds the total ascent & descent to the drive table. data is calculated when closing a drive based on positions data.

having data within drive table allows for performant calculation of a new "slope-adjusted efficiency" metric as proposed in: https://github.com/teslamate-org/teslamate/discussions/4789

the drives dashboard has been modified:
  - additional variable (Efficiency) with two options (by distance, slope-adjusted)
  - slope-adjusted takes ascent & descent into account when calculating efficiency of the drive
    vehicle mass has been hard-coded to 2100kg, regen breaking efficiency is hard-coded to 85%


swiffer:
@JakobLichterfeld - could you review this one, not so sure with the elixir related changes. once merged into main i'll work on the grafana related changes. in a follow-up PR.

swiffer:
went ahead and added the metric as well - that way it's one full PR to review and the outcome can be seen already.

swiffer:
just some thoughts

- having the ascent / descent available per drive allows looking at additional statistics (e.g. how is average elevation influencing consumption (wh/km)) - similar to what we report in the efficiency dashboard.

- adding the info to drives for performant analysis was important as positions is only holding elevation info for streamed data, it's missing in non streamed data sampled every 15 sec. looking at elevation changes on 15s granularity would be sufficient but as it's not available seq scan on positions would be needed.

- also do like the visualizations in TeslaMateCustomDashboards (https://github.com/jheredianet/Teslamate-CustomGrafanaDashboards?tab=readme-ov-file#speed--temperature) - something we could do for bucketed elevation (up/down) as well.

brianmay:
Assuming it is all tested and working, including the migration, it looks good to me.

One possible concern, wondering how long that migration is likely to take to run if you have lots of data. People might panic and think something has gone wrong if it takes too long... If it is indeed to long, might be better to have some way to do the data migration by hand.

Something also to think about, what happens if you have to revert to a previous version for some reason? I believe this should be OK:

* New db fields are nullable (assuming this deliberate) and will get null values.
* When moving back to a new version the new fields will still be null, and have to be updated by running the migration by hand.

swiffer:
@JakobLichterfeld - do you still have slow hardware to test?

migration was quickly executed here. we might could add it to the release notes / upgrade guide again.

JakobLichterfeld:
> @JakobLichterfeld - do you still have slow hardware to test?

Not in production, but planning to test it there, that's why I didn't merged the PR yet.

JakobLichterfeld:
> One possible concern, wondering how long that migration is likely to take to run if you have lots of data. People might panic and think something has gone wrong if it takes too long... If it is indeed to long, might be better to have some way to do the data migration by hand.

Test on slow HW (Raspberry Pi 3 B+, DB size ~ 1GB)

```bash
teslamate               | 2025-07-17 11:57:28.122 [info] == Running 20250613133700 TeslaMate.Repo.Migrations.AddAndCalculateElevationChanges.up/0 forward
teslamate               | 2025-07-17 11:57:28.133 [info] alter table drives
teslamate               | 2025-07-17 11:57:28.222 [info] execute "WITH elevation_changes AS (\n  SELECT\n    drive_id,\n    SUM(CASE WHEN elevation_diff > 0 THEN elevation_diff ELSE 0 END) as ascent,\n    SUM(CASE WHEN elevation_diff < 0 THEN ABS(elevation_diff) ELSE 0 END) as descent\n  FROM (\n    SELECT\n      drive_id,\n      elevation - LAG(elevation) OVER (PARTITION BY drive_id ORDER BY date) as elevation_diff\n    FROM positions\n    WHERE elevation IS NOT NULL\n  ) as changes\n  GROUP BY drive_id\n)\nUPDATE drives d\nSET\n  ascent = ec.ascent,\n  descent = ec.descent\nFROM elevation_changes ec\nWHERE d.id = ec.drive_id;\n"
(...)
teslamate               | 2025-07-17 11:58:40.539 [info] == Migrated 20250613133700 in 72.3s
```

--> Migration took 72.3 seconds, imo a little borderline but still manageable. What do you think, @brianmay and @swiffer ?

swiffer:
Looks good to me - what could be added is some log output or a hint to the next release notes.

JakobLichterfeld:
> or a hint to the next release notes.

Yeah, I would definitely include it in the Changelog for 2.1.0 (increase Minor). Edit: done

slashhesham:
I'm getting an error in this dashborad:

db query error: pq: column "ascent" does not exist

JakobLichterfeld:
> I'm getting an error in this dashborad:
> 
> db query error: pq: column "ascent" does not exist

Interesting, with TeslaMate v2.1.0? Did the migration run successfully? If so, please open an Issue, as I can't reproduce, nor have all the info about your setup, host etc.

slashhesham:
> > I'm getting an error in this dashborad:
> > db query error: pq: column "ascent" does not exist
> 
> Interesting, with TeslaMate v2.1.0? Did the migration run successfully? If so, please open an Issue, as I can't reproduce, nor have all the info about your setup, host etc.

Thanks, I just did a docker pull and it's now working.

PR #4799, opened 2025-06-13
Grafana 12.0.1+security-01

swiffer:


JakobLichterfeld (changes requested):
lgtm, ty!

May you elaborate on the changes in the visited dashboard?

swiffer:
I don't know why - saved the dashboard in the latest version and that was the result...

PR #4806, opened 2025-06-21
Revert "docs: Update README.md to fix typo"

JakobLichterfeld:
Reverts teslamate-org/teslamate#4804

as sneak peak is actually correct

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Sample screenshots of bundled dashboards can be seen by clicking the links below
 
 ## Screenshots
 
-Sneak peek into TeslaMate interface and bundled dashboards. See [the docs](https://docs.teslamate.org/docs/screenshots) for additional screenshots.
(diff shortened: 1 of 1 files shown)


brianmay:
Are you sure sneak peak is correct?

Here is a cartoon I found:

https://theoatmeal.com/comics/sneak_peek

JakobLichterfeld:
At least all Tools told it. Originally I know it from cinema.

brianmay:
Both spellings can be correct, just the meaning is different.

JakobLichterfeld:
Yeah, and in tech cases, sneak peak seems to be common sense. But we can avoid both and just call it snapshots or first look.


>Sneak peek” has evolved beyond its literal meaning. While it originally referred to a glimpse of something not yet released or still under wraps, it's now widely used to describe a short, enticing preview — especially of something visual, like a UI or a product feature.

>In this context, even though TeslaMate is publicly available, calling it a “sneak peek” still makes sense. It signals that you’re offering a quick, curated glimpse into the interface, highlighting the look and feel before users dive deeper themselves.

>It's less about secrecy, and more about creating interest and a sense of discovery.

PR #4817, opened 2025-07-02
feat(mqtt): always publish healthy status and disable retain to prevent stale healthy status via mqtt

allivshits:
This update improves MQTT health signaling by ensuring the :healthy topic is always published and never retained.

🔧 Key changes:

1. Introduced @do_not_retain for topics like :healthy that:
    - Must be published on every poll (even if value unchanged)
    - Must not be retained by the MQTT broker
2. Renamed @always_published to @publish_if_nil for clarity

✅ Benefit:
Prevents stale healthy: true messages from giving a false sense of system status. This makes TeslaMate easier to monitor using MQTT liveness tools like Home Assistant’s expire_after.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - feat: preserve the folders when restoring dashboards (#4715 - @IngmarStein)
 - feat: use Grafana 12.0.1+security-01 (#4799 - @swiffer)
 - feat: use Grafana 12.0.2 (#4805 - @swiffer)
+- feat(mqtt): always publish healthy status and disable retain to prevent stale healthy status via mqtt (#4817 - @allivshits)
 
 #### Build, CI, internal
 
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -22,6 +22,19 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
     GenServer.start_link(__MODULE__, opts)
   end
 
+  @do_not_retain ~w(healthy)a
+
+  defp clear_retained(car_id, namespace, publisher) do
+    for key <- @do_not_retain do
+      topic =
+        ["teslamate", namespace, "cars", car_id, key]
+        |> Enum.reject(&is_nil(&1))
+        |> Enum.join("/")
+
+      call(publisher, :publish, [topic, "", [retain: true, qos: 1]])
+    end
+  end
+
   @impl true
   def init(opts) do
     car_id = Keyword.fetch!(opts, :car_id)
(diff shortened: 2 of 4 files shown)


brianmay:
I agree, sounds like we should not be using retain here.

Also sounds like we should be using the LWT (Last Will and Testament) feature of MQTT that can automatically publish an unhealthy message if we get disconnected from MQTT. I don't think we do that, although only had a quick glance to check.

See:https://hexdocs.pm/tortoise/connecting_to_a_mqtt_broker.html#last-will-message.

Not tested :-)

brianmay:
With the LWT support, disabling retain may not be so important. But have no firm opinions here. I guess Teslamate could stop working but still be connected to MQTT.

brianmay:
Is publishing a message without retain sufficient to delete the retain message? I have a vague feeling the previous retained message might stay around (which has caused endless confusion to me in the past), but not 100% sure here right now.

allivshits:
> I have a vague feeling the previous retained message might stay around (which has caused endless confusion to me in the past), but not 100% sure here right now.

you're right. i'll update the code to cleanup old retained values

allivshits:
As for LWT, I considered it. My take - it's only going to help with crash, etc. If docker was properly shut down - it'll not trigger so broker will retain healthy=true.

JakobLichterfeld:
Thanks for your PR. The stale unhealthy message is indeed unfortunate.

Only source of health of TeslaMyte is the url endoint`<your ip>:3000/health_check` (see discussion here: https://github.com/teslamate-org/teslamate/discussions/4782)

MQTT health does not have any info about TeslaMate health in general. Monitoring TeslaMate via MQTT is not advised at all.

allivshits:
re: @JakobLichterfeld 
> MQTT health does not have any info about TeslaMate health in general. Monitoring TeslaMate via MQTT is not advised at all.

With this update, it may be worth rethinking the advice against MQTT monitoring. With this change TeslaMate publishes a dedicated healthy topic on every poll, you can rely on a single MQTT sensor (in case of Home Assistant with expire_after) for liveness checks — no need to juggle an HTTP health_check endpoint and a separate MQTT 'healthy' sensor setup + implementing custom logic to merge their states.

JakobLichterfeld:
> With this update, it may be worth rethinking the advice against MQTT monitoring.

e.g. TeslaMate runs without exposing anything via MQTT, for example by config, or because of misconfigured MQTT broker or unhealthy Mosquito docker image. So a MQTT health topic does not say anything about TeslaMate health in general. For sure you can use the MQTT topic, but it will always be misleading and only act as an indicator for a small portion of it.

allivshits:
Seems like all the checks are passing now. I've been also using this locally - works as described. @brianmay and @JakobLichterfeld anything else I can do to get PR accepted?
PS: thnx for the feedback and relevant context information!

brianmay (approved):

PR #4827, opened 2025-07-16
chore(issue-template): enhance bug report template with additional fields and guidance

JakobLichterfeld:
- Added a dropdown for reproducibility confirmation.
- Improved descriptions and labels for clarity.
- Updated installation options with links and support notes.
- Included reverse proxy details for better issue context.

Issue #4825, opened 2025-07-15
If blocking third party network traffic the Maps on Home Page don't show as expected 


SemoTech:
> @SemoTech Are you accessing it through a proxy? If so, did you redirect `live/websocket` as descibed in https://docs.teslamate.org/docs/guides/apache 

 _Originally posted by @DrMichael in [#4824](https://github.com/teslamate-org/teslamate/issues/4824#issuecomment-3072523562)_

Hello @DrMichael and thanks for trying to help.  Unfortunately I was unable to reply to you in the original thread since @JakobLichterfeld locked it to Collaborators only, so I had to create this follow-up thread. 

Yes, I am using a proxy, Caddy V2 Proxy to be exact, since it handles Let's Encrypt directly.  Below are my _sanitized_ Docker-Compose.yaml and the relevant part of the Caddy configuration for Teslamate files.  
Caddy needs WAN access to get and validate the SSL but will only allow access to Teslamate and Grafana on their internal 3000 and 4000 ports to requests from the LAN, and any WAN access is redirected to the FBI :-) 

The interesting thing is that I have restarted and pulled the container image countless times and Maps have no issues under Grafana Dashboards. This has been my setup for a very long time. 

Only after TM v2.0.0 came out did I notice the Home page Maps stopped working, so it's hard to believe this is a configuration issue on my part, but I am happy to troubleshoot further if you have any suggestions...

Docker-Compose.yaml Code for Teslamate:

`services:
  teslamate:
    container_name: teslamate
    image: teslamate/teslamate:latest
    restart: unless-stopped
    environment:
      - ENCRYPTION_KEY=${TM_ENCRYPTION_KEY}
      - DATABASE_USER=${TM_DB_USER}
      - DATABASE_PASS=${TM_DB_PASS}
      - DATABASE_NAME=${TM_DB_NAME}
      - DATABASE_HOST=database
      - MQTT_HOST=mosquitto  
      - VIRTUAL_HOST=${FQDN_TM}
      - CHECK_ORIGIN=true
      - NTP_SERVERS=${TM_NTP}
      - TZ=${TM_TZ}
    labels:
      - "com.centurylinklabs.watchtower.enable=true"
    networks:
      tm-net:
      caddy-net:
    volumes:
      - ${DATA_PATH}/import:/opt/app/import
    cap_drop:
      - all

  database:
    container_name: teslamate-postgres
    image: postgres:16
    restart: unless-stopped
    environment:
[...]

PR #4829, opened 2025-07-16
docs: streamline installation guides

JakobLichterfeld:
Doc improvements, as discussed in #4827:

- [x] clearly state which installation methods are supported and which not (last mentioned here: #4825)
  - [X] update Issue Template accordingly 
- [x] removing the information about manual installation steps from the Docker installation guide
- [x] rephrase reverse proxy and hardening
- [X] rename guides and move the Fleet API Docs to Advanced Configuration instead
- [X] add nix develop to run treefmt to support non nix direnv users
- [x] include #4758 
- [X] mention dedicated version for MCU2 upgraded cars in docker install as well


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -122,6 +122,7 @@ body:
         - Docker (https://docs.teslamate.org/docs/installation/docker/)
         - NixOS (https://docs.teslamate.org/docs/installation/nixos)
         - Manual (no support)
+        - Unraind (no support, if not using docker compose)
```
swiffer, line 125: Unraid

swiffer (commented):
spotted that typo - otherwise, thanks! great work

PR #4830, opened 2025-07-16
docs: Updated documentation for the manually_fixing_data.mdx

zhaoyi48:
Correct quote format to resolve "-bash: !: event not found" error.


JakobLichterfeld:
Thanks for your contribution!

The quoting is correct and battled tested.

If you encounter errors, its due to copy pase issue or encoding issues in your terminal

Issue #4706, opened 2025-04-29
need to simplify manual process of manually closing drives under nixos


brianmay:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

I think this process could be simplified. Not sure yet what to do, just want to document what I had to do for now. I don't often have to do this often, so it gets irritating that I have to try to remember what to do each time.

Find the un-closed drives/charges:

```console
[root@iot2:~]# psql --user postgres teslamate
psql (16.8)
Type "help" for help.

teslamate=# select id from charging_processes where end_date is null and start_date <  now() +
 interval '-2 days' order by start_date;
 id
----
(0 rows)

teslamate=# select id from drives where end_date is null and start_date <  now() + interval '-2
 days' order by start_date;
  id
------
 8770
 9500
 9864
(3 rows)

teslamate=#
\q
```

Find the executable:

```console
[root@iot2:~]# cat /etc/systemd/system/teslamate.service  | grep ExecStart
ExecStart=/nix/store/wf5rsxwgw4a0h387cx0ya69vwms507db-teslamate-2.0.0/bin/teslamate start
ExecStartPre=/nix/store/wf5rsxwgw4a0h387cx0ya69vwms507db-teslamate-2.0.0/bin/teslamate eval "TeslaMate.Release.migrate"
```

Here the executable is /nix/store/wf5rsxwgw4a0h387cx0ya69vwms507db-teslamate-2.0.0/bin/teslamate.

Attempt to use this:

```console
[root@iot2:~]# /nix/store/wf5rsxwgw4a0h387cx0ya69vwms507db-teslamate-2.0.0/bin/teslamate  rpc "TeslaMate.Repo.get(TeslaMate.Log.Drive,9864) |> TeslaMate.Log.close_drive()"
cat: /nix/store/wf5rsxwgw4a0h387cx0ya69vwms507db-teslamate-2.0.0/releases/COOKIE: No such file or directory
```

This isn't going to work, because we need the cookie.

Lookup Cookie value and use:

```
[root@iot2:~]# cat /etc/systemd/system/teslamate.service  | grep EnvironmentFile
EnvironmentFile=/run/agenix/teslamate

[root@iot2:~]# cat /run/agenix/teslamate
[censored]
RELEASE_COOKIE=[censored]

[...]

PR #4849, opened 2025-07-23
feat(nix): add idiomatic maintenance scripts

JakobLichterfeld:
closes: #4706

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 #### Documentation
 
 - docs: add reference to idiomatic nix backup and restore scripts (#4849 - @JakobLichterfeld)
+- docs: add reference to idiomatic nix maintenance scripts (#4849 - @JakobLichterfeld)
 
 ## [2.1.0] - 2025-07-21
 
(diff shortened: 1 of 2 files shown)


brianmay:
I was just thinking something like this would be really good :-)

JakobLichterfeld:
Close drive tested successfully with an already closed drive(In my case I needed to write my RELASE_COOKIE like='...' to prevent issues due to special characters)

```bash
[root@MainServer:/etc/nixos]# teslamate-close-drive aa
Error: Drive ID must be an integer.

[root@MainServer:/etc/nixos]# teslamate-close-drive 3033
Attempt to close the drive with ID 3033.

18:34:44.601 [info] GET https://nominatim.openstreetmap.org/reverse -> 200 (224.591 ms)

18:34:45.569 [info] GET https://nominatim.openstreetmap.org/reverse -> 200 (961.187 ms)
```

brianmay (approved):

JakobLichterfeld:
Thanks, will add some small hints in the docs

PR #4870, opened 2025-08-05
feat(nix): use datasources.settings.datasources to allow merging Grafana sources from multiple modules, correctly set default Grafana dashboard path and allow disabling default Grafana dashboard with setDefaultDashboard option

JakobLichterfeld:
- [X] feat(nix): use datasources.settings.datasources to allow merging Grafana sources from multiple modules
- [X] fix(nix): correctly set default Grafana dashboard path
- [X] feat(nix): allow disabling default Grafana dashboard with setDefaultDashboard option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - feat(webview): show offline duration in extended view on mobile as well to improve UX (#4848 - @JakobLichterfeld)
 - feat: use Grafana 12.1.0 (#4855 - @swiffer)
+- feat(nix): use datasources.settings.datasources to allow merging Grafana sources from multiple modules (#4870 - @JakobLichterfeld)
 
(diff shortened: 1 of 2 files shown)


brianmay (approved):
Assuming it is tested and works, looks good to me.

brianmay:
I really wish the grafana nixos module supported multiple instances of grafana. Grafana isn't good for the case when users should not have access to all data (e.g. granting every family member complete history of car location is likely to lead to undesirable outcomes).

JakobLichterfeld:
> Assuming it is tested and works, looks good to me.

Yeah it is.

> I really wish the grafana nixos module supported multiple instances of grafana.

Yes, same here. At least, this change makes it possible to use additional data sources and dashboards alongside TeslaMate.

PR #4882, opened 2025-08-09
set descent and ascent to 0 when out of bounds (> 32767 meters)

### LINKED ISSUE #4876: Migration to add total ascent & descent to the drive table should be improved

### What happened?

#4798 added total ascent & descent to the drive table via Migration: [20250613133700_add_and_calculate_elevation_changes ](https://github.com/teslamate-org/teslamate/blob/92c7a6a36d55f9592933aef3b299ea71cf40b78d/priv/repo/migrations/20250613133700_add_and_calculate_elevation_changes.exs)

This was tested with > 400 000 km of recorded data before releasing.

Due to GPS issues, one user had more than 32 767 m in elevation (in his case 41 473 m), which resulted in the migration to fail with  `Postgrex.Error) ERROR 22003 (numeric_value_out_of_range) smallint out of range` (#4868 and #4871).

Edit: more users #4847, #4877

### Workaround (edited)

Downgrade to v2.0.0

### Possible solutions:

1. Using integer instead of smallint
2. Clamping:
```sql
ascent = LEAST(ec.raw_ascent, 32767),
descent = LEAST(ec.raw_descent, 32767)
```

### Open Points

- [ ] Migration adjustment to avoid such issues
- [ ] Avoid overflow of `ascent`and `descent` in `drives` table during normal operation
- [ ] automated test

### Expected Behavior

- Migration succeed even if users messed with their DB or had outstanding GPS issues.
- No overflow during normal data logging

Yes

GPS elevation issues during a drive.

### Relevant log output

```shell
n.a.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

v2.1.0

### PostgreSQL version

_No response_

- [X] I run the latest major supported PostgreSQL version

No

n.a.

swiffer:
fixes #4876

i for now opted into setting the elevation to 0 if out of bounds.

changing the migration that way is acceptable imho as outcome is only changing in cases the migration did not succeed anyhow before the change.

leaves up the question if smallint was reasonable choice but driving more than 32767 without charging in between seems unlikely.
if we want to change column type it needs to be done in another migration anyhow.


JakobLichterfeld:
Ty! It wasn't meant that you need to fix it.

> i for now opted into setting the elevation to 0 if out of bounds. doing so ensures dashboard queries continue to work without any changes.

Great choice imo.

> leaves up the question if smallint was reasonable choice but driving more than 32767 without charging in between seems unlikely.

Imo it was reasonable, pikes peak 22,75 time in a row without charging would be unbelievable :-) 

Ty for fixing the compile issues. I will test next week and then release the bugfix version.

USAFPride:
I've noticed when starting a drive inside a structure with no GPS signal, that the start point and elevation are significantly off.  In my case, the elevation was 40k feet, which last I checked was impossible.  There should be some error checking to prevent recording any data if the elevation is over x' (or meters)

JakobLichterfeld:
> There should be some error checking to prevent recording any data if the elevation is over x' (or meters)

Thats what is added in this PR, when over the smallint limit it will be set to 0

USAFPride:
> > There should be some error checking to prevent recording any data if the elevation is over x' (or meters)
> 
> Thats what is added in this PR, when over the smallint limit it will be set to 0

Is it just the elevation data set to 0? Or all data related to that reading at that specific point in time?  IMO, it should be the later with the all the retrieved data discarded

JakobLichterfeld:
> Is it just the elevation data set to 0?

It is "only" the elevation data that would otherwise cause an overflow. 

> Or all data related to that reading at that specific point in time?

It is difficult to say with 100% certainty whether the data is invalid.

> with the all the retrieved data discarded

This would invalidate the term 'data logger'. :-)

swiffer:
Data logged in positions remains complete untouched. This is just setting the new aggregated columns calculated at the end of a drive to 0 if unplausible.

JakobLichterfeld (approved):
lgtm and tested on my end. I added a comment, as smallint goes to 32767 and on first look, code is not as easy to read.

JakobLichterfeld:
still one user with issues, #4895 -> #4896

PR #4898, opened 2025-08-18
feat(webview): add a log view to simplify log access and support

### LINKED ISSUE #4891: feat(webview): add a log view to simplify log access and support
_As a user_, accessing TeslaMate logs can be difficult. Providing simple access to logs from Webview could improve self-support.

_As a maintainer_, it is time-consuming when users do not provide logs when reporting issues. If access to logs were easier, this would reduce the number of back-and-forth exchanges.

JakobLichterfeld:
- Introduced a new setting for enabling file logging in the global settings.
- Implemented a live view for displaying logs and a UI toggle for file logging.
- Updated router and layout to include the new logs page.
- Added migration for the new logging setting.

closes #4891

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -306,3 +306,13 @@ main.full-height {
     background-color: initial;
   }
 }
+
+.logs {
+  background-color: #f5f5f5;
+  border: 1px solid #ccc;
+  padding: 10px;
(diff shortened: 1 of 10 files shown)


brianmay:
My recent issue #4902 might be significant here.

PR #4896, opened 2025-08-18
fix(migration): simplify ascent and descent calculations and ensure smallint bounds are never exceeded

JakobLichterfeld:
fixes #4895

PR #4900, opened 2025-08-18
fix(migration): cast final result to smallint

swiffer:
The SUM function in the CTE produces a NUMERIC or BIGINT result, and even though you cap it with LEAST, the final value may not be explicitly cast to SMALLINT before being assigned to the drives table's ascent and descent columns

### DIFF
--- a/lib/teslamate/log.ex
+++ b/lib/teslamate/log.ex
@@ -304,13 +304,13 @@ defmodule TeslaMate.Log do
            select: %{
              elevation_gains:
                fragment(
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
As the issues was not solved, neither with this (see https://github.com/teslamate-org/teslamate/pull/4896#issuecomment-3200622061), or with #4896, I would close both PRs. What do you think?

swiffer:
Agreed

PR #4920, opened 2025-08-26
Update italian translation

giovaorama:
Hi,
I update the translation for Italian.
In details:
- add some missing translation
- correct the gender of some words
- use more uniform translations


giovaorama:
> Ty for your contribution!

Should I correct something? I saw that the some test are failed

JakobLichterfeld:
> Should I correct something? I saw that the some test are failed

I don't think so, this looks like the random test failure due to too many tests run in parallel. I will re trigger the test.

JakobLichterfeld:
Now all tests are fine

PR #4956, opened 2025-09-15
i18n: add thai car location translation

tomzt:
Update default.po for Thai language


JakobLichterfeld:
Ty for your contribution!

Issue #4902, opened 2025-08-18
should not log tesla token


brianmay:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

I just updated to the latest mcu2-upgraded-cars, restarted, and while looking at my logs saw my access token staring back at me. This is not good from a security point of view, as logs are typically stored non-encrypted in plain text.

See the `[...XXX...]` in the logs below. Curiously it only happened once, the subsequent 408 failures did not do this.

The `reason: :unknown` might be significant.

I have a vague recollection that Elxir/Pheonix might have a way to filter secrets, but I can't recall right now off the top of my head.

The errors are probably because the car is asleep when I restarted; will keep an eye on this just to make sure.

### Expected Behavior

Secrets need to be filtered before logging.

### Can the issue reliably be reproduced?

No

### Steps To Reproduce the issue

No idea.

### Relevant log output

```shell
Aug 19 07:48:29 iot2 systemd[1]: Starting TeslaMate...
Aug 19 07:48:31 iot2 teslamate[293433]: 2025-08-19 07:48:31.491 [info] Migrations already up
Aug 19 07:48:31 iot2 systemd[1]: Started TeslaMate.
Aug 19 07:48:33 iot2 teslamate[293714]: 2025-08-19 07:48:33.710 [info] No tzdata release files found in custom data dir. Copying release file from tzdata priv dir.
Aug 19 07:48:33 iot2 teslamate[293714]: 2025-08-19 07:48:33.794 [info] System Info: Erlang/OTP 26 (jit)
Aug 19 07:48:33 iot2 teslamate[293714]: 2025-08-19 07:48:33.794 [info] Version: 2.1.1-mcu2-upgraded
Aug 19 07:48:33 iot2 teslamate[293714]: PostgreSQL version 16.9 is compatible (16.x series).
Aug 19 07:48:34 iot2 teslamate[293714]: 2025-08-19 07:48:34.818 [info] POST https://auth.tesla.com/oauth2/v3/token -> 200 (806.958 ms)
Aug 19 07:48:34 iot2 teslamate[293714]: 2025-08-19 07:48:34.818 [info] Refreshed api tokens
Aug 19 07:48:34 iot2 teslamate[293714]: 2025-08-19 07:48:34.868 [info] Scheduling token refresh in 6 h
[...]

Issue #4897, opened 2025-08-18
database "teslamate" has a collation version mismatch - after upgrading debian 12 to 13


kongomongo:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

Hi there, after upgrading from debian 12 to 13 I had the OOM repeatedly kill beam.smp and started investigating. I am running 3 instances of teslamate so I thought maybe its time to increase the 2GB RAM to 4DB. didnt help. So i stopped all instances except one.

Now I am seeing this in my log:

database-1   | 2025-08-18 11:57:57.244 UTC [77] WARNING:  database "teslamate" has a collation version mismatch
database-1   | 2025-08-18 11:57:57.244 UTC [77] DETAIL:  The database was created using collation version 2.36, but the operating system provides version 2.41.
database-1   | 2025-08-18 11:57:57.244 UTC [77] HINT:  Rebuild all objects in this database that use the default collation and run ALTER DATABASE teslamate REFRESH COLLATION VERSION, or build PostgreSQL with the right library version.

I was running the latest 1.x of teslamate (1.33.0), upgraded that to 2.1.1. Didnt help.

I am seeing no other issues so asking here. 

This might be related?
https://github.com/teslamate-org/teslamate/issues/3583

### Can the issue reliably be reproduced?

Yes

### Relevant log output

```shell
database-1   | 2025-08-18 11:57:57.244 UTC [77] WARNING:  database "teslamate" has a collation version mismatch
database-1   | 2025-08-18 11:57:57.244 UTC [77] DETAIL:  The database was created using collation version 2.36, but the operating system provides version 2.41.
database-1   | 2025-08-18 11:57:57.244 UTC [77] HINT:  Rebuild all objects in this database that use the default collation and run ALTER DATABASE teslamate REFRESH COLLATION VERSION, or build PostgreSQL with the right library version.
```

### Type of installation

Docker (https://docs.teslamate.org/docs/installation/docker/)

### Version

2.1.1

### PostgreSQL version

postgresql-16/now 16.10-1.pgdg13+1

### Are you running latest major supported PostgreSQL version?

- [ ] I run the latest major supported PostgreSQL version

### Are you using a reverse Proxy for TeslaMate?

No

kongomongo:
seems like that was causing the memory usage:

https://github.com/teslamate-org/teslamate/discussions/3045#discussioncomment-8757861

[...]

PR #4968, opened 2025-09-25
fix: tokens to private schema and readonly grafana user

brianmay:
This change allows configuring grafana with the db user that has restricted access rights to the database.

Note: User by default doesn't have any password set, admin needs to set the password after running migration.

This shouldn't affect the teslamate user which by default has full admin access.


brianmay:
This have been bothering me for a while.

But no effect unless user changes grafana config. Wonder if there is anyway to make this more idiot proof?

Can split this up into two if required.

JakobLichterfeld (approved):

JakobLichterfeld:
> This have been bothering me for a while.

Very nice and neat solution.
This will close #4902, innit?

> But no effect unless user changes grafana config. Wonder if there is anyway to make this more idiot proof?

@swiffer : Do you have any ideas on this?

brianmay:
I want to revise those permissions when more awake.

Ideally grafana user should have read-only access to only public schema on teslamate database.

Possible this is what I have got. I believe that requiring public schema implies requiring access to the database that owns the schema. But need to double check this.

brianmay:
This all looks good to me:

```
temp=> \conninfo
You are connected to database "temp" as user "grafana_user" on host "localhost" (address "127.0.0.1") at port "7000".
temp=> select * from public.students;
ERROR:  permission denied for table students
temp=> \c teslamate
Password:
You are now connected to database "teslamate" as user "grafana_user".
teslamate=> select * from public.
public.addresses                  public.charging_processes         public.settings
public.addresses_id_seq           public.charging_processes_id_seq  public.settings_id_seq
public.cars                       public.drives                     public.states
public.car_settings               public.geofences                  public.states_id_seq
public.car_settings_id_seq        public.geofences_id_seq           public.trips_id_seq
public.cars_id_seq                public.positions                  public.updates
public.charges                    public.positions_id_seq           public.updates_id_seq
public.charges_id_seq             public.schema_migrations
teslamate=> select * from public.cars;
 id | eid | vid | model | efficiency | inserted_at | updated_at | vin | name | trim_badging | settings_id | exterior_color | spoiler_type | wheel_type | display_priority | marketing_name
----+-----+-----+-------+------------+-------------+------------+-----+------+--------------+-------------+----------------+--------------+------------+------------------+----------------
(0 rows)

teslamate=> select * from public.token;
ERROR:  relation "public.token" does not exist
LINE 1: select * from public.token;
                      ^
teslamate=> select * from private.token;
ERROR:  permission denied for schema private
LINE 1: select * from private.token;
                      ^
teslamate=> \c temp
Password:
You are now connected to database "temp" as user "grafana_user".
temp=> select * from public.students;
ERROR:  permission denied for table students
```

temp with the students database was a dummy database I created for this test:

```
$ teslamate_psql
psql (17.5)
Type "help" for help.

teslamate=# create database temp;
ERROR:  source database "template1" is being accessed by other users
DETAIL:  There is 1 other session using the database.
teslamate=# create database temp;
CREATE DATABASE
teslamate=# \c temp
You are now connected to database "temp" as user "teslamate".
temp=# CREATE TABLE students (
    student_id SERIAL PRIMARY KEY,
    first_name VARCHAR(50) NOT NULL,
    last_name VARCHAR(50) NOT NULL,
    email VARCHAR(100) UNIQUE,
    enrollment_date DATE DEFAULT CURRENT_DATE
);
CREATE TABLE
temp=#
```

Initially I was concerned that I could log in as `grafana_user` with any password. I think this is just the devenv environment gives me a very open config using "trust":

```
$ cat .devenv/state/postgres/pg_hba.conf
[...]

# CAUTION: Configuring the system for local "trust" authentication
# allows any local user to connect as any PostgreSQL user, including
# the database superuser.  If you do not trust all your local users,
# use another authentication method.

# TYPE  DATABASE        USER            ADDRESS                 METHOD
# "local" is for Unix domain socket connections only
local   all             all                                     trust
# IPv4 local connections:
host    all             all             127.0.0.1/32            trust
# IPv6 local connections:
host    all             all             ::1/128                 trust
# Allow replication connections from localhost, by a user with the
# replication privilege.
local   replication     all                                     trust
host    replication     all             127.0.0.1/32            trust
host    replication     all             ::1/128                 trust
```

I think most people would never have this for anything other then development.

swiffer:
Like the idea of separating tokens in private schema and also restricting rights of grafana user to read-only on public (did the latter in my setup anyhow).

Not sure if roles should be part of database migrations - also consider that roles and users are not included in backups created via pg_dump - we either would need to rerun the migration on restore or change pg upgrade / backup docs.

Alternatively instructions could be part of a advanced / hardening guide? Still how do we guide users when pg upgrades are required?

swiffer:
https://github.com/docker-library/docs/tree/master/postgres#initialization-scripts

Bringing us back to: "creating our own postgres image?"

Otherwise we could create idempotent bash script that is run on teslamate container startup creating roles and users (if newly introduced env vars are specified).

Like this approach the most.

brianmay:
Yes, have mixed feeling on including the roles also. On the plus side, it means we can update the roles if required in a later migration. But is this really in the scope? What if users don't like or want that role for some reason?

Or if you are like me and teslamate doesn't have superuser access, and the migration will fail. Yuck.

An advanced / hardening guide would probably be a good start. Unfortunately, I think the user's who need it most are likely to be the same users who ignore it.

Creating our own postgres image might be something worth investigating, not sure I can really think this through right now. And would anybody use it? Maybe new users, existing users might be harder to migrate.

brianmay:
I think I will split this PR up.

The first part, moving the token table into the private schema I think is relatively straightforward, and easy. Like I said I think won't break for most people either (my setup is probably different!).

Might try to change it so it won't fail if the schema already exists. That way I can manually create the schema and fix the permissions before doing migration.

Then we can discuss the next part, trying to get our users to benefit from this.

brianmay:
Now done, this PR is the schema change only.

JakobLichterfeld:
> Creating our own postgres image might be something worth investigating, not sure I can really think this through right now. And would anybody use it? Maybe new users, existing users might be harder to migrate.

We discussed this several times in the past, see here for latest example with links to older discussions: https://github.com/teslamate-org/teslamate/issues/4897#issuecomment-3197585607

TL;DR: Mirroring the PostgreSQL image and add customization like we do with the `teslamate:grafana`image.

JakobLichterfeld (approved):

PR #4980, opened 2025-10-01
Rename object_id to default_entity_id

MrPaulAR:
When I upgraded Home Assistant to 2024.10 I began to see depreciation notices that we should be using default_entity_id instead of object_id.

After updating my local HA sensors the warnings went away.  Issuing a PR to update TeslaMate examples.

### DIFF
--- a/website/docs/integrations/home_assistant.md
+++ b/website/docs/integrations/home_assistant.md
@@ -60,7 +60,7 @@ Don't forget to replace `<teslamate url>`, `<your tesla model>` and `<your tesla
 ```yml title="mqtt_sensors.yaml"
 - sensor:
     name: Display Name
-    object_id: tesla_display_name # entity_id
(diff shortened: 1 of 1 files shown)


ayonix:
According to https://www.home-assistant.io/integrations/device_tracker.mqtt/#default_entity_id the entities should be with the domain, i.e. `default_entity_id: device_tracker.tesla_location`?

MrPaulAR:
Looks like the domain is inherited from the parent context in the yaml.  With the below this shows up as `device_tracker.tesla_pixel_location` in my Home Assistant.

``` yaml
- device_tracker:
    name: Location
    default_entity_id: tesla_pixel_location
    unique_id: teslamate_pixel_location
    availability: *teslamate_pixel_availability
    device: *teslamate_pixel_device_info
    json_attributes_topic: "teslamate/cars/2/location"
    icon: mdi:crosshairs-gps
```

PR #4985, opened 2025-10-02
Update README.md: link for MQTT

DanCard:
- Vehicle data is published to a local **[MQTT](https://en.wikipedia.org/wiki/MQTT)** Broker


JakobLichterfeld (approved):
This is a useful addition for those unfamiliar with MQTT. Ty for your contribution!

PR #4983, opened 2025-10-02
fix: charging stats now correctly calculate cost when set to miles

DrMichael:
Updated SQL query to convert distance using the specified length unit.

PR #4986, opened 2025-10-03
fix: cast to numeric instead of integer when converting from km to miles to avoid rounding issues in all dashboards

swiffer:
to avoid rounding issues and keep in line with other dashboards

- while reviewing #4983 - @DrMichael, thanks quickly addressing this

might not change a lot but now we are casting to numeric in all cases of convert_km for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,9 +23,10 @@
 
 #### Dashboards
 
-- fix: charging stats now correctly calculate cost when set to miles (#4983 -@DrMichael)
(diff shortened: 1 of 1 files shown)


DrMichael (approved):
Good point. What is the difference between numeric and integer?

swiffer:
Integer has no decimals while numeric has

JakobLichterfeld (approved):
Nice find. I do love those standardizations.
Ty!

PR #4988, opened 2025-10-03
fix: correctly determine charging phases in charge detail dashboard

### LINKED ISSUE #414: Incorrectly displays 2 phase charging (while really there are 3)
When charging the car (Tesla Model 3 SR+) at home on a 3-phase 230V connection (in the Netherlands), the Internal - Charge Details shows 2 phases where there are really 3 phases. Charging is done at 8A max (out of 16) and is correctly displayed as charging at 6kW (just as shown in the car and in the Tesla app). I checked the kWh-meter connected to the charger and this has all three phases delivering power.

I saw in the release notes that a change has been made for this in version 1.14.0 but it looks like it is not working for me in version 1.15.1.

**Operating environment**

- TeslaMate Version: v1.15.1
- OS: Synology NAS DSM version: 6.2.2-24922-4
chargedetails

### LINKED ISSUE #4987: Phase Correction in Charging Details Dashboards leads to wrong data

### What happened?

I've charged on 2 phases only - the data shown in the charges dashboard is not adding up correctly

Image

the data gathered and logged to the charges table

Image

the change of phases shown was applied in #414 / #2527

the charge correction applied in dashboards is to simple (compared to what is done in elixir)

https://github.com/teslamate-org/teslamate/blob/0b11c412b18f2ed6dfe507a1575237aa9b7f6f06/lib/teslamate/log.ex#L533

### Expected Behavior

_No response_

Yes

_No response_

### Relevant log output

```shell
-
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

v2.1.2-dev

### PostgreSQL version

v17.6

Yes (specify in next field)

Traefik

swiffer:
this fixes #4987

the idea is to not "blindly" change phases (i actually charged with two phases lately) but use the same approach that is used in elixir for phase correction.

https://github.com/teslamate-org/teslamate/blob/0b11c412b18f2ed6dfe507a1575237aa9b7f6f06/lib/teslamate/log.ex#L533

the original case statement was added in #2527 and intended to fix #414


JakobLichterfeld (approved):
Perfect, ty!

swiffer:
@JakobLichterfeld - do you know if it's possible that number of phases change within a charging process?

as in contrast to teslamate's elixir code we want to visualize data over time i might need to rethink this a little (the way it's solved now phases are constant within a charging process).

JakobLichterfeld:
> @JakobLichterfeld - do you know if it's possible that number of phases change within a charging process?

From my understanding, it is not possible. You can change the charging current, but not the number of phases used. If the phases change, the charge is completed and a new charge starts.

swiffer:
Ok, i guess in that case we can keep as is and fine-tune if any users notices data beeing off / comes up with a different example.

sdwalker:
Had to revert this one to fix errors with the Charge Details and Power panels on the dashboard

swiffer:
@sdwalker could you elaborate whats broken for you / share charges and charging process data that is affected?

sdwalker:
Any DC supercharging (US only?) session is broken
Looking through all of the old supercharging charges with this commit reverted the phases are always zero

AC (split-phase) charging sessions are fine

swiffer:
Same here - will come up with a fix tomorrow morning!

PR #4989, opened 2025-10-03
feat: add shared buffers size to db info dashboard

swiffer:
this adds info about shared_buffers size - by default it's 128MB for docker containers and can be tuned if hardware allows for an increase to increase query performance.

general recomendation is to have it set to 1/4 of available RAM on a dedicated instance.

https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-SHARED-BUFFERS


JakobLichterfeld (approved):
Nice addition, ty!

JakobLichterfeld:
What came to my mind during merging: What do you think about adding a note to explain it with a link to the PostgreSQL doc (similar to what we do with the indexes)?

Issue #4940, opened 2025-09-06
Cannot allocate 2147483711 bytes of memory


a386572631:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

`ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").`
When the database and teslamate start simultaneously, teslamate needs to allocate 2G of memory, which is too large, isn't it

### Can the issue reliably be reproduced?

Yes

### Relevant log output

```shell
[root@iv-ye3wyt0bnkbw80cg565r ~]# docker logs -f e5fe0eb663d3
waiting for postgres at database:5432
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
ll_alloc: Cannot allocate 2147483711 bytes of memory (of type "port_tab").
```

### Type of installation

Docker (https://docs.teslamate.org/docs/installation/docker/)

### Version

2.1.1

### PostgreSQL version

17-trixie

### Are you running latest major supported PostgreSQL version?

- [x] I run the latest major supported PostgreSQL version

### Are you using a reverse Proxy for TeslaMate?

No

JakobLichterfeld:
> When the database and teslamate start simultaneously, teslamate needs to allocate 2G of memory, which is too large, isn't it

[...]

PR #4999, opened 2025-10-13
docs: Avoid memory bloat in some misconfigured versions of Docker/containerd Hosts (e.g. on Debian 13)

dyxyl:
As found in #3045 and #4940, some versions of Docker/containerd set the "nofiles" limit very high, which causes beam.smp to use > 2 GB of memory.  In particular, this happens in the current Debian stable (13/Trixie).

I've checked that this reduces the memory size in the default docker install, but not the Apache or Traefik installs.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -7,7 +7,7 @@ set -e
 
 # prevent memory bloat in some misconfigured versions of Docker/containerd
 # where the nofiles limit is very large. 0 means don't set it.
-if test "${ULIMIT_MAX_NOFILE}" != 0 && test `ulimit -n` -gt "${ULIMIT_MAX_NOFILE}"; then
+if test "${ULIMIT_MAX_NOFILE}" != 0 && test "$(ulimit -n)" -gt "${ULIMIT_MAX_NOFILE}"; then
 	ulimit -n "${ULIMIT_MAX_NOFILE}"
 fi
 


JakobLichterfeld:
Thanks for your contribution.
Generally speaking: configuring the host or docker is not in the scope of TeslaMate. There are a ton of useful docker options, but they would blow up the docker guides.

brianmay:
I think it might be worth while here considering how many people affected. And we already do document the docker-compose file.

JakobLichterfeld:
I tend to agree since the beginning.
From a beginner's perspective, these values seem to appear out of nowhere without any explanation. This does not inspire confidence.

website/docs/advanced_guides/traefik.md
```diff
@@ -58,6 +58,8 @@ services:
       traefik.http.routers.teslamate.tls.certresolver: "tmhttpchallenge"
     cap_drop:
       - ALL
+    ulimits:
+      nofile: 65536
```
JakobLichterfeld: To explain this to newcomers, a comment would greatly enhance UX.

dyxyl:
I agree about the cryptic nature of these settings.

Please consider the change I've just made to `entrypoint.sh` instead, so the docker-compose file changes are not needed.

swiffer:
If the change to the entrypoint alone is good enough i would vote for that.

Lower overall maintenance for us and less complexity exposed to users.

Could you update the pr and remove the other changes?

JakobLichterfeld:
> If the change to the entrypoint alone is good enough i would vote for that.

I do see it differently. This is hiding a setting which is host specific and even worse make it hard coded and not changeable. In my instance the ulimits is much higher and this change will decrease performance.

So I do see several options:

- an env variable and default to `65536`. Which will decrease performance for some users
- using docker setting, as it is most likely caused by misconfigured docker hosts (and add proper comment). This has the drawback of more maintenance for us and more complexity exposed to users.
- do not change anything, as it is not TeslaMate related

swiffer:
is that really affecting performance?

 - i read it as "maximum number of simultanously open ports" - and ERL_MAX_PORTS 65536 sounds a lot.

JakobLichterfeld:
To add: I strongly vote against using a deprecated env variable:

> "Previously the environment variable ERL_MAX_PORTS was used for setting the maximum number of simultaneously existing ports. **This environment variable is deprecated**, and scheduled for removal in OTP-R17, but can still be used."

source: https://erlang.org/documentation/doc-7.2/erts-7.2/doc/html/erl.html

swiffer:
> To add: I strongly vote against using a deprecated env variable:
> 
> > "Previously the environment variable ERL_MAX_PORTS was used for setting the maximum number of simultaneously existing ports. **This environment variable is deprecated**, and scheduled for removal in OTP-R17, but can still be used."
> 
> source: https://erlang.org/documentation/doc-7.2/erts-7.2/doc/html/erl.html

fully agree, wasn't aware. in that case also voting for nothing or adding it to the docs.

dyxyl:
Ah, I didn't realize the env var was deprecated.  I've just pushed a change to set the ulimit directly.  I don't see how this will decrease performance for anyone (assuming no one has thousands of cars connected to the same instance).

dyxyl:
I've refined entrypoint.sh again with a configuration variable and also to avoid an error if the system's limit is smaller.

entrypoint.sh
```diff
@@ -1,8 +1,15 @@
-#!/usr/bin/env sh
+#!/usr/bin/env dash
```
JakobLichterfeld: `#!/usr/bin/env sh` is a link to dash, no need to change this

JakobLichterfeld: ENTRYPOINT should be dash as well...

JakobLichterfeld:
As OP left the PR for the moment, #5025 is the successor with all findings integrated.

PR #5009, opened 2025-10-27
fix for regression introduced in #4988

swiffer:
this ensures the dashboard to render correctly for DC / fast charger based charging processes


swiffer:
@sdwalker - could you test this one?

sdwalker:
> @sdwalker - could you test this one?

Worksforme

PR #4979, opened 2025-10-01
fix: allow using different PostgreSQL port than default when using socket_dir connection

jaypark0006:
fix https://github.com/teslamate-org/teslamate/issues/4720 

when config DATABASE_SOCKET_DIR, runtime.exs will not pass port: System.getEnv(DATABASE_PORT) to TeslaMate.Repo, so Ecto will fallback to s.PGSQL.5432

PR #5026, opened 2025-11-07
docs: fix the formatting of the `DATABASE_SOCKET_DIR` description

IngmarStein:
Also remove `DATABASE_PORT` from the list of mutually exclusive settings as the port is also used for UNIX sockets as of #4979.


IngmarStein:
Is there a place to see the formatter's output?

From the build log, I only see that it changes the file, but not how:
```
ERRO file has changed path=website/docs/configuration/environment_variables.md prev_size=27052 prev_mod_time="2025-11-07 12:40:10 +0000 UTC" current_size=27058 current_mod_time="2025-11-07 12:45:24 +0000 UTC"
```

PR #5074, opened 2025-12-23
docs: fix external image rendering with Traefik SSL

swiffer:
as traefik is terminating ssl the protocol is http for grafana - however, the root_url (externally) is https://

(it worked before - however is required to be https for the external image renderer docker service to work correctly).

i plan adding an advanced guide for it to allow for image functionality - the plugin within grafana no longer exists.

https://github.com/grafana/grafana-image-renderer


JakobLichterfeld (approved):
lgtm, ty!

PR #5086, opened 2026-01-01
docs(home_assistant): default tesla_active_route_distance_to_arrival sensor to kilometers

kenni:
For consistency, update the tesla_active_route_distance_to_arrival sensor to default to kilometers instead of miles, as all other Teslamate sensors exposed to Home Assistant uses kilometers by default.


JakobLichterfeld (approved):
Thanks for your contribution!

PR #5094, opened 2026-01-06
docs: add Matedroid Android app reference

vide:
Small PR to add reference in the website about my new project MateDroid which is basically an Android native client for the stats collected by Teslamate. Fully opensource (GPLv3) and currently only installable via side-loading.

PR #4922, opened 2025-08-27
docs: add Fastcruise to projects using TeslaMate

darox:

PR #5097, opened 2026-01-08
docs: add T-Buddy iOS app reference

JakobLichterfeld:
as discussed in: https://github.com/teslamate-org/teslamate/pull/5094#issuecomment-3723947334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ feat: add optional dark mode feature (#5065 - @wooter)
 - docs: fix external image rendering with Traefik SSL (#5074 - @swiffer)
 - docs(home_assistant): default tesla_active_route_distance_to_arrival sensor to kilometers (#5086 - @kenni)
 - docs: Remove hints to now deprecated Grafana VS Code Extension in contribution guide (#5110 - @swiffer)
+- docs: add Matedroid Android app reference (#5094 - @vide)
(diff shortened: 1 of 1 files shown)


LukStankovic:
Hi @JakobLichterfeld, do we want to add only open source apps to docs? I don't know if T-Buddy will be open source or not (currently it isn't open source but only in TestFlight). But we can add add https://hedgiemate.com/ and https://fastcruise.app/ as well for iOS. 🙌🏻 

I found that PR for Fastcruise already exists here: https://github.com/teslamate-org/teslamate/pull/4922. I can edit it in this PR and add all apps at once if you want.

What do you think?

JakobLichterfeld:
> I don't know if T-Buddy will be open source or not (currently it isn't open source but only in TestFlight)

It is MIT, innit? https://github.com/garanda21/t-buddy/blob/main/LICENSE

LukStankovic:
I was actually basing my comment on the [README](https://github.com/garanda21/t-buddy/blob/main/README.md#-license), where it mentions that the app is proprietary software. That's why I assumed it was closed source.

If I misunderstood or if the info is outdated, sorry about that! 🙏🏻

JakobLichterfeld:
You are right, thanks for pointing it out.

So it violates TeslaMates license and trademark.md and obviously would not be added.

PR #5036, opened 2025-11-15
build: use Elixir 1.19.5 OTP 26

swiffer:
Hi @JakobLichterfeld - see you closed #5016 - let's give it another try with this image and see if things changed

@s56oa / @leewillis77 - could you try this one to bee CPU usage

PR #5127, opened 2026-02-05
refactor: improve pattern matching to fix Elixir 1.19 warnings

JakobLichterfeld:
see discussion in #5036

### DIFF
--- a/lib/tesla_api/stream.ex
+++ b/lib/tesla_api/stream.ex
@@ -200,7 +200,8 @@ defmodule TeslaApi.Stream do
   end
 
   @impl true
-  def handle_disconnect(%{reason: reason, attempt_number: n}, state) when is_number(n) do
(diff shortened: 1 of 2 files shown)


brianmay:
Is there any way we can silence the warnings in the 3rd party libraries?

brianmay:
I don't have high hopes, but lets see what happens :-)

@copilot Can you fix these compile time warnings?

brianmay:
@copilot Can you see the warnings that were produced in the failing actions? The warnings are being turned into errors.

brianmay:
It came up with a solution.... #5129.

Don't think I like it though, too many useless runtime checks. I asked copilot to fix this, but it keeps crashing.

JakobLichterfeld:
Thanks for the try. I reverted Elixir 1.19 for now as these warnings keep us from going forward.
Lets handle all together in #5130

PR #5062, opened 2025-12-05
feat(nix): use nixos-25.11

JakobLichterfeld:

PR #5130, opened 2026-02-06
build: use Elixir 1.19.5 OTP 26

JakobLichterfeld:
Open Points:
- [x] new warnings about pattern matching needs to be solved, otherwise CI will fail
- [X] NixOS build need to use Elixir 1.19.5 OTP 26 as well. As Elixir 1.19 is only available in NixOS >= 25.11 this will handled in #5062
- [X] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Let's start with a quote: "Why do programmers like dark mode? Because light attr
 
 As always, lots of improvements have been made. Memory usage has been reduced, while performance has been increased — it sounds too good to be true, but it's the reality.
 Battery heating is now also shown on the overview dashboard, and the drives dashboard now always shows the consumption.
-To make your TeslaMate experience even better, we have also made more than 36 other improvements.
+To make your TeslaMate experience even better, we have also made more than 37 other improvements.
 
 Enjoy!
 
@@ -50,6 +50,7 @@ Enjoy!
 - build(nix): use elixir 1.19 to align with dockerfile (#5062 - @JakobLichterfeld)
 - chore: reorganize CLI definition to avoid elixir 1.19 warning (#5062 - @JakobLichterfeld)
 - build: update db_connection to version 2.9.0 (#5062 - @JakobLichterfeld)
+- build: use Elixir 1.19.5 OTP 26 (#5130 - @swiffer, @brianmay and @JakobLichterfeld)
 
 #### Dashboards
 


brianmay:
See changes in https://github.com/teslamate-org/teslamate/pull/5129 - looks a lot better now - probably can copy to this pr. 

(Assuming it actually fixes warnings - not checked that yet)

swiffer:
@brianmay - haven't seen your commit / branch - fixed all warnings within this branch already, currently running on my instance - would be great to get some feedback / review the changes

brianmay:
Great!

It appears that `--warnings-as-errors` means that warnings only get converted to errors in our code, if I am reading the logs correctly. Warnings in third party libraries appear to not matter. Good.

More concerning, there are two test failures. Need to investigate.

swiffer:
> Great!
> 
> It appears that `--warnings-as-errors` means that warnings only get converted to errors in our code, if I am reading the logs correctly. Warnings in third party libraries appear to not matter. Good.

yes, this is itentional -> https://hexdocs.pm/mix/1.19.5/Mix.Tasks.Compile.html#module-command-line-options

> More concerning, there are two test failures. Need to investigate.

🤔

swiffer:
I think the Github Workflows could be optimized to support (Matrix Testing (PG 16.7 / 18 | Elixir 1.18-otp26 / 1.19-otp29). Also build caches for the Docker Image could be optimized but this is a whole other story.

Back to this PR. I've matrix tested in my local DevEnv. The code still runs successfully on 1.18.4-otp26 / 1.18.4-otp28. The two tests failing happens only on 1.19.

JakobLichterfeld:
This PR is tested with CI on main, meaning currently with Elixir 1.18 OTP 26. I assume the errors are a result thereof.
Once we merged this PR, the new CI will be used, which we may imrove with matrix testing different Elixir versions, but we officially only support one.

swiffer:
According to the logs it is tested with 1.19 otp 26 - isnt it?

JakobLichterfeld:
Oh yeah, that's strange, no explanation for me, I need to dive deeper into it when on computer.

brianmay:
That first test failure; when I run it locally I get:

```
       pattern: {:start_state, ^car, :online, [date: ^d1]}
       value:   {
                  :start_state,
                  %TeslaMate.Log.Car{
                    __meta__: #Ecto.Schema.Metadata<:built, "cars">,
                    charging_processes: #Ecto.Association.NotLoaded<association :charging_processes is not loaded>,
                    drives: #Ecto.Association.NotLoaded<association :drives is not loaded>,
                    efficiency: nil,
                    eid: 0,
                    exterior_color: nil,
                    id: 5098,
                    inserted_at: nil,
                    marketing_name: nil,
                    model: "3",
                    name: nil,
                    positions: #Ecto.Association.NotLoaded<association :positions is not loaded>,
                    settings: %TeslaMate.Settings.CarSettings{
                      __meta__: #Ecto.Schema.Metadata<:built, "car_settings">,
                      car: #Ecto.Association.NotLoaded<association :car is not loaded>,
                      enabled: true,
                      free_supercharging: false,
                      id: nil,
                      lfp_battery: false,
                      req_not_unlocked: true,
                      suspend_after_idle_min: 15,
                      suspend_min: 21,
                      use_streaming_api: true
                    },
                    settings_id: nil,
                    spoiler_type: nil,
                    trim_badging: nil,
                    updated_at: nil,
                    vid: 1000,
                    vin: "1000",
                    wheel_type: nil
                  },
                  :online,
                  [date: ~U[2026-02-09 22:04:17.361Z]]
                }
```

The ^d1 is highlighted red, which indicates that the date didn't match.

In the above case, it was expecting a date of "2026-02-09T21:49:17.360Z".

Which is out exactly by 15 minutes - which is the offset used in the test. Need to take a break, will come back later.

If you skip the date check it fails with another error.

brianmay:
Might be getting myself confused, with my debugging.

Time in message: 2026-02-09 23:07:47.428Z
Expected time:     2026-02-09T23:07:47.427Z

Looks like an off by 0.001 seconds error.

Times should be exact to the millisecond though, so that shouldn't be happening.

brianmay:
Somewhere the timestamp (milliseconds; integer) is being incremented by 15min + 1ms between the first :start_state and the 2nd :state_state message. The test expects it to be incremented by 15min == 900,000ms only.

Not sure if this is expected or not.

     10:11:40.785 car_id=59256 [error] Start / :online
     10:11:40.785 [error] 1770678700767
     10:11:40.813 car_id=59256 [warning] Vehicle went offline while driving
     10:11:41.049 car_id=59256 [error] Start / :online
     10:11:41.049 [error] 1770679600768

swiffer:
i think the message you receive is caused by driving_test.exs#393

from what i understood so far the drive_event in driving_test.exs#389 caused a :start_state message as well in 1.18 but isn't in 1.19 anymore?

brianmay:
:start_state is still there - at least we get two events as expected, but the date doesn't match so the pattern match fails.

brianmay:
The test has:

```elixir
      events =
        [
          {:ok, online_event()},
          drive_event(now_ts, 0.1, 30, 20, 200, nil),
          drive_event(now_ts + 1, 0.1, 30, 20, 200, nil)
        ] ++
          List.duplicate({:ok, %TeslaApi.Vehicle{state: "offline"}}, 20) ++
          [
            drive_event(now_ts + :timer.minutes(15), 0.2, 20, 19, 190, nil),
            {:ok,
             online_event(
               drive_state: %{
                 timestamp: now_ts + :timer.minutes(15) + 1,
                 latitude: 0.3,
                 longitude: 0.3
               }
             )}
          ]
```

Note, in particular, near the end:

```rust
now_ts + :timer.minutes(15) + 1
```

So I think this line:

```elixir
d1 = DateTime.from_unix!(now_ts + :timer.minutes(15), :millisecond)
```

Should really be:

```elixir
d1 = DateTime.from_unix!(now_ts + :timer.minutes(15) + 1, :millisecond)
```

copilot says the difference in checking date times that are off by one millisecond between Elixir 1.18 and Elixir 1.19 is expected; but I am still trying to understand this.

The change fixes one problem, but the test still fails due to an apparently missing `:start_drive` message. I haven't looked at that yet.

brianmay:
That suggested change above breaks Elixir 1.18 which I don't understand. Going to keep investigating.

brianmay:
Forget what I said above.

For Elixir 1.18, we jump back to Start / :online sooner. Which in turn means we consider the next drive_state part of a new drive.

For Elixir 1.19, this doesn't happen. Which in turn means we consider the next drive_state as part of the existing drive.

This in turn means that the Elixir 1.18 is generating additional events that the 1.19 is not generating, which in turn is leading to the errors.

Still investigating.

brianmay:
Very suspicious this line is the culprit:

```elixir
offline_min = DateTime.diff(offline_end, offline_start, :second) / 60
```

Under 1.18, this gives something like:

(2026-02-10 22:49:49.406Z - 2026-02-10 22:34:49.407Z) / 60 = 900 / 60 = 15.0

i.e. we have rounded the up the time to the nearest second.

Under 1.19 we get something like:

(2026-02-10 22:49:49.406Z - 2026-02-10 22:34:49.407Z) / 60 = 899 / 60 = 14.983333333333333

i.e. we have truncated the time down to the previous second.

Later on in the code we have:

```elixir
offline_min >= @drive_timeout_min
```

Where @drive_timeout_min is 15. In Elixir 1.18 this will be true. In Elixir 1.19 this will be false.

Don't know the fix yet. Need to take a break before I think about that.

Think this is an issue that only effects the tests.

As a test, I kludged the line to be:

```elixir
offline_min = DateTime.diff(offline_end, offline_start, :second) / 60 + 1
```

Now the tests pass both on 1.18 and 1.19

Haven't looked at the other failing test, but wouldn't surprise me if it is the same thing.

I might have made mistakes, feel free to verify my results :-)

swiffer:
Thanks for your deep dive!

In https://github.com/teslamate-org/teslamate/blob/f3c73f25ae2c613353a7dea0baa241005ce9b27b/test/teslamate/vehicles/vehicle/driving_test.exs#L385 we are adding one microsecond so the last drive event is not 15 seconds prior.

In line 389 / 393 we are of by one microsecond for what we are trying to test (now_ts + timer instead of now_ts + 1 + timer...)

Behavior of DateTime.diff changed with 1.19 to fix an issue we relied upon in the testsuite: https://github.com/elixir-lang/elixir/issues/14273

swiffer:
@brianmay - thanks investigating and finding the root cause! I've updated the PR within commit based on the findings.

https://www.github.com/teslamate-org/teslamate/pull/5130/commits/d46e35125bc28a5f23749eb3837e7fb7da1e0564

please have a look.

CI failing because of code style issue now, fill have that fixed later.

swiffer:
@JakobLichterfeld - tests passing and rebased on main 🥳

JakobLichterfeld:
Thank you both for diving into this and solving it. Very much appreciated.

PR #5132, opened 2026-02-07
fix: remove reference to legacy pkgs.system

brianmay:
This is the one and only reference I could find so far...


JakobLichterfeld:
ty, I already have it done in #5062

brianmay:
OK, probably better to merge it there then.

PR #5142, opened 2026-02-14
build(deps): bump floki & gettext, remove override on ranch, require elixir 1.19

swiffer:
bumped
- floki, which required one test case modification
- gettext, which required an override case timex has not updated their constraint yet (see [PR](https://github.com/bitwalker/timex/pull/783)) - feeling save to do so because the changelog for 1.0 is not listing any breaking changes: https://github.com/elixir-gettext/gettext/blob/main/CHANGELOG.md

require Elixir 1.19 in mix
- @JakobLichterfeld, this currently breaks ensure_linting workflow (Elixir 1.18.4 is used there) - do you know why?

remove override from ranch, no longer needed


JakobLichterfeld:
> * @JakobLichterfeld, this currently breaks ensure_linting workflow (Elixir 1.18.4 is used there) - do you know why?

Damm, looked into it as I couldn't explain. It seems like treefmt-nix use `package = "elixir";` https://github.com/numtide/treefmt-nix/blob/337a4fe074be1042a35086f15481d763b8ddc0e7/programs/mix-format.nix#L8

which resolves to 1.18.4. We use for your package `pkgs.beam.packages.erlang_26.elixir_1_19` https://github.com/teslamate-org/teslamate/blob/876aa3db8e941d1c9d32ba54bc245c8f26d084d9/nix/flake-modules/package.nix#L10

I will try to override treefmt-nix package

swiffer:
According to the docs it looks like the program can be overwritten?

https://github.com/numtide/treefmt-nix?tab=readme-ov-file#nix-classic-without-flakes

JakobLichterfeld:
> According to the docs it looks like the program can be overwritten?

Yeah, that's what I mean with:

> I will try to override treefmt-nix package

I was a bit confused at first, but after some time off, I figured it out.

JakobLichterfeld (approved):
Nice one!

JakobLichterfeld (approved):

sdwalker:
Backed down the elixir to 1.18 as Debian doesn't have 1.19 either

swiffer:
Haven't really thought about the unsupported install methods.

Looks like FreeBSD and Debian have only 1.17 / 1.18 available.

Enforcing 1.19 is now effectively blocking install.

We could relax the requirement to e.g. 1.17 - not sure if any other version bumps are causing issues.

brianmay:
What would be idea is if we can run the tests against both 1.18 and 1.19 then.

(unless there is any good reason for 1.17 over 1.18 that I missed)

swiffer:
FreeBSD has 1.17 only.

Running the Testsuite with a Matrix (Elixir 1.17 / Postgres 16.7 || Elixir 1.19 / Postgres 18) should be fine.

Will relax the Elixir constraint later today and open a new issue to propose some changes to the current CI/CD Flow.

swiffer:
see #5145

PR #5143, opened 2026-02-14
feat(grafana-home-dashboard): change cors proxy

swiffer:
as most instances are selfhosted we likely do no longer apply to the free development pricing model for corsproxy.io: https://corsproxy.io/pricing/

cors.lol is offering a free plan that works for us: https://cors.lol/#pricing


swiffer:
To be more specific (i think we are but as domains aren't in the allowlist provided by corsproxy we practically aren't).

cors.lol has rate limiting only but no domain filters.

I was getting 403 responses on corsproxy.io myself.

JakobLichterfeld (approved):
Ty!

PR #5150, opened 2026-02-23
Feature/ios phase2 improvements

drewpost:


brianmay:
I think we would be open to consider incorporating a better json API in Teslamate. The IOS stuff might be better off as a separate project though.

Why was this closed?

Issue #5067, opened 2025-12-11
Geo-Fence to disable API calls


Latif17:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

My Tesla MS is parked underground with no signal and no access to WiFi. Due to this, the car is seen as offline, Teslamate tries to ping it, doesn't get a response and then shows fuse blown. But due to the poor signal, the car seems to show behaviour of waking up following this and thus I am in a cycle of the car being woken up when it shouldn't. This behaviour is intermittent, some periods of the night there are no issues, other times the car is awake with brief moments of offline for 5 hours straight. 

### Expected Behavior

Better sleeping behaviour. 

Is it possible to use the geo-fence feature to recognise the car is at Home Parked and in an area of poor signal to improve the sleep. 

### Can the issue reliably be reproduced?

No

### Relevant log output

```shell
2025-12-11 00:00:47.640 car_id=1 [warning] Vehicle went offline while driving
2025-12-11 00:00:47.703 [info] Streaming API: Vehicle offline
2025-12-11 00:00:47.703 car_id=1 [warning] Stream reports vehicle as offline, fetching vehicle state ...
2025-12-11 00:00:47.752 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (48.434 ms)
2025-12-11 00:00:52.856 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (60.548 ms)
2025-12-11 00:00:58.140 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (68.042 ms)
2025-12-11 00:01:03.244 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (60.236 ms)
2025-12-11 00:01:08.351 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (63.054 ms)
2025-12-11 00:01:13.449 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/1492930878669475/vehicle_data -> 408 (56.337 ms)
2025-12-11 00:01:17.840 [info] Streaming API: Vehicle offline
[...]

PR #5151, opened 2026-02-23
fix(vehicle): update state transition to handle offline scenario after a drive end with no network

jlestel:
What happens when a vehicle is in an underground parking garage, [I've noticed](https://github.com/teslamate-org/teslamate/issues/5067#issuecomment-3938543417) that the drive completes correctly but the status is still online and the requests continue to run every 30 seconds :

1. The vehicle transitions to `{:driving, {:unavailable, 0}, drive}` → incrementing `n` every 5 seconds.
1. After `n=15` (75 seconds), it transitions to `{:driving, {:offline, last}, drive}`.
1. After 15 minutes (`@drive_timeout_min`), `timeout_drive` closes the trip → transitions to `{:driving, {:offline, last}, nil}`.
1. From this point on, `{:keep_state}` loops every 30 seconds → never transitions to `:offline`.

The fix: when the trip has already been closed (`drive = nil`) and offline responses are still being received, it must switch to `:start` for the normal offline cycle to begin.

I have also updated the "times out a drive" test to reflect this change, and I have added a specific regression test for the underground parking scenario to ensure that this case is handled correctly in the future.

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -1017,7 +1017,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
         {:driving, {:offline, _last}, nil},
         %Data{} = data
       ) do
-    {:keep_state, %{data | last_used: DateTime.utc_now()}, schedule_fetch(30, data)}
+    {:next_state, :start, %Data{data | last_used: DateTime.utc_now()}, schedule_fetch(data)}
   end
 
   def handle_event(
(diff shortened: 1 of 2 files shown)


brianmay:
Thanks for debugging this.

This state machine is rather complicated, I am not surprised there are errors.

brianmay (approved):

jlestel:
I thought the CI would provide me with an image that could be tested.

I haven't tested it in real-world conditions yet...

brianmay:
> I thought the CI would provide me with an image that could be tested.

Yes, I believe this is correct.

jlestel:
GHCR build is skipped for PRs from outside repo

JakobLichterfeld:
> GHCR build is skipped for PRs from outside repo

Yeah this is for security reasons, if you don't mind, we can replicate your changes in a branch in this repo and the CI will build the image.

JakobLichterfeld:
--> 5152

PR #5152, opened 2026-02-24
fix(vehicle): update state transition to handle offline scenario after a drive end with no network

JakobLichterfeld:
successor of #5151 from @jlestel, so CI can build test images

> What happens when a vehicle is in an underground parking garage, [I've noticed](https://github.com/teslamate-org/teslamate/issues/5067#issuecomment-3938543417) that the drive completes correctly but the status is still online and the requests continue to run every 30 seconds :
> 
> 1. The vehicle transitions to `{:driving, {:unavailable, 0}, drive}` → incrementing `n` every 5 seconds.
> 2. After `n=15` (75 seconds), it transitions to `{:driving, {:offline, last}, drive}`.
> 3. After 15 minutes (`@drive_timeout_min`), `timeout_drive` closes the trip → transitions to `{:driving, {:offline, last}, nil}`.
> 4. From this point on, `{:keep_state}` loops every 30 seconds → never transitions to `:offline`.
> 
> The fix: when the trip has already been closed (`drive = nil`) and offline responses are still being received, it must switch to `:start` for the normal offline cycle to begin.
> 
> I have also updated the "times out a drive" test to reflect this change, and I have added a specific regression test for the underground parking scenario to ensure that this case is handled correctly in the future.


jlestel:
Seems to resolve this bug. The state changed to offline after the Drive timeout, at 19:22:48  : 

```
2026-02-25 19:22:07.624 car_id=1 [warning] Stream reports vehicle as offline, fetching vehicle state ...
2026-02-25 19:22:07.624 car_id=1 [info] Fetch already in progress ...
2026-02-25 19:22:08.090 [info] GET https://nominatim.openstreetmap.org/reverse -> 200 (231.447 ms)
2026-02-25 19:22:08.248 [info] GET https://nominatim.openstreetmap.org/reverse -> 200 (149.654 ms)
2026-02-25 19:22:08.278 car_id=1 [info] Driving / Timeout / 16 km – 27 min
2026-02-25 19:22:37.857 [info] Streaming API: Vehicle offline
2026-02-25 19:22:37.857 car_id=1 [warning] Stream reports vehicle as offline, fetching vehicle state ...
2026-02-25 19:22:37.939 [info] GET https://owner-api.teslamotors.com/api/1/vehicles/xxx/vehicle_data -> 408 (81.141 ms)
2026-02-25 19:22:48.127 car_id=1 [info] Start / :offline
2026-02-25 19:22:48.134 car_id=1 [info] Stream disconnecting ...
```

Capture d’écran 2026-02-25 à 7 50 42 PM

I will confirm further underground arrivals in the coming days.

jlestel:
I can confirm that I behave the same way in the days following an underground arrival 👍
Capture d’écran 2026-02-28 à 12 56 28 AM
Capture d’écran 2026-02-28 à 12 56 43 AM

PR #5175, opened 2026-03-04
feat: show bar indicators to statistics dashboard

pencilheart:
Xnip2026-03-04_23-04-48

Hi TeslaMate team,

I’d like to propose a Grafana dashboard focused on daily statistics (“Day Trips”). The goal is to provide a clear per-day aggregation view of driving sessions, so users can quickly compare different days without drilling into individual trip details.

What it shows (daily level):
	•	Start/End time of the day’s trip records
	•	Total distance and duration
	•	Average speed and outside temperature
	•	Energy consumed (gross)
	•	Net consumption (Wh/km) and net efficiency
	•	Gross consumption (Wh/km) and gross efficiency
(With bar-style visuals to make day-to-day comparison easy.)

Why this is useful:
Many users primarily want a day-by-day overview (daily routine tracking, commuting analysis, high-consumption days, seasonal temperature impact, etc.). A dedicated daily dashboard makes it much easier to spot trends and outliers than browsing individual trips.

If your team finds this useful, I’d really appreciate it if you could review it and share any feedback or suggestions for improvement.

Thanks!


DrMichael:
Well, we have the Statistics dashboard with a daily view. Looks quite similar. It might be better to integrate the changes into that? (I like the coloured bars...)

pencilheart:
> Well, we have the Statistics dashboard with a daily view. Looks quite similar. It might be better to integrate the changes into that? (I like the coloured bars...)

Oh, my bad. Statistics dashboard is great (Forgive me miss it before). It is true that the two conflicted. 

If integration is needed, I will do it.

swiffer:
hi @pencilheart - thanks for your contribution - how is that different from looking at the statistics dashboard with setting the period to day?

swiffer:
sorry, somehow missed the reply by DrMichael - integration would be needed, yes

agree that having a bar for distance would be great. as energy used most likely correlates i wouldn't add a graph for that considering the space available.

@pencilheart - do you want to to update this PR ?

pencilheart:
@swiffer Alright, I'll update it.

pencilheart:
Before:
Xnip2026-03-10_22-00-11

After:
Xnip2026-03-10_21-59-07

DrMichael:
Ah, then please also for "Driving Efficiency" :-)

Perhaps narrow some columns also... So all fit better on one screen.

swiffer:
i would keep driving efficiency as is (same visuals are used in Drives and Efficiency Dashboards)

DrMichael:
> i would keep driving efficiency as is (same visuals are used in Drives and Efficiency Dashboards)

Well, then let's make it consistent and the bar for distance same style as with driving efficiency. (Although the other style looks better to me.)

DrMichael:
> i would keep driving efficiency as is (same visuals are used in Drives and Efficiency Dashboards)

Valid point. What about changing all efficiency displays to gradient? I played around a bit:
image
image
image
image

Opinions?

swiffer:
Not really - personally like the retro style and the color palette being a little less "poppy" - but open changing it if others like it more

JakobLichterfeld:
Is the only thing missing the screenshot and link in README according to https://docs.teslamate.org/docs/development#making-changes-to-grafana-dashboards?

JakobLichterfeld:
I would really love to include it in next release. @pencilheart may I ask you to add the screenshot and link in README?

pencilheart:
@JakobLichterfeld Done!

JakobLichterfeld (changes requested):
Hi hanheihei, thanks for the contribution and the nice idea — bar indicators make the distance column much easier to scan! 🙏

A few points before we can merge:

1. **Hardcoded `max: 400` saturates the gauge for longer periods.** This table aggregates per `$period` (day/week/month/year). With month or year selected, virtually every row exceeds 400 km, so all bars are pegged at 100 % and the comparison the bars are meant to enable no longer works. Please drop the `max` and `min` overrides entirely — Grafana then auto-scales the gauge to the data range of the current query, which works for every period.

2. **km and mi overrides should be identical.** The `sum_distance_km` override sets `"valueDisplayMode": "text"` and `"decimals": 0`, while the `sum_distance_mi` override has neither — users with miles configured would get a visibly different (and unrounded) rendering. Please apply the same cell options to both. (With `max` removed, the "same 400 for km and mi" scale mismatch goes away as well.)

3. Minor: the PR description talks about a new "Day Trips" dashboard, but the diff modifies the existing Statistics dashboard. Could you update the description to match the actual change?

Once the overrides are cleaned up, please also regenerate the screenshot so README/website show the final look.

Thanks again — looking forward to the update!

PR #5180, opened 2026-03-06
i18n: updated Catalan default.po

pellix:
Added 'view car location' string and minor changes


JakobLichterfeld (changes requested):
ty

JakobLichterfeld (approved):

PR #5187, opened 2026-03-07
fix(grafana): use FLOOR/CEIL over ROUND for timestamps used in dashboard links to avoid timeranges becoming to narrow

### LINKED ISSUE #5173: Upgrade to version 3.0 18 database, select the time period in trip to display normal, click from the timeline will be abnormal (no track, there is an altitude), manually change the start time forward for a few seconds to display normal

### What happened?

Upgrade to version 3.0 18 database, select the time period in trip to display normal, click from the timeline will be abnormal (no track, there is an altitude), manually change the start time forward for a few seconds to display normal

### Expected Behavior

_No response_

Yes

_No response_

### Relevant log output

```shell
Upgrade to version 3.0 18 database, select the time period in trip to display normal, click from the timeline will be abnormal (no track, there is an altitude), manually change the start time forward for a few seconds to display normal
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

3.0

### PostgreSQL version

database:   image: postgres:18-trixie   ...   volumes:     - teslamate-db:/var/lib/postgresql

No

_No response_

swiffer:
fixes #5173 

so far we've used ROUND() to calculate the start_date_ts / end_date_ts.

This is causing trouble in dashboards that are using the timerange as a filter to determine the drives / charges to show (as ROUND() will round up the start date depending on it's value).

some dashboards mitigated that by adding +10 / -10.

consequently use FLOOR / CEIL instead to ensure we aren't narrowing timeranges down to much.


swiffer:
@DrMichael could you maybe review this one?

DrMichael:
> @DrMichael could you maybe review this one?

Hmmm, not sure, how to see a difference. Anyway, I still get round instead of floor/ceil with "image: ghcr.io/teslamate-org/teslamate:pr-5187"?

swiffer:
Within the Timeline dashboard there should have been drives which when clicking on the Link in the Start column haven't shown any data in the Trips dashboard.

After this change the data should be shown correctly.

DrMichael:
lgtm I can confirm that my "no data" cases are being resolved by this PR.

swiffer:
ty, lgtm - if i understood correctly we want to have a formal approval via the review functionality - would be great if you could give that as well!

PR #5190, opened 2026-03-08
docs: drop private schema before restore

brianmay:
Backups assume that the public schema already exists before restore, and that the private schema does not exist and should be created.


brianmay:
There is a test failure, but nothing to do with this pull request :-(

swiffer:
yes, looks like one test is flaky - thanks! merging now

PR #5193, opened 2026-03-09
feat(webview): Make icons inherit text color for better contrast in dark mode

olsoybakk:
This pull request enhances the dark mode theme in TeslaMate by making icons inherit the text color, resulting in better contrast and improved readability in dark mode environments. Previously, icons may have had inconsistent coloring that didn't align with the surrounding text, potentially reducing usability.

### Before
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - feat(webview): Sort vehicles by display_priority (#5188 - @olsoybakk and @swiffer)
 - fix(grafana): use FLOOR/CEIL over ROUND for timestamps used in dashboard links to avoid timeranges becoming to narrow (#5187 - @swiffer)
+- feat(webview): Make icons inherit text color for better contrast in dark mode (#5193 - @olsoybakk)
 
 #### Build, CI, internal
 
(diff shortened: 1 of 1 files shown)


olsoybakk:
@swiffer Is this ok? Icons are hard to see in dark-mode

swiffer:
Hi @olsoybakk - text color is currently set to grey-dark via css class in `lib/teslamate_web/live/car_live/summary.html.heex`

I would suggest changing the class to `has-text-inherit` instead of `has-text-grey-dark` there instead of using a theme specific overwrite later. This will slightly change colors in light theme as well.

Would you like to update this PR accordingly and add a changelog entry referencing this PR ?

olsoybakk:
> Hi @olsoybakk - text color is currently set to grey-dark via css class in `lib/teslamate_web/live/car_live/summary.html.heex`
> 
> I would suggest changing the class to `has-text-inherit` instead of `has-text-grey-dark` there instead of using a theme specific overwrite later. This will slightly change colors in light theme as well.
> 
> Would you like to update this PR accordingly and add a changelog entry referencing this PR ?

@swiffer Thanks for the input. I've changed my PR and added a changelog.

swiffer:
Thanks! LGTM

PR #5199, opened 2026-03-10
fix: ensure Charges / Drives dashboards load correctly if no Geofence exists

### LINKED ISSUE #5191: invalid geofence selected if no geofences defined

### What happened?

Same people are reporting that if they have no geofences defined, then the default geofence filter is selected is invalid, and must be manually changed to "All" in order to display data.

* https://github.com/teslamate-org/teslamate/issues/5157#issuecomment-4017094621
* https://github.com/teslamate-org/teslamate/discussions/5189

Probably other references I missed.

I have not personally seen this - I have geofence defined - but enough people are reporting this I think a bug report is warranted.

### Expected Behavior

All should be the default geofence, even if no geofences defined.

Yes

Go to Charges or Drives dashboard without any geofences defined.

### Relevant log output

```shell
N/A
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

v3.0.0

### PostgreSQL version

_No response_

No

_No response_

swiffer:
If no Geofence exists Grafana v12.4.0 fails to initialize the "Include All option" correctly causing "No Data" errors in Charges / Drives dashboards (see grafana/grafana#5191).

Until fixed upstream add a workaround by disabling the "Include All option" in these dashboards and add an entry via SQL that is imitating this option instead.

Should be reverted once fixed upstream in https://github.com/grafana/grafana-postgresql-datasource/issues/125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 
 #### Dashboards
 
-- fix(grafana): use FLOOR/CEIL over ROUND for timestamps used in dashboard links to avoid timeranges becoming to narrow (#5187 - @swiffer)
+- fix: use FLOOR/CEIL over ROUND for timestamps used in dashboard links to avoid timeranges becoming to narrow (#5187 - @swiffer)
+- fix: ensure Charges / Drives dashboards load correctly if no Geofence exists (#5199 - @swiffer)
 
 #### Translations
 
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -1416,18 +1416,19 @@
       },
       {
         "allValue": "-1",
+        "allowCustomValue": false,
         "current": {},
         "datasource": {
           "type": "grafana-postgresql-datasource",
           "uid": "TeslaMate"
         },
-        "definition": "SELECT name AS __text, id AS __value FROM geofences ORDER BY name COLLATE \"C\" ASC;",
-        "includeAll": true,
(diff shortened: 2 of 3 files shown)


JakobLichterfeld (approved):
Ty, lgtm

DrMichael:
Hi @swiffer, Drives behaves a liitle strange now, if there is a geofence "smaller" than `All`.
E.g. when clicking the drives value in Statistics.

Maybe somethin like this works...
```
SELECT name AS __text, id AS __value
FROM geofences
UNION ALL
SELECT 'All', 0
WHERE NOT EXISTS (SELECT 1 FROM geofences) ORDER by __text;
```

swiffer:
@DrMichael - this one should ensure the "fake" entry is always on top, isn't it?

```sql
-- The \"Include All option\" built into Grafana is not initializing correctly if no value is returned by the SQL query anymore
-- Create the option via SQL as a temporary workaround until fixed (https://github.com/grafana/grafana-postgresql-datasource/issues/125)

with geofences_incl_all_option as (
  select 'All' as __text, -1 as __value, null as name
  union all
  SELECT name AS __text, id AS __value, name from geofences
 )
 
 SELECT __text, __value from geofences_incl_all_option order by name asc nulls first
```

DrMichael:
> @DrMichael - this one should ensure the "fake" entry is always on top, isn't it?

Right. Looks good, but how would I test it?

swiffer:
think the savest approach is what i've done in grafana/grafana#5335, tested in my instance with and without geofences

DrMichael:
> think the savest approach is what i've done in grafana/grafana#5335, tested in my instance with and without geofences

Fixes my case with https://github.com/teslamate-org/teslamate/pull/5335#issuecomment-4460208296

Issue #192, opened 2019-10-05
ERROR: type "earth" does not exist at character 136


mapchen:
Hi,

I get following error when starting new version 1.10 from docker. I was migrating from 1.19 and had no problem before.

@adriankumpf any idea?

br

pi@raspberrypi:~/teslamate $ sudo docker-compose up
Starting teslamate_teslamate_1 ... done
Starting teslamate_grafana_1   ... done
Starting teslamate_db_1        ... done
Starting teslamate_mosquitto_1 ... done
Attaching to teslamate_teslamate_1, teslamate_db_1, teslamate_mosquitto_1, teslamate_grafana_1
db_1         | 2019-10-05 16:10:32.669 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
db_1         | 2019-10-05 16:10:32.669 UTC [1] LOG:  listening on IPv6 address "::", port 5432
db_1         | 2019-10-05 16:10:34.138 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1         | 2019-10-05 16:10:35.178 UTC [24] LOG:  database system was shut down at 2019-10-05 16:09:42 UTC
db_1         | 2019-10-05 16:10:35.182 UTC [25] LOG:  incomplete startup packet
teslamate_1  | waiting for postgres at db:5432 (2sec)
mosquitto_1  | 1570291833: mosquitto version 1.6.7 starting
mosquitto_1  | 1570291833: Config loaded from /mosquitto/config/mosquitto.conf.
mosquitto_1  | 1570291833: Opening ipv4 listen socket on port 1883.
mosquitto_1  | 1570291833: Opening ipv6 listen socket on port 1883.
db_1         | 2019-10-05 16:10:35.233 UTC [1] LOG:  database system is ready to accept connections
grafana_1    | /run.sh: line 75: /usr/share/grafana/bin/grafana-cli: No such file or directory
teslamate_1  | 18:10:45.447 [info] Already up
teslamate_grafana_1 exited with code 127
db_1         | 2019-10-05 16:11:05.768 UTC [34] ERROR:  type "earth" does not exist at character 136
db_1         | 2019-10-05 16:11:05.768 UTC [34] QUERY:  SELECT cube(cube(cube(earth()*cos(radians($1))*cos(radians($2))),earth()*cos(                                                                                                        radians($1))*sin(radians($2))),earth()*sin(radians($1)))::earth
[...]

PR #5205, opened 2026-03-13
fix: restore earthdistance function search paths

nebhale:
## Summary
Reapply the public search path for the ll_to_earth and earth_box functions after the earthdistance extension upgrade.

## Why
TeslaMate previously added search_path = public for these functions in the PostgreSQL 17 compatibility work because earthdistance functions without an explicit search path can fail to resolve the earth type correctly.

Later, the earthdistance v1.2 upgrade migration updated the extension but did not restore that custom function metadata. That appears to drop the expected search path from the extension-managed functions.

This follow-up migration restores the public search path for both functions.

## Context
- Original qualification/search-path issue: #192
- Search-path hardening added in: #4231
- Extension upgrade that likely reset the function metadata: #4648

## Validation
- mix format priv/repo/migrations/20260312120000_set_search_path_for_earthdistance_functions.exs
- mix compile

### DIFF
--- a/nix/backup_and_restore.nix
+++ b/nix/backup_and_restore.nix
@@ -25,6 +25,8 @@ let
       create schema public;
       CREATE EXTENSION cube WITH SCHEMA public;
       CREATE EXTENSION earthdistance WITH SCHEMA public;
+      ALTER FUNCTION ll_to_earth SET search_path = public;
+      ALTER FUNCTION earth_box SET search_path = public;
     .
 
     # Restore
--- a/priv/repo/migrations/20260502120000_set_search_path_for_earthdistance_functions.exs
+++ b/priv/repo/migrations/20260502120000_set_search_path_for_earthdistance_functions.exs
@@ -1,0 +1,12 @@
+defmodule TeslaMate.Repo.Migrations.SetSearchPathForEarthdistanceFunctions do
+  use Ecto.Migration
+
+  def up do
+    execute("ALTER FUNCTION ll_to_earth SET search_path = public")
+    execute("ALTER FUNCTION earth_box SET search_path = public")
+  end
+
+  def down do
+    :ok
+  end
+end
--- a/website/docs/installation/unsupported/unraid.md
+++ b/website/docs/installation/unsupported/unraid.md
@@ -187,6 +187,8 @@ DROP SCHEMA private CASCADE;
 CREATE SCHEMA public;
 CREATE EXTENSION cube WITH SCHEMA public;
 CREATE EXTENSION earthdistance WITH SCHEMA public;
+ALTER FUNCTION ll_to_earth SET search_path = public;
+ALTER FUNCTION earth_box SET search_path = public;
 SQL
 
 if [ $? -ne 0 ]; then
--- a/website/docs/maintenance/backup_restore.mdx
+++ b/website/docs/maintenance/backup_restore.mdx
@@ -80,6 +80,8 @@ DROP SCHEMA private CASCADE;
 CREATE SCHEMA public;
 CREATE EXTENSION cube WITH SCHEMA public;
 CREATE EXTENSION earthdistance WITH SCHEMA public;
+ALTER FUNCTION ll_to_earth SET search_path = public;
+ALTER FUNCTION earth_box SET search_path = public;
 .
 
 # Restore


swiffer:
this isn't needed anymore, is it ?

nebhale:
@swiffer I back my teslamate instance with a Google Cloud SQL PG instance.  I've been looking to upgrade form 17 to 18 and it fails some pre-upgrade checks (I have two schemas on the same database):

image

My guess is that you're right, PG itself doesn't require this but at least Google's managed instances do.

swiffer:
Ok, seems to be problematic if using inplace upgrade / pg_restore without previously creating the extensions agian like we do in the docs.

```
# Stop the teslamate container to avoid write conflicts
docker compose stop teslamate

# Drop existing data and reinitialize (Don't forget to replace first teslamate if using different TM_DB_USER)
docker compose exec -T database psql -U teslamate teslamate << .
DROP SCHEMA public CASCADE;
DROP SCHEMA private CASCADE;
CREATE SCHEMA public;
CREATE EXTENSION cube WITH SCHEMA public;
CREATE EXTENSION earthdistance WITH SCHEMA public;
.

# Restore
docker compose exec -T database psql -U teslamate -d teslamate < teslamate.bck

# Restart the teslamate container
docker compose start teslamate
```

You have been able to upgrade correctly after setting the search_path?

nebhale:
Yes, I was able to run those two commands manually and upgrade the database.  Restarting Teslamate against the modified and upgraded database does not appear to have any problems.

swiffer:
Good to hear - if we set this via migration - are we able to drop creating the schema and extensions in our own restore?

nebhale:
My hypothesis is yes, but I've never upgraded this way, so I defer to others on an authoritative answer.

swiffer:
I will give it a try over the weekend ;)

brianmay:
Disclaimer: I don't really understand these functions, or why they are required. Or why setting the search path is important. So I might have got details wrong.

I can say that the migration will only run if it has not run before. So if you restore an old backup that hasn't had this migration, it will run the migration. But if you restore a recent backup that already has had this migration applied, it will not run this migration again. Because it is already recorded in the schema_migrations table.

Normally this wouldn't matter, because the backup would also have all the values from all the applied migrations.

If I look at my backup, after manually applying the changes in this migration, the backup does not contain the new search_path values, all I see in my backup is this line:

```
SELECT pg_catalog.set_config('search_path', '', false);
```

Which I think might be insufficient. ?

nebhale:
I'd like to keep this alive and eventually get it merged if possible.  The change does appear to be necessary for some installations and the change has been validated.

brianmay:
As above, it makes me nervous that this value does not appear to be saved in backups. Do we need to update the restore documentation to include a step to manually fix this?

nebhale:
> Do we need to update the restore documentation to include a step to manually fix this?

Copy that.  I'll update the PR.

nebhale:
Documentation updated and ready for re-review.

JakobLichterfeld:
> Do we need to update the restore documentation to include a step to manually fix this?

No, the issue is only present when someone does not follow the restore docu:

> Ok, seems to be problematic if using inplace upgrade / pg_restore without previously creating the extensions agian like we do in the docs.

So imo this PR is not needed


> If I look at my backup, after manually applying the changes in this migration, the backup does not contain the new search_path values, all I see in my backup is this line:
> 
> ```
> SELECT pg_catalog.set_config('search_path', '', false);
> ```
> 
> Which I think might be insufficient. ?

And as you observed does not provide any meaningful value.

brianmay:
I do think we need more data to justify the need for this. i.e. what steps are required to reproduce the search_path being unset.

For the record, AI is telling me the solution is not to rely on the search path at all, and prefix everything with "public.", e.g. `public.earth_box(public.ll_to_earth(...))`.

JakobLichterfeld:
Thanks for digging into this, @nebhale, and for the patient back-and-forth — I went through the extension internals before deciding.

**The search path is no longer load-bearing.** With earthdistance 1.2 (shipped in the Nov 2024 minor releases: 16.5+/17.1+), all functions were rewritten as new-style SQL bodies (`RETURN …`) with cross-extension references qualified via `@extschema:cube@`. Those bodies are parsed and dependency-tracked at definition time, so the runtime `search_path` no longer affects name resolution. That is exactly why `CREATE OR REPLACE` inside `ALTER EXTENSION earthdistance UPDATE` (#4648, migration `20250407155134`) dropped the `SET` attribute: the workaround from #4231 became obsolete rather than being lost. We already enforce PostgreSQL >= 16.7 / 17.3 / 18.0 in `lib/teslamate/database_check.ex`, so every supported install is on 1.2.

Empirically this matches: `20240929084639_recreate_geo_extensions` drops and recreates both extensions without re-applying the `ALTER FUNCTION`, so **no** installation — new or upgraded — has had this setting since September 2024. Roughly two years of production, zero reports of geofence lookups failing.

A few more reasons not to merge:

- It would not fully fix your case anyway. earthdistance has seven functions; this sets two. We call `earth_distance()` in `lib/teslamate/locations.ex` and `lib/teslamate/custom_expressions.ex`, and `earth()`, `sec_to_gc()`, `gc_to_sec()`, `latitude()`, `longitude()` are equally unqualified — the Cloud SQL pre-upgrade check would keep flagging them.
- It is not durable, as @brianmay observed: `pg_dump` does not dump attributes of extension-member functions, the migration never re-runs (it is recorded in `schema_migrations`), and any future `ALTER EXTENSION … UPDATE` resets it again. We would be carrying it in the docs forever.
- It adds a new failure mode: `ALTER FUNCTION` on extension-owned functions requires ownership. On managed Postgres where the extensions were created by a superuser role, the migration would fail hard at startup — for everyone, to satisfy a vendor lint that affects one setup.

So what you hit is a Google Cloud SQL pre-upgrade check (a CVE-2018-1058-style heuristic about mutable search paths), not a TeslaMate defect. Applying those two statements manually before the 17 → 18 upgrade, as you did, is the right place for that fix: it is a one-off, vendor-specific step, and it does not change TeslaMate's behaviour.

Closing this one — not because the investigation was wrong, but because the underlying issue was already fixed properly by the extension upgrade. Really appreciate the detailed reproduction and the validation work.

nebhale:
FWIW, I disagree with Claude’s assessment here.  For example, altering the two extensions, and not all seven _did_ allow the upgrade. This means that the other five extensions are different in some way that wasn’t property chased down.

Additionally, as a long-time OSS maintainer, I think Claude’s judgement that a behavior on a major cloud provider that prevents users from upgrading without manual intervention was fine, is short-sighted.  Anything that adds friction to upgrades simply elongates an OSS team’s maintenance window.  How long will PG17 need to be supported since the Telsmate can’t make the jump to 18 without human intervention on a provider representing 14% of the market?  Claude’s analysis here was focused on runtime behavior (“so no installation — new or upgraded — has had this setting since September 2024. Roughly two years of production, zero reports of geofence lookups failing.”) and not upgrade behavior.  This wasn’t the appropriate lens to judge this by.

My problem is solved so no hard feelings on closing the PR.

JakobLichterfeld:
Thanks for the follow-up, @nebhale — and fair point on one thing: my claim that the pre-upgrade check would keep flagging the remaining functions was a prediction, not something we tested, and your result disproves it. Most likely the check only inspects functions actually referenced by objects in the database — `ll_to_earth` and `earth_box` are exactly the two TeslaMate has used in index expressions — but that's a property of Google's checker, not of TeslaMate.

On the broader point, I see the upgrade-friction argument, but I don't think it changes the outcome. What blocked you was a vendor-specific heuristic in Cloud SQL's pre-upgrade check, applied to a database that had been restored outside our documented procedure. TeslaMate itself behaves correctly on every supported PostgreSQL version, the upgrade path is documented, and it works without manual intervention on earthdistance plugin for the overwhelming majority of installations — including plenty on managed providers.

We also deliberately don't maintain provider-specific upgrade instructions: we can't test or keep pace with each vendor's pre-upgrade tooling, and documenting one provider's workaround creates an implicit promise we can't honor for the others. For cases like this one, the vendor's own check output points at the exact statements needed — as it did for you.

So: closing stands, but the two-vs-seven observation was a genuinely useful data point. Thanks again for reporting back.

PR #5209, opened 2026-03-14
ci: ensure not running 2 workflows when pushing to a branch that has a PR open

swiffer:
the DevOps workflow is currently triggered on push an pr - this triggers multiple runs for branches that have a PR open.

I assume we want to keep the DevOps workflow for both. to avoid running it twice I've added concurrency configuration which cancels the workflow triggered by the push and allows the one triggered by updating the pr to run through.

In addition this will cancel any workflow runs still running if additional push events occure meanwhile.


JakobLichterfeld (approved):
Nice one! Ty

PR #5210, opened 2026-03-14
ci: fix concurrency group in DevOps workflow

swiffer:
ensure to use source branch name (pull request) and fallback to branch name if not set (push)

PR #5211, opened 2026-03-14
ci: limit DevOps workflow runs of type push to branch main to prevent duplicate runs on PR

swiffer:
Previous attemts (#5209 and #5210) to ensure not running the DevOps workflow twice for pushes to branches with existing PRs failed as workflows / jobs cancelled via concurrency settings count as failed.

- Limit the workflow to run on events of type PR targeting main and pushes to main instead.
- Keep Concurrency config for canceling any outdated workflow runs.

### DIFF
--- a/.github/workflows/devops.yml
+++ b/.github/workflows/devops.yml
@@ -3,6 +3,7 @@ name: DevOps
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


JakobLichterfeld (approved):
lgtm, ty!

PR #5213, opened 2026-03-14
ci: bump actions to avoid warnings for node 20

swiffer:


JakobLichterfeld:
> Regarding checkout: #5083 and #5118 from dependabot had broken the CI in the past, that's why I reverted and have chosen to stay on v5.

Seems like still same issue

swiffer:
Seems to be a workaround to not persist the auth data:

https://github.com/actions/checkout/issues/2299

Will try tomorrow

swiffer:
@JakobLichterfeld - the "Setup Elixir and Cache Dependencies" composite action was using "actions/checkout"

as this action is already used in workflows it's executed twice.

"actions/checkout" v5 persist auth leading to duplicate auth headers. I've removed the duplicated "actions/checkout" step from "Setup Elixir and Cache Dependencies"

JakobLichterfeld:
Ah, thanks for looking deeper into it. Atm my concentration is on another level, and not in a good way.

swiffer:
No problem at all! Trying to at least here support as good as I can.

JakobLichterfeld:
> Trying to at least here support as good as I can.

Are you kidding? You do an amazing job with the PR comments, reviews, changelog, overview, CI, bugs etc. 💚 After checking in yesterday, I felt relaxed seeing how you manage all this atm.

PR #5215, opened 2026-03-15
fix(ci): ensure security checks pass for push events, fix repository for purges

swiffer:


swiffer:
@JakobLichterfeld - as I'm touching security related code here it would be great if you could have a quick look.

the build for main failed as the "Check if PR is from outside repo" failed (as the event is no longer a pull_request in that case).

JakobLichterfeld:
Please double check with an external PR after merge

swiffer:
looks good to me:

https://github.com/teslamate-org/teslamate/pull/5216

however noticed that when closing an external PR we are running the ghcr_purge workflow which makes no sense.

JakobLichterfeld:
Perfect.

> however noticed that when closing an external PR we are running the ghcr_purge workflow which makes no sense.

It was meant to build for external PRs, that's why it was run. As someone was able to open a PR and then add malicious commits step by step we disabled it. So it is a leftover, but goal was to re-enable with bulletproof sec.

brianmay:
Wonder if something like this might help, require manual approval before building ghcr for external PRs: https://github.com/marketplace/actions/manual-workflow-approval

I think we don't get that many.

JakobLichterfeld:
Wonderif/Doubt this will cover all edge-cases. The only successful attacker needed 3 attempts to get it "right".


The solution I had in mind is based on GitHub recommendations for PR target:
https://github.com/teslamate-org/teslamate/pull/4637#issuecomment-2785657023

brianmay:
Yes, if you can get that working, then that would be better,

Basically ensure that the untrusted build code never gets access to our secrets.

swiffer:
👍got it - has the repo setting for "Approval for running fork pull request workflows from contributors" been set to "Require approval for all external contributors" already?

would that already prevent any issues (besides someone from our own repo approving without fully checking for any malicious changes)?

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

brianmay:
Regardless of who we allow to make/approve requests, we should also fix the CI not to supply secrets to the build process.

It is not particularly hard, I have a vague recollection of doing something similar for another project (can't remember which one...). It requires time for fiddling to get the details exactly right.

Would make some suggestions, but right now, not exactly comprehending how our CI works. And I should be doing something else right  now :-)

JakobLichterfeld:
> has the repo setting for "Approval for running fork pull request workflows from contributors" been set to "Require approval for all external contributors" already?
> 
> would that already prevent any issues (besides someone from our own repo approving without fully checking for any malicious changes)?

It has, but this does not dolcecall cases. Someone do a normal PR, CI approved manual, which gets merged and the another malicious, no approval needed to run CI.

PR #5218, opened 2026-03-16
i18n: Update Korean translations in default.po

dongbum:
Update Korean translations in default.po


JakobLichterfeld (approved):
Thanks for your contribution!

PR #5219, opened 2026-03-16
ci: ensure setting the output for check_paths correctly

swiffer:
as per: https://docs.github.com/de/actions/reference/workflows-and-actions/workflow-syntax#example-of-onworkflow_calloutputs and system logs of github actions i think that check_paths never really limited execution of jobs ?

looking at the evaluation of `if: needs.check_paths.outputs.githubfolder != true` in old jobs:

https://github.com/teslamate-org/teslamate/actions/runs/23088937058/job/67070159871

```
2026-03-14T13:31:18.8750000Z Evaluating teslamate_build.if
2026-03-14T13:31:18.8750000Z Evaluating: (success() && (((needs.check_paths.outputs.githubfolder != 'true') && (needs.check_if_pr_from_outside_repo.outputs.is_pr_from_outside_repo == 'false'))))
2026-03-14T13:31:18.8750000Z Expanded: (true && (('' != 'true') && ('false' == 'false')))
```
the expanded expression should have been 
```
2026-03-14T13:31:18.8750000Z Expanded: (true && (('false' != 'true') && ('false' == 'false')))
```
note that false is missing here!

### DIFF
--- a/.github/workflows/buildx.yml
+++ b/.github/workflows/buildx.yml
@@ -26,7 +26,7 @@ jobs:
 
   teslamate_build:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder != 'true' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
     strategy:
       fail-fast: false
       matrix:
@@ -62,7 +62,7 @@ jobs:
     needs:
       - check_paths
       - teslamate_build
-    if: needs.check_paths.outputs.githubfolder != 'true' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -83,7 +83,7 @@ jobs:
             type=edge
   grafana:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder != 'true' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     steps:
--- a/.github/workflows/cleanup_caches.yml
+++ b/.github/workflows/cleanup_caches.yml
@@ -22,7 +22,7 @@ jobs:
   cleanup:
     name: Delete caches when PR is closed
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder != 'true'
+    if: needs.check_paths.outputs.githubfolder == 'false'
     runs-on: ubuntu-24.04
 
     steps:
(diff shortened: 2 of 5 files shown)


swiffer:
@brianmay - could you may have a look on this one?

JakobLichterfeld:
As far as I remember we use the trick '!= true' as GitHub does not return false but an empty string. It was working correctly in the past with this "trick".

swiffer:
Mh - the used action states otherwise and the workflow outputs is missing an outputs. Key so far for getting the job values per docs.

https://github.com/dorny/paths-filter?tab=readme-ov-file#outputs

Do you mind getting it merged and giving it a try in a follow up pr?

I stumpled upon it as the job step for checking if pr is from outside repo is not working anymore since i moved it to a separate workflow (by copying the code from check_paths)

JakobLichterfeld:
That's what I remember right now, took some iterations from what I recall.
Sure, I can only assist on weekend right now, but if you feel like giving it a try and multiple tests, I am not stopping you.

swiffer:
Will do - just wanted to get a review for this one as its touching github actions code again.

JakobLichterfeld:
Basically you changes should change anything innit? So == false is same as !=true but for an empty string the current solution is more robust and therefore choosen.

swiffer:
I think the condition currently always evaluates to true as the workflow is not outputting any value at all.

Changing this to == false just hardens it against any unexpected values.

The more relevant part is in the change of the check_paths workflow.

JakobLichterfeld:
I'm not convinced. The check_paths workflow does not need outputs syntax, it takes his value. But happy to be proofed otherwise.
!= true is more robust for empty strings

swiffer:
the ghcr_build workflow is now running correctly: https://github.com/teslamate-org/teslamate/actions/runs/23160446020

it hasn't before this change: https://github.com/teslamate-org/teslamate/actions/runs/23117150814/job/67144603731

a PR that tries to modify any files in the github folder is not executing the workflow at all:

https://github.com/teslamate-org/teslamate/pull/5220

this is not caused by check_paths but the file exclusions in the trigger rule from what i understand.

JakobLichterfeld:
Sounds perfect.
Runner details are not expandable on mobile so can't look in the details.

Yeah, the exclusion rule is the first level safety net. A PR with multiple commits needs to be checked on all commits and the exclusion rule does not from what I recall, therefore the double safety with dorny path filter, but need to look in the old PR comments.

swiffer:
Ok, will do another test from external to be sure!

JakobLichterfeld:
Ty. Basically a PR with non GitHub folder changes and consecutive commit with changes in GitHub folder.

brianmay:
Looking at the docs for https://github.com/dorny/paths-filter the recommended approach is `== 'true'`. However this is different, we are checking for false not true.

It does depend what you want to happen if we do get an empty string. I believe this could happen if there is a mistake in our CI (e.g. the check step wasn't run).

This will return false:

```
'true' == 'false' => false
'false' == 'false' => true
'' == 'false' => false
```

This will return true:

```
'true' != 'true' => false
'false' != 'true' => true
'' != 'true' => true
```

I think failing false might be safer, it makes it obvious something went wrong.

swiffer:
> Ty. Basically a PR with non GitHub folder changes and consecutive commit with changes in GitHub folder.

https://github.com/teslamate-org/teslamate/pull/5222

did that here - works as expected

PR #5226, opened 2026-03-18
docs: add optional MQTT username and password to Docker installation instructions

P6g9YHK6:
Comment out optional MQTT username and password.


JakobLichterfeld (changes requested):
Thanks for your contribution!

The installation docs will need to be as simple as possible. This is a potential source of error which will lead to negative user experience.
Therefore, I will close this PR without merging.

All available env variables are already documented: https://docs.teslamate.org/docs/configuration/environment_variables

PR #5229, opened 2026-03-21
fix: handle incomplete data correctly in Statistics dashboard

swiffer:
incomplete data is breaking first row stats in statistics dashboard.

any incomplete drives charges have no end date set - by default postgres is sorting nulls as:

https://www.postgresql.org/docs/current/queries-order.html#QUERIES-ORDER
```
By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.
```
we therefore get nulls on top (current day / month / ...) - that works if we only incomplete datapoint is the current drive / charge but fails for anything else.


swiffer:
@DrMichael / @AlexWhitehouse - could you please check (using Grafana Tag `pr-5229`) if this fixes the statistics dashboard in your case ?

https://docs.teslamate.org/docs/development#testing-with-our-ci-which-builds-the-docker-images-automatically-per-pr

before merging this one i will update other dashboards relying on this query as well.

DrMichael (approved):
Yes, looks good to me.

PR #5233, opened 2026-03-25
fix: fix folder creation and bash 3.2 compatibility in dashboards.sh

svennergr:
## Summary

- **Fix `create_folder` missing `metadata.generateName`**: The Grafana Kubernetes-style API (`/apis/folder.grafana.app/...`) requires either `metadata.name` or `metadata.generateName` in the request payload. The `create_folder` function was only sending `{spec: {title: ...}}`, which caused `422 Unprocessable Entity` errors when restoring dashboards to Grafana Cloud.
- **Fix bash 3.2 compatibility**: Replace `${array[-1]}` (negative array subscript, requires bash 4.2+) with `${array[${#array[@]}-1]}`, since macOS ships with bash 3.2 by default.

## How to reproduce

1. Set up a Grafana Cloud instance and create a service account token
2. Run `URL=https://<instance>.grafana.net GRAFANA_API_TOKEN=<token> GRAFANA_ORG_NAMESPACE=<namespace> ./grafana/dashboards.sh restore`
3. Before this fix: folder creation fails with 422, and the status echo fails on macOS with `bad array subscript`

## Test plan

- [x] Verified folder creation succeeds against Grafana Cloud with the `metadata.generateName` fix
- [x] Verified full `restore` completes successfully (23 dashboards across 3 folders)
- [x] Verified the bash 3.2 array subscript fix works on macOS (bash 3.2.57)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/grafana/dashboards.sh
+++ b/grafana/dashboards.sh
@@ -271,7 +271,7 @@ restore() {
 				-d "$final_payload" \
 				"$api_endpoint"
 
-			echo "RESTORED $(basename "$dashboard_path") into Grafana folder '${target_grafana_folder_titles_array[${#target_grafana_folder_titles_array[@]} - 1]}' (UID: $leaf_folder_to_assign_dashboard_uid) (Conceptual path: $(
+			echo "RESTORED $(basename "$dashboard_path") into Grafana folder '${target_grafana_folder_titles_array[${#target_grafana_folder_titles_array[@]}-1]}' (UID: $leaf_folder_to_assign_dashboard_uid) (Conceptual path: $(
 				IFS=/
 				echo "${target_grafana_folder_titles_array[*]}"
 			))"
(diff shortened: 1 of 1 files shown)


swiffer:
Thanks! Could we make use of the name property and use the same uids we use for docker based provisioning?
The uid seems to be scoped to the org, therefore that seems to be cleaner to me.

svennergr:
> Thanks! Could we make use of the name property and use the same uids we use for docker based provisioning? The uid seems to be scoped to the org, therefore that seems to be cleaner to me.

Added in 04bcffc496a3182d0d73bfb5a36885547e974004

svennergr:
@brianmay fixed CI in 0bc7b0fd. PTAL

swiffer:
😬 i think the overall script could be simplified a lot by making use of yq over jq and use the dashboards.yml file as a reference to backup / restore the dashboards but that might would be a larger rewrite.

could we may revert to the simple fix for now and open an issue / separate PR for switching to yq ?

svennergr:
Like this @swiffer ?

PR #5237, opened 2026-03-26
feat(webview): Add dark mode support for background and buttons in the map

olsoybakk:
- Added dark background on all maps (visible when you zoom out)
- ZoomControl has dark buttons on dashboard (not implemented in the geolocation map)

image


olsoybakk:
Dark buttons in dashboard map. Ok to implement, @swiffer ?

swiffer:
ok, to implement, thanks for your contribution!

given that the dark-theme sass is already scoped to the dark theme and only applied in that case i think the code can be simplified and we can use class based overwrites instead of adding additional classes for dark-mode (which would also present in light mode based on the initial implementation).

opened PR #5240 

reference: https://dev.to/deepakdevanand/leaflet-map-dark-theme-5ej0

PR #5240, opened 2026-03-29
feat(webview): Add dark mode support for background and buttons in the map

swiffer:


swiffer:
merging after review and concensus in #5237 - thanks @olsoybakk for your contribution!

PR #5259, opened 2026-04-05
refactor(vehicle): simplify the state machine: plain atom states, DB records moved into the state data

brianmay:
## Summary

This PR completes the state machine simplification by replacing compound state tuples like `{:driving, status, %Log.Drive{}}`, `{:charging, %Log.ChargingProcess{}}`, and `{:updating, %Log.Update{}}` with simple atoms (`:driving`, `:charging`, `:updating`). DB records and sub-states are stored in `Data` struct fields.

### Commit 1: Add preparatory fields to Data struct
Adds `current_drive`, `current_charging_process`, `current_update`, and `driving_status` fields.

### Commit 2: Populate Data struct fields in parallel with compound state tuples
Every transition into a compound state now also updates the corresponding struct fields.

### Commit 3: Migrate compound state reads to struct field guards
Replaces pattern-match reads with guards on the struct fields.

### Commit 4: Replace compound state tuples with simple atoms throughout state machine
**This is the main simplification commit.** After this:
- The GenState state is always a simple atom
- `{:driving, status, drive}` → `:driving` (status in `data.driving_status`, drive in `data.current_drive`)
- `{:charging, cproc}` → `:charging` (cproc in `data.current_charging_process`)
- `{:updating, update}` → `:updating` (update in `data.current_update`)

Key changes in this commit:
- All 12 `next_state` positions now use simple atoms
- Driving sub-state machine (available → unavailable → offline) is tracked in `data.driving_status` instead of the state tuple's second element
- All driving handlers match `:driving` with guards on `driving_status`
- `suspend_logging`, `fetch/can_fall_asleep`, and `fetch/reachable` all use simple atom matching
- `Summary` module updated to handle the new state format

## Refactoring sequence (complete!)

1. ✅ Replace `fake_online_state` integers with `pre_online_check` atoms
2. ✅ Add struct fields + populate them + migrate reads
3. ✅ **This PR**: Replace compound states with simple atoms

## Before vs After

**Before (hard to read):**
```elixir
def handle_event(:internal, {:update, {:offline, _}}, {:driving, {:unavailable, n}, drv}, ...)
def handle_event(:internal, {:update, {:offline, _}}, {:driving, {:offline, last}, drive}, ...)
def handle_event(:info, {:stream, %Stream.Data{} = stream_data}, {:driving, status, drv}, ...)
```

**After (clean):**
```elixir
def handle_event(:internal, {:update, {:offline, _}}, :driving, %Data{driving_status: {:unavailable, n}}) do ...
def handle_event(:internal, {:update, {:offline, _}}, :driving, %Data{driving_status: {:offline, last}}) do ...
def handle_event(:info, {:stream, %Stream.Data{}}, :driving, %Data{driving_status: :available, current_drive: drv}) do ...
```

## Testing
All commits are no-behaviour-change mechanical refactors. Existing tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,7 @@ Enjoy!
 - fix(charging): fall back to charger_power when phase detection fails (#5592 - @JakobLichterfeld)
 - fix(charges): enforce positive charger phases at the database (#5592 - @JakobLichterfeld)
 - fix(charging): recalculate charge_energy_used for existing processes (#5592 - @JakobLichterfeld)
+- refactor(vehicle): simplify the state machine: plain atom states, DB records moved into the state data (#5259 - @brianmay, @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,6 +188,7 @@ Enjoy!
 - build(deps): update flake.lock (#5560)
 - build(deps-dev): bump sass from 1.101.0 to 1.102.0 in /assets (#5566)
 - build(deps): bump leaflet-control-geocoder from 3.3.1 to 4.0.0 in /assets (#5565)
+- test(vehicle): make the store-position interval configurable and lock the state-machine field lifecycle with regression tests (#5259 - @JakobLichterfeld)
 
 #### Dashboards
 
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -468,7 +468,9 @@ defmodule TeslaMate.Vehicles.Vehicle do
         Logger.info("Vehicle is currently in service", car_id: data.car.id)
 
         case state do
-          {:driving, _, %Log.Drive{} = drive} ->
+          {:driving, _, _} when data.current_drive != nil ->
+            drive = data.current_drive
+
             {:ok, %Log.Drive{distance: km, duration_min: min}} =
               call(data.deps.log, :close_drive, [drive])
 
@@ -478,7 +480,8 @@ defmodule TeslaMate.Vehicles.Vehicle do
               car_id: data.car.id
             )
 
-            {:next_state, :start, %Data{data | last_used: DateTime.utc_now()},
+            {:next_state, :start,
+             %Data{data | last_used: DateTime.utc_now(), driving_status: nil, current_drive: nil},
              [
                broadcast_fetch(false),
                broadcast_summary(),
@@ -525,11 +528,18 @@ defmodule TeslaMate.Vehicles.Vehicle do
         end
 
         interval =
-          case state do
-            {:driving, _, _} -> 10
-            {:charging, _} -> 15
-            :online -> 20
-            _ -> 30
+          case {data.current_drive != nil, data.current_charging_process != nil} do
+            {true, _} ->
+              10
+
+            {_, true} ->
+              15
+
+            _ ->
+              case state do
+                :online -> 20
+                _ -> 30
+              end
           end
 
         {:keep_state, data,
@@ -1206,7 +1216,8 @@ defmodule TeslaMate.Vehicles.Vehicle do
         {:driving, {:offline, _last}, nil},
         %Data{} = data
       ) do
-    {:next_state, :start, %Data{data | last_used: DateTime.utc_now()}, schedule_fetch(data)}
+    {:next_state, :start, %Data{data | last_used: DateTime.utc_now(), driving_status: nil},
+     schedule_fetch(data)}
   end
 
   def handle_event(
@@ -1269,13 +1280,13 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
         Logger.info("Vehicle was charged while being offline: #{added} kWh", car_id: data.car.id)
 
-        {:next_state, :start, %{data | last_used: DateTime.utc_now()},
+        {:next_state, :start, %{data | last_used: DateTime.utc_now(), driving_status: nil},
          {:next_event, :internal, {:update, {:online, now}}}}
 
       not has_gained_range? and offline_min >= @drive_timeout_min ->
         unless is_nil(drv), do: timeout_drive(drv, data)
 
-        {:next_state, :start, %{data | last_used: DateTime.utc_now()},
+        {:next_state, :start, %{data | last_used: DateTime.utc_now(), driving_status: nil},
          {:next_event, :internal, {:update, {:online, now}}}}
 
       not is_nil(drv) ->
@@ -1289,7 +1300,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
   def handle_event(:internal, {:update, {:asleep, _vehicle}}, {:driving, _, drv}, data) do
     unless is_nil(drv), do: timeout_drive(drv, data)
-    {:next_state, :start, data, schedule_fetch(data)}
+    {:next_state, :start, %{data | driving_status: nil}, schedule_fetch(data)}
   end
 
   #### msg: :online
@@ -1345,7 +1356,8 @@ defmodule TeslaMate.Vehicles.Vehicle do
         Logger.info("End of drive initiated by: #{inspect(vehicle.drive_state)}")
         Logger.info("Driving / Ended / #{km && round(km)} km – #{min} min", car_id: data.car.id)
 
-        {:next_state, :start, %{data | last_used: DateTime.utc_now(), geofence: geofence},
+        {:next_state, :start,
+         %{data | last_used: DateTime.utc_now(), driving_status: nil, geofence: geofence},
          {:next_event, :internal, {:update, {:online, vehicle}}}}
 
       %Vehicle{drive_state: nil} ->
@@ -1513,13 +1525,13 @@ defmodule TeslaMate.Vehicles.Vehicle do
             :online ->
               true
 
-            {:driving, _, _} ->
+            {:driving, _, _} when data.current_drive != nil ->
               true
 
-            {:updating, _} ->
+            {:updating, _} when data.current_update != nil ->
               true
 
-            {:charging, _} ->
+            {:charging, _} when data.current_charging_process != nil ->
               true
 
             :start ->
@@ -1545,9 +1557,9 @@ defmodule TeslaMate.Vehicles.Vehicle do
         reachable? =
           case expected_state do
             :online -> true
-            {:driving, _, _} -> true
-            {:updating, _} -> true
-            {:charging, _} -> true
+            {:driving, _, _} when data.current_drive != nil -> true
+            {:updating, _} when data.current_update != nil -> true
+            {:charging, _} when data.current_charging_process != nil -> true
             :start -> false
             {:offline, _} -> false
             {:asleep, _} -> false
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -1534,13 +1534,13 @@ defmodule TeslaMate.Vehicles.Vehicle do
             :online ->
               true
 
-            :driving when data.current_drive != nil ->
+            :driving ->
               true
 
-            :updating when data.current_update != nil ->
+            :updating ->
               true
 
-            :charging when data.current_charging_process != nil ->
+            :charging ->
               true
 
             :start ->
@@ -1566,9 +1566,9 @@ defmodule TeslaMate.Vehicles.Vehicle do
         reachable? =
           case expected_state do
             :online -> true
-            :driving when data.current_drive != nil -> true
-            :updating when data.current_update != nil -> true
-            :charging when data.current_charging_process != nil -> true
+            :driving -> true
+            :updating -> true
+            :charging -> true
             :start -> false
             {:offline, _} -> false
             {:asleep, _} -> false
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -525,18 +525,11 @@ defmodule TeslaMate.Vehicles.Vehicle do
         end
 
         interval =
-          case {data.current_drive != nil, data.current_charging_process != nil} do
-            {true, _} ->
-              10
-
-            {_, true} ->
-              15
-
-            _ ->
-              case state do
-                :online -> 20
-                _ -> 30
-              end
+          case state do
+            :driving -> 10
+            :charging -> 15
+            :online -> 20
+            _ -> 30
           end
 
         {:keep_state, data,
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -36,7 +36,17 @@ defmodule TeslaMate.Vehicles.Vehicle do
               #   :probing          – stream connected, waiting for first power reading
               #   :confirmed_fake   – stream reported power=nil, treating as fake online
               #   :confirmed_real   – stream reported numeric power, treating as real online
-              pre_online_check: :idle
+              pre_online_check: :idle,
+              # Future fields for extracting DB records from state tuples (not used yet)
+              # These will allow moving from {:driving, status, %Log.Drive{}} to :driving + data fields
+              # %Log.Drive{} | nil (will replace drive in {:driving, _, drive})
+              current_drive: nil,
+              # %Log.ChargingProcess{} | nil (will replace cproc in {:charging, cproc})  
+              current_charging_process: nil,
+              # %Log.Update{} | nil (will replace update in {:updating, update})
+              current_update: nil,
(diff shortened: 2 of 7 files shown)


brianmay:
This is only the first phase so far, we haven't seen the simplification to the compound state tuples yet. That will come later.

brianmay:
As before, this was done with AI. It was going to make one big pull request, then we agreed that breaking it up might be better.

brianmay:
I decided to do the full refactor in one pull request. While this the PR more complicated, don't think you can really review this probably unless you can see the entire picture.

The steps are still there, but as separate commits. Which probably should stay as separate commits.

Key questions before merging:

1. Does this make things simpler and easier to understand?

I think by making the state an enum only and putting state data into the data variable we have made it easier to understand. If anyone disagrees, feel free to speak out :-)

2. Is it correct?

It looks OK to me... Although certainly is scope for errors to creep in. Guess next step will be to test this.

brianmay:
I spotted a place where the logic had been changed and was more complicated then required. Fixed now.

brianmay:
All tests are expected to pass.

Still need to manually review the "Fix compound state tuple handling after dce479c refactoring" commit - haven't done that yet.

brianmay:
Might make sense to merge the last three commits into the "Migrate compound state reads to struct field guards" commit, seeing as all they do is fix problems caused by this commit.

Will decide when more awake :-)

brianmay:
I rebased this against the main branch, now that mcu2-upgraded-cars has been merged into main.

brianmay:
I got Claude Fable 5 to review this (while I still had access to Fable5!). I haven't had a chance to seriously evaluate its feedback yet, so will just post it here in its entirety so I don't accidentally lose it.

>   PR 5259 Review: extract DB records from compound state tuples
> 
>   Overview
> 
>   Final phase of the state-machine refactor: {:driving, status, drive},
>   {:charging, cproc}, {:updating, update} become plain atoms, with the records
>   moved to new Data fields (current_drive, current_charging_process,
>   current_update, driving_status). 2 files, +124/−75. CI is green (lint job
>   still pending).
> 
>   Correctness
> 
>   I traced every converted clause against the old behavior and found no
>   functional regressions:
> 
>   - The tricky {:driving, {:offline, last}, nil} drive-timeout case maps
>   correctly to driving_status: {:offline, last}, current_drive: nil and still
>   routes to :start on the next offline event (vehicle.ex:1218 vs 1236).
>   - Summary.format_state(:driving, {:offline, _}) → :offline preserves the old
>   special case, and the {state, _} clause still covers the remaining tuple
>   states ({:offline, _}, {:asleep, _}, {:suspended, _}).
>   - No leftover compound-state patterns anywhere in lib/ or test/, and both
>   Summary.into call sites pass the new required driving_status key.
> 
>   Main issue: stale record fields on exit (latent, worth fixing now)
> 
>   The fields are cleared on some exits but not others, so the invariant "field
>   mirrors state" — the whole point of the refactor — doesn't actually hold:
> 
>   - current_charging_process is never cleared when charging completes (:charging
>   → :start, vehicle.ex:1170). Only the asleep-while-charging path clears it
>   (line 1132).
>   - current_update is never cleared on either :updating → :start path (cancel at
>   ~1412, finish at ~1434).
>   - current_drive survives most :driving → :start exits: end-of-drive (1364),
>   gained-range (1283), offline-timeout (1289), and asleep (1308) all clear
>   driving_status but leave the closed drive in current_drive. Only the
>   service-mode path clears both (481).
> 
>   This is safe today because every entry into a state overwrites its field and
>   all readers guard on the state atom — but it's exactly the kind of footgun
>   this refactor exists to remove. The service-mode handler already reads
>   data.current_drive != nil as a guard (line 468), so a future reader doing the
>   same in another state would see a stale record. Suggestion: clear all three
>   fields (and driving_status) at every transition to :start — a small helper
>   like reset_activity(data) would make this impossible to forget.
> 
>   Smaller points
> 
>   - driving_status sentinel is inconsistent. The struct default is :available,
>   the comment documents :available | {:unavailable, n} | {:offline, 
>   last_vehicle}, but every exit path sets nil. Pick one "not driving" value —
>   nil with a default of nil reads best — and update the comment to include it.
>   - Redundant self-assignment at line 1154: %{data | current_charging_process: 
>   cproc} where cproc was just extracted from data — it's a no-op; plain data is
>   clearer. Same pattern would apply if you keep the keep-charging clause as is.
>   - Summary now silently couples to the caller: into/2's attrs map pattern-match
>   makes driving_status required, which is fine for the two internal callers,
>   but a brief @doc note or typespec update on Summary.into would help.
> 
>   Test coverage
> 
>   No test changes, and no tests reference the old tuples or new fields — they
>   exercise the public API, so the passing suite is a genuine
>   behavior-preservation signal. However, the field lifecycle (the staleness
>   issue above) is exactly the part with zero coverage. A test asserting the
>   records are cleared after a drive/charge/update ends would lock in the
>   invariant.
> 
>   Verdict
> 
>   Solid mechanical refactor with a clear before/after win in readability, and
>   the commit sequencing (populate → migrate reads → swap states) made it easy to
>   verify. I'd fix the stale-field cleanup and the nil/:available inconsistency
>   before merging, since they undermine the invariant the new design depends on;
>   the rest is optional polish.
> 
> ✻ Cooked for 2m 29s
> 
> ※ recap: I reviewed TeslaMate PR 5259, the state-machine refactor replacing compound state tuples with atoms. It's correct overall, but stale record fields aren't cleared on most state exits. Next step: decide whether to fix that cleanup before merging. (disable recaps in /config)
>

JakobLichterfeld:
> I got Claude Fable 5 to review this (while I still had access to Fable5!).

TY!

> I traced every converted clause against the old behavior and found no
> functional regressions:

Nice!


> The fields are cleared on some exits but not others, so the invariant "field
> mirrors state" — the whole point of the refactor — doesn't actually hold:

Not so nice :-)

> Suggestion: clear all three
> fields (and driving_status) at every transition to :start — a small helper
> like reset_activity(data) would make this impossible to forget.

Sounds reasonable.

> I'd fix the stale-field cleanup and the nil/:available inconsistency
> before merging, since they undermine the invariant the new design depends on;
> the rest is optional polish.

Agree

JakobLichterfeld:
Thanks again Brian! I did another in-depth pass on the current head (c3e626ba), building on the review posted in June — the points agreed there are still open, and the deep pass surfaced two things that review missed, one of them a real regression.

The good news first: the pattern-match conversion itself is complete — no leftover compound-state patterns in any match, and the clause ordering of the converted `:driving` handlers checks out.

**1. Regression: positions are no longer stored while charging (`vehicle.ex:921-922`)**

The `store_position` timeout guard was not converted:

```elixir
def handle_event({:timeout, :store_position}, :store_position, state, data)
    when state == :online or (is_tuple(state) and elem(state, 0) == :charging) do
```

`:charging` is a plain atom now, so `is_tuple(state)` is false and the charging branch can never match. When the 5-minute timer fires during a charging session, the no-op clause below consumes the event without re-arming it — no positions are stored for the entire session (previously one every 5 min), and the timer chain stays dead until the next `:online` entry. The fix is a one-liner (`when state in [:online, :charging]`), but please add a test with it — the suite being green despite this regression is exactly the coverage gap flagged in June.

**2. Behavior change: stream frames while driving-unavailable are now ignored (`vehicle.ex:668`)**

The driving stream handler now only matches `driving_status: :available`. The old clause matched any status and scheduled an immediate fetch for the non-available ones. Now, while `:driving` with `{:unavailable, n}` or `{:offline, last}` (polling API says offline, stream still delivering — a real scenario), every stream frame falls through to the generic catch-all, which logs "Received stream data: …" at `:info` (roughly one frame per second while driving) and no longer triggers `schedule_fetch(0)` — recovery waits for the 5s/30s polling cycle instead of reacting to the stream. Please either restore a clause for the non-available statuses or make ignoring them explicit and quiet.

**Still open from the June review (as agreed):**

- Clear `current_drive` / `current_charging_process` / `current_update` plus `driving_status` on every transition to `:start` (the reset-helper approach), so the "field mirrors state" invariant actually holds — e.g. `current_charging_process` survives charging completion (`vehicle.ex:1170`), `current_update` both `:updating` exits, `current_drive` most driving exits.
- `driving_status` sentinel: struct default is `:available`, all exits set `nil` — pick `nil` as the only "not driving" value and update the struct comment.
- Nit: `%{data | current_charging_process: cproc}` at `vehicle.ex:1154` is a self-assignment; plain `data` is clearer.
- Tests locking the field lifecycle (records cleared after a drive/charge/update ends).

One heads-up for the rebase (the branch currently conflicts with main): main has moved in exactly this area — e.g. #5535 (mid-drive offline + stream reconnect), #5515 (geofence updates while driving), #5417 (streaming-aware interval on drive start) — so the converted clauses will need care there, particularly around point 2.

To get this over the line, I'll push commits addressing the points above directly to this branch (starting with the rebase onto current main, then the fixes). 

---

🤖 Review drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
Update: I rebased this onto current main and pushed commits addressing all points from the review above.

**Rebase:** The conflicts were the expected ones — main moved in exactly this area. The semantics of #5535 (stream reconnect on drive resume), #5515 (geofence updates while driving) and #5417 (streaming-aware interval) are preserved in the converted clauses. Main's new `Summary` unit tests (#5487) now pass the required `driving_status` key; that adaptation is folded into the "Migrate compound state reads" commit to keep the branch bisectable.

**Fixes on top:**

- 474786f9 `fix(vehicle): keep periodic position storing while charging` — the `store_position` guard still expected the `{:charging, _}` tuple and never matched the new atom, so no positions were stored during a charging session. Comes with a regression test (verified red against the broken guard) and a test-configurable `:store_position_interval` start option.
- 68fe7fee `fix(vehicle): restore immediate fetch on stream data while unavailable` — stream frames while `:driving` in `{:unavailable, n}` / `{:offline, last}` fell through to the generic catch-all (info-log spam, no immediate fetch). Behavior is back to pre-refactor: a frame while unavailable triggers an immediate fetch.
- e941f5f4 `refactor(vehicle): reset activity fields on every transition to :start` — the agreed `reset_activity/1` helper now runs on all 17 transitions to `:start`, `nil` is the single "not driving" sentinel for `driving_status`, the no-op self-assignment is gone, and the drive/charge/update flow tests assert the fields are cleared once the vehicle is back `:online`.

Full suite is green (461 tests). @brianmay from my side this is now ready — please give it a look, and shout if you disagree with any of the calls above (especially restoring the immediate-fetch behavior rather than dropping it).

---

🤖 Fixes drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld (approved):

brianmay:
Review by opencode/Big Pickle:

> ## Review Summary: PR 5259 (simplify vehicle state machine to plain atoms)
> 
> **Verdict: no bugs found.** Mechanical, well-executed refactor; verified by applying the diff, compiling with `--warnings-as-errors`, and running the 124-test vehicle suite (all pass).
> 
> **Verification highlights:**
> - All 17 `{:next_state, :start, ...}` transitions call `reset_activity/1` (needed since DB sub-state left the dropped state tuple) — none missed.
> - No stale `{:driving, _, _}` / `{:charging, _}` / `{:updating, _}` patterns remain.
> - Split driving-stream handler (vehicle.ex:707-759) is behavior-equivalent to the old `case`; `driving_status: :available` always implies `current_drive != nil`.
> - Sub-state clause ordering (e.g. `{:unavailable, n}` with `n < 15` before catch-all) preserves old match order.
> - `Summary.format_state/2` covers all remaining state shapes; dropping the 3-tuple clause is correct.
> 
> **Minor issues:**
> 1. **Test efficacy (low):** the new `keeps storing positions periodically while charging` test asserts only two `insert_position` events, but two independent store-position timers (online-entry + charging-entry) can each fire once to satisfy it — a broken reschedule wouldn't be caught. Assert a third event to actually exercise "periodically."
> 2. **Leftover tuple state in summary_test.exs:** `state: {:online, nil}` no longer occurs in production (states are now atoms); harmless but tests an unreachable shape.
> 3. **Pre-existing observation:** the double store-position timer (online + charging both armed) causes doubled position inserts while charging; predates this PR but the new test makes it visible.
> 
> No further action needed — nothing blocking in the PR.

brianmay (commented):
Looks good to me.

Looks like I can't approve my own pull request :-)

I think this needs to be carefully tested, simply because of the invasive changes into the state machine logic.

d1stru3t0r:
So move it to 4.2 to have a surprise 4.1 release these hours? :P

JakobLichterfeld:
Thanks for running another review over it! The verification section matches my own checks. On the minor issues:

**1 & 3 rest on a wrong premise:** there is no second store-position timer. Generic timeouts in `gen_statem` are keyed by name — setting `{:timeout, :store_position}` again *replaces* the running timer (see the `gen_statem` docs: "Setting a timer with the same Name while it is running will restart it with the new time-out value.", see https://www.erlang.org/doc/apps/stdlib/gen_statem.html#t:generic_timeout/0). So entering `:charging` replaces the timer armed at `:online` entry, and there are no doubled inserts — before or after this PR. The new test also drains the mailbox *after* `:charging` is established, so the two asserted inserts can only come from that single timer, and the second one requires the handler to have re-armed — a broken reschedule fails the test.

**2 is fair:** `state: {:online, nil}` in `summary_test.exs` was never a real state shape (it came in with #5487, independent of this PR). I'll change it to `:online` in passing.

Nothing blocking from my side either.

---

🤖 Reply drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
> I think this needs to be carefully tested, simply because of the invasive changes into the state machine logic.

Agreed on careful testing — the state machine is the heart of the logger. When running this branch for a few days, specifically watching: positions being stored every 5 min during charging (the regression fixed here), recovery behavior on mid-drive connectivity loss, clean field resets after drive/charge/update (`:sys.get_state`), and the logs for any `FunctionClauseError` or stream-data spam. If anything surfaces report here.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
> So move it to 4.2 to have a surprise 4.1 release these hours? :P

Ok, as requested :-) https://github.com/teslamate-org/teslamate/releases/tag/v4.1.0 and https://github.com/teslamate-org/teslamate/releases/tag/v4.1.1

d1stru3t0r:
> > So move it to 4.2 to have a surprise 4.1 release these hours? :P
> 
> Ok, as requested :-) https://github.com/teslamate-org/teslamate/releases/tag/v4.1.0 and https://github.com/teslamate-org/teslamate/releases/tag/v4.1.1


haha ty, I was mid trip when I tried to update my docker and I bricked everything xd

JakobLichterfeld:
> haha ty, I was mid trip when I tried to update my docker and I bricked everything xd

Quoting myself: 
> yeah, I can hear it myself as I type this.

Same for you 😀

d1stru3t0r:
Ok, i think it got extremele fricked up. I've read the stuff about backup but at this point I feel like i can't do anything with it. Do you know the simple way to backup the database, as I'd want to delete my entire docker, reinstall and then reuse my backup.

I can either use teslamate without database now OR sync database but then teslamate doesn't work. No idea what got so ruined and I'm tired of so much debugging nonsense on termux

brianmay:
@d1stru3t0r This probably deserves a [new discussion item](https://github.com/teslamate-org/teslamate/discussions/new/choose).

d1stru3t0r:
> @d1stru3t0r This probably deserves a [new discussion item](https://github.com/teslamate-org/teslamate/discussions/new/choose).

Small noobie question, my 3000 work now but reports a very old status, do i have to regenerate tesla tokens? Or what tf is happening xd

JakobLichterfeld:
Please open a discussion as mentioned by Brian or an Issue, we do not know anything about what you did, what your logs show etc. This PR is not the right place to discuss your issues. Thank you.

JakobLichterfeld:
Rebased onto `main` after Phase 1 of the rework (#5652). The conflicts were all of one kind: `main` moved the vehicle's view of time onto the clock seam (#5688, #5689, #5706) on the same lines this branch adds the struct fields and `reset_activity/1`; resolved by keeping the clock expressions and the branch's changes. The two changelog lines had landed in the 4.1.0 section by context and were moved to `[unreleased]` inside their commits.

Proof: the characterization suite compares unchanged — 212 tests, 167 goldens, none re-recorded — so the refactoring is behavior-neutral against current `main`. The vehicle mock tests pass as well (126).

---

Issue #5231, opened 2026-03-23
High CPU load near 100% constant.


georgemaxim:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

Hello,

Since couple of weeks back teslaMate has been hammering my poor old cpu. I do not know exactly since when. It gotten so bad that I had to stop the stack until other updated, I tried to look it other people have the same issue but could not find any. I am running it on a Synology DSM.

The problem is the DB container, taking 98% - 100% of CPU.

Ill attach my compose file.

version: "3"

services:
  teslamate:
    image: teslamate/teslamate:latest
    # or pin to a current tag, e.g.:
    # image: teslamate/teslamate:2.1.1    container_name: TeslaMate
    container_name: TeslaMate
    networks:
      - teslamate
    restart: always
    environment:
      - ENCRYPTION_KEY=*
      - DATABASE_USER=teslamate
      - DATABASE_PASS=*
      - DATABASE_NAME=teslamate
      - DATABASE_HOST=database
      - MQTT_HOST=mosquitto
    ports:
      - 4000:4000
    volumes:
      - ./import:/opt/app/import
    cap_drop:
      - all

  database:
    image: postgres:17
    # optionally pin a minor if you want the release-note minimum:
    # image: postgres:17.3    container_name: TeslaMate_Database
    container_name: TeslaMate_Database
    networks:
      - teslamate
    restart: always
    environment:
      - POSTGRES_USER=teslamate
      - POSTGRES_PASSWORD=*
      - POSTGRES_DB=teslamate
    volumes:
      - teslamate-db:/var/lib/postgresql/data

  grafana:
    image: teslamate/grafana:latest
    container_name: TeslaMate_Grafana
    networks:
      - teslamate
    restart: always
    environment:
      - DATABASE_USER=teslamate
      - DATABASE_PASS=*
      - DATABASE_NAME=teslamate
      - DATABASE_HOST=database
    ports:
      - 3000:3000
    volumes:
      - teslamate-grafana-data:/var/lib/grafana
      - /volume1/docker/teslamate/custom_dasboards/customdashboards.yml:/etc/grafana/provisioning/dashboards/customdashboards.yml
      - /volume1/docker/teslamate/custom_dasboards/dashboards:/TeslamateCustomDashboards

  mosquitto:
    image: eclipse-mosquitto:2
    container_name: TeslaMate_MQTT
    networks:
      - teslamate
[...]

PR #5075, opened 2025-12-25
perf: Replace positions indexes from BTREE to BRIN to reduce memory usage

ilya-y-synth:
I noticed that the indexes in the position table take up the same amount of space as the data, although they are already sorted by date and new data is only added to the end, which means BTREE can be replaced with a more memory efficient BRIN index.

Raw data size: 50.5 MiB
Indexes: 
**Before:** 

|   |   |
|---|---|
| positions_drive_id_date_index  | 15 MB  |
| positions_date_index  | 11 MB  |
| positions_pkey  | 11 MB  |
| positions_car_id_index  | 3512 kB  |
| positions_car_id_date__ideal_battery_range_km_IS_NOT_NULL_index  | 592 kB  |

Total: 41 MB

**After:**

|   |   |
|---|---|
| positions_drive_id_date_brin_index  | 24 kB  |
| positions_date_brin_index  | 24 kB |
| positions_pkey  | 11 MB  |
| positions_car_id_index  | 3512 kB  |
| positions_car_id_date__ideal_battery_range_km_IS_NOT_NULL_index  | 592 kB  |

Total: 15.1 MB (-63%)

PR #5276, opened 2026-04-11
fix: improve brin indexes (autosummarize + timestamp_minmax_multi_ops)

swiffer:
# Tune BRIN indexes on `positions` table for better long-term performance

## Related Issues / PRs

- Replaces indexes changed in #5075 (BTREE → BRIN on `positions` table)
- Addresses performance degradation reported in #5231

## Description

In #5075 we switched two indexes on the `positions` table from BTREE to BRIN to achieve significant storage savings. While this worked well initially, BRIN indexes can degrade over time as new data is inserted (especially with unsummarized ranges), leading to slower query performance.

This PR tunes the BRIN indexes to improve resilience against degradation:

- **Enable `autosummarize = true`**: New ranges at the end of the table are automatically summarized by autovacuum, keeping the index effective without manual intervention.
- **Use `timestamp_minmax_multi_ops` for date column**: Stores multiple min/max values per block instead of a single pair. This helps handle outliers better while still benefiting from the extremely high correlation observed on the date column. The index remains very small, so the modest size increase is acceptable.
- **Set `pages_per_range = 64`**: Reduces the number of rows per range (roughly from ~10,000 to ~5,000 on typical instances), providing a better balance between index precision and size.

### Impact on index size
On a test instance with ~10 million rows, the index grew from ~54 KB to ~568 KB. This is still negligible and considered a worthwhile trade-off for sustained query performance.

The goal of these changes is to prevent the indexes from degrading as quickly as they did before, while preserving most of the storage benefits of BRIN.

## Changes

- Updated the index definition via migration with the new storage parameters.
- No changes to application code or queries.

## How to Test / Review

1. On your current TeslaMate instance (before switching to this PR's image), run the following queries and note the execution times:

```sql
SELECT d0."id" FROM "drives" AS d0 
WHERE (((d0."start_date" > '2026-04-01') 
AND EXISTS((SELECT 1 FROM "positions" AS sp0 
WHERE (sp0."drive_id" = d0."id") AND (sp0."id" > 0))))
AND NOT (EXISTS((SELECT 1 FROM "positions" AS sp0 
WHERE ((sp0."drive_id" = d0."id") AND (sp0."id" > 0)) 
AND (sp0."odometer" IS NOT NULL) 
AND (sp0."ideal_battery_range_km" IS NULL)))));
```

```sql
SELECT attname, correlation, n_distinct
FROM pg_stats
WHERE tablename = 'positions'
  AND attname IN ('drive_id', 'date')
ORDER BY attname;
```
2. Switch to the Docker image built from this PR (or apply the migration).
3. Let the database migration run (it will recreate/rebuild the affected indexes).
4. Re-run the queries above and compare execution times.

**Note:** Recreating the indexes should immediately improve performance (observed improvement from ~2.5s to ~0.04s on one instance). The new settings are designed to maintain that performance longer as data grows.
You can also manually trigger summarization for testing:

```sql
SELECT brin_summarize_new_values('positions_drive_id_date_index'::regclass)
       AS unsummarized_ranges_summarized;
```

## Review Request

@ilya-y-synth — as the original author of #5075, could you please take a look?

## Checklist

- [x] Migration successfully applies and rebuilds the indexes
- [x] Queries show improved or stable performance after migration
- [x] Index size remains reasonable for large tables (~10M+ rows)
- [x] No unintended impact on insert/update performance

Feel free to share your before/after timings and any observations!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - feat(webview): Make icons inherit text color for better contrast in dark mode (#5193 - @olsoybakk)
 - feat(webview): Add dark mode support for background and buttons in the map (#5240 - @olsoybakk and @swiffer)
 - fix(webview): Prevent rounding of map tiles via Bulma CSS (#5265 - @swiffer)
+- perf: ensure BRIN indexes don't degrade over time (#5276 - @swiffer)
 
 #### Build, CI, internal
 
--- a/priv/repo/migrations/20260411070212_improve_brin_indexes_on_positions.exs
+++ b/priv/repo/migrations/20260411070212_improve_brin_indexes_on_positions.exs
@@ -1,0 +1,24 @@
+defmodule TeslaMate.Repo.Migrations.ImproveBrinIndexesOnPositions do
+  use Ecto.Migration
+
+  def change do
+    # Drop existing BRIN indexes
+    drop_if_exists(index(:positions, [:drive_id, :date]))
+    drop_if_exists(index(:positions, [:date]))
+
+    # Create new BRIN indexes with optimized options
+    create(
+      index(:positions, ["date timestamp_minmax_multi_ops"],
+        using: "brin",
+        options: "autosummarize = true, pages_per_range = 64"
+      )
+    )
+
+    create(
+      index(:positions, [:drive_id, "date timestamp_minmax_multi_ops"],
+        using: "brin",
+        options: "autosummarize = true, pages_per_range = 64"
+      )
+    )
+  end
+end


swiffer:
@georgemaxim confirmed in #5231 that manually updating the index by running 

`SELECT brin_summarize_new_values('positions_drive_id_date_index'::regclass) AS unsummarized_ranges_summarized;`

greatly improved performance in his case. all instances will be affected sooner or later without autosummarize enabled. this change seems ready to be merged.

ilya-y-synth (approved):
Hi @swiffer, oops, I missed that😅 
Looks good to me, thanks for fixing this

PR #5289, opened 2026-04-18
feat: add service mode to webview and reduce log when car is `Unlocked` at service mode

NirKli:
### Description

TeslaMate currently logs repeated `[warning] Unlocked` messages at a high frequency (~every 30 seconds) when the vehicle remains unlocked.

This becomes especially noisy when the vehicle is in **service mode**, where it is expected to stay unlocked for extended periods.

---

### Changes

- Add `service_mode` field to `TeslaApi.Vehicle.State` (parsed from Tesla API)
- Reduce repeated `Unlocked` warning logs when `service_mode = true`
- Keep existing behavior when the vehicle is not in service mode
- Add service mode indicator (🔧) to the web UI

---

### Behavior

- `Unlocked` warnings are logged normally when not in service mode  
- When `service_mode` is enabled, repeated logs are suppressed to avoid noise  

---

### Screenshots

docker_loh
Untitled


brianmay:
I imagine this might (?) conflict with #4453. Shouldn't be hard to resolve either way, just wanted to flag this as a possible issue.

JakobLichterfeld:
> I imagine this might (?) conflict with #4453. Shouldn't be hard to resolve either way, just wanted to flag this as a possible issue.

@NirKli may I ask you to double check and resolve the ocnfilic?

NirKli:
@JakobLichterfeld — just to clarify, are you referring to a git conflict or a potential logic conflict? On the git side, I've rebased onto the latest main and it's now conflict-free. 
On the logic side, the service mode changes (suppressing the unlocked warning and the `service_mode?` helper) are in a separate area of `vehicle.ex` from the fake online detection, so I don't see an overlap — but happy to double-check if you had something specific in mind.

brianmay:
I am wondering if this "Unlocked" message should just get completely deleted or downgraded to Debug level?

Does it actually provide any benefit?

Actually maybe some sort of rate limited for all  the messages in this function would be ideal. Only needs to be logged once (every x minutes?), not every time try_to_suspend is called.

NirKli:
@brianmay — I think keeping it but downgrading to `Logger.debug` makes sense — it's still useful for troubleshooting without cluttering normal logs. I've made that change locally. 
What do you think?

JakobLichterfeld:
> just to clarify, are you referring to a git conflict or a potential logic conflict?

I was referring to a potential logic conflict with the new fake wakeup detection for MCU2 upgraded cars. Exactly what you assumed:

> On the logic side, the service mode changes (suppressing the unlocked warning and the `service_mode?` helper) are in a separate area of `vehicle.ex` from the fake online detection, so I don't see an overlap — but happy to double-check if you had something specific in mind.

So I just want to be extra safe here, as Brian pointed out.

JakobLichterfeld:
> downgrading to `Logger.debug` makes sense

On the one hand I do agree, on the other hand, I do like to see service mode in normal log. Why? There are some users which complain about no data collected but in fact the car was in service mode which suppress data collection.

NirKli:
@JakobLichterfeld @brianmay — I see both points here. 
A possible middle ground: 
keep `Logger.warning` (or `Logger.info`) when the car is in service mode so users can see why data isn't being collected, and downgrade to `Logger.debug` (or remove entirely) for the plain unlocked case when not in service mode.
That way the service mode case stays visible in normal logs while reducing noise for the unlocked case. 
What do you think?

JakobLichterfeld:
We do have a setting "Vehicle must be locked". So as a user, my expecation would be:
- Vehicle must be locked set to true: normal log entry with warning, car is unlocked
- Vehicle must be locked set to false: only debug log entry
- Vehicle is in service mode: In logs I do see car is in service mode, no data collected, once at beginning once at end.

NirKli:
@JakobLichterfeld I implemented the three cases you described:

- **"Vehicle must be locked" = true** → `Logger.warning` "Unlocked ..." (unchanged behavior)
- **"Vehicle must be locked" = false** → `Logger.debug` "Unlocked ..." only
- **Service mode** → `Logger.info` once when entering ("Car is in service mode, data collection suspended") and once when leaving ("Car left service mode, data collection resumed")

One design note on the service mode part: detecting *entering/leaving* requires comparing each fetch to the previous one, so the check inherently has to run on every `vehicle_data` fetch. I placed it at the point where the fetch result is handled, because that's the single spot where both the previous response (`data.last_response`) and the current `vehicle` are available side by side — everywhere else `last_response` has already been overwritten. The check itself is just a couple of pattern matches and only emits a log when the value actually flips, so it logs exactly once per transition rather than every cycle.

The alternative was to track a `service_mode?` flag in the state struct and detect inside `try_to_suspend`, but several branches there return `{:keep_state_and_data, ...}` and don't carry data forward, so the flag would need threading through each one (and missing one would log every cycle). I went with the simpler, single choke-point approach — happy to refactor to the flag-based version if you'd prefer.

JakobLichterfeld:
> I implemented the three cases you described:

Awesome, ty!

> I went with the simpler, single choke-point approach — happy to refactor to the flag-based version if you'd prefer.

That sounds reasonable to me. Let's see what Brian thinks about it.

brianmay:
AI had some conerns:

## Review of PR #5289 — Service mode indicator + Unlocked log noise reduction

### Bug — Misleading log message text (medium)

`lib/teslamate/vehicles/vehicle.ex` logs "Car is in service mode, **data collection suspended**" / "...resumed", but the code does not actually suspend polling. Data continues to be fetched; only the *warning* log is suppressed. Operators reading this will believe polling has stopped. Recommend rewording to reflect actual behavior (e.g. "suppressing unlocked warning").

### Issue — Wrong translations for "Service Mode" (low)

Most locale `.po` files add `msgid "Service Mode"` with a **fuzzy** `msgstr` that actually translates *sleep mode* (e.g. `de` → "Schlafmodus", `ja` → "スリープモード", `zh_Hans` → "休眠模式"). These should be cleared to empty msgstrs (matching `en`/`default.pot`) to avoid surfacing "Sleep Mode" next to the service-mode wrench.

### Notes (informational)

- **Transition logging placement**: `log_service_mode_transition/3` is only invoked from the `vehicle_data` fetch clause (`vehicle.ex:343`); not from the lightweight `vehicle/id` endpoint or streaming-only updates. Transitions during streaming-only polling won't be logged until the next full fetch.
- **New debug log**: A `Logger.debug("Unlocked ...")` now fires on every poll cycle when the car is unlocked but `req_not_unlocked` is `false`. Harmless at default log levels, but previously absent.

### Looks good
- `TeslaApi.Vehicle.State` struct/parser additions correctly handle `nil`/absent `service_mode`.
- `Summary` struct, MQTT `@simple_values`, and the heex icon follow existing patterns.
- `service_mode?(nil)` falls through to catch-all → `false`, so first fetch (nil `last_response`) won't crash.

**Recommend before merge**: fix the misleading "data collection suspended/resumed" log strings, and clear the fuzzy sleep-mode translations for "Service Mode".

NirKli:
Thanks for the review @brianmay!

Addressed both items in the latest commit:

- **Misleading log messages**: Renamed "Car is in service mode, data collection suspended/resumed" → "Car entered/left service mode". Polling continues during service mode — the old wording implied it stopped, which was wrong.
- **Wrong Service Mode translations**: Cleared the fuzzy sleep-mode `msgstr` in all 16 locale `.po` files.
They now fall back to the English msgid "Service Mode" until proper translations are added by native speakers.

brianmay (approved):

JakobLichterfeld:
Thanks! Really love those additions! And thanks for review Brian!

PR #5300, opened 2026-04-24
feat: add car setting to hide vehicles from dashboards

git-anish:
## Summary
- Add a per-car `show_in_dashboards` setting, defaulting to enabled for existing behavior.
- Expose the setting under Data Collection on the Settings page.
- Filter built-in Grafana dashboard car selectors so cars with this setting disabled are not shown in dashboard car dropdowns.
- Add a dashboard JSON contract test so every built-in `car_id` variable must include the visibility filter.

## Rationale
Users may want to keep a disabled car in TeslaMate so Settings remains usable and TeslaMate can remember that future data collection is disabled, while also hiding that car from dashboards.

## Grafana dashboard notes
This PR changes dashboard variable queries, not panel layout or visualizations. No updated screenshots are included because the rendered dashboards are unchanged except for hidden cars being absent from car selectors.

## Verification
- `git diff --check`
- Parsed all dashboard JSON files with Node
- Confirmed all 22 dashboard `car_id` variables include the `show_in_dashboards` filter

Focused `mix` verification was not run locally because this environment does not have Elixir/Mix installed and the local Docker daemon is not running.


git-anish:
Hey folks, I made this PR to get around the following issues

https://github.com/teslamate-org/teslamate/pull/3993#issuecomment-3478219550
https://github.com/teslamate-org/teslamate/pull/3993#issuecomment-3482442094

git-anish:
The latest commit should address the previous DevOps failure from `mix gettext.extract --check-up-to-date` by updating `priv/gettext/default.pot` for the new `Show in dashboards` label.

I also enabled Actions on my fork and ran the same checks against a fork-local PR with the latest branch head:

- DevOps: passed
  - linting/treefmt passed
  - POT check passed
  - dependency/static analysis/Dialyzer passed
  - tests passed
- Build GHCR images: passed
  - Grafana image build passed
  - amd64, arm/v7, arm64 builds passed
  - GHCR manifest merge passed

Fork validation runs:
- DevOps: https://github.com/git-anish/teslamate/actions/runs/24921062633
- Build GHCR images: https://github.com/git-anish/teslamate/actions/runs/24921062635

swiffer:
thanks for your contribution!

 - i'm not a fan of testing dashboard semantics from the teslamate testsuite.

we do have another dashboard setting (currently unexposed in the settings page) for controlling the order of cars within grafana

- it would be great to have one settings group "Visualization" and boths settings (Show in Dashboards, Rank/Order) listed as part of this group

git-anish:
> thanks for your contribution!
> 
> * i'm not a fan of testing dashboard semantics from the teslamate testsuite.

Could you recommend how to better test this?
 
> we do have another dashboard setting (currently unexposed in the settings page) for controlling the order of cars within grafana
> 
> * it would be great to have one settings group "Visualization" and boths settings (Show in Dashboards, Rank/Order) listed as part of this group

Good call, I've actually used that feature before :D 

I can look at doing this as a separate PR? or would you recommend just updating this PR to contain that?

swiffer:
> Could you recommend how to better test this?

no, as far as i know we currently don't test dashboards at all - open for suggestions but adding it to the elixir based testsuite seems wrong to me - therefore i would vote for dropping that test for now.

> Good call, I've actually used that feature before :D
> 
> I can look at doing this as a separate PR? or would you recommend just updating this PR to contain that?

would be great to have it in one so the grouping and structure of settings can be checked at once.

git-anish:
Latest fixes are now on this PR branch at `f87f972` (same head SHA validated on my fork).\n\nWhat changed since the earlier failing runs:\n- fixed the Settings LiveView regression from the visualization changes\n- updated `priv/gettext/default.pot` after the latest copy/template changes\n\nFork validation on the same head SHA passed:\n- DevOps: success\n- Build GHCR images: success\n\nFork PR used for validation: https://github.com/git-anish/teslamate/pull/1

git-anish:
Following up on @swiffer's feedback: I updated this PR rather than splitting it and reworked the settings UI around the existing visualization controls.

What's in the PR now:
- added a new per-car `Visualization` group in Settings
- moved `Show in dashboards` into that group
- exposed the existing car ordering setting there as `Display order` with up/down controls
- kept `Data Collection` focused on the `Enabled` toggle
- reused the existing `cars.display_priority` field for ordering rather than adding a new dashboard-only field
- dropped the Elixir-side dashboard semantics test, per the feedback about not testing Grafana dashboard JSON from the TeslaMate test suite

Test coverage in the PR is now app-side only:
- settings context tests for `show_in_dashboards` and car reordering
- LiveView tests for the new `Visualization` section and reorder controls
- vehicle summary update coverage for reordered cars
- Home page regression coverage for disabled cars

Latest fixes are on this PR branch at `f87f972`:
- fixed the Settings LiveView regression introduced while regrouping the visualization controls
- updated `priv/gettext/default.pot` after the latest template/copy changes

I validated the same head SHA on my fork before pushing the follow-up fixes here:
- DevOps: success
- Build GHCR images: success

Fork validation PR: https://github.com/git-anish/teslamate/pull/1

JakobLichterfeld:
Not sure how to proceed with this AI slop and bloat. I tend to close this pr.

git-anish:
@JakobLichterfeld that's fair

The original thing I was trying to solve was narrower than where this PR ended up:

- keep disabled cars in TeslaMate settings so they can remain configured/disabled
- hide those disabled cars from Grafana/dashboard selectors
- avoid deleting the car row or its settings row, because that breaks the Settings experience for this use case

"disabled car should still exist in Settings, but not keep showing up in dashboards."

I think the initial version of this PR was closer to that contained goal. The scope grew after I tried to also incorporate the existing car ordering/priority setting into the same UI structure, plus the follow-up test and CI fixes. At that point this became broader and noisier than I intended, and I agree it’s not in a good shape for review as-is.

I don’t think I’m the right person to keep pushing this version further. If it helps, I’d be happy to close this PR and instead turn the original problem into a focused issue/request along these lines:

- disabled cars should remain configurable in Settings
- disabled cars should optionally be excluded from Grafana/dashboard car selectors
- this should not require deleting the car or `car_settings` rows
- any solution should consider the visualization/order behavior as that is related to this effort

If you’d prefer, I can write that up as a clean issue summary and close this PR.

JakobLichterfeld:
Don't get me wrong; I can see how users would benefit from the proposal. 

The changes are somewhat bloated and based on trial and error rather than architecture and clean code.
My concern is that we want to have clean, maintainable code.

git-anish:
closing this in favour of https://github.com/teslamate-org/teslamate/pull/5321

PR #5316, opened 2026-05-01
fix: update GitHub Actions configuration for Dependabot to include action directories

swiffer:
the current depenabot config for github actions is not bumping actions reused in local actions

add the actions folder and group the updates per action

e.g. see #5307 (not covering the build action)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,6 +33,12 @@ updates:
       interval: "monthly"
 
   - package-ecosystem: "github-actions"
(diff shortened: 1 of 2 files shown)


JakobLichterfeld (approved):
Nice solution. Ty!
Always hated the behavior. `/`should be sufficient to catch all, but is obviously not.

swiffer:
found that workaround here: https://github.com/dependabot/dependabot-core/issues/6345

let's see if it works correctly

PR #5321, opened 2026-05-03
Make Display Priority 0 Hide Vehicles

shivam-sh:
## Summary

This exposes the existing per-car `display_priority` setting in Settings as a small “Visualization” section and reuses that same field for Grafana dashboard visibility.

Cars with `display_priority > 0` continue to appear in dashboard car selectors, ordered by `display_priority`. Setting the car to hidden stores `display_priority = 0`, so no new setting column or migration is needed.

## Rationale

The goal is to let users keep a car in TeslaMate and continue managing it in Settings, while hiding it from Grafana dashboard selectors when it should not appear in visualizations.

This keeps dashboard visibility and dashboard ordering represented by one existing field instead of introducing a separate `show_in_dashboards` setting.

## Changes

- Add a “Visualization” section to per-car Settings.
- Add controls for:
  - Show in dashboards
  - Display priority
- Filter built-in Grafana `car_id` variables with `display_priority > 0`.
- Preserve existing dashboard ordering by `display_priority`.
- Add Settings LiveView coverage for updating visibility and priority.

## Verification

- `mix test`
- `MIX_ENV=test mix ci`
- `git diff --check`
- Parsed all built-in dashboard JSON files and verified all 22 `car_id` variables use the visibility filter.


git-anish:
tested and verified locally. Both new settings work, hiding the cars is possible and so is reordering them.

Screenshot 2026-05-04 at 9 53 26 PM
Screenshot 2026-05-04 at 9 53 10 PM

git-anish:
@swiffer  @JakobLichterfeld   thoughts on this?

swiffer:
not sure i like mixing display_priority and show_in_dashboards by reusing the display_priority field.

in #5188 we've started sorting the vehicles in the teslamate settings by display_priority - that means all vehicles with prio 0 will be shown first (instead of last which would be expected).

This PR looks good to me UI-wise (exposing display_priority, new toggle and grouped in Visualization) - the backend related changes still look larger than needed and overall this PR now is in between two different to approaches overall which makes it hard to review.

JakobLichterfeld:
> not sure i like mixing display_priority and show_in_dashboards by reusing the display_priority field.

Same here, and with tendency to dislike it, as this are separate use cases.

> the backend related changes still look larger than needed and overall this PR now is in between two different to approaches overall which makes it hard to review.

Only can echo this and repeat from here:
https://github.com/teslamate-org/teslamate/pull/5300#issuecomment-4364273876

> Not sure how to proceed with this AI slop and bloat.

Combined with switching git usernames I tend to close and block.

git-anish:
> Combined with switching git usernames I tend to close and block.

Hey Jacob, this PR is from my brother.... him and I decided to work on this together since we both use teslamate. I'll wait for him to chime in.

https://www.linkedin.com/in/anish5/

https://www.linkedin.com/in/shivam-sh/

brianmay:
> not sure i like mixing display_priority and show_in_dashboards by reusing the display_priority field.

I am uneasy with this also.

We are reusing one attribute for two different purposes. Which means for example that if you want to hide a value can no longer store the display_priority anymore. Maybe trivial here, but makes me uneasy.

My LLM said "When a car is hidden, the priority input is disabled (not submitted). When re-enabling, the fallback logic returns 1, silently discarding any previously set custom priority."

I asked my LLM, and it pointed out some other potential issues:

> High — NULL display_priority breaks all dashboards on upgrade
> The new WHERE display_priority > 0 filter excludes NULL values (SQL treats NULL > 0 as false). No migration backfills existing rows, so any car created before this column was introduced will disappear from all Grafana dashboards after upgrading. Fix: add a migration to backfill display_priority = 1, or change the filter to WHERE display_priority > 0 OR display_priority IS NULL.
>
> Medium — LiveView bypasses the Vehicles GenServer
> Log.update_car/2 is called directly from the LiveView (index.ex:195, index.ex:207) rather than going through the Vehicles GenServer, which also calls Log.update_car/2 and caches the car struct. This creates a potential state coordination gap where a subsequent Tesla API sync could overwrite the setting.

I haven't investigated these in too much detail, but I think they do deserve some attention.

JakobLichterfeld:
Thanks for the work here, and for keeping the underlying use case on the table — being able to keep a car in TeslaMate while hiding it from dashboards is legitimate, and I don't want that idea to get lost.

I'm closing this PR nonetheless.

**Why**

This is the second attempt at the same feature (#5300), and it still sits between two approaches: a dedicated visibility setting on one side, and overloading `display_priority` on the other. Reusing `display_priority` means one column carries two independent invariants, and that has concrete consequences in this diff:

- `Vehicles.list/0` sorts by `{display_priority, id}`, so hidden cars (`0`) sort to the **top** of TeslaMate's own car list — the opposite of what "hidden" should do. This is exactly the regression @swiffer pointed at after #5188.
- Hiding writes `display_priority: 0` and unhiding falls back to `1`, silently discarding whatever ordering the user had configured.
- The `1..32767` range check lives as a private function in the LiveView, while `Log.Car.changeset/2` still casts `:display_priority` without any validation. If `0` gains meaning, that meaning belongs in the schema — with a migration and a documented constraint — not in a form handler.
- The LiveView calls `Log.update_car/2` directly, bypassing the `Vehicles` process that caches the car struct and builds summaries from it.

On top of that, 26 changed files across 22 dashboards for what is conceptually one setting, plus no documentation for the new semantics and no upgrade note for users who already set `display_priority = 0` by hand, make this hard to review as a whole.

**For the record on the review comments in this thread**

The NULL concern doesn't apply: `display_priority` has been `NOT NULL DEFAULT 1` since migration `20210831153305`, so `WHERE display_priority > 0` cannot drop existing rows. The "a Tesla API sync could overwrite the setting" concern is also overstated — the identify attributes don't include `display_priority`, so the stored value is safe; what does go stale is the cached car struct in the vehicle process. Please verify LLM-generated findings against the code before posting them; unverified ones cost reviewers more time than they save.

That applies to PR descriptions too. Generated summaries and verification checklists that no human has re-checked are the main reason both attempts were hard to assess.

**What a mergeable version would look like**

- One narrow PR: a dedicated boolean on `CarSettings` for dashboard visibility, with migration, changeset validation and context tests — `display_priority` keeps meaning "order" only.
- Dashboard JSON changes as a separate, mechanical follow-up once the backend semantics are agreed on.
- Discuss the design in an issue first, hand-written, before opening a PR.

I'm happy to review that. Not this shape.

PR #5330, opened 2026-05-10
feat: classify offline charges using newly geofence charge_type hint

### LINKED ISSUE #4401: Offline Charging sessions - Missing Added Energy + Default Charge Type

### What happened?

At home (geo-fenced defined location) the car is parked and charged in an underground garage without internet connection.
The charge session are correctly identified in duration and location, but there are to recurrent issues

1. Every few charging session, the amount of total added energy cannot be calculated and is left blank, although the before and after SOC are correctly identified.
Sometimes it works, sometimes not... anybody has an idea why that could be ?

2. The charging session are recorded as DC, while it is actually AC charging. As it is offline, there is no way for teslamate to know which type it is.
Is there a way to add a default charge type to a Geo-Fenced location to cover the case of offline charges ?

Here a screenshot of some past home charges:
Screenshot 2024-11-25 at 09 00 23

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
And here a log of the past 3 charges: [teslamate_log_25-11-2024_charges.log](https://github.com/user-attachments/files/17900113/teslamate_log_25-11-2024_charges.log)
```

### Screenshots

_No response_

### Additional data

[charges.csv](https://github.com/user-attachments/files/17900196/charges.csv)
[charging_processes.csv](https://github.com/user-attachments/files/17900197/charging_processes.csv)

### Type of installation

Docker

### Version

v1.31.1

finikwashere:
## Problem

Closes #4401

When a car charges in a location with no internet (underground garage, gym, etc.), TeslaMate synthesises the charging session from the last known state before going offline and the first state after reconnecting. All `charges` rows for that session have `NULL` `charger_phases` and `fast_charger_present`, so every Grafana dashboard query that uses:

```sql
CASE WHEN NULLIF(mode() within group (order by charger_phases),0) IS NULL
  THEN 'DC' ELSE 'AC' END
```

incorrectly labels the session as **DC** — even for a plain home AC wallbox.

Affected dashboards: **Charges**, **Charging Stats**, **Battery Health**, **Trip** (9 SQL queries total).

## Solution

Introduce an optional **Default charge type** hint on Geofences (AC / DC / auto-detect). The Grafana SQL falls back to it only when it cannot determine the type from live telemetry.

### Priority order in the new CASE expression

1. `charger_phases IS NOT NULL` → **AC** (normal online session)
2. `fast_charger_present = true` → **DC** (online DC session without phase data)
3. `geofence.charge_type IS NOT NULL` → **use configured hint** (offline sessions at known locations)
4. Fallback → **DC**

### Changes

| File | What changed |
|---|---|
| `priv/repo/migrations/20260510111557_add_charge_type_to_geofences.exs` | Add nullable `charge_type varchar` column to `geofences` |
| `lib/teslamate/locations/geo_fence.ex` | Add `charge_type` Ecto.Enum field and include in changeset |
| `lib/teslamate_web/live/geofence_live/form.html.heex` | Add "Default charge type" dropdown to geofence edit form |
| `grafana/dashboards/charges.json` | Fix charge_type CASE expression |
| `grafana/dashboards/charging-stats.json` | Fix 4 CASE expressions + add geofences JOIN |
| `grafana/dashboards/battery-health.json` | Fix CASE expression + add geofences JOIN |
| `grafana/dashboards/trip.json` | Fix 3 CASE expression variants + add geofences JOIN |

### Why this fixes existing data retroactively

The fix is applied at **query time** in Grafana. No `UPDATE` to historical rows is needed — setting `charge_type = 'AC'` on a geofence immediately corrects all past charging sessions at that location.

## Testing

Tested on a live TeslaMate instance with an underground home garage (no cell reception during charging). Before the fix every Home charge showed as DC. After setting `charge_type = 'AC'` on the Home geofence, all historical home charges immediately show as AC while Supercharger and public DC fast-charger sessions remain correctly classified as DC.


JakobLichterfeld:
Thanks for your contribution.

> immediately corrects all past charging sessions at that location.

This is not the best practice for changing historical charging types. It is okay to change them if they were offline, but otherwise, the correct data may become corrupted.

finikwashere:
> This is not the best practice for changing historical charging types. It is okay to change them if they were offline, but otherwise, the correct data may become corrupted.

Should it be optional, triggered by the user after upgrade?

brianmay:
> immediately corrects all past charging sessions at that location.

Not sure, there might be some confusion here in how this was documented.

The wording implies that we change (correct) the database at query time. "correct" would mean we write new values to the database.

That would be bad, queries have to remain read only. My grafana only has read only access to the database.

But my reading of the queries seems to indicate that they are read only and do not change any data.

Or am I totally confused here?

brianmay:
I strongly suspect the tests will need fixing...

finikwashere:
Apologies for the confusion.
The tests are updated.
I've also updated the translations.

On "corrects all past charging sessions"** (@brianmay): Grafana queries are read-only, and this is what the SQL here also does. The wording in the PR description was misleading. What I meant is:

> Setting `charge_type = 'AC'` on a geofence causes all *future queries* to return 'AC' for historical sessions at that location — no data is written or modified. The SQL purely reads `geofences.charge_type` as an additional input to the `CASE` expression at query time.

The new `CASE` expression uses the geofence hint only as step 3 of 4, after the live telemetry signals have already been checked:

```sql
CASE
  WHEN NULLIF(mode() within group (order by charger_phases), 0) IS NOT NULL THEN 'AC'  -- online AC session
  WHEN bool_or(fast_charger_present) = true THEN 'DC'                                  -- online DC session
  WHEN MAX(g.charge_type::text) IS NOT NULL THEN MAX(g.charge_type::text)              -- offline fallback
  ELSE 'DC'
END
```

So for any session that was **online** during charging (i.e. `charger_phases` or `fast_charger_present` has real data), the geofence hint is never reached. It only kicks in for sessions where the car had no connectivity and both signals are NULL — which is exactly the scenario described in #4401.

PR #5332, opened 2026-05-12
Fix OSV vulnerabilities: update cowlib, decimal, and phoenix to patched versions

brianmay:
Think this might be a good idea, with the number of supply chain vulnerabilities that have been going around.

Note that this package installed by nix in stable is older then the version run by CI; this will be fixed when nix unstable is released.

Cannot fix currently:

- cowlib (3 CVEs: 1 High, 1 Medium, 1 Low) — The High CVE-2026-7790 fix (2.16.1) isn't published on Hex.pm yet (CVEs were disclosed yesterday, 2026-05-11). The Medium and Low CVEs have no fix version at all yet.
- decimal (1 Medium) — CVE-2026-32686 is fixed in 3.0.0, but ecto, postgrex, and ex_cldr all constrain to < 3.0. Those upstream deps need to update their constraints first.

Might want to silence these alerts, not sure.


brianmay:
When I run locally I get:

```
╭────────────────────────────────────┬──────┬───────────┬─────────┬─────────┬───────────────┬───── ≈
│ OSV URL                            │ CVSS │ ECOSYSTEM │ PACKAGE │ VERSION │ FIXED VERSION │ SOUR ≈
├────────────────────────────────────┼──────┼───────────┼─────────┼─────────┼───────────────┼───── ≈
│ https://osv.dev/EEF-CVE-2026-43968 │ 6.3  │ Hex       │ cowlib  │ 2.16.0  │ --            │ mix. ≈
│ https://osv.dev/EEF-CVE-2026-43969 │ 2.1  │ Hex       │ cowlib  │ 2.16.0  │ --            │ mix. ≈
│ https://osv.dev/EEF-CVE-2026-7790  │ 8.7  │ Hex       │ cowlib  │ 2.16.0  │ 2.16.1        │ mix. ≈
│ https://osv.dev/EEF-CVE-2026-32686 │ 6.9  │ Hex       │ decimal │ 2.4.1   │ 3.0.0         │ mix. ≈
╰────────────────────────────────────┴──────┴───────────┴─────────┴─────────┴───────────────┴───── ≈
```

The CI won't show pre-existing issues apparently.

brianmay:
Not 100% convinced this is working. Every other repo I tried this on will show pre-existing emails.

brianmay:
Probably is working, I didn't use pull requests on the other repos, but merged straight into main. Should see the alerts if merged.

JakobLichterfeld:
Thanks, Brian!
I'm not sure if we should use our CI guards for this new action as well. I tend to say: we do need our filtering. @swiffer what's your take on this?

brianmay:
My understanding is while the risks are not 0 they are extremely low. Partly because this has restricted privileges anyway and only for the duration of the job. And partly because this is scan only (for elixir anyway) and should not be executing any untrusted code.

swiffer:
someone could add additional jobs to the jobs: section which would be executed as part of the workflow run (also for external PRs), right?

JakobLichterfeld:
Exactly, that's my concern, and why I vote to include safeguards as well.

JakobLichterfeld (approved):

brianmay:
My understanding is (correct me if I am wrong) that somebody creates a PR from a fork repo, then the CI runs from the target repository.

https://devactivity.com/insights/boost-your-github-productivity-understanding-ci-cd-workflow-behavior-in-forked-pull-requests/

Otherwise somebody could just create a new workflow file that steals all your secrets. And nothing we do could stop it.

Which means it isn't possible to add extra jobs to the workflow without it being approved and merged.

(this is github, others like github have different rules)

JakobLichterfeld:
Yes and a "pull_request_target" trigger for a workflow punches big holes into it. We had it in the past to generate Docker images to test for each PR, even for those from forks. This could be used to extract secrets, in 1,5 years or so one attacker was successful in overcome our sec measures against it by combining 3 steps, and he needed several attempts to do so.

So yes, the actions run from main and malicious code needed to be merged first but there are known loopholes and protection was never a bad idea.

PR #5334, opened 2026-05-12
feat(grafana): add Trip Cost and Cost/100km panels to Drive Details

### LINKED ISSUE #5333: feat(grafana): Add estimated Trip Cost panels to Drive Details dashboard

### What happened?

`trip.json` shows `Total Charging Cost` and `Ø Cost per 100 km/miles`
When clicking on the single drive, `drive-details.json` does not show this information for individual trip.

### Expected Behavior

I expected two stat panels on the `drive-details.json`, giving a per-trip cost estimate based on recent charging price data, in particular, a 3-tier fallback lookup is used to find the most relevant energy price:

1. 24h window — average cost/kWh across all charging sessions that completed within the 24 hours immediately before the drive started
2. Last charge fallback — if no sessions in that window have cost data, uses the single most recent charge before the trip
3. N/A — if no prior charge has any cost data at all

Yes

1. Open any Drive Details
2. No cost information visible

### Relevant log output

```shell
none
```

### Screenshots

I would love to have something like this:

Image

### Additional data

_No response_

### Type of installation

### Version

latest

### PostgreSQL version

v17.7

No

_No response_

finikwashere:
Closes #5333 

Added two new stat panels to the Drive Details dashboard that estimate the cost of a trip based on recent charging price data:

- Trip Cost (est.): total estimated cost for the drive
- Ø Cost / 100 <unit>: estimated cost per 100 km or 100 mi

Cost/kWh is determined by a 3-tier lookup:
  1. Average cost/kWh across all charging sessions that ended within the 24 hours before the drive started (the most relevant window)
  2. Fallback: cost/kWh of the single most recent charge before the drive, if no sessions exist in the 24h window
  3. N/A if no prior charge has cost data at all

Both panels are placed at y=28 alongside the existing Distance driven and Elevation Summary panels (all resized to w=3 to fit on one row).

### DIFF
--- a/grafana/dashboards/internal/drive-details.json
+++ b/grafana/dashboards/internal/drive-details.json
@@ -1816,7 +1816,7 @@
       },
       "gridPos": {
         "h": 3,
-        "w": 6,
+        "w": 3,
         "x": 0,
         "y": 28
       },
@@ -1898,8 +1898,8 @@
       },
       "gridPos": {
         "h": 3,
-        "w": 6,
-        "x": 6,
+        "w": 3,
+        "x": 3,
         "y": 28
       },
       "id": 20,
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Thanks for your idea. Cost per drive is not a metric which is valuable, even for a trip it is misleading, see https://github.com/teslamate-org/teslamate/discussions/2433#discussioncomment-13402292 and https://github.com/teslamate-org/teslamate/pull/4678#issuecomment-2830284946

JakobLichterfeld:
Sorry for the terse close — let me give the actual reasoning, since the #4678 argument (SOC delta over a period) doesn't apply to your approach.

Your energy figure is per-drive and fine. The problem is the price side: there is no link in the data between a charging session and the energy a given drive consumed. The 24h window is an arbitrary cut-off, and the fallback has no time limit at all — a drive can be priced with a charge from months earlier. Two identical drives get different costs depending on whether a Supercharger stop happened to precede one of them, even though most of the energy came from home charging.

Without FIFO attribution of charged kWh to drives, a per-drive cost isn't derivable from what TeslaMate stores, and a stat panel with 2 decimals implies a precision we don't have. That's different from the trip dashboard, which shows money actually spent in the selected period rather than a synthesized per-drive price.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)](https://claude.com/claude-code) (Opus 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5324, opened 2026-05-05
feat: use Grafana 13.0.1

swiffer:
Grafana 13 brings support for disabling feature toggles enabled by default [via env](https://github.com/grafana/grafana/pull/121271)

I've used that to disable 2 big new features for now:
- Provisioning (Git Sync) -> https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v13-0/#git-sync
- New Dashboard Layouts (and Dashboard Schema v2) -> https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v13-0/#dynamic-dashboards-is-now-generally-available

In addition I've disabled two toggles:
- splashScreen (showing highlights of v13 release)
- newPanelPadding (which caused some panels (e.g. in Battery Health to be cut off)
  grafik

PR #5335, opened 2026-05-15
fix: create placeholder geofence if no geofence exists to correctly load charges and drives dashboard

swiffer:
this is a revised workaround for the Charges & Drives Dashboard not loading if no Geofence exists.

Instead of trying to create the "All" Option this creates a Dummy / Placeholder Geofence if non exists.

Doing so instead of creating an "All" Option via SQL (#5199) is safer as we rely on the logic & sorting around the Grafana Built-In All Option in all cases and is less likely to introduce side effects.

If no Geofence exists -> Dashboard renders correctly, Geofence selector looks like this:

grafik

If a Geofence exists -> Dashboard renders correctly, no Placeholder entry is created.


swiffer:
@JakobLichterfeld - will rebase one of the Grafana PRs (incl. the security patch).

Personally leaning towards 13.0.1-security-01 as this gives control over feature toggles, features introduced in 13.0 have been disabled and doing so avoids one PR down the road but open to stay on 12.4.3 for TeslaMate 3.1.0.

DrMichael:
Works good for my case with Geofence smaller than "All".

JakobLichterfeld (approved):
Nice workaround, ty!
As this really fixes the issue and we already have #5199 in the changelog I will just add it

JakobLichterfeld (approved):

JakobLichterfeld:
> Personally leaning towards 13.0.1-security-01 as this gives control over feature toggles, features introduced in 13.0 have been disabled and doing so avoids one PR down the road but open to stay on 12.4.3 for TeslaMate 3.1.0.

OK, let's go with  13.0.1-security-01 -> #5324

PR #5336, opened 2026-05-15
feat: improve filters in Statistics and Charges dashboards

DrMichael:
Add charge_energy_used to the Statistics and Charges and changing the filter in charges from 0.1 to 0.


swiffer:
@DrMichael - why is the condition incl. "IS NULL" filters in drives and isn't in statistics?

DrMichael:
> @DrMichael - why is the condition incl. "IS NULL" filters in drives and isn't in statistics?

That was, what Copilot proposed. Still thinking, what NULL really means. I was stumbling across this:

I have this month a NULL for charge_energy_used, but 0.2 for added:
image

But Charges shows (with different time):
image

Wierd...

swiffer:
- i do not have a single charge with `charge_energy_added is null` - there should not be any completed charges with charge_energy_added is null that occured after https://github.com/teslamate-org/teslamate/commit/99b71128cbf82bcbe7cb694a662ae8d769de889e - guess makes sense to still have that condition for historical reasons and potential null values
- incomplete charges are currently filtered via end_date is (not) null - in Charges -> Incomplete Charges
- charge_energy_used is null if charging via AC and phases cannot be determined via determine_phases (which could be the case for really short charging sessions) - in these cases dashboard queries fallback to charge_energy_added for charge_energy_used

i therefore suggest to go with:

```sql
where
  charge_energy_added is null -- include charges where calculating charge_energe_added failed prior to commit 99b7112
  or charge_energy_added > 0 -- exclude any charges which haven't added any energy or aren't completed yet
```

do not add any filtering for charge_energy_used (if null -> fallback to charge_energy_added via greatest())

what i do find weird is how you're ending up with 1.16 for the energy used - shouldn't that be 0.02 it's null ? 🤔

DrMichael:
> * i do not have a single charge with `charge_energy_added is null` - there should not be any completed charges with charge_energy_added is null that occured after [99b7112](https://github.com/teslamate-org/teslamate/commit/99b71128cbf82bcbe7cb694a662ae8d769de889e) - guess makes sense to still have that condition for historical reasons and potential null values

Neither do I, only charge_energy_added 100 times NULL.

> i therefore suggest to go with:
> 
> ```sql
> where
>   charge_energy_added is null -- include charges where calculating charge_energe_added failed prior to commit 99b7112
>   or charge_energy_added > 0 -- exclude any charges which haven't added any energy or aren't completed yet
> ```
> 
> do not add any filtering for charge_energy_used (if null -> fallback to charge_energy_added via greatest())

Ok, will do that. So, change just the 0.1 to 0 and leave Charges as it is. https://github.com/teslamate-org/teslamate/pull/5340

> what i do find weird is how you're ending up with 1.16 for the energy used - shouldn't that be 0.02 it's null ? 🤔

I will open a discussion for that. When I use the above filter in Statistics, I get a reasonable value for added (as in Charges). And I cannot see, why...

DrMichael:
Replaced by https://github.com/teslamate-org/teslamate/pull/5340

PR #5337, opened 2026-05-15
refactor(nix,postgres): provision database declaratively, connect via socket

### LINKED ISSUE #4904: teslamate-close-charge: syntax error near unexpected token `newline'

### What happened?

```
[root@iot2:~]# teslamate-close-charge
/run/current-system/sw/bin/teslamate-close-charge: line 3: 1: Please provide a charge ID to close

[root@iot2:~]# teslamate-close-charge 2790
/run/agenix/teslamate: line 1: syntax error near unexpected token `newline'

[root@iot2:~]# bash teslamate-close-charge 2790
/run/agenix/teslamate: line 1: syntax error near unexpected token `newline'
```

To me it looks like the shell doesn't like my secrets, and wants the values to be quoted.

But the same file is also read by the systemd service file, and pretty sure environment variables cannot be quoted or you get the quotes included in the value.

Yuck.

I think this file needs to be parsed, and we cannot really use "source" to do it.

I notice my db password has a greater then symbol suspect this is the problem:

```
DATABASE_PASS=xyzxyz>
```

(This isn't my password)

### Expected Behavior

No errors.

Yes

_No response_

### Relevant log output

```shell
N/A
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

NixOS (https://docs.teslamate.org/docs/installation/nixos)

### Version

mcu2-upgraded-cars

### PostgreSQL version

_No response_

Yes (specify in next field)

nginx

JakobLichterfeld:
Replace the single `initialScript` with declarative `ensureDatabases` and `ensureUsers`, freeing the `services.postgresql.initialScript` option, which is single-valued and could not be combined with other modules that need it (e.g. to `CREATE EXTENSION`).

TeslaMate now connects to a local PostgreSQL via the unix socket using peer authentication, so no password is sent over TCP and startup no longer depends on the password being set first. The role password is still applied at runtime from DATABASE_PASS via `postStart` (kept out of the Nix store) because Grafana and the TCP fallback still require it.

Additionally read `RELEASE_COOKIE` without sourcing the env file 
Sourcing the environment file executed it as a shell script (arbitrary code execution if a secret contained $(...) or backticks) and broke on values containing quotes. Extract RELEASE_COOKIE literally with sed
instead and strip one optional surrounding pair of double quotes, so the value is handled safely whether or not it is quoted in the file.
This closes #4904

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - fix(nix,maintenance): read RELEASE_COOKIE without sourcing the env file (#5337 - @JakobLichterfeld)
 - fix(nix,postgres): set role password safely for any value (#5337 - @JakobLichterfeld)
 - fix(nix): drop schemas in the configured database during restore (#5337 - @JakobLichterfeld)
+- fix(nix,grafana): disable the periodic plugin update check (#5337 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - fix: reconnect stream when a drive resumes after mid-drive offline phase to avoid missing elevation (#5535 - @JakobLichterfeld)
 - refactor(nix,postgres): provision database declaratively, connect via socket (#5337 - @JakobLichterfeld)
 - fix(nix,maintenance): read RELEASE_COOKIE without sourcing the env file (#5337 - @JakobLichterfeld)
+- fix(nix,postgres): set role password safely for any value (#5337 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - fix(mqtt): return publish errors without crashing (#5514 - @magrathean-uk)
 - fix(geofences): increase cost precision (#5508 - @magrathean-uk)
 - fix: reconnect stream when a drive resumes after mid-drive offline phase to avoid missing elevation (#5535 - @JakobLichterfeld)
+- refactor(nix,postgres): provision database declaratively, connect via socket (#5337 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - fix(geofences): increase cost precision (#5508 - @magrathean-uk)
 - fix: reconnect stream when a drive resumes after mid-drive offline phase to avoid missing elevation (#5535 - @JakobLichterfeld)
 - refactor(nix,postgres): provision database declaratively, connect via socket (#5337 - @JakobLichterfeld)
+- fix(nix,maintenance): read RELEASE_COOKIE without sourcing the env file (#5337 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - refactor(nix,postgres): provision database declaratively, connect via socket (#5337 - @JakobLichterfeld)
 - fix(nix,maintenance): read RELEASE_COOKIE without sourcing the env file (#5337 - @JakobLichterfeld)
 - fix(nix,postgres): set role password safely for any value (#5337 - @JakobLichterfeld)
+- fix(nix): drop schemas in the configured database during restore (#5337 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/lib/teslamate/database_check.ex
+++ b/lib/teslamate/database_check.ex
@@ -8,6 +8,8 @@ defmodule TeslaMate.DatabaseCheck do
     defstruct [:version_string, :version_num, :major]
   end
 
+  # When changing these requirements, also update the eval-time assertion in
+  # nix/module.nix (services.teslamate.postgres.package), which mirrors them.
   @version_requirements %{
     1600 => %{min_version: "16.7", min_version_num: 160_007},
     1700 => %{min_version: "17.3", min_version_num: 170_003},
--- a/nix/backup_and_restore.nix
+++ b/nix/backup_and_restore.nix
@@ -20,7 +20,7 @@ let
     systemctl stop teslamate.service
 
     # Drop existing data and reinitialize
-    sudo -u teslamate psql -U ${databaseUser} << .
+    sudo -u teslamate psql -U ${databaseUser} -d ${databaseName} << .
       DROP SCHEMA IF EXISTS public cascade;
       DROP SCHEMA IF EXISTS private CASCADE;
       CREATE SCHEMA public;
--- a/nix/flake-modules/package.nix
+++ b/nix/flake-modules/package.nix
@@ -27,7 +27,7 @@
       nodePackages = pkgs.buildNpmPackage {
         name = "${pname}-assets";
         src = "${src}/assets";
-        npmDepsHash = "sha256-b3b5yEVM3E3X9Gl7FheXTnNgs2U9Vwtr9O8IjKro5hA="; # if you change the npm deps, you need to update this hash
+        npmDepsHash = "sha256-x0/2/j1nsWhJHo1f+7KDdI2ks4rB3th1Ty/ad76cRts="; # if you change the npm deps, you need to update this hash
         # npmDepsHash = pkgs.lib.fakeHash;
         dontNpmBuild = true;
         inherit nodejs;
--- a/nix/maintenance.nix
+++ b/nix/maintenance.nix
@@ -42,7 +42,7 @@ let
 
     # load RELEASE_COOKIE from the env file
     ${loadFromEnvFile "RELEASE_COOKIE"}
-    : ''${RELEASE_COOKIE?'RELEASE_COOKIE must be set in the environment file'}
+    : ''${RELEASE_COOKIE:?'RELEASE_COOKIE must be set in the environment file'}
 
     echo "Attempt to close the drive with ID ''${1}."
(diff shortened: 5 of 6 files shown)


JakobLichterfeld:
`initialDatabases` is not in NixOS 25.11, therefore we need to wait with this till it is releases

brianmay:
I would really like to update this sometime to use the socket connection to the database by default rather then the TCP connection. Not only is there likely to be a tiny performance gain, but means there is one less password to worry about.

Not sure about grafana though, is that supports socket connections or not. And it probably should have a dedicated postgresql user too.

Don't know if any of this is in scope for this change or not. But maybe something to think about.

JakobLichterfeld:
Exactly. I wonder the same, would it be possible to use socket connection and grafana (yes I do it similar with Prometheus in my homelab) and TeslaMate as well to connect to the DB at the same time? Why not.

As it seems like the options I tried to use won't be available (old links in mynixos.com) I will look more into it.

JakobLichterfeld:
> As it seems like the options I tried to use won't be available (old links in mynixos.com) I will look more into it.

Replaced with the familiar once and use `ensureClauses` for superuser and password set. The latter needs testing.

JakobLichterfeld:
> Replaced with the familiar once and use `ensureClauses` for superuser and password set. The latter needs testing.

testing showed it didn't work as expected for password, so I solved it correctly.

> I would really like to update this sometime to use the socket connection to the database by default rather then the TCP connection

I refactored to use socket connection for TeslaMate in default.

JakobLichterfeld:
I also fixed the RELEASE_COOKIE sourcing issue (from #4904)

JakobLichterfeld:
## Self-review notes (maintainer hat on)

Since this touches provisioning and secrets handling, here is a summary of what I verified before asking for review:

**Verified:**
- `DATABASE_SOCKET_DIR` support already exists in `config/runtime.exs`; the socket branch is only enabled when `postgres.enable_server && postgres.user == "teslamate"`, which is exactly the case where peer auth works (the systemd unit runs as OS user `teslamate`). Any other role name or an external DB falls back to TCP with password, unchanged.
- `ensureDatabases`/`ensureUsers` are idempotent, so existing installs migrate cleanly: DB and role already exist, `ensureDBOwnership` merely transfers ownership. Nice side effect: the password is now (re-)applied on every activation, so rotating `DATABASE_PASS` in the secrets file finally takes effect — `initialScript` only ever ran on first init.
- `\getenv` + `:'password'` keeps the secret off the command line and quotes it correctly for any value (including single quotes). Requires psql >= 14; our default is `postgresql_17` and the requirement is documented in a comment.
- The `sed`-based `RELEASE_COOKIE` extraction handles values containing quotes, `$(...)` and backticks safely; `tail -n1` mirrors systemd's last-assignment-wins.

**Known minor follow-ups (happy to fix in this PR if reviewers agree):**
- The `: ${RELEASE_COOKIE?...}` guard should become `: ${RELEASE_COOKIE:?...}` — the sed extraction always sets the variable (possibly empty), so the unset-only form can no longer fire.
- If `DATABASE_PASS` is missing from the secrets file, the `postStart` fails with a rather opaque psql syntax error; an explicit `test -n` guard with a clear message would be friendlier.

Tested on my own NixOS install (fresh provisioning and upgrade from the `initialScript` state). CI is green. Review welcome, especially on the systemd ordering (`after = postgresql-setup.service`).

JakobLichterfeld:
Both follow-ups from my self-review notes above are now fixed:

- `: ${RELEASE_COOKIE?...}` → `: ${RELEASE_COOKIE:?...}` in both maintenance scripts, so the guard also catches an empty value (the sed extraction always sets the variable, possibly to the empty string).
- The `postStart` of `postgresql-setup.service` now fails early with a clear message pointing at `services.teslamate.secretsFile` when `DATABASE_PASS` is missing, instead of an opaque psql syntax error from the unsubstituted `:'password'`.

While at it, the `secretsFile` option docs now mention that the file uses systemd `EnvironmentFile` syntax, which is why one surrounding pair of double quotes is stripped when extracting values.

brianmay:
This mostly looks good to me.

AI Report:

> ## PR #5337 Review Summary
> 
> Refactor of NixOS module's PostgreSQL setup, moving password configuration from `initialScript` (which sourced env files as shell scripts) to a `postgresql-setup.postStart` step using `\getenv`. Overall a security improvement.
> 
> **Action recommended before merge:**
> 1. **Standby/replica breakage (low severity)** — `nix/module.nix` `postStart`: `ExecStartPost` runs even when `ExecStart` short-circuits on ``. On a read-only standby, `ALTER USER ... WITH ENCRYPTED PASSWORD` will fail with `ON_ERROR_STOP=1`, blocking the `teslamate` service. Guard with `[[ -f ${cfg.dataDir}/standby.signal ]] && exit 0` (or `pg_is_in_recovery`).
> 
> **Informational:**
> 2. `\getenv` requires PostgreSQL ≥ 14, but `postgres.package` is user-overridable. An assertion (`lib.versionAtLeast cfg.postgres.package.version "14"`) would surface misconfig earlier — not a bug with defaults.
> 3. Maintenance delete scripts (`deleteDrive`/`deleteCharge`) won't work with non-default `postgres.user` due to peer auth — pre-existing, but the PR widens the asymmetry with the new TCP fallback branching. Worth a docs note.
> 
> **Verified correct** — quote-stripping escapes, `sed | tail -n1` last-wins semantics, `RELEASE_COOKIE` fail-fast, `ensureDBOwnership` conditional, `After=` no-op when server disabled, `\getenv`/`:'password'` SQL literal safety.
> 
> Want me to implement the standby guard fix?

Most of these are minor issues I think.

Not quite comprehending the "Standby/replica breakage" stuff right now. Possibly because I have never done this with postgresql :-)

JakobLichterfeld:
Thanks for the thorough look, Brian!

Quick explainer on the standby point, since it's a bit obscure: a standby is a streaming-replication replica running in recovery mode — it is read-only, so *any* write statement fails, including `ALTER USER`. With `ON_ERROR_STOP=1` that would have marked `postgresql-setup.service` as failed. In practice this is a mostly theoretical concern for this module (TeslaMate itself cannot run against a read-only database — migrations and inserts would fail immediately, and the role password is replicated from the primary anyway), but the guard is cheap, so I added it: `postStart` now checks `SELECT pg_is_in_recovery()` and exits cleanly on a standby. I went with `pg_is_in_recovery()` rather than checking for `standby.signal`, since the module doesn't reference the data directory.

Also addressed:

- Added a version assertion, but stricter than suggested: TeslaMate's startup check (`lib/teslamate/database_check.ex`) already requires >= 16.7 (16.x), >= 17.3 (17.x) or >= 18, so the module now asserts exactly that at eval time — which subsumes the PostgreSQL >= 14 requirement for `\getenv`. Newer majors are allowed, mirroring the runtime check which only warns for those.
- The peer-auth asymmetry in the delete scripts (`deleteDrive`/`deleteCharge` only work with the default role name) is pre-existing; I'll track it as a follow-up rather than widening this PR. --> #5533

brianmay (approved):

JakobLichterfeld:
Pushed one more fix found while testing the restore path: `fix(nix): drop schemas in the configured database during restore`.

Without `-d`, psql connects to the database named after the *user*, so the `DROP SCHEMA ... CASCADE` / `CREATE EXTENSION` step ran against the wrong database whenever `databaseName` differs from `databaseUser` — worst case dropping schemas in an unrelated database that happens to share the user's name — while the subsequent restore targeted the correct database and then failed on the existing data. The restore invocation below already passed `-d ${databaseName}`; this makes the drop step consistent.

Two related observations, deliberately left for the follow-up issue (#5533) on the maintenance scripts: the backup/restore scripts share the same peer-auth limitation as the delete scripts (they only work with the default role name), and the restore psql invocations run without `ON_ERROR_STOP`, so individual failing statements don't abort the script.

JakobLichterfeld:
One more small hardening found while reviewing the Grafana defaults: the NixOS module disables Grafana's update checks by default, *except* `analytics.check_for_plugin_updates`, which defaults to `declarativePlugins == null` — so with our nixpkgs-managed plugins it phoned grafana.com every 10 minutes for updates that can never be applied. Now explicitly disabled, matching the existing `reporting_enabled = false`.

JakobLichterfeld:
Found a real bug when deploying this branch to my production server (NixOS 26.05) — and fixed it, plus a related hardening.

**The bug:** the `postgresql-setup` `postStart` set the role password with two separate `-c` options:

```
psql ... -c '\getenv password DATABASE_PASS' -c "ALTER USER ... :'password'"
```

This fails: psql only performs `:'var'` interpolation when reading from stdin or a file, and a variable set by `\getenv` in one `-c` is **not** available for interpolation in a following `-c` — so `:'password'` was sent verbatim and the service failed with:

```
ERROR:  syntax error at or near ":"
LINE 1: ALTER USER "teslamate" WITH ENCRYPTED PASSWORD :'password'
```

Fix: feed both commands through a single stdin stream (`printf '%s\n' ... | psql`). Reproduced the failure and verified the fix against PostgreSQL 17, including a password containing `:`, `'`, `\`, `$` and spaces (SCRAM login round-trip succeeds).

**Related hardening (secret in logs):** the interpolated `ALTER USER` carries the cleartext password. With default config nothing is logged, but if the operator has enabled `log_statement`/`log_min_duration_statement` or error-statement logging, the cleartext would land in the PostgreSQL server log. I confirmed this leak with `log_statement=all`, then closed it by turning `log_statement`, `log_min_duration_statement` and `log_min_error_statement` off for the session before the `ALTER` (allowed since we connect as the postgres superuser). After the change the secret is absent from a `log_statement=all` server log; only the harmless `SET log_statement = 'none';` line remains. Note this log exposure was equally present in the previous `initialScript` approach — not a regression, but worth closing while we're here.

The password value never appears in `argv` or the Nix store in either version.

🤖 Comment drafted with [Claude Code](https://claude.com/claude-code) (Opus 4.8 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5339, opened 2026-05-16
refactor: use frame key when specifying window to improve readability

swiffer:
both work, but as a frame key exists let's use that for better readability?

https://hexdocs.pm/ecto/Ecto.Query.html#windows/3-frame


JakobLichterfeld (approved):
lgtm

JakobLichterfeld (approved):

PR #5352, opened 2026-05-29
fix(nix): make grafana.secretKeyFile optional with old insecure grafana default fallback

JakobLichterfeld:
Add a warning when secretKeyFile is not set to inform users that the insecure grafana built-in default key is being used. This avoids a breaking change for existing installations while encouraging users to set a secure key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,7 +72,8 @@
 - build(deps): bump webpack-dev-server from 5.2.3 to 5.2.4 in /website (#5343)
 - sec(deps): add ws override to version 8.20.1 in /website (#5344 - @JakobLichterfeld)
 - feat(nix): update NixOS version to 26.05 (#5350 - @JakobLichterfeld)
-- fix(nix): correct URL format for PostgreSQL connection(#5351 - @JakobLichterfeld)
+- fix(nix): correct URL format for PostgreSQL connection in Grafana 13+ (#5351 - @JakobLichterfeld)
+- fix(nix): make grafana.secretKeyFile optional with old insecure grafana default fallback (#5352 - @JakobLichterfeld)
 
 #### Dashboards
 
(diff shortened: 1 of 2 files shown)


brianmay:
How do you migrate an existing setup to use the new encryption key?

brianmay:
OK, procedure is (adapted from instructions AI supplied):

1. SSH in, stop Grafana
    ```
    systemctl stop grafana
    ```
2. Run rotation (INI still has the old insecure key). Note: ini path was from my system.
    ```
    nix run github:erooke/grafana-secretkey-rotation-tool -- update -db /var/lib/grafana/data/grafana.db -ini /nix/store/3n2fckmpi6q94xd7cmf72xi0vd8q4z3w-config.ini
    ```
    prints: New secret_key: abc123def456...  <-- copy this
3. Encrypt that key with agenix
    ```
    echo -n "abc123def456..." | nix run .#agenix -- -e secrets/grafana_secret_key.age
    ```
4. Update config to use new key, and deploy.
5. Start Grafana
    ```
    systemctl start grafana
    ```

Getting the path to the ini file in step 3 correct is a bit fiddly, that path is from my system. As an alternative, this will work too, the tool only needs the old secret.

```
echo "[security]" > /tmp/grafana.ini
echo "secret_key = SW2YcwTIb9zpOOhoPsMm" >> /tmp/grafana.ini
```

The the command is:

```
nix run github:erooke/grafana-secretkey-rotation-tool -- update -db /var/lib/grafana/data/grafana.db -ini /tmp/grafana.ini
```

The tool never modifies the INI — it reads the old key from it, does the DB migration, and gives you the replacement. You then wire that into your config.

brianmay:
Possibly this is overkill for teslamate module, which I think supplies its own config, not sure.

JakobLichterfeld:
I went the same route and then realized as I migrated our module to deploy as grafa datasouce no migration needed.

PR #5355, opened 2026-05-31
feat: add geofence filter to Charging Stats dashboard with default to all

faekz0r:
## Summary

- add a multi-select `Geofence` variable to the Charging Stats dashboard
- default to `All` so the existing dashboard view is preserved
- apply the selector consistently to charge-based stats, maps, curves, and station tables
- reuse the Charges dashboard placeholder workaround for installations without geofences

## Why

Charging Stats currently aggregates charging sessions from every location, unlike the Charges dashboard. This prevents users from answering location-specific questions such as average cost per 100 km when charging at Home.

The `Ø Cost per 100 km` panel intentionally keeps its driving-consumption estimate period-based and filters only the charging-cost-rate CTE. Selecting Home therefore means "cost per 100 km using Home charging prices" without changing the driving estimate.

For the mixed drive/charge SOC panels, only the charging-process branch is filtered. Drive rows remain available as sequence boundaries.

## Validation

- `python -m json.tool grafana/dashboards/charging-stats.json`
- `git diff --check`
- structural JSON verification for the selector, placeholder query, 19 filtered SQL targets, qualified aliases, and the special cost-per-100-km placement
- smoke-tested `All`, `Home`, and `Home + another location` selections against a TeslaMate PostgreSQL deployment
- representative Home average-cost aggregation: `EXPLAIN (ANALYZE, BUFFERS)` execution time `1.245 ms`

### DIFF
--- a/grafana/dashboards/charging-stats.json
+++ b/grafana/dashboards/charging-stats.json
@@ -113,7 +113,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT\n\tcount(*)\nFROM\n\tcharging_processes\nWHERE\n\t$__timeFilter(end_date)\n\tAND duration_min >= $min_duration\n\tAND car_id = $car_id;",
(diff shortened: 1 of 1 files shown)


JakobLichterfeld (approved):
Nice addidion, ty for your contribution!

PR #5376, opened 2026-06-08
feat: enable parquet buffer migration to prevent unified-storage secrets error on Grafana startup

RafaelPaulovic:
## What

Bump Grafana base image `13.0.1-security-01` → `13.0.2`.

Set `GF_UNIFIED_STORAGE_MIGRATION_PARQUET_BUFFER=true`.

## Why

`13.0.2` is the regular release superseding the out-of-band `13.0.1-security-01` patch. No `-security` variant exists for `13.0.2`. Tag is multi-arch (amd64, arm64, arm/v7).

Per the [13.0 upgrade guide](https://grafana.com/docs/grafana/latest/upgrade-guide/upgrade-v13.0/#sqlite-deployments), SQLite deployments can hit `database is locked` errors during unified storage migration. Grafana states it automatically retries using the Parquet buffer, but `migration_parquet_buffer` defaults to `false` — this PR enables it explicitly via env var to make the behavior deterministic rather than relying on error-triggered auto-retry.

The `SQLITE_BUSY` error is visible in startup logs without the flag:

```
logger=secrets level=error msg="Failed to get current data key" error="database is locked (5) (SQLITE_BUSY)"
```

## Testing

Built and ran the image locally (arm64, Apple Silicon):

- `version=13.0.2`, `/api/health` → `database: ok` ✓
- Datasource + all 23 dashboards provisioned, 0 parse errors ✓
- DB migrations: 710 executed, 0 errors ✓
- `Config overridden from Environment variable var="GF_UNIFIED_STORAGE_MIGRATION_PARQUET_BUFFER=true"` confirmed in startup logs ✓

`pluginVersion` strings left at `13.0.1+security-01` — a proper export against a live 13.0.2 instance is out of scope for this PR.

Follows up on #5324.

### DIFF
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -1,6 +1,6 @@
 # Ensure selecting a tag that is available for arm/v7, arm64, and amd64
 # https://hub.docker.com/r/grafana/grafana/tags
-FROM grafana/grafana:13.0.1-security-01
+FROM grafana/grafana:13.0.2
 
 ENV GF_ANALYTICS_REPORTING_ENABLED=false \
     GF_ANALYTICS_CHECK_FOR_UPDATES=false \
@@ -26,6 +26,7 @@ ENV GF_ANALYTICS_REPORTING_ENABLED=false \
     GF_FEATURE_TOGGLES_splashScreen=false \
     GF_FEATURE_TOGGLES_newPanelPadding=false \
     GF_FEATURE_TOGGLES_dashboardNewLayouts=false \
+    GF_UNIFIED_STORAGE_MIGRATION_PARQUET_BUFFER=true \
     DATABASE_PORT=5432 \
     DATABASE_SSL_MODE=disable
 
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -96,7 +96,7 @@
         "textMode": "value_and_name",
         "wideLayout": true
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -239,7 +239,7 @@
         "textMode": "value_and_name",
         "wideLayout": true
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -328,7 +328,7 @@
         "textMode": "value_and_name",
         "wideLayout": true
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -538,7 +538,7 @@
           "sort": "none"
         }
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -645,7 +645,7 @@
         "sparkline": false,
         "textMode": "auto"
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -745,7 +745,7 @@
         "sizing": "auto",
         "valueMode": "color"
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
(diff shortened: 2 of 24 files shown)


RafaelPaulovic:
From: https://www.reddit.com/r/grafana/comments/1tzujre/grafana_13_broke_my_teslamate_dashboards_after/

@JakobLichterfeld could you please take a look?

JakobLichterfeld:
Thanks for your / your Claude subscriptions contribution.

I do not like mixing two different things in one PR.

> From: https://www.reddit.com/r/grafana/comments/1tzujre/grafana_13_broke_my_teslamate_dashboards_after/

From Grafana https://grafana.com/whats-new/2026-04-14-unified-storage-for-folders-and-dashboards/

> The bug only affects self-managed instances with Git Sync feature flags

TeslaMate grafana image does not use this and the env is not needed, from what I currently oversee.

RafaelPaulovic:
Hi @JakobLichterfeld, thanks for taking a look. The unified storage migration happens regardless of the Git Sync (provisioning) feature toggle. The reported issue in the Reddit post is happening in main when running the docker image locally.

To fix the issue, it requires bumping the Grafana version and setting the parquet buffer for the migration instead (or using a config file instead of env. var. override to enable the unified storage parquet migrator).

JakobLichterfeld (approved):
Thanks for explanation.

lgtm. I could not find a corresponding setting for NixOS, so module.nix keeps unchanged. (`"unified.storage.migration.parquet".buffer = false;` in `services.grafana.settings` would be a bit hacky imo)

@swiffer may I ask you to take a look as well?

swiffer:
haven't encountered the issue myself but that workaround is described here:

https://grafana.com/docs/grafana/latest/upgrade-guide/upgrade-v13.0/#sqlite-deployments

wonder why setting it manually is required given that quote -> "Grafana automatically retries using the Parquet buffer"

comment on bumping the pluginVersion to "13.0.2" - i haven't simply bumped the version string and instead exported the dashboards after upgrade to ensure syntax is correct and exports have been done with the version stated in the files. while changes are minor it's most often the case that new version bring to options to panel config.

i therefore recommend to either do that or revert the pluginVersion string to "13.0.1-security-01".

swiffer:
@JakobLichterfeld - have been kept quite busy the last weeks, I plan to start working on grafana and db-performance related issues again mid of next week!

JakobLichterfeld:
> wonder why setting it manually is required given that quote -> "Grafana automatically retries using the Parquet buffer"

Good question, which brings us back to the point, why we do not have any other report of this issue.

> i haven't simply bumped the version string and instead exported the dashboards after upgrade to ensure syntax is correct and exports have been done with the version stated in the files. while changes are minor it's most often the case that new version bring to options to panel config.
> 
> i therefore recommend to either do that or revert the pluginVersion string to "13.0.1-security-01".

Totally right. I assumed the Rafael did this and not just change the string.

JakobLichterfeld:
> @JakobLichterfeld - have been kept quite busy the last weeks, I plan to start working on grafana and db-performance related issues again mid of next week!

That didn’t go unnoticed 😄 I briefly wondered if my ‘comeback’ came across a bit too strong after my time away :-)
Jokes aside, we’ll keep the Grafana/db gremlins contained until you’re back mid-next week 😄

RafaelPaulovic:
Thanks for the review @swiffer!

On SSD, the auto-retry resolves and the migration completes, but in slow disks (Reddit user had Synology DS918+, RAID5, Btrfs), the issue did not auto-resolve and required the user to troubleshoot. In that case, the docs recommends enabling `migration_parquet_buffer`. Assuming TeslaMate typical deployment is hardware with slow-disk (Synology NAS, Raspberry Pi, ...), the flag is a meaningful safeguard for the users. We are also considering making parquet the default migration path in a future version.

Regarding pluginVersion, I reverted dashboard versions to 13.0.1+security-01.

grafana/Dockerfile
```diff
@@ -1,6 +1,6 @@
-# Ensure selecting a tag that is available for arm64 and amd64
+# Ensure selecting a tag that is available for arm/v7, arm64, and amd64
 # https://hub.docker.com/r/grafana/grafana/tags
-FROM grafana/grafana:13.0.1-security-01
+FROM grafana/grafana:13.0.2
```
JakobLichterfeld, line 3: > Regarding pluginVersion, I reverted dashboard versions to 13.0.1+security-01.

You don't please revert correctly

JakobLichterfeld (changes requested):

PR #5378, opened 2026-06-09
fix: restore vehicle_unavailable fallback for streaming fetches to prevent unwanted vehicle online state

ciyahu:
## Summary

This restores the existing `fetch_with_reachable_assumption/2` fallback for `use_streaming_api: true` fetches when vehicle data is allowed.

Before v3.1.0, reachable states used `fetch_with_reachable_assumption/2`, so `get_vehicle_with_state` returning `:vehicle_unavailable` would fall back to `get_vehicle`. After the MCU2/fake-online streaming changes, the streaming branch calls `get_vehicle_with_state` directly and no longer falls back.

This can leave a vehicle stuck in `:online` while `/vehicle_data` keeps returning 408, even though the lightweight vehicle endpoint reports `asleep` or `offline`.

## Change

Use `fetch_with_reachable_assumption/2` in the streaming branch whenever `allow_vehicle_data?` is true.

## Test

Added a regression test covering:

`online -> vehicle_data unavailable -> lightweight vehicle state asleep -> state transitions to asleep`

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -1507,7 +1507,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
           end
 
         if allow_vehicle_data? do
-          call(deps.api, :get_vehicle_with_state, [car.eid])
+          fetch_with_reachable_assumption(car.eid, deps)
         else
           call(deps.api, :get_vehicle, [car.eid])
         end
--- a/test/support/mocks/api.ex
+++ b/test/support/mocks/api.ex
@@ -32,12 +32,12 @@ defmodule ApiMock do
   @impl true
   def handle_call({action, _id}, _from, %State{events: [event | []]} = state)
       when action in [:get_vehicle, :get_vehicle_with_state] do
-    {:reply, exec(event), state}
+    {:reply, exec(event, action), state}
   end
 
   def handle_call({action, _id}, _from, %State{events: [event | events]} = state)
       when action in [:get_vehicle, :get_vehicle_with_state] do
-    {:reply, exec(event), %State{state | events: events}}
+    {:reply, exec(event, action), %State{state | events: events}}
   end
 
   def handle_call({:sign_in, _tokens} = event, _from, %State{pid: pid} = state) do
@@ -50,6 +50,17 @@ defmodule ApiMock do
     {:reply, {:ok, pid}, state}
   end
 
-  defp exec(event) when is_function(event), do: event.()
-  defp exec(event), do: event
+  # Events tagged with :get_vehicle or :get_vehicle_with_state may only be
+  # consumed by that API call, allowing tests to pin which endpoint was used.
+  defp exec({expected_action, event}, action)
+       when expected_action in [:get_vehicle, :get_vehicle_with_state] do
+    if expected_action != action do
+      raise "expected #{inspect(expected_action)} to be called, but got #{inspect(action)}"
+    end
+
+    exec(event, action)
+  end
+
+  defp exec(event, _action) when is_function(event), do: event.()
+  defp exec(event, _action), do: event
 end
--- a/test/teslamate/vehicles/vehicle_test.exs
+++ b/test/teslamate/vehicles/vehicle_test.exs
@@ -241,6 +241,32 @@ defmodule TeslaMate.Vehicles.VehicleTest do
     end
 
     @tag :capture_log
+    test "falls back to vehicle state when vehicle data is unavailable while streaming", %{
+      test: name
+    } do
+      now_ts = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
+
+      events = [
+        {:ok, online_event(now_ts)},
+        {:error, :vehicle_unavailable},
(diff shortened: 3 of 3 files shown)


JakobLichterfeld (commented):
Nice catch! Ty for your contribution.

@brianmay may I ask you to take a look as well?

brianmay:
I will have a look when I get a chance. Might be several days, not sure right now.

brianmay:
LLM says:

"One thing to note about the test infrastructure: ApiMock does not distinguish between get_vehicle and get_vehicle_with_state — both consume from the same event queue. This means the test implicitly assumes the correct call order but does not explicitly verify which API function is called for which event slot. It's not a bug, but it means the test would pass even if the calls were swapped in some hypothetical future refactor."

Which is a pre-existing issue before this PR, but means the tests unfortunately will fail to any detect issues where the wrong function is called.

Not sure how to fix this right now though.

lib/teslamate/vehicles/vehicle.ex
```diff
@@ -1507,7 +1507,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
           end
 
         if allow_vehicle_data? do
-          call(deps.api, :get_vehicle_with_state, [car.eid])
+          fetch_with_reachable_assumption(car.eid, deps)
         else
           call(deps.api, :get_vehicle, [car.eid])
```
brianmay, line 1512: Should this line be replaced also?

```
fetch_with_unreachable_assumption(car.eid, deps)
```

Then it will match the older non-streaming condition.

Or is there something I missed here?

brianmay (commented):
Looks good to me. ~~But wondering if the other line should be changed also.~~

(these two lines being different has always bothered me for a while now).

lib/teslamate/vehicles/vehicle.ex
brianmay: Upon consulting with LLM, I realize this would be wrong, we would bypass the fake wakeup detection.

brianmay:
LLM says there is a minor behavior change:

>When the car is genuinely awake but /vehicle_data transiently 408s while driving or charging, the old code retried vehicle_data in 10s/15s (error handler intervals). The new code gets {:ok, online} from the fallback and lands in the catch-all at vehicle.ex:415, which schedules the next fetch at @asleep_interval (30s). While driving the stream covers position data, so impact is small, but during charging this stretches the polling gap for charge data from 15s to 30s on flaky connections. Pre-3.1.0 behaved the same way, so this is a restoration, not a regression — just worth knowing.

brianmay (approved):

brianmay:
LLM suggests the following to strengthen the tests, and make sure teslamate uses the correct call.

```diff
diff --git a/test/support/mocks/api.ex b/test/support/mocks/api.ex
index 820829d16e..39bbd6569f 100644
--- a/test/support/mocks/api.ex
+++ b/test/support/mocks/api.ex
@@ -32,12 +32,12 @@
   @impl true
   def handle_call({action, _id}, _from, %State{events: [event | []]} = state)
       when action in [:get_vehicle, :get_vehicle_with_state] do
-    {:reply, exec(event), state}
+    {:reply, exec(event, action), state}
   end

   def handle_call({action, _id}, _from, %State{events: [event | events]} = state)
       when action in [:get_vehicle, :get_vehicle_with_state] do
-    {:reply, exec(event), %State{state | events: events}}
+    {:reply, exec(event, action), %State{state | events: events}}
   end

   def handle_call({:sign_in, _tokens} = event, _from, %State{pid: pid} = state) do
@@ -50,6 +50,17 @@
     {:reply, {:ok, pid}, state}
   end

-  defp exec(event) when is_function(event), do: event.()
-  defp exec(event), do: event
+  # Events tagged with :get_vehicle or :get_vehicle_with_state may only be
+  # consumed by that API call, allowing tests to pin which endpoint was used.
+  defp exec({expected_action, event}, action)
+       when expected_action in [:get_vehicle, :get_vehicle_with_state] do
+    if expected_action != action do
+      raise "expected #{inspect(expected_action)} to be called, but got #{inspect(action)}"
+    end
+
+    exec(event, action)
+  end
+
+  defp exec(event, _action) when is_function(event), do: event.()
+  defp exec(event, _action), do: event
 end
diff --git a/test/teslamate/vehicles/vehicle_test.exs b/test/teslamate/vehicles/vehicle_test.exs
index 4cbae3b443..393a58da3d 100644
--- a/test/teslamate/vehicles/vehicle_test.exs
+++ b/test/teslamate/vehicles/vehicle_test.exs
@@ -249,7 +249,7 @@
       events = [
         {:ok, online_event(now_ts)},
         {:error, :vehicle_unavailable},
-        {:ok, %TeslaApi.Vehicle{state: "asleep"}}
+        {:get_vehicle, {:ok, %TeslaApi.Vehicle{state: "asleep"}}}
       ]

       :ok = start_vehicle(name, events)
```

ciyahu:
Thanks for taking a look and for the helpful suggestion.

That makes sense to me. I updated `ApiMock` to support tagging expected API calls, and adjusted the regression test so it now explicitly verifies that the fallback consumes the `:get_vehicle` event.

JakobLichterfeld:
Awesome, ty!

PR #5379, opened 2026-06-10
fix: properly clean up state and delete tokens on sign out

jlestel:
When you sign out via the Settings page, the Tesla API tokens are only removed from the in-memory cache (ETS), but they remain stored in the private.tokens database table. As a result, if your application restarts the tokens are read back from the DB, refreshed, and vehicle logging resumes automatically (effectively logging you back in even though you explicitly signed out).

**To reproduce**
Sign in with your Tesla account.
Confirm a token pair exists: SELECT count(*) FROM private.tokens; → returns 1.
Go to Settings and click Sign out.
Check again: SELECT count(*) FROM private.tokens; → still returns 1 ❌.
Restart TeslaMate.
The app boots already signed in and resumes recording, instead of showing the sign-in screen.

**Expected behavior**
After signing out, the tokens should be removed from the database (and recording should stop). A restart should land on the sign-in screen, with no recording resuming.

**Root cause**
TeslaMate.Api.sign_out/1 only calls :ets.delete(name, :auth) and it never touches the database.
TeslaMate.Auth has no token-deletion function (get and save only).
On boot, TeslaMate.Api.init/1 restores the tokens from private.tokens, refreshes them, and restarts the vehicle loggers.

### DIFF
--- a/lib/teslamate/api.ex
+++ b/lib/teslamate/api.ex
@@ -81,10 +81,10 @@ defmodule TeslaMate.Api do
   end
 
   def sign_out(name \\ @name) do
-    true = :ets.delete(name, :auth)
-    :ok
-  rescue
-    _ in ArgumentError -> {:error, :not_signed_in}
+    case fetch_auth(name) do
+      {:error, :not_signed_in} -> {:error, :not_signed_in}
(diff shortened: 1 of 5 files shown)


JakobLichterfeld (approved):
What a catch. Ty for fixing!

PR #5380, opened 2026-06-10
ci: sign published images with SLSA provenance + SBOM attestations

oivindoh:
Publishes signed SLSA build provenance for every TeslaMate image (Docker Hub + GHCR), plus SPDX SBOMs for teslamate/teslamate, generated keylessly via Sigstore using the workflow's GitHub OIDC identity.

Consumers can prove an image was built by this repo's CI from a specific commit, with no shared keys to manage:

```sh
gh attestation verify oci://teslamate/teslamate:latest --repo teslamate-org/teslamate
```

## What's attested

- teslamate/teslamate (Docker Hub + GHCR): SLSA provenance on the manifest-list and per-platform digests, plus an SPDX SBOM attached to each per-platform digest.
- teslamate/grafana (Docker Hub + GHCR): SLSA provenance (built as a single multi-arch image).
- Attestations are skipped on pull_request builds, so unmerged code can't produce a signed image that verifies against teslamate-org/teslamate and reads as official.

## How

- Uses actions/attest-build-provenance and actions/attest (SPDX), pushed to the registry as referrers; SBOMs generated by anchore/sbom-action (syft).
- Disables BuildKit-side attestations (provenance and sbom off) so the build digest is the plain image digest that appears in the multi-arch manifest list. BuildKit's own provenance wraps the image in an OCI index whose digest tag-based verification would miss.
- Requires id-token: write and attestations: write.
- REGISTRY_IMAGE stays repointable; the attestation subject-name is qualified with docker.io only when unqualified, so building test images on another registry is unaffected.

## Docs

Verification how-to lives in the docs site (website/docs/advanced_guides/verifying_images.md)

## Tested

Validated end-to-end on a fork: provenance and per-platform SBOMs for teslamate/teslamate verify with gh attestation verify, the per-platform SBOMs are confirmed arch-specific, and a pull_request build was confirmed to produce no attestation.

### DIFF
--- a/.github/actions/grafana/action.yml
+++ b/.github/actions/grafana/action.yml
@@ -41,20 +41,6 @@ runs:
         sbom: false
         cache-from: type=gha
         cache-to: type=gha,mode=max
-    - name: Resolve attestation subject name
-      # See build action: qualify Docker Hub short names with docker.io so
-      # actions/attest-* can find the registry, without pinning the image input.
-      id: subject
-      if: github.event_name != 'pull_request'
-      shell: bash
-      env:
-        IMAGE: ${{ inputs.image }}
-      run: |
-        case "${IMAGE%%/*}" in
-          *.* | *:* | localhost) name="$IMAGE" ;;
-          *) name="docker.io/$IMAGE" ;;
-        esac
-        echo "name=$name" >> "$GITHUB_OUTPUT"
     - name: Attest grafana provenance
       # grafana is a single multi-arch build, so steps.build.outputs.digest is
       # the manifest-list digest that `gh attestation verify oci://...:tag`
@@ -62,6 +48,10 @@ runs:
       if: github.event_name != 'pull_request'
       uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
       with:
-        subject-name: ${{ steps.subject.outputs.name }}
+        # See build action: qualify Docker Hub short names with docker.io.
+        subject-name: >-
+          ${{ startsWith(inputs.image, 'ghcr.io/')
+              && inputs.image
+              || format('docker.io/{0}', inputs.image) }}
         subject-digest: ${{ steps.build.outputs.digest }}
         push-to-registry: true
--- a/.github/actions/grafana/action.yml
+++ b/.github/actions/grafana/action.yml
@@ -27,11 +27,41 @@ runs:
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
     - name: Build and push
+      id: build
       uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
       with:
         context: grafana
         push: true
         platforms: linux/amd64,linux/arm/v7,linux/arm64
         tags: ${{ steps.docker_meta.outputs.tags }}
+        # See build action for why BuildKit-side attestations are disabled:
+        # we want outputs.digest to be the multi-arch manifest list digest so
+        # that signed attestations land where `gh attestation verify` looks.
+        provenance: false
+        sbom: false
         cache-from: type=gha
         cache-to: type=gha,mode=max
+    - name: Resolve attestation subject name
+      # See build action: qualify Docker Hub short names with docker.io so
+      # actions/attest-* can find the registry, without pinning the image input.
+      id: subject
+      if: github.event_name != 'pull_request'
+      shell: bash
+      env:
+        IMAGE: ${{ inputs.image }}
+      run: |
+        case "${IMAGE%%/*}" in
+          *.* | *:* | localhost) name="$IMAGE" ;;
+          *) name="docker.io/$IMAGE" ;;
+        esac
+        echo "name=$name" >> "$GITHUB_OUTPUT"
+    - name: Attest grafana provenance
+      # grafana is a single multi-arch build, so steps.build.outputs.digest is
+      # the manifest-list digest that `gh attestation verify oci://...:tag`
+      # resolves to. Skip on pull_request: see build action for the rationale.
+      if: github.event_name != 'pull_request'
+      uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
+      with:
+        subject-name: ${{ steps.subject.outputs.name }}
+        subject-digest: ${{ steps.build.outputs.digest }}
+        push-to-registry: true
--- a/.github/workflows/buildx.yml
+++ b/.github/workflows/buildx.yml
@@ -19,6 +19,8 @@ env:
 permissions:
   contents: read
   packages: write
+  id-token: write
+  attestations: write
 
 jobs:
   check_paths:
--- a/.github/workflows/ghcr_build.yml
+++ b/.github/workflows/ghcr_build.yml
@@ -22,6 +22,8 @@ env:
 permissions:
   contents: read
   packages: write
+  id-token: write
+  attestations: write
 
 jobs:
   check_paths:
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -68,11 +68,62 @@ runs:
         platforms: ${{ matrix.platform }}
         labels: ${{ steps.meta.outputs.labels }}
         tags: ${{ steps.docker_meta.outputs.tags }}
+        # Disable BuildKit-side attestations so outputs.digest is the bare image
+        # manifest digest. With provenance enabled, BuildKit wraps the image in
+        # an OCI index whose digest does NOT appear in the multi-arch manifest
+        # list produced by `imagetools create`, which would make tag-based
+        # attestation verification fail. Signed SLSA provenance is added by
+        # actions/attest-build-provenance below.
+        provenance: false
+        sbom: false
         cache-from: |
           type=registry,ref=ghcr.io/${{ inputs.repository }}:buildcache-${{ matrix.cache_id }}-${{ steps.normalize_version.outputs.normalized_version }}
           type=registry,ref=ghcr.io/${{ inputs.repository }}:buildcache-${{ matrix.cache_id }}-main
         cache-to: type=registry,ref=ghcr.io/${{ inputs.repository }}:buildcache-${{ matrix.cache_id }}-${{ steps.normalize_version.outputs.normalized_version }},mode=max
         outputs: type=image,name=${{ env.REGISTRY_IMAGE }},push-by-digest=true,name-canonical=true,push=true
+    - name: Generate SBOM for pushed image
+      # Skip on pull_request: PR builds are unreviewed code. Attesting them
+      # would yield signed attestations that pass `gh attestation verify
+      # --repo teslamate-org/teslamate`, which consumers read as "official".
+      if: github.event_name != 'pull_request'
+      uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+      with:
+        image: ${{ env.REGISTRY_IMAGE }}@${{ steps.build.outputs.digest }}
+        format: spdx-json
+        output-file: /tmp/sbom-${{ matrix.cache_id }}.spdx.json
+        upload-artifact: false
+        upload-release-assets: false
+    - name: Resolve attestation subject name
+      # attest-* reads the registry from subject-name's first path segment, so a
+      # bare "teslamate/teslamate" fails. Qualify short names with docker.io;
+      # leave already-qualified names (ghcr.io/...) so REGISTRY_IMAGE stays repointable.
+      id: subject
+      if: github.event_name != 'pull_request'
+      shell: bash
+      env:
+        IMAGE: ${{ env.REGISTRY_IMAGE }}
+      run: |
+        case "${IMAGE%%/*}" in
+          *.* | *:* | localhost) name="$IMAGE" ;;
+          *) name="docker.io/$IMAGE" ;;
+        esac
+        echo "name=$name" >> "$GITHUB_OUTPUT"
+    - name: Attest build provenance
+      if: github.event_name != 'pull_request'
+      uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
+      with:
+        subject-name: ${{ steps.subject.outputs.name }}
+        subject-digest: ${{ steps.build.outputs.digest }}
+        push-to-registry: true
+    - name: Attest SBOM
+      if: github.event_name != 'pull_request'
+      uses: actions/attest@59d89421af93a897026c735860bf21b6eb4f7b26 # v4.1.0
+      with:
+        subject-name: ${{ steps.subject.outputs.name }}
+        subject-digest: ${{ steps.build.outputs.digest }}
+        predicate-type: "https://spdx.dev/Document"
+        predicate-path: /tmp/sbom-${{ matrix.cache_id }}.spdx.json
+        push-to-registry: true
     - name: Export digest
       shell: bash
       run: |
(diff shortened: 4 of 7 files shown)


brianmay:
I ran this change through an LLM (Claude Code with Fable 5), and got the following feedback. Not sure if I agree or not with all these points, but probably worth discussion.


> Main concern: PR builds get official signed provenance
> 
> ghcr_build.yml runs on pull_request (same-repo PRs pass the check_if_pr_from_outside_repo gate and do build/push). With id-token: write now granted and the attest steps unconditional, every same-repo PR build produces a signed attestation that passes gh attestation verify --repo teslamate-org/teslamate — including unreviewed, unmerged code. SECURITY.md tells consumers verification proves the image "was built by this repo's CI from a specific commit," which is technically accurate, but most users will read a passing verification as "this is an official image." The predicate does record the ref, but nobody checks that by hand.
> 
> Suggestion: gate the three attest/SBOM steps in .github/actions/build/action.yml (and the merge-action step) with if: github.event_name != 'pull_request' — composite action steps support if: with the github context — or add an action input so the workflows opt in explicitly. Cheap to do, removes the ambiguity entirely.
> 
> Other findings
> 
> - Grafana SBOM subject is misleading (.github/actions/grafana/action.yml). The grafana build pushes three platforms in one build-push-action call, so outputs.digest is the manifest list digest. syft scanning image@<list-digest> SBOMs only the runner's platform (amd64), yet the attestation binds that SBOM to the multi-arch digest — and contradicts SECURITY.md's "SBOMs are attached to each per-platform image digest." Either scope the SBOM with syft's --platform, or note the grafana exception in SECURITY.md.
> - Manifest digest capture works but is roundabout (.github/actions/merge/action.yml). Hashing imagetools inspect --raw output is correct (raw bytes = digest preimage), but docker buildx imagetools inspect "$TAG" --format '{{json .Manifest}}' | jq -r .digest gets the digest directly without re-hashing — less to go wrong if buildx ever alters raw output framing.
> - Shell interpolation nit (same step): TAG="${{ env.REGISTRY_IMAGE }}:${{ steps.meta.outputs.version }}" injects an expression into bash. The version derives from ref names and only same-repo refs reach this code (and line 46 already does the same), so risk is low — but passing via an env: block is the hardened pattern for new code.
> - Dispatch fallback tag pollutes Docker Hub (buildx.yml). type=raw,value=${{ github.ref_name }} on dispatch publishes e.g. teslamate/teslamate:main — a new public tag users may pull and which only updates on manual dispatch, so it goes stale. A branch named like 1.30 would also collide with the semver tags. Consider namespacing it (value=dispatch-${{ github.ref_name }}) or pushing dispatch tests to a separate repo/tag prefix.
> - Build job timeouts: SBOM generation pulls and scans the just-pushed image inside jobs with timeout-minutes: 10. The arm builds are already the slow path; worth watching the first few runs and bumping the timeout if syft adds 2–3 minutes.
> - workflow_dispatch gating looks consistent across both workflows; check_if_pr_from_outside_repo correctly falls to false for non-PR events, and dorny/paths-filter handles dispatch-from-branch with an explicit base. No issues found there.

The "Main concern"; I have mixed feelings here, only trusted people can do this; but then again only changes to main are considered official.

The first point looks like a genuine bug.

ON the 2nd point, another LLM said the line:

```
DIGEST="sha256:$(docker buildx imagetools inspect "$TAG" --raw | sha256sum | awk '{print $1}')"
```

Should be replaced with:

```
DIGEST=$(docker buildx imagetools inspect "$TAG" | awk '/^Digest:/{print $2}')
```

3rd point, should be easy to fix I think. I have got caught by this issue myself. If this followed shell script rules, it would be OK, but this reference gets expanded in place, and does not follow shell script rules. If that is confusing let me know and I will elaborate.

4th - 5th point, no immediate action required.

oivindoh:
Nice catch, I believe I've addressed all the findings now, though the grafana solution may not be optimal. I haven't touched timeouts yet - everything built fine during tests on my side, including the full run when bumping main.

brianmay (approved):

.github/actions/grafana/action.yml
```diff
@@ -11,7 +11,7 @@ inputs:
   image:
     description: "Image target"
     required: true
-    default: "teslamate/grafana"
+    default: "docker.io/teslamate/grafana"
```
JakobLichterfeld: This will break building gh test images without changing .github/workflows/buildx.yml accordingly

.github/actions/grafana/action.yml
```diff
@@ -27,11 +27,101 @@ runs:
...
+    # platform's layers regardless of the runner architecture.
+    - name: Resolve per-platform digests
+      id: platforms
+      if: github.event_name != 'pull_request'
+      shell: bash
+      env:
+        IMAGE: ${{ inputs.image }}
+        DIGEST: ${{ steps.build.outputs.digest }}
+      run: |
+        manifests="$(docker buildx imagetools inspect "${IMAGE}@${DIGEST}" --format '{{json .Manifest.Manifests}}')"
```
JakobLichterfeld: This looks way too hacky. Either refactor to use matrix build or drop entierly

.github/actions/grafana/action.yml
```diff
@@ -27,11 +27,101 @@ runs:
...
+        amd64="$(jq -r '.[] | select(.platform.architecture=="amd64") | .digest' <<<"$manifests")"
+        armv7="$(jq -r '.[] | select(.platform.architecture=="arm" and .platform.variant=="v7") | .digest' <<<"$manifests")"
+        arm64="$(jq -r '.[] | select(.platform.architecture=="arm64") | .digest' <<<"$manifests")"
+        echo "amd64=${amd64}" >> "$GITHUB_OUTPUT"
+        echo "armv7=${armv7}" >> "$GITHUB_OUTPUT"
```
JakobLichterfeld: We do not support armv7 anymore

.github/actions/grafana/action.yml
```diff
@@ -27,11 +27,101 @@ runs:
...
+    - name: Generate SBOM (amd64)
+      if: github.event_name != 'pull_request'
+      uses: anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+      with:
+        image: ${{ inputs.image }}@${{ steps.platforms.outputs.amd64 }}
+        format: spdx-json
+        output-file: /tmp/grafana-sbom-amd64.spdx.json
+        upload-artifact: false
+        upload-release-assets: false
+    - name: Generate SBOM (arm/v7)
```
JakobLichterfeld: We do not support armv7 anymore

.github/workflows/buildx.yml
```diff
@@ -14,19 +14,26 @@ on:
       - "!.github/workflows/**" # Important: Exclude PRs related to .github/workflows from auto-run
       - "!.github/actions/**" # Important: Exclude PRs related to .github/actions from auto-run
 env:
-  REGISTRY_IMAGE: teslamate/teslamate
+  # Fully qualified: `actions/attest-build-provenance` looks up registry
+  # credentials by the first slash-separated component of subject-name. The
+  # bare `teslamate/teslamate` form makes it search for credentials under a
+  # registry named "teslamate" and fail; "docker.io/teslamate/teslamate" is
+  # unambiguous and resolves to the existing Docker Hub login.
+  REGISTRY_IMAGE: docker.io/teslamate/teslamate
```
JakobLichterfeld: This will break gh test image build without changing .github/workflows/buildx.yml accordingly.

.github/workflows/buildx.yml
```diff
@@ -14,19 +14,26 @@ on:
...
+  attestations: write
 
 jobs:
   check_paths:
     uses: ./.github/workflows/check_paths.yml
 
   teslamate_build:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/buildx.yml
```diff
@@ -62,7 +69,7 @@ jobs:
     needs:
       - check_paths
       - teslamate_build
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/buildx.yml
```diff
@@ -81,9 +88,10 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=edge
+            type=raw,value=dispatch-${{ github.ref_name }},enable=${{ github.event_name == 'workflow_dispatch' }}
   grafana:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/ghcr_build.yml
```diff
@@ -1,6 +1,7 @@
 name: Build GHCR images
 
 on:
+  workflow_dispatch:
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/ghcr_build.yml
```diff
@@ -22,22 +23,24 @@ env:
...
+  attestations: write
 
 jobs:
   check_paths:
     uses: ./.github/workflows/check_paths.yml
 
   check_if_pr_from_outside_repo:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder == 'false'
+    if: github.event_name == 'workflow_dispatch' || needs.check_paths.outputs.githubfolder == 'false'
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/ghcr_build.yml
```diff
@@ -22,22 +23,24 @@ env:
...
     uses: ./.github/workflows/check_if_pr_from_outside_repo.yml
 
   teslamate_build:
     name: Build images
     needs:
       - check_paths
       - check_if_pr_from_outside_repo
-    if: needs.check_paths.outputs.githubfolder == 'false' && needs.check_if_pr_from_outside_repo.outputs.is_pr_from_outside_repo == 'false'
+    if: github.event_name == 'workflow_dispatch' || (needs.check_paths.outputs.githubfolder == 'false' && needs.check_if_pr_from_outside_repo.outputs.is_pr_from_outside_repo == 'false')
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/ghcr_build.yml
```diff
@@ -95,7 +98,7 @@ jobs:
...
```
JakobLichterfeld: workflow_dispatch will not be merged

.github/workflows/ghcr_build.yml
```diff
@@ -73,7 +76,7 @@ jobs:
     needs:
       - check_paths
       - teslamate_build
-    if: needs.check_paths.outputs.githubfolder == 'false'
+    if: github.event_name == 'workflow_dispatch' || needs.check_paths.outputs.githubfolder == 'false'
```
JakobLichterfeld: workflow_dispatch will not be merged

SECURITY.md
```diff
@@ -3,3 +3,53 @@
 ## Reporting a Vulnerability
 
 For reporting a security vulnerability, please contact `security AT teslamate DOT org`.
+
+## Verifying published container images
```
JakobLichterfeld: Thanks for writing detailed instruction!

Not quite sure if this is the right place for the doc and instruction. Doc lives in ./website/docs.

JakobLichterfeld (changes requested):
Thanks for this idea and your contribution oivindoh! I has been a while.

Please find below my findings from first look into it. Happy to be proved otherwise.

.github/workflows/buildx.yml
JakobLichterfeld: Or you found a bug in current dispatch, can't remember to test it. Perhaps we should drop it.

SECURITY.md
oivindoh: Moved to website

.github/workflows/ghcr_build.yml
oivindoh: Dropped

.github/workflows/ghcr_build.yml
oivindoh: Dropped

.github/workflows/ghcr_build.yml
oivindoh: Dropped

.github/workflows/ghcr_build.yml
oivindoh: Dropped

.github/workflows/buildx.yml
oivindoh: Just left behind for my testing, dropped

.github/workflows/buildx.yml
oivindoh: Dropped

.github/workflows/ghcr_build.yml
oivindoh: Dropped

.github/actions/grafana/action.yml
oivindoh: Dropped grafana SBOM, gets too messy for my liking

.github/actions/grafana/action.yml
oivindoh: Dropped - I branched prior to this

.github/actions/grafana/action.yml
oivindoh: Dropped

.github/workflows/buildx.yml
oivindoh: Hope [this](https://github.com/teslamate-org/teslamate/compare/e0006ee5e7f87b9f9f086a88f27bd72f19d34a24..2ee5b83499e8935597bac02b30b5635259aab7fa#diff-ec444582781b87b7476f7da1be059db3757d1b8a5c33cbe8fd1e72490844cd2fR105) is more palatable

.github/actions/grafana/action.yml
oivindoh: Unbroke this now

.github/workflows/buildx.yml
JakobLichterfeld: I don't like forced push after review, as it destroy all review references

oivindoh: Preference noted 👍

JakobLichterfeld (approved):
Thanks for the rework and detailed code comments!

Since TeslaMate only publishes to Docker Hub and GHCR, would it make sense to avoid the extra Resolve attestation subject name step? Either by storing `REGISTRY_IMAGE` fully-qualified from the start (`docker.io/... /` `ghcr.io/...`) or by inlining the Docker Hub qualification via a GitHub expression. That would remove the duplicated shell snippet from all three composite actions.

As I assume the tagging logic does not handle the fully-qualified name, I would suggest inlining with:

```yaml
subject-name: >-
  ${{ startsWith(env.REGISTRY_IMAGE, 'ghcr.io/')
      && env.REGISTRY_IMAGE
      || format('docker.io/{0}', env.REGISTRY_IMAGE) }}
```

JakobLichterfeld (approved):

JakobLichterfeld:
[Attestations](https://github.com/teslamate-org/teslamate/attestations) work, `artifact-metadata: write` missing, will fix

https://github.com/teslamate-org/teslamate/actions/runs/28796169905

JakobLichterfeld:
> `artifact-metadata: write` missing, will fix

Fixed with #5484 --> https://github.com/teslamate-org/teslamate/actions/runs/28800176722

PR #5388, opened 2026-06-13
fix: force HTTP/2 for Tesla auth token refresh requests

NirKli:
Refreshing tokens via auth.tesla.com over HTTP/1.1 causes Tesla to issue access tokens scoped to the Fleet API only, which then get a 403 "forbidden, see fleet-api" on legacy Owner API endpoints. Forcing HTTP/2 for the auth host avoids this downgrade.


longzheng:
This doesn't seem to fix the issue for me, I'm still getting `[warning] TeslaApi.Error / %{"error" => "forbidden, see https://developer.tesla.com/docs/fleet-api", "error_description" => "", "response" => nil}`

NirKli:
@longzheng Im closing this PR, I think I found the issue

PR #5397, opened 2026-06-13
Minor docs & code changes after bumping to OTP 28

### LINKED ISSUE #5402: no mention of supported platform in requirements
In https://docs.teslamate.org/docs/installation/docker#requirements there is no mention that we only support aarch64 and amd64 - and since #5391 - we no longer support arm7. This seems to be an unfortunate omission.

We probably need more documentation to anticipate expected issues:

* How do I tell if I am using arm7?
* What should I do if I am using arm7?
* I tried upgrading and didn't get the latest version.
* etc.

swiffer:
Did some minor refinements since we're requiring OTP 28 now. Noticed that we forgot to the supported Grafana version in the docs as well.

FreeBSD & Debian (Trixie) ship with Erlang 26 / 27 - therefore installs most likely do not work anymore when relying on those Erlang versions as per #5391.

We currently don't list any requirements (architecture wise) in the docker requirements - should we add them?

-> https://docs.teslamate.org/docs/installation/docker#requirements

### DIFF
--- a/.github/workflows/ghcr_purge.yml
+++ b/.github/workflows/ghcr_purge.yml
@@ -56,7 +56,7 @@ jobs:
         with:
           packages: "teslamate"
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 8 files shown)


JakobLichterfeld (approved):
TY! I added a detailed changelog.

Issue #5323, opened 2026-05-05
use of driving_interval inconsistant


brianmay:
Look at the code, I noticed when starting drive we use `driving_interval()` unconditionally:

```rust
      %V{drive_state: %Drive{shift_state: shift_state}} when shift_state in ~w(D N R) ->
        Logger.info("Start of drive initiated by: #{inspect(vehicle.drive_state)}")

        {drive, data} = start_drive(create_position(vehicle, data), data)

        {:next_state, {:driving, :available, drive}, data,
         [
           broadcast_summary(),
           schedule_fetch(driving_interval(), data)
         ]}
```

But when updating a driving the time varies depending on if streaming is enabled or not.

```rust
  def handle_event(
        :internal,
        {:update, {:online, vehicle}},
        {:driving, :available, drv},
        %Data{} = data
      ) do
    interval = if streaming?(data), do: default_interval(), else: driving_interval()
```

I might be mistaken, but this looks wrong...

Fixing this is low priority I think compared with other issues, but just wanted to document it regardless.

JakobLichterfeld:
Thanks for documenting.

JakobLichterfeld:
Fixed by #5417. Closing.

PR #5417, opened 2026-06-15
fix(vehicle): use streaming-aware interval when a drive starts

evanclan:
## Summary
Aligns the polling interval used when a drive is first detected with the interval already used for subsequent driving updates. Issue #5323 noted that drive start always scheduled `driving_interval()`, while the `:driving` update handler switches to `default_interval()` when streaming is active.

## Problem
When a drive begins via the polling path, the first scheduled fetch always uses `driving_interval()`:

```elixir
# lib/teslamate/vehicles/vehicle.ex (before)
schedule_fetch(driving_interval(), data)
```

During an active drive, updates already use a streaming-aware interval:

```elixir
interval = if streaming?(data), do: default_interval(), else: driving_interval()
```

With streaming enabled, that meant the first fetch after drive start polled at the faster driving cadence and then immediately switched to the default interval on the next update.

## Solution
Compute the same `interval` at drive start and pass it to `schedule_fetch/2`, matching the existing driving-update logic.

## Out of scope
- Other drive-start paths that originate from stream events (`schedule_fetch(0, data)`) — those already rely on stream delivery rather than polling cadence.
- Broader polling-interval refactors or new tests for the state machine.

## Test plan
- [ ] Verified the change mirrors the existing `handle_event/4` interval selection at `vehicle.ex` line ~1287.
- [ ] Elixir toolchain not available locally; CI to run the project test suite.

## Related
- #5323 — documents the inconsistent `driving_interval()` usage at drive start

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -1035,11 +1035,12 @@ defmodule TeslaMate.Vehicles.Vehicle do
         Logger.info("Start of drive initiated by: #{inspect(vehicle.drive_state)}")
 
         {drive, data} = start_drive(create_position(vehicle, data), data)
+        interval = if streaming?(data), do: default_interval(), else: driving_interval()
 
         {:next_state, {:driving, :available, drive}, data,
          [
            broadcast_summary(),
-           schedule_fetch(driving_interval(), data)
+           schedule_fetch(interval, data)
          ]}
 
       %V{charge_state: %Charge{charging_state: charging_state, battery_level: lvl}}


JakobLichterfeld:
Thanks for your contribution!

@brianmay May I ask you to review this, as it should solve your findings in #5323? Currently, test is not successful.

brianmay:
That test failure is weird:

```
** (UndefinedFunctionError) function System.convert_time_unit/3 is undefined or private
    (elixir 1.19.5) System.convert_time_unit(-576460718593916575, :native, :millisecond)
    (db_connection 2.10.1) lib/db_connection/connection_pool.ex:398: DBConnection.ConnectionPool.start_idle/2
** (EXIT from #PID<0.95.0>) shutdown
    (db_connection 2.10.1) lib/db_connection/connection_pool.ex:242: anonymous fn/5 in DBConnection.ConnectionPool.drop_idle/6

    (stdlib 7.3) gen_server.erl:2434: :gen_server.try_handle_info/3
    (stdlib 7.3) gen_server.erl:2420: :gen_server.handle_msg/3
    (stdlib 7.3) proc_lib.erl:333: :proc_lib.init_p_do_apply/3
Last message: {:timeout, #Reference<0.1718860511.67371009.181595>, -576460719593882020}
```

Didn't change anything here.

And I can't reproduce this failure locally :-(

brianmay:
Tests worked after retry...

JakobLichterfeld:
> Tests worked after retry...

That's wired, as I already retried them before.

brianmay (approved):

PR #5429, opened 2026-06-21
fix(mqtt): retry failed value publications

ciyahu:
## Summary

- Cache MQTT values only after they have been published successfully.
- Retry failed values when the next vehicle summary arrives.
- Distinguish missing cache entries from successfully published `nil` values.

## Problem

`VehicleSubscriber` previously replaced `last_values` with the complete
current value map regardless of individual MQTT publish results.

When a publication failed, the value was still treated as published.
Subsequent summaries containing the same value were filtered out, so a
transient MQTT failure could leave a retained topic missing or stale until
the value changed or the subscriber restarted.

## Solution

Associate each asynchronous publish result with its topic and value, then
add only successful publications to `last_values`.

Failed publications remain uncached and are retried with the next summary.
`Map.has_key?/2` is used to distinguish a value that has never been
published from a successfully published `nil` value.

## Test plan

- Verify that a failed `display_name` publication is retried.
- Verify that a failed `nil` `shift_state` publication is retried.
- Verify that successfully published unchanged values are not published again.
- Full DevOps workflow passed, including formatting, tests, and static analysis.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -67,37 +67,36 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
       |> add_geofence(summary)
       |> add_active_route(summary)
 
-    publish_values(values, state)
-    {:noreply, %{state | last_values: values}}
-  end
-
-  defp publish_values(values, %State{last_values: values} = state) do
-    values
-    |> Map.take(@do_not_retain)
-    |> Enum.each(fn {key, value} ->
-      publish({key, value}, state)
-    end)
+    last_values = publish_values(values, state)
+    {:noreply, %{state | last_values: last_values}}
   end
 
   defp publish_values(values, state) do
+    last_values = state.last_values || %{}
+
     values
     |> Stream.reject(&match?({_key, :unknown}, &1))
     |> Stream.filter(fn {key, value} ->
       ((key in @publish_if_nil or value != nil) and
-         (state.last_values == nil or Map.get(state.last_values, key) != value)) or
+         (not Map.has_key?(last_values, key) or Map.get(last_values, key) != value)) or
         key in @do_not_retain
     end)
-    |> Task.async_stream(&publish(&1, state),
+    |> Task.async_stream(fn entry -> {entry, publish(entry, state)} end,
       max_concurrency: 10,
       on_timeout: :kill_task,
       ordered: false
     )
-    |> Enum.each(fn
-      {_, reason} when reason != :ok ->
-        Logger.warning("MQTT publishing failed: #{inspect(reason)}")
+    |> Enum.reduce(last_values, fn
+      {:ok, {{key, value}, :ok}}, acc ->
+        Map.put(acc, key, value)
 
-      _ok ->
-        nil
+      {:ok, {_entry, reason}}, acc ->
+        Logger.warning("MQTT publishing failed: #{inspect(reason)}")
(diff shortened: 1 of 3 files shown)


ciyahu:
This is essentially harmless for frequently-changing fields (speed, power, battery level, etc.), since the next summary a few seconds later re-publishes and corrects the value. The real problem is with effectively-static fields like `display_name`, which is carried in every summary but, due to deduplication, only actually published on the first frame and never again (since it doesn't change). If that first publish happens to fail because MQTT isn't connected yet, the current code still caches the value as "published," so deduplication sees "no change" and never re-publishes — leaving subscribers without the value for the entire runtime until a restart. This change only caches a value after a successful publish and retries on the next summary after a failure, which eliminates the issue.

brianmay:
Have you tested this with an error condition and found it to work?

Hopefully lib/teslamate/mqtt/publisher.ex should have the correct logic to deal with Tortoise errors, even with qos > 0 (we use qos = 1). Not sure if this has been tested recently...

ciyahu:
Thanks for taking a look, and good question!

You're right that this isn't a recent regression — the trigger is quite narrow, which is probably why it rarely comes up.

To clarify the failure mode: a plain network outage isn't actually the problem here. While the broker is briefly unreachable the connection process stays alive, so a QoS 1 message goes into the in-flight store and is re-sent on reconnect — nothing is lost in that case, and you're right that publisher.ex + QoS 1 handle it well.

The one case where a value is genuinely lost is when Tortoise311.publish returns {:error, :unknown_connection} — i.e. the connection process isn't registered at the moment of publishing (its crash/restart window), so the message never enters the in-flight store. In that path publisher.ex's non-zero-qos branch (we use qos: 1) hard-matches {:ok, ref} = Tortoise311.publish(...), so the error ends up crashing the Publisher rather than being handled — which is what this PR guards against on the subscriber side. You're also right to be unsure whether that path is well-exercised today; from what I can tell it isn't.

This is also why it's so easy to miss: frequently-changing fields self-correct on the very next summary, so only effectively-static fields like display_name (published just once on the first frame after dedup) stay missing for the whole runtime if that single publish happens to land in the window. It tends to surface only on consumers that rebuild their state from retained topics — Home Assistant being the typical one, especially when it restarts and reloads from retained — where a car's display_name shows up as permanently empty.

On testing under an error condition: I think this is already covered by test "retries failed values until they are published successfully" in vehicle_subscriber_test.exs — using a publisher mock that returns {:error, ...} on the first publish of display_name/shift_state and :ok afterwards, it checks that those values are re-published on the next summary while already-published fields are not. The reduce also handles the task-crash case ({:exit, reason}, the path where publisher.ex crashes on unknown_connection) the same way, so neither writes to last_values.

brianmay:
I asked AI to review:

> Verdict: Ready to merge, no blocking issues.
> The PR fixes a real bug where failed MQTT publishes were cached as if successful, preventing retries until the value changed. Approach is sound: only cache successful publishes, use Map.has_key?/2 to distinguish missing-from-cache vs cached nil.
> Minor observations (non-blocking):
> 1. :unknown values no longer cached — positive behavioral change, efficiency improvement
> 2. Lost fast-path for no-change case (removed Enum.each clause) — negligible overhead at ~1 summary/sec/car
> 3. Log spam on persistent broker failures — pre-existing, not a regression
> 4. Reduce patterns robustly handle :ok, {:error, _}, and {:exit, _}
> 5. Test coverage adequate — validates retry scenarios correctly

brianmay (approved):

JakobLichterfeld:
Thank you both!

PR #5436, opened 2026-06-22
fix(webview): show charging finish time in browser local time

Ashok28:
The "Expected Finish Time" was formatted server-side with Timex.local/1, which uses the server's timezone (UTC in typical Docker deployments) rather than the user's. 

Render the UTC instant as ISO-8601 and convert it client-side with a new LocalDateTime hook, matching the existing pattern used by the Scheduled Charging row.

### DIFF
--- a/assets/js/hooks.js
+++ b/assets/js/hooks.js
@@ -43,6 +43,27 @@ export const LocalTime = {
   },
 };
 
+export const LocalDateTime = {
+  render() {
+    const dateStr = this.el.dataset.date;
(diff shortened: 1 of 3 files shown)


JakobLichterfeld (approved):
Nice finding and ty for your contribution!

JakobLichterfeld (changes requested):
Please follow: https://docs.teslamate.org/docs/development#update-pot-files-extract-messages-for-translation to update the pot files

JakobLichterfeld (approved):
TY.

PR #5438, opened 2026-06-24
fix(performance): use existing indexes for last-inserted / latest complete position lookups

### LINKED ISSUE #5306: Add (car_id, date) btree on positions to speed up bundled Grafana dashboards
## Summary

Several built-in Grafana dashboards (and presumably the Phoenix UI) issue queries against `positions` of the form:

```sql
SELECT … FROM positions WHERE car_id = ? ORDER BY date DESC LIMIT 1;
SELECT … FROM positions WHERE car_id = ? AND date >= NOW() - INTERVAL '…' …;
```

There is no btree index that covers this access pattern in the general case. The existing relevant indexes on `positions` are:

| Index | Covers |
|---|---|
| `positions_pkey (id)` | row lookups by id |
| `positions_car_id_index (car_id)` | car_id only — no help for `ORDER BY date DESC` |
| `positions_car_id_date__ideal_battery_range_km_IS_NOT_NULL_index (car_id, date) WHERE ideal_battery_range_km IS NOT NULL` | only queries that explicitly include `ideal_battery_range_km IS NOT NULL` |
| `positions_date_index` | BRIN on date alone |
| `positions_drive_id_date_index` | BRIN on (drive_id, date) |

The result: any "latest-by-time-per-car" or "recent-window-per-car" query that does **not** include `ideal_battery_range_km IS NOT NULL` falls back to a Parallel Seq Scan over the entire `positions` table.

## Reproduction

On a single-car deployment with 842,860 rows in `positions` (84 MB table), PostgreSQL 17.3:

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT usable_battery_level, date
FROM positions
WHERE car_id = 1 AND usable_battery_level IS NOT NULL
ORDER BY date DESC
LIMIT 1;
```

(This is the exact query from the bundled `battery-health.json` dashboard, panel "State of Health".)

**Before** (no `(car_id, date DESC)` index):
```
Limit  (actual time=1794..1894 rows=1)
  -> Gather Merge ... Workers Planned: 2
       -> Sort (Sort Method: top-N heapsort)
            -> Parallel Seq Scan on positions (rows=280953/loop x 3)
                 Filter: (car_id = 1)
Execution Time: 1894.561 ms
```

**After** adding `CREATE INDEX CONCURRENTLY idx_positions_car_id_date_desc ON positions (car_id, date DESC)`:
```
Limit  (actual time=0.055..0.056 rows=1)
  Buffers: shared hit=7
  -> Index Scan using idx_positions_car_id_date_desc on positions
       Index Cond: (car_id = 1)
       Filter: (usable_battery_level IS NOT NULL)
Execution Time: 0.085 ms
```

**Roughly 22,000× faster on this query.** Index size on a single-car dataset: 25 MB.

## Affected built-in dashboards / queries

Greppable matches in `grafana/dashboards/*.json` for `WHERE car_id = $car_id … ORDER BY date DESC LIMIT 1` (or range variants without the partial-index filter):

- `battery-health.json` — "State of Health" (2 queries)
- `overview.json` — outside_temp last-60m latest reading; several other latest-row lookups
- `trip.json` — `WHERE car_id = ? AND date BETWEEN …`
- `trackingdrives.json` — same shape, multiple panels
- `drive-stats.json` — speed-section aggregations
- `visited.json` — date-bucketed lat/long over windows
- `drive-details.json` — multiple `ORDER BY date ASC` over `WHERE car_id = ? AND timefilter`

Queries that already include `ideal_battery_range_km IS NOT NULL` (charge-level, projected-range, parts of overview) are correctly served by the existing partial index and would be unaffected.

The Phoenix UI's "current state" view is also a likely beneficiary — any "latest position for car X" path goes through this shape — though I have not profiled the Elixir side directly.

## Proposed change

Add a regular composite btree to the schema via a new Ecto migration:

```elixir
defmodule TeslaMate.Repo.Migrations.AddPositionsCarIdDateDescIndex do
  use Ecto.Migration

  @disable_ddl_transaction true
  @disable_migration_lock true

  def change do
    create_if_not_exists index(
      :positions,
      [:car_id, :date],
      name: :positions_car_id_date_desc_index,
      concurrently: true
    )
  end
end
```

Notes:
- `CONCURRENTLY` so existing installations can migrate without locking the `positions` table (running deployments keep ingesting).
- DESC vs ASC: PostgreSQL btrees can scan in either direction, so the column-order index serves both `ORDER BY date DESC LIMIT N` and `ORDER BY date ASC` ranges. (I used `DESC` in the manual SQL above only as a hint to the planner — the migration above using default ASC works equivalently for the planner.)
- Build time on a 842k-row / single-car dataset: ~10 s. Larger / multi-car installs will scale roughly linearly.
- Index size estimate: 25 MB on this dataset (single-car, n_distinct=1 keeps it very compact). Multi-car installs will be larger but still bounded by the row count.
- Write amplification: append-mostly workload (~1 row/sec while driving, near-zero when parked) makes the maintenance cost negligible.
- Optional follow-up: drop `positions_car_id_index` once the new composite is in place — it's redundant for any car_id lookup (the new index has the same leading column) and `pg_stat_user_indexes` showed 0 scans against it on this install. Out of scope for this PR; would want broader confirmation first.

## Related issues / prior art

- [PR #5075](https://github.com/teslamate-org/teslamate/pull/5075) (merged 2026-01-08, *perf: Replace positions indexes from BTREE to BRIN to reduce memory usage*) — converted `positions_drive_id_date_index` and `positions_date_index` from BTREE to BRIN. Kept `positions_car_id_index` (single-column on `car_id`) and the `(car_id, date) WHERE ideal_battery_range_km IS NOT NULL` partial index. **This proposal does not undo that work** — the `(car_id, date)` composite proposed here never existed; it's a new index, not a regression of [#5075](https://github.com/teslamate-org/teslamate/pull/5075).
- [Issue #5172](https://github.com/teslamate-org/teslamate/issues/5172) (closed 2026-03-14, *Severe UI Lag and High IO on Grafana Pages After Upgrade to v3.0.0*) — user reported Battery Health and Visited dashboards loading much slower after the v3.0.0 upgrade (which shipped [#5075](https://github.com/teslamate-org/teslamate/pull/5075)). Comment from @vincep5 on 2026-03-08 specifically asks whether the BRIN/BTREE swap is responsible. The issue was closed as a duplicate of [#5178](https://github.com/teslamate-org/teslamate/issues/5178), but [#5178](https://github.com/teslamate-org/teslamate/issues/5178) is an unrelated `accept-language`/BEAM CPU lock — so the actual Battery Health slowness reported in [#5172](https://github.com/teslamate-org/teslamate/issues/5172) was not addressed. The query plans above explain the symptom and the proposed index resolves it.
- [PR #5276](https://github.com/teslamate-org/teslamate/pull/5276) (merged 2026-04-11, *fix: improve brin indexes (autosummarize + timestamp_minmax_multi_ops)*) — followup tuning on the BRIN swap. Helpful for date-range scans, but doesn't help `WHERE car_id = ?` queries because BRIN on `date` alone has no notion of car_id and `positions_car_id_index` is single-column. The composite is the missing piece.
- [PR #4964](https://github.com/teslamate-org/teslamate/pull/4964) (merged 2025-09-24, *perf: Optimize Grafana query for trip view to leverage indexes more effectively*) — precedent for fixing this class of slowness at the query level. The proposal here addresses the index side instead, which avoids touching every dashboard.
- [PR #3801](https://github.com/teslamate-org/teslamate/pull/3801) (closed 2024-04-05, *perf: Positions table index on Odometer and Car_ID …*) — earlier index proposal. Different columns (odometer + car_id), no EXPLAIN evidence; closed quickly. Mentioned only for completeness.

## Environment

- TeslaMate 1.32.x (current release line)
- PostgreSQL 17.3 (CloudNativePG-managed cluster on k3s)
- 842k positions, single-car install, ~3.5 years of history
- Discovered while profiling a downstream consumer that runs the same access pattern; rolled out the index locally and saw broad latency drops across the bundled Grafana dashboards as a side benefit.

swiffer:
## Summary

Avoid sequential scans on large `positions` tables after the BRIN migration by using existing indexes — **no new btree**.

### Elixir
- Rename `get_latest_position/0` → `get_last_inserted_position/0`: `ORDER BY id DESC` (PK scan) for geofence map center only
- `get_latest_position(%Car{})`: require `ideal_battery_range_km IS NOT NULL` so restore uses the partial `(car_id, date)` index and skips incomplete streaming rows

### Grafana
- Battery Health / Overview latest-value panels: same complete-row filter (and safer UNION ordering where needed)
- Built on top of Grafana 13.1.1 (#5559)

### i18n
- Refresh POT source refs after unused `Logger` cleanup in charge cost LiveView

Closes #5306

---

🤖 Assisted by Grok 4.5 (xAI) via Grok Build (planning, code edits, PR description).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@
 - feat(grafana): make data health summary actionable (#5526 - @magrathean-uk)
 - fix(grafana): drop the Releases panel from the home dashboard to end the CORS proxy dependency (#5548 - @JakobLichterfeld)
 - fix(dashboards): filter latest-value position panels on complete rows so they use the partial index (#5438 - @swiffer)
+- fix(grafana): Battery Health latest SOC/kWh panels pick the newest UNION row and use `usable_battery_level` on charges (#5438 - @swiffer)
 - fix(grafana): use local calendar for Statistics period end boundaries (#5562 - @wjsall)
 
 #### Translations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - fix(vehicle): identify base Model 3 from model year 2022 as RWD instead of SR+ (#5551 - @magrathean-uk)
 - fix(mqtt): avoid blocking startup on retained cleanup (#5549 - @magrathean-uk)
 - feat: use Grafana 13.1.1 (#5559 - @swiffer)
+- fix(performance): use existing indexes for last-inserted / latest complete position lookups (#5438 - @swiffer)
 
 #### Build, CI, internal
 
@@ -80,6 +81,7 @@
 
 - feat(grafana): make data health summary actionable (#5526 - @magrathean-uk)
 - fix(grafana): drop the Releases panel from the home dashboard to end the CORS proxy dependency (#5548 - @JakobLichterfeld)
+- fix(dashboards): filter latest-value position panels on complete rows so they use the partial index (#5438 - @swiffer)
 - fix(grafana): use local calendar for Statistics period end boundaries (#5562 - @wjsall)
 
 #### Translations
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -1093,7 +1093,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT * FROM ((SELECT usable_battery_level, date\r\nFROM positions\r\nWHERE car_id = $car_id AND usable_battery_level IS NOT NULL\r\nORDER BY date DESC\r\nLIMIT 1)\r\nUNION\r\n(SELECT usable_battery_level, date\r\nFROM charges c\r\nJOIN charging_processes p ON p.id = c.charging_process_id\r\nWHERE p.car_id = $car_id AND usable_battery_level IS NOT NULL\r\nORDER BY date DESC\r\nLIMIT 1)) AS last_usable_battery_level LIMIT 1",
+          "rawSql": "SELECT * FROM ((SELECT usable_battery_level, date\r\nFROM positions\r\nWHERE car_id = $car_id AND ideal_battery_range_km IS NOT NULL AND usable_battery_level IS NOT NULL\r\nORDER BY date DESC\r\nLIMIT 1)\r\nUNION ALL\r\n(SELECT c.usable_battery_level, date\r\nFROM charges c\r\nJOIN charging_processes p ON p.id = c.charging_process_id\r\nWHERE p.car_id = $car_id AND c.usable_battery_level IS NOT NULL\r\nORDER BY date DESC\r\nLIMIT 1)) AS last_usable_battery_level ORDER BY date DESC LIMIT 1",
           "refId": "SOC",
           "sql": {
             "columns": [
@@ -1370,7 +1370,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT * FROM ((SELECT usable_battery_level * ('$aux'::json ->> 'CurrentCapacity')::float / 100 as kWh, date, ('$aux'::json ->> 'CurrentCapacity')::float as Total\nFROM positions\nWHERE car_id = $car_id  AND usable_battery_level IS NOT NULL\nORDER BY date DESC\nLIMIT 1)\nUNION\n(SELECT battery_level * ('$aux'::json ->> 'CurrentCapacity')::float / 100 as kWh, date, ('$aux'::json ->> 'CurrentCapacity')::float as Total\nFROM charges c\nJOIN charging_processes p ON p.id = c.charging_process_id\nWHERE p.car_id = $car_id  AND usable_battery_level IS NOT NULL\nORDER BY date DESC\nLIMIT 1)) AS last_usable_battery_level LIMIT 1",
+          "rawSql": "SELECT * FROM ((SELECT usable_battery_level * ('$aux'::json ->> 'CurrentCapacity')::float / 100 as kWh, date, ('$aux'::json ->> 'CurrentCapacity')::float as Total\nFROM positions\nWHERE car_id = $car_id AND ideal_battery_range_km IS NOT NULL AND usable_battery_level IS NOT NULL\nORDER BY date DESC\nLIMIT 1)\nUNION ALL\n(SELECT c.usable_battery_level * ('$aux'::json ->> 'CurrentCapacity')::float / 100 as kWh, date, ('$aux'::json ->> 'CurrentCapacity')::float as Total\nFROM charges c\nJOIN charging_processes p ON p.id = c.charging_process_id\nWHERE p.car_id = $car_id  AND c.usable_battery_level IS NOT NULL\nORDER BY date DESC\nLIMIT 1)) AS last_usable_battery_level ORDER BY date DESC LIMIT 1",
           "refId": "A",
           "sql": {
             "columns": [
--- a/grafana/dashboards/overview.json
+++ b/grafana/dashboards/overview.json
@@ -1601,7 +1601,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT\r\n\t$__time(date),\r\n\tconvert_celsius(driver_temp_setting, '$temp_unit') as \"Driver Temperature [°$temp_unit]\",\r\n\tconvert_celsius(inside_temp, '$temp_unit') AS \"Inside Temperature [°$temp_unit]\"\r\nFROM positions\r\nWHERE driver_temp_setting IS NOT NULL AND inside_temp IS NOT NULL AND car_id = $car_id AND date >= (TIMEZONE('UTC', NOW()) - INTERVAL '60m')\r\nORDER BY date DESC\r\nLIMIT 1;",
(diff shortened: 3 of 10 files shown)


JakobLichterfeld:
> I recognized that Positions with ideal_battery_range_km have no elevation - is that by design?

yes, that is by design. Positions with `ideal_battery_range_km` come from `vehicle_data` polling (REST), and that endpoint does not provide altitude at all. `elevation` only ever comes from the streamed positions (websocket), which in turn carry only the small streaming field set — so the two are mutually exclusive per row. The SRTM backfill doesn't fill the polled rows either, as it only processes drives without any streamed positions.

Full write-up in https://github.com/teslamate-org/teslamate/discussions/5532#discussioncomment-17716008

JakobLichterfeld:
> I put the small POT refresh in #5554. It fixes the current `mix gettext.extract --check-up-to-date` failure and only updates the two generated source references. It targets this branch directly, so it can be merged without pulling in anything else.

Ty, as said, this is not helpful and was closed.

I will take over this PR in the comming days

swiffer:
> The SRTM backfill doesn't fill the polled rows either, as it only processes drives without any streamed positions.

ah, i see - i have none non-streamed drives - having elevation filled for all positions logged would may allow some elevation queries / long-term stats to use the positions with ideal_battery_range_km only (for streamed drives last objective carried forward could be a viable alternative to the SRTM path). but this is out of scope here.

will continue working on this PR (on top of #5559) now

JakobLichterfeld (approved):
Thanks @swiffer — really glad you picked this one up again and are carrying it over the finish line yourself! 🙏

Did a full review pass, the approach is exactly right: no new btree, just making the existing partial index `(car_id, date) WHERE ideal_battery_range_km IS NOT NULL` and the PK do the work. Once it's rebased on top of #5559 as you planned, the diff should shrink to a nicely focused core. A few notes on the substance — all good:

- `get_last_inserted_position/0` with `ORDER BY id DESC` is semantically fine for the geofence map center, and the `ideal_battery_range_km IS NOT NULL` filter in `get_latest_position/1` is arguably *more* correct for `restore_last_known_values`: streaming rows don't carry the range/battery values the restore needs anyway. The only trade-off is that the restored lat/lon may be slightly older than the newest streamed row — negligible in practice.
- The Battery Health query change is a real bugfix beyond performance: the old outer `LIMIT 1` without `ORDER BY` on a `UNION` picked an arbitrary branch, and the kWh panel's charges branch used `battery_level` instead of `usable_battery_level`. Nice catch — maybe worth reflecting in the changelog entry so it's visible as a fix, not just perf.
- The extra filter on the Overview inside-temp panel is safe since `inside_temp` only comes from polled `vehicle_data` rows anyway.

Nothing blocking. Happy to merge once it's green after your rebase as #5559 already landed. 🚀


JakobLichterfeld:
One more thought: #5483 explored the same approach and was closed in favor of this PR, but it included regression tests for the log queries (`get_last_inserted_position/0` / `get_latest_position/1`) and a dashboard-query detector that flags latest-position blocks missing the partial-index predicate. Porting that coverage over here would lock in the fix and prevent future dashboard edits from silently reintroducing seq scans — feel free to cherry-pick from there.

JakobLichterfeld:
sorry my bad I thought It would be a good idea to just overtake it :-(

JakobLichterfeld:
To untangle: the branch is now rebased onto current main (your three commits, authorship preserved — head is 06ecaed). If your local branch has new work on top, just force-push. Won't touch the branch again, promise 😄

swiffer:
Thanks @JakobLichterfeld — rebased onto current `main` (including #5559), addressed your review notes, and this should be ready for a final look / merge when CI is green.

### Work completed since the last pass
- Rebased onto `main` after #5559 landed; branch is up to date on `swiffer/optimize-index-usage` (latest: `393a354b`)
- Changelog: called out the Battery Health SOC/kWh changes as a **correctness fix**, not only performance
- Tests: added `test/teslamate/log/log_position_test.exs` covering `get_latest_position/1` and `get_last_inserted_position/0`

### Responses to your review notes
- **`get_last_inserted_position/0` (`ORDER BY id DESC`)** — agreed; kept as-is for geofence map center / PK scan. Semantics stay “last inserted”, not “latest by date”.
- **`get_latest_position/1` (`ideal_battery_range_km IS NOT NULL`)** — agreed; complete polled rows only so restore uses the partial `(car_id, date)` index. The slightly older lat/lon vs newest streaming row trade-off is accepted as negligible (and streaming rows lack the range/battery fields restore needs anyway).
- **Battery Health as a real bugfix** — reflected in the changelog under Dashboards: outer `ORDER BY date DESC` on the UNION so we pick the newest source row, and the charges branch uses `usable_battery_level` (not `battery_level`) for the kWh panel.
- **Overview inside-temp filter** — agreed; left as-is (`inside_temp` only comes from polled `vehicle_data` rows).
- **#5483 regression coverage** — ported the **Log** query tests only (`get_latest_position/1` / `get_last_inserted_position/0`). Skipped the Grafana dashboard-query detector: static SQL linting of dashboard JSON does not belong in the Elixir/ExUnit suite; the dashboard SQL changes themselves remain in this PR. Happy to revisit as a separate lightweight CI check if useful later.
- **Elevation / SRTM on polled rows** — still out of scope here (as discussed).

### Ask
Please take another look when you can and merge if you’re happy — nothing blocking from my side.

---

🤖 Assisted by Grok 4.5 (xAI) via Grok Build (planning, changelog, tests, this comment).

JakobLichterfeld (approved):
Took another look after the latest commits — this is exactly where I hoped it would land. Thank you so much!

- Diff is now nicely focused after the rebase, all the Grafana 13.1.1 churn is gone
- Thanks for the separate changelog entry for the Battery Health fix (deterministic `UNION ALL` ordering + `usable_battery_level` on charges) — good that it's visible as a bugfix, not just perf
- The tests pin down both behavior changes precisely: `get_latest_position/1` skipping incomplete streaming rows (incl. the `nil` case and per-car isolation), and the `get_last_inserted_position/0` test asserting that the last *inserted* row wins over a later-dated one — exactly the `ORDER BY id` semantics

CI is green, approving and merging. Thanks for seeing this through, @swiffer! 🚀

One optional follow-up (no need to hold this PR for it): #5483 also had a dashboard-query detector that flags latest-position panels missing the partial-index predicate — porting that would keep future dashboard edits from silently reintroducing seq scans. @magrathean-uk, feel free to bring that in separately if you like.

PR #5440, opened 2026-06-25
ci: speed up check_linting by running treefmt in a lean app

JakobLichterfeld:
The check_linting job took ~17m because `nix develop` realised the full devenv shell (elixir-ls, postgres, nodejs, mosquitto, osv-scanner, …).
That closure is several GB and exceeds the cache GC budget, so it was garbage-collected before saving and rebuilt on every run.

Add a lean `.#lint` flake app (treefmt wrapper + the release Elixir) that runs `mix deps.get` and then `exec treefmt "$@"`, so CI calls `nix run .#lint -- --ci` and `--ci` is just a pass-through arg. Also bump gc-max-store-size to 2G so the small closure is actually cached.

Expose `teslamate.rebar3` from package.nix as the single source of truth and use it in both the devenv and lint setups, replacing the hardcoded `beam27Packages.rebar3` (erlang 27) that did not match the project's erlang_28 toolchain.

### DIFF
--- a/.github/workflows/ensure_linting.yml
+++ b/.github/workflows/ensure_linting.yml
@@ -44,5 +44,5 @@ jobs:
         # (and fail to cache) the multi-GB development closure on every run.
         # It runs `mix deps.get` (so mix-format finds its plugin deps) and then
         # treefmt, passing through `--ci`.
-        run: nix run --override-input devenv-root "file+file://"<(printf %s "$PWD") .#lint -- --ci
+        run: nix run .#lint -- --ci
         # or use: https://github.com/cachix/git-hooks.nix?tab=readme-ov-file
--- a/.github/workflows/ensure_linting.yml
+++ b/.github/workflows/ensure_linting.yml
@@ -24,8 +24,10 @@ jobs:
           restore-prefixes-first-match: nix-${{ runner.os }}-
           # collect garbage until Nix store size (in bytes) is at most this number
           # before trying to save a new cache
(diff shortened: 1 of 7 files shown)


.github/workflows/ensure_linting.yml
```diff
@@ -37,5 +39,10 @@ jobs:
           purge-primary-key: never
 
       - name: Run treefmt in CI mode
-        run: nix develop --override-input devenv-root "file+file://"<(printf %s "$PWD") . --command treefmt --ci
+        # Use the lean `.#lint` app instead of the full devenv shell: it only
+        # provides the treefmt wrapper + Elixir, so CI does not have to realize
+        # (and fail to cache) the multi-GB development closure on every run.
+        # It runs `mix deps.get` (so mix-format finds its plugin deps) and then
+        # treefmt, passing through `--ci`.
+        run: nix run --override-input devenv-root "file+file://"<(printf %s "$PWD") .#lint -- --ci
```
brianmay: Hopefully that `--override-input` parameter is no longer required. Unless I missed something here?

JakobLichterfeld: Nice catch and fixed now.

brianmay (approved):

JakobLichterfeld:
TY! I rebased to resolve changelo conflicts.

JakobLichterfeld:
Yiha, down from ~17min to 3 min!

PR #5452, opened 2026-06-28
feat: add redacted support diagnostics bundle

magrathean-uk:
## Summary

- Add a Settings support bundle download endpoint that returns redacted diagnostics JSON.
- Include app/runtime/database summary data, open record counts, and car settings by internal car id only.
- Harden a few async test waits that were timing-sensitive during local full-suite validation.

## Privacy

The bundle is allowlist-based and intentionally omits Tesla tokens, passwords, MQTT credentials, raw environment variables, VINs, car display names, exact coordinates, full configured URLs, file paths, and raw logs.

## Test Plan

- `mix ci`
- `mix test test/teslamate/support_diagnostics_test.exs test/teslamate_web/controllers/support_controller_test.exs test/teslamate_web/live/settings_test.exs`
- Real-world restored database smoke test for `GET /settings/support-bundle.json`
- Privacy scan of the generated JSON against car names, VINs, URLs, token bytes, and coordinates


JakobLichterfeld:
Ty for your contribution. May I ask you to elaborate on the use-case a bit more?

> * Harden a few async test waits that were timing-sensitive during local full-suite validation.

Thats a nice one. Ty!

To solve the CI issue: please follow https://docs.teslamate.org/docs/development#update-pot-files-extract-messages-for-translation

JakobLichterfeld:
Thanks for explaining the use case.

From our experience as maintainers, this doesn't really match how support usually works. We rarely ask users for database state or runtime metadata, and I can't recall cases where users accidentally exposed secrets when providing the information we requested (we always state something like “please provide your `docker-compose.yml` and remove any secrets before posting”).

Most support issues can be resolved using the existing logs and the relevant parts of the configuration, so I'm not yet convinced that introducing and maintaining a dedicated diagnostics endpoint provides enough value.

Do you have examples of issues where this would have made a noticeable difference?

I'm also wondering about the implementation choice. Why does this need to be an unauthenticated HTTP endpoint instead of an explicit command that the user runs when requested by a maintainer? If this is primarily intended for support, a command, or one-off report generation seems like it would achieve the same goal without adding another always-available endpoint.

magrathean-uk:
Fair point. I do not want to add a new support surface if it does not really fit how TeslaMate maintainers handle support.

I took the endpoint concern seriously and reworked the PR in that direction in `74c5eb7b`. The web part is gone now: no Settings button, no controller, no route, no always-available HTTP endpoint. What is left is just explicit report generation from the same redacted/allowlisted payload.

So the shape is now much narrower:

- maintainer asks for it only when useful
- user explicitly generates the JSON
- no logs
- no env/config dump
- no credentials
- no VINs, car names, URLs, coordinates, or raw paths
- no repair action or database writes

On examples: I do not want to overclaim that this would solve the normal support flow. I was thinking about cases around open drive/charging rows, migration/version confusion, runtime mismatch, or “something is stuck” reports where a tiny state summary can save a few rounds of back-and-forth. But if those are not common enough from the maintainer side, that is a fair reason not to carry it.

The revised PR is basically an opt-in diagnostic report helper now, not a UI feature. If that still feels like too much maintenance surface, I am fine closing it cleanly. If the explicit-command shape is closer to what you had in mind, I can keep it there and adjust naming/command wording however you prefer.

Local checks after the rework:

- `MIX_ENV=test mix ci` → 342 passed
- `mix gettext.extract --check-up-to-date` → passed
- `git diff --check` → passed

JakobLichterfeld:
I appreciate that you reworked the implementation based on the feedback. Removing the HTTP endpoint definitely addresses one of my concerns.

That said, my main concern wasn't the implementation alone but whether this is solving a problem we actually have.

Looking at the issues we handle, we usually end up asking for logs rather than runtime or database metadata. In fact, we've already discussed making log collection easier (see #4891), because that's where most of the friction is today.

So I do think there is room for improving the support experience, but I'm not convinced that a diagnostic state report is the area that provides the biggest benefit atm.

Ironically, I actually found the original proposal more compelling from a feature perspective. The current version feels much more limited, yet it still introduces something we would have to maintain going forward.

I really appreciate your willingness to adapt the implementation based on feedback. At the same time, I'd prefer that we first establish the need for a feature and then discuss the best implementation, rather than iterating on the implementation before we've agreed that the feature itself solves a recurring problem.

As a first quick win, would you mind extracting the test improvements into a separate PR? I think they stand well on their own.

PR #5455, opened 2026-06-30
docs: add PWA installation instructions to README and create dedicated PWA documentation

JakobLichterfeld:


brianmay:
I have mixed feelings here...

One one hand any new documentation is probably a good thing.

On the other hand, if it encourages people to expose the insecure port to the Internet (so that they can access it on their mobile phones when away from home) that might be a bad thing.

JakobLichterfeld:
> On the other hand, if it encourages people to expose the insecure port to the Internet (so that they can access it on their mobile phones when away from home) that might be a bad thing.

That's a good point. Do you have any suggestions for improvement? Should we add the note from the installation guide again?

brianmay:
I think this is going to be challenging. 

Maybe we need another page under Installation "Securing the web port" or something like that that talks about the dangers of exposing the port to the Internet, and various ways of securing it (from local only blocking through to various authentication proxys).

The other installation pages could in turn link to this page with a big warning about insecure access.

Issue #3736, opened 2024-03-12
lots of test failures


brianmay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Noticed lots of test failures when running tests locally. The found I have to start tests with clean database. Oops.

But even then, was getting errors - up to 7. Then I found if I ran test files individually, was more like to work.

At a hunch I made the following change:

```diff
diff --git a/test/teslamate/vehicles/vehicle/charging_test.exs b/test/teslamate/vehicles/vehicle/charging_test.exs
index 106d64df..26fb4f65 100644
--- a/test/teslamate/vehicles/vehicle/charging_test.exs
+++ b/test/teslamate/vehicles/vehicle/charging_test.exs
@@ -1,5 +1,5 @@
 defmodule TeslaMate.Vehicles.Vehicle.ChargingTest do
-  use TeslaMate.VehicleCase, async: true
+  use TeslaMate.VehicleCase, async: false

   alias TeslaMate.Vehicles.Vehicle.Summary
   alias TeslaMate.Log.ChargingProcess
diff --git a/test/teslamate/vehicles/vehicle/streaming_test.exs b/test/teslamate/vehicles/vehicle/streaming_test.exs
index 20f54723..c43db381 100644
--- a/test/teslamate/vehicles/vehicle/streaming_test.exs
+++ b/test/teslamate/vehicles/vehicle/streaming_test.exs
@@ -1,5 +1,5 @@
 defmodule TeslaMate.Vehicles.Vehicle.StreamingTest do
-  use TeslaMate.VehicleCase, async: true
+  use TeslaMate.VehicleCase, async: false

   import ExUnit.CaptureLog

diff --git a/test/teslamate/vehicles/vehicle/suspend_logging_test.exs b/test/teslamate/vehicles/vehicle/suspend_logging_test.exs
index ae3d6677..f282debe 100644
--- a/test/teslamate/vehicles/vehicle/suspend_logging_test.exs
+++ b/test/teslamate/vehicles/vehicle/suspend_logging_test.exs
@@ -1,5 +1,5 @@
 defmodule TeslaMate.Vehicles.Vehicle.SuspendLoggingTest do
-  use TeslaMate.VehicleCase, async: true
+  use TeslaMate.VehicleCase, async: false

   alias TeslaMate.Vehicles.Vehicle.Summary
   alias TeslaMate.Vehicles.Vehicle
[...]

PR #5456, opened 2026-06-30
test: harden async waits

magrathean-uk:
## Summary

- replace a fixed sleep with the existing `TestHelper.eventually/1` helper
- avoid timing-sensitive receive assertions in a few async tests
- keep mailbox silence checks bounded where needed

## Context

Extracted from #5452 after maintainer feedback that the test hardening stands well on its own.

## Checks

- `mix test test/teslamate/terrain_test.exs test/teslamate/updater_test.exs test/teslamate/vehicles/vehicle_test.exs test/teslamate_web/live/car_summary_live_test.exs` -> 50 passed
- `mix test --warnings-as-errors` -> 339 passed
- `mix ci` -> 339 passed

### DIFF
--- a/test/teslamate/terrain_test.exs
+++ b/test/teslamate/terrain_test.exs
@@ -100,14 +100,9 @@ defmodule TeslaMate.TerrainTest do
       assert Terrain.get_elevation(name, {0, 0}) == nil
 
       # circuit broke after 3 attempts
-      TestHelper.eventually(
-        fn ->
-          assert_received {SRTM, {:get_elevation, 0, 0, _opts}}
-          assert_received {SRTM, {:get_elevation, 0, 0, _opts}}
-          assert_received {SRTM, {:get_elevation, 0, 0, _opts}}
-        end,
-        attempts: 25
-      )
+      assert_receive {SRTM, {:get_elevation, 0, 0, _opts}}, 1_000
+      assert_receive {SRTM, {:get_elevation, 0, 0, _opts}}, 1_000
+      assert_receive {SRTM, {:get_elevation, 0, 0, _opts}}, 1_000
 
       refute_receive _
     end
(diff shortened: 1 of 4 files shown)


JakobLichterfeld:
Ty! #3736 could be resolved by this, @brianmay, innit?

brianmay:
More AI feedback :-)

Looks good, although some minor concerns also.


## Review Summary for PR #5456 ("test: harden async waits")

**No bugs found.** The changes correctly harden async test waits. Observations:

### terrain_test.exs
Switching from `eventually`+`assert_received` to three sequential `assert_receive ..., 1_000` is sound — the 3 SRTM calls land in mailbox order before the circuit opens. Note: the new form is strictly sequential (order-sensitive if messages were ever differentiated).

### updater_test.exs
**Inconsistency**: only the `"4.99.0"` case was converted to `eventually`; the first two cases (`"5.1.3-dev"`, `"5.1.2"`) still use fixed `Process.sleep(100) + assert` with the original flakiness. Consider converting them too.

### vehicle_test.exs
Dropping explicit `100` timeout for the default `assert_receive_timeout` (300ms) is strictly more lenient — fine. `refute_receive _, 100` at line 554 strengthens the assertion by waiting 100ms — good change.

### car_summary_live_test.exs
The `eventually` block works correctly (AssertionError is caught and retried). Caveat: with `attempts: 10`, `delay: 5`, the retry window is only ~45-50ms — potentially still flaky if LiveView re-render after `resume_logging` is slower.

---

The main actionable item is the **updater_test.exs asymmetry** — the PR's stated intent (harden async waits) isn't fully applied there. Want me to convert the remaining two cases?

brianmay (approved):

JakobLichterfeld (approved):
TY!

PR #5458, opened 2026-07-01
refactor: send TOKEN as Bearer header via FleetAuth middleware

jlestel:
Replace TOKEN query-param injection (?token=...) with a dedicated FleetAuth middleware that sends the token as an Authorization: Bearer header on all API and auth requests, including the streaming subscribe message. 

Legacy ?token= format remains supported for backward compatibility. 

Update docs accordingly.


jlestel:
Is it possible to get an image build result to try out for this branch, please? @JakobLichterfeld

JakobLichterfeld:
> Is it possible to get an image build result to try out for this branch, please? @JakobLichterfeld

Yes --> #5470

Thank you!

PR #5472, opened 2026-07-02
docs(mqtt): add service mode topic to MQTT integration documentation

JakobLichterfeld:
as introduced by #5289


JakobLichterfeld:
@NirKli you introduced it via ##5472, innit?

NirKli:
> @NirKli you introduced it via ##5472, innit?

 Yep, that was me in #5289 — added the service_mode field to TeslaApi.Vehicle.State plus the log-noise reduction and the 🔧 UI indicator.
Thanks for documenting the MQTT topic! 🙌

PR #5470, opened 2026-07-01
refactor: send TOKEN as Bearer header via FleetAuth middleware

JakobLichterfeld:
Replace TOKEN query-param injection (?token=...) with a dedicated FleetAuth middleware that sends the token as an Authorization: Bearer header on all API and auth requests, including the streaming subscribe message.

Legacy ?token= format remains supported for backward compatibility.

Update docs accordingly.

This is successcor of #5458 from @jlestel as repo branch, so test images can be generated

PR #5481, opened 2026-07-05
refactor: send TOKEN as Bearer header via FleetAuth middleware (rebased)

jlestel:
Rebased version of #5470 onto the current upstream `main` (was 16 commits behind).

Content is identical — a single commit adding `TeslaApi.Middleware.FleetAuth` that sends the token as an `Authorization: Bearer` header (legacy `?token=` still supported). The only rebase conflict was in `lib/tesla_api.ex`, resolved by keeping the improved upstream `Tesla.Middleware.Logger` config and adding the `FleetAuth` plug on top.

Targeting the `jlestel/custom_token_as_bearer` branch so this can be merged into PR #5470 without maintainers having to force-push.


JakobLichterfeld (approved):
ty for rebasing, but the conflict still exists :-)

JakobLichterfeld (changes requested):
ty for rebasing, but the conflict still exists :-)

JakobLichterfeld:
I rebased #5470

PR #5487, opened 2026-07-07
feat: show and color the software update icon through the update lifecycle

NirKli:
# Show and color the software update icon through the whole update lifecycle

## Problem (current behavior)

When a software update finishes downloading and is waiting to be installed, the update (gift) icon disappears from the car card. 
There is no indication that an update has been downloaded and is ready/scheduled to install.

currect_state

The root cause is that `update_available` only considered a few statuses:

```elixir
status when status in ["available", "downloading", "downloading_wifi_wait"] -> true
```

Once Tesla reports the download as complete, the status becomes `scheduled` (and later `installing`), which were not in that list, so the icon was not displayed.
On top of that, the summary didn't expose any progress information, so the tooltip was always just "Software Update available (version)" and the icon had no visual state.

The Tesla API already reports the full software update state — a `status` together with `download_perc` / `install_perc` progress. The `status` was already read, but only some of its values were handled, and the progress percentages weren't surfaced at all:

tesla_api

Note that `install_perc` is already `10` here even though the update is only *scheduled* (downloaded and waiting), not actively installing — so the label has to be driven by `status`, not by `install_perc`, otherwise it would wrongly read "Installing 10%".

## What changed

- **Keep the icon visible for the full update lifecycle.** `update_available`  now also includes `scheduled` and `installing`, so the icon stays on the card from "available" all the way through "installing".
- **Surface the progress/status.** The vehicle summary now carries `update_status`, `download_perc` and `install_perc` (none of which existed before), so the UI can show progress.
- **Add status‑driven tooltips** (previously the tooltip was always just "Software Update available"):
  - `downloading` → `Downloading X%`
  - `installing` → `Installing X%`
  - `scheduled` / `available` → `Software Update available (version)`
- **Color the icon to reflect the state, matching the car's own UI:**
  - Downloading (in progress) → default color
  - Download finished / ready to install → **orange** (`#e8a33d`)
  - Installing → **green** (`#4caf50`)

### Downloading

Default color, percentage shown in the tooltip.

downloading

### Finished downloading (ready / scheduled to install)

The icon turns **orange** and the tooltip reads "Software Update available".
The orange is intentional — it matches the Tesla car's own UI/UX, which also shows the update indicator in orange once the download is complete and the update is ready to install.

finished_downloading

### Installing

The icon turns **green** while the update is being installed.

installing

## Tests

Added LiveView and summary tests covering the `downloading`, `installing` and `scheduled` states (tooltip text and icon color), plus the `update_status` / `download_perc` / `install_perc` mapping in the summary.


JakobLichterfeld:
Love this addition!

May I ask you to add the new MQTT values to [website/docs/integrations/mqtt.md](https://github.com/teslamate-org/teslamate/blob/da09dc3428743bfd11cbe2b922b0b286cfa0962b/website/docs/integrations/mqtt.md)?

NirKli:
Updated the MQTT integration docs to document the two new topics published by this PR: `teslamate/cars/$car_id/download_perc` and `teslamate/cars/$car_id/install_perc` (see `website/docs/integrations/mqtt.md`).

JakobLichterfeld (approved):
Perfect, thank you!

sdwalker:
Need to linkify the gift icon to the notateslaapp release notes for the new version :)

JakobLichterfeld:
> Need to linkify the gift icon to the notateslaapp release notes for the new version :)

PR welcome :-)

PR #5490, opened 2026-07-08
feat: link the software update icon to the notateslaapp release notes

NirKli:
## What

Makes the software-update (gift) icon on the car card a link to the **notateslaapp release notes** for the update version, e.g.:

`https://www.notateslaapp.com/software-updates/version/2026.20.6.1/release-notes`

This mirrors the existing behavior of the **Version** row (which already links the current version to notateslaapp), now applied to the *available update's* version.

## Why

Follow-up to #5487, as requested there:

> Need to linkify the gift icon to the notateslaapp release notes for the new version :) @sdwalker 

## Details

- The icon is wrapped in a link built from `@summary.update_version` (already the clean version string with the build hash stripped), using the same `target="_blank"` / `rel="noopener noreferrer"` pattern as the Version row.
- Falls back to a plain `<span>` if there is ever no version, to avoid emitting a broken URL.
- Tooltip text and the lifecycle icon coloring (orange when ready to install, green while installing) are unchanged.

Icon is clickable.

https://github.com/user-attachments/assets/0dfc5325-5313-47fb-a4ba-61e8a0e85440

Link to notateslaapp
image

## Tests

Updated the LiveView summary tests to assert the icon now renders as an `<a>` with the correct `href`, tooltip, `target`, and `rel`, across the downloading / installing / scheduled states.

### DIFF
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -153,34 +153,43 @@
           </span>
         <% end %>
         <%= if @summary.update_available do %>
-          <span
-            class="icon has-text-inherit has-tooltip-top has-tooltip-left-mobile"
-            data-tooltip={
-              cond do
-                @summary.update_status == "installing" and (@summary.install_perc || 0) > 0 ->
-                  gettext("Installing %{percent}%", percent: @summary.install_perc)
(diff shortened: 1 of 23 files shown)


JakobLichterfeld (approved):
Nice, ty!

PR #5495, opened 2026-07-09
feat: add fullscreen mode to vehicle summary map

hakong:
## Summary

- add a fullscreen toggle to the vehicle summary map
- keep the compact map passive by default and enable map interactions while fullscreen
- add light and dark themed button styling and update gettext catalogs

## Why

I wanted a way of displaying the map in fullscreen! Simple as that.

## Testing

- `npx --yes prettier --write assets/css/app.scss assets/css/themes/dark.scss assets/js/hooks.js`
- `mix gettext.extract --merge`
- `mix format --check-formatted lib/teslamate_web/live/car_live/summary.html.heex`
- `mix ci` (360 tests, 0 failures)
- tested locally with an isolated Docker/Podman stack on port 4400

## Note
I had Codex do this. I did review the diff and tested it locally.


hakong:
How it looks dark and light mode:
image

image

After maximizing, the "minimize map" button looks like this:
image

JakobLichterfeld:
Thanks for your idea and contribution. May I ask you/or your agent to elaborate about the usecase a bit more? In style: From a user perspective …

Thank you.

hakong:
Yes, sure!
Sometimes I want to display where the car is on a bigger screen and/or see the map bigger, but zooming in on the overview page scales all elements up, so that doesn't really help.
I did look at the Grafana dashboards but couldn't find anything that displays the cars location live.

I just sometimes want to see a "bigger picture" of where the car is and keep it live on a big display/TV/monitor.

Zooming out on the small map window helps but obviously makes the entire map smaller. So instead of zooming out I want to keep the zoom level but "see more map around the car" if that makes sense?

An example would be if the car is driving on a long, straight road. The default map will only show a small area around it so I would have no idea where on that long straight road the car is. It might be anywhere on a 10 or 20km long stretch of road. Maximizing the map gives immediate context.

Anyway I just thought this would be a small but nice quality of life feature.

hakong:
One thing I'm thinking of changing is: re-center the map on minimizing it. If I drag it somewhere far away from the car while maximized, then minimize, it stays on that spot, possibly with the car off-map.

hakong:
Map now reverts to default view after leaving fullscreen (centers and resets zoom level).

assets/css/themes/dark.scss
```diff
@@ -52,4 +52,14 @@
   .leaflet-container {
     background-color: var(--bulma-card-background-color);
   }
+
+  .map-fullscreen-button {
+    --map-fullscreen-button-border-color: rgba(204, 204, 220, 0.14);
```
JakobLichterfeld, line 57: reafactoring tle magic colors and duplication would be another pr

JakobLichterfeld (changes requested):
Thanks for elaborating on the use case. I do see some users benefiting from this.

Please add a test for the button comparable ot the others.

assets/css/themes/dark.scss
hakong: I think I understand your comment more as a "this could be refactored later"-note to self, rather than you wanting me to change anything. Is that correct?

JakobLichterfeld: correct :-)

JakobLichterfeld (changes requested):
Thanks for adding the test.

Please update the Pot files, follow: https://docs.teslamate.org/docs/development#update-pot-files-extract-messages-for-translation

JakobLichterfeld (approved):
Perfect, ty!

PR #5499, opened 2026-07-11
fix(grafana): time-weighted Ø Speed in Drive Details (avoids standstill sampling bias)

### LINKED ISSUE #5497: Drive Details 'Ø Speed' overstates average speed: point average is biased by standstill downsampling
### Summary

The **Drive Details** internal dashboard's **Ø Speed** stat (`grafana/dashboards/internal/drive-details.json`) computes average speed as a plain point average:

```sql
SELECT convert_km(avg(speed)::numeric, '$length_unit') AS speed_${length_unit}h
FROM positions
WHERE car_id = $car_id AND $__timeFilter(date)
```

Because streaming inserts positions much more densely while moving than while stopped, this systematically **overstates average speed**, especially in stop-and-go city driving.

### Evidence (measured on real data)

A user provided a second-level export of one 24-minute city drive (7.51 km, panel showed **27 km/h**):

- Sampling density: ~**0.29 s/point while moving** vs ~**1.6 s/point at standstill** (≈5.5× difference)
- Standstill was **30.6 % of wall time** but only **7.5 % of samples**
- Plain point average of all samples = **26.84 → displayed 27 km/h**
- True average (distance ÷ duration) = **18.8 km/h** → the panel overstated by **+43 %**

Reproduced independently on a second installation (different drive): point average **33.7** vs distance/duration **29.7** (+13 %). The bias grows with the share of stopped time, and numerically the point average happens to approach "distance ÷ moving time", which confuses users into thinking stopped time is intentionally excluded.

### Suggested fix

Time-weighted average (equivalent to distance ÷ elapsed time within the window), e.g.:

```sql
WITH pts AS (
  SELECT speed,
         EXTRACT(EPOCH FROM (LEAD(date) OVER (ORDER BY date) - date)) AS dt
  FROM positions
  WHERE car_id = $car_id AND $__timeFilter(date) AND speed IS NOT NULL
)
SELECT convert_km((SUM(speed * dt) / NULLIF(SUM(dt), 0))::numeric, '$length_unit') AS speed_${length_unit}h
FROM pts
WHERE dt BETWEEN 0 AND 60  -- cap gaps so data holes don't skew the weighting
```

On the datasets above this returns 18.9 and 30.0 respectively, matching distance ÷ duration.

We ship a localized fork of the dashboards and have applied this fix in [wjsall/teslamate-chinese-dashboards v1.8.1](https://github.com/wjsall/teslamate-chinese-dashboards/releases/tag/v1.8.1) (analysis in [issue #31](https://github.com/wjsall/teslamate-chinese-dashboards/issues/31)). Happy to open a PR here if the approach looks right to you.

wjsall:
Fixes #5497

### Problem

The Drive Details **Ø Speed** stat computes `AVG(positions.speed)` — a plain point average. The stream inserts positions much more densely while moving (~0.29 s/point) than at standstill (~1.6–2 s/point), so stopped time is underweighted and the panel **overstates average speed**, especially in stop-and-go city driving.

Measured on real data (details in #5497):

| dataset | point average (panel) | distance ÷ duration (truth) | error |
|---|---|---|---|
| 24-min city drive, 7.51 km (second-level CSV) | 27 km/h | 18.8 km/h | **+43 %** |
| 48-min drive, 23.8 km (second installation) | 33.7 km/h | 29.7 km/h | +13 % |

The bias also numerically approaches "distance ÷ moving time", which misleads users into thinking stopped time is intentionally excluded.

### Fix

Time-weighted average (equivalent to distance ÷ elapsed time within the window), with inter-sample gaps capped at 60 s so data holes don't skew the weighting. On the datasets above it returns 18.9 and 30.0 km/h respectively, matching distance ÷ duration.

Single-panel change in `grafana/dashboards/internal/drive-details.json`; `$__timeFilter`, `$length_unit` conversion and the field alias are unchanged.

### DIFF
--- a/grafana/dashboards/internal/drive-details.json
+++ b/grafana/dashboards/internal/drive-details.json
@@ -2382,7 +2382,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "SELECT\n\tconvert_km(avg(speed)::numeric, '$length_unit') AS speed_${length_unit}h\nFROM positions\nWHERE car_id = $car_id AND $__timeFilter(date)",
+          "rawSql": "WITH pts AS (\n\tSELECT speed, EXTRACT(EPOCH FROM (LEAD(date) OVER (ORDER BY date) - date)) AS dt\n\tFROM positions\n\tWHERE car_id = $car_id AND $__timeFilter(date) AND speed IS NOT NULL\n)\n-- Time-weighted average: the stream downsamples at standstill, so a plain point\n-- average underweights stopped time and overstates speed. Cap gaps at 60s.\nSELECT\n\tconvert_km((SUM(speed * dt) / NULLIF(SUM(dt), 0))::numeric, '$length_unit') AS speed_${length_unit}h\nFROM pts\nWHERE dt BETWEEN 0 AND 60",
           "refId": "A",
           "sql": {
             "columns": [
--- a/grafana/dashboards/internal/drive-details.json
+++ b/grafana/dashboards/internal/drive-details.json
@@ -2382,7 +2382,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
-          "rawSql": "WITH pts AS (\n\tSELECT speed, EXTRACT(EPOCH FROM (LEAD(date) OVER (ORDER BY date) - date)) AS dt\n\tFROM positions\n\tWHERE car_id = $car_id AND $__timeFilter(date) AND speed IS NOT NULL\n)\n-- Time-weighted average: the stream downsamples at standstill, so a plain point\n-- average underweights stopped time and overstates speed. Cap gaps at 60s.\nSELECT\n\tconvert_km((SUM(speed * dt) / NULLIF(SUM(dt), 0))::numeric, '$length_unit') AS speed_${length_unit}h\nFROM pts\nWHERE dt BETWEEN 0 AND 60",
+          "rawSql": "SELECT convert_km((distance / NULLIF(EXTRACT(EPOCH FROM end_date - start_date), 0) * 3600)::numeric, '$length_unit') AS speed_${length_unit}h\nFROM drives\nWHERE id = $drive_id",
           "refId": "A",
           "sql": {
             "columns": [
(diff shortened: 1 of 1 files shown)


JakobLichterfeld (changes requested):
Thanks for the thorough analysis in #5497 and for the PR — the measurements clearly show the current point average is misleading.

 As said, as a user I'd expect Drive Details to show two figures (like most GPS/sports apps do):

- **Ø Speed** — distance ÷ total duration, including stops
- **Ø Moving Speed** — average while actually moving

Your time-weighted query effectively converges to the first one. But since this dashboard already has `$drive_id`, and distance and duration are stored on `drives`, we can compute it exactly and consistently with the Drives dashboard (which already shows `distance / duration` as avg speed):

```sql
SELECT convert_km((distance / NULLIF(EXTRACT(EPOCH FROM end_date - start_date), 0) * 3600)::numeric, '$length_unit') AS speed_${length_unit}h
FROM drives
WHERE id = $drive_id
 ```

This also makes the stat consistent with the other per-drive stats in this dashboard (duration, distance, consumption), which don't react to time-range zooming either.

For the second figure, your time-weighted CTE is exactly the right tool: with standstill excluded (`AND speed > 0`) it yields a proper moving average — which the current point average only approximates by accident.

Would you be up for reworking the PR in that direction: "Ø Speed" from `drives`, plus a new "Ø Moving Speed" stat panel next to it?

**Comment by wjsall** [comment]
Thanks for the great review — you're right, `distance / duration` from `drives` is the cleaner, exact source. Reworked into two stats as you suggested:

- **Ø Speed** (panel 37) — `distance / (end_date − start_date)` via `$drive_id`, matching the on-page **Drive Duration** stat (so the two line up).
- **Ø Moving Speed** (new stat, panel 41) — the time-weighted CTE with `AND speed > 0`, scoped by `drive_id` like the other per-drive stats so it doesn't react to the time range either.

Verified on a real 23.8 km / 48.1 min drive (~3.8 min stopped):
- Ø Speed = **29.7 km/h** (distance ÷ elapsed time ✓)
- Ø Moving Speed = **32.5 km/h** — above the total, as expected once standstill is excluded

One note: I initially scoped Ø Moving Speed by `car_id + $__timeFilter` (leftover from the old panel) — that broke your own "per-drive shouldn't follow the time range" point (widening the range pulled in adjacent drives). Fixed to `drive_id` in the second commit. Ready for another look 🙏

**Review by JakobLichterfeld — CHANGES_REQUESTED** [review]
Thanks, this is exactly the direction I had in mind — Ø Speed from `drives` looks good.

One correctness issue in the new **Ø Moving Speed** panel: `speed 0` sits in the CTE's `WHERE`, and Postgres evaluates `WHERE` *before* window functions. So `LEAD(date)` runs over the already-filtered set, and the last moving sample before a stop gets the whole standstill period as its `dt` (capped at 60 s) instead of its real sample spacing. A 45 s traffic light assigns 45 s of weight to a ~3 km/h coasting sample — same class of weighting bias as the original bug, just in the opposite direction.

Moving the filter to the outer query fixes it:

```sql
WITH pts AS (
	SELECT speed, EXTRACT(EPOCH FROM (LEAD(date) OVER (ORDER BY date) - date)) AS dt
	FROM positions
	WHERE drive_id = $drive_id
)
SELECT
	convert_km((SUM(speed * dt) / NULLIF(SUM(dt), 0))::numeric, '$length_unit') AS speed_${length_unit}h
FROM pts
WHERE speed 0 AND dt BETWEEN 0 AND 60
```

(`speed 0` also filters out NULLs, so the separate `speed IS NOT NULL` isn't needed.)

Two nits:
- The trailing newline at the end of the file got dropped — please restore it.
- The new panel sits at `x: 9, y: 34`, so it renders alone below Ø Speed with empty space to its left. Worth a look in Grafana; `x: 0` might read better.

With that fixed I'm happy to merge.

Issue #5500, opened 2026-07-11
docs: upgrading PostgreSQL guide should link directly to restore section


isis59:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

The link to "restore" database in your documentation here https://docs.teslamate.org/docs/maintenance/upgrading_postgres send to the backup script, leading user to loose all the data because we erase the previous backup with the empty db ! 

### Can the issue reliably be reproduced?

Yes

### Relevant log output

```shell
no log
```

### Type of installation

Docker (https://docs.teslamate.org/docs/installation/docker/)

### Version

v4

### Are you running latest major supported PostgreSQL version?

- [ ] I run the latest major supported PostgreSQL version

### Are you using a reverse Proxy for TeslaMate?

No

JakobLichterfeld:
Thanks for reporting.

Not sure what you mean

<img width="947" height="101" alt="Image" src="https://github.com/user-attachments/assets/11cd6463-0836-448c-a2c5-8ef01bfcff90" />

and

<img width="959" height="73" alt="Image" src="https://github.com/user-attachments/assets/942cab9e-91f6-4986-8f8c-64ac25142382" />

So it is clearly stated to do a backup.


Do you mean that backup and restore should be two separate files? Since today everybody was able to read the corresponding section, and it starts with backup on purpose.

If you have ideas for improvements, they very much welcomed.

isis59:
Hi,

The last step of the doc is missing in your screenshot... 

<img width="1983" height="224" alt="Image" src="https://github.com/user-attachments/assets/3f20b70e-da1c-4a49-bb06-90870cc04018" />

When you click on "restore", instead of pushing you to https://docs.teslamate.org/docs/maintenance/backup_restore#restore , the backup script is displayed. 

I copied the script, modified the dbname to be aligned with my newly created backup, and loosed 1 year of data... just because the script is the backup one, and not the restore ...

JakobLichterfeld:
Ah nice catch. So reading restore and then landing on backup and restore and you need to scroll to restore.
Let's see if this can be handled better.

isis59:
I just gave you the right link to use in your doc ;)
https://docs.teslamate.org/docs/maintenance/backup_restore#restore

JakobLichterfeld:
> I just gave you the right link to use in your doc ;) https://docs.teslamate.org/docs/maintenance/backup_restore#restore

[...]

PR #5502, opened 2026-07-12
docs: split the backup and restore guides into two separate guides and highlight that you should transfer your backup of the host

JakobLichterfeld:
- [x] split the backup and restore guides into two separate guides
- [x] highlight that you should transfer your backup of the host
- [x] Correct backup links in various documentation files to point to the new backup.mdx.
- [x] redirect old `backup_restore` path to `backup`


JakobLichterfeld:
As this prevent issues like in #5500 I merge directly

PR #5503, opened 2026-07-12
fix: honor DATABASE_USER/DATABASE_PASS with DATABASE_SOCKET_DIR

### LINKED ISSUE #5286: Teslamate crashes when using postgres with unix socket connection: connect raised KeyError exception: key :username not found

### What happened?

I am trying to run teslamate on a system that already hosts a Postgres database for several other applications and is also backed up frequently.
That server only allows connections to the unix socket and doesn't open a TCP socket / port, as all applications using the server are running on the same machine.

I've configured Teslamate as follows:
```ini
DATABASE_SOCKET_DIR=/run/postgresql
DATABASE_USER=teslamate
DATABASE_NAME=teslamate
DATABASE_PASS=<password>
```

Upon starting teslamate, the postgrex connect call fails with `connect raised KeyError exception: key :username not found.` (full error below).
It seems that postgrex is expecting a username (and likely password, too) for its `connect` call, which [are not provided](https://github.com/teslamate-org/teslamate/blob/5bf8f826aba9b65f2e3382cb9e32a4d63b846202/config/runtime.exs#L112-L115).

I am aware that one _usually_ uses peer authentication for database connections via the socket, however, that's not always possible.
One example is running applications, such as `teslamate` in this case, within docker and mounting the postgresql socket into the container.
In this case, peer authentication will not work and one needs to fall back to username and password.
However, teslamate / postgrex doesn't even start to attempt a connection to postgresql here.

### Expected Behavior

Teslamate doesn't crash but starts normally.

Yes

_No response_

### Relevant log output

```shell
** (RuntimeError) connect raised KeyError exception: key :username not found. The exception details are hidden, as they may contain sensitive data such as database credentials. You may set :show_sensitive_data_on_connection_error to true when starting your connection if you wish to see all of the details
    (elixir 1.19.5) lib/keyword.ex:608: Keyword.fetch!/2
    (postgrex 0.22.0) lib/postgrex/protocol.ex:864: Postgrex.Protocol.startup/2
    (postgrex 0.22.0) lib/postgrex/protocol.ex:760: Postgrex.Protocol.handshake/3
    (postgrex 0.22.0) lib/postgrex/protocol.ex:229: Postgrex.Protocol.connect_endpoints/6
    (db_connection 2.9.0) lib/db_connection/connection.ex:80: DBConnection.Connection.handle_event/4
    (stdlib 5.2.3.5) gen_statem.erl:1397: :gen_statem.loop_state_callback/11
    (stdlib 5.2.3.5) proc_lib.erl:241: :proc_lib.init_p_do_apply/3
2026-04-16 22:22:03.739 [notice] Application teslamate exited: exited in: TeslaMate.Application.start(:normal, [])
    ** (EXIT) exited in: DBConnection.Holder.checkout(#PID<0.2481.0>, [log: #Function<14.97393993/1 in Ecto.Adapters.SQL.with_log/4>, pool: DBConnection.ConnectionPool, repo: TeslaMate.Repo, pool_size: 10, timeout: 60000])
        ** (EXIT) killed
Kernel pid terminated (application_controller) ("{application_start_failure,teslamate,{bad_return,{{'Elixir.TeslaMate.Application',start,[normal,[]]},{'EXIT',{killed,{'Elixir.DBConnection.Holder',checkout,[<0.2481.0>,[{log,#Fun<Elixir.Ecto.Adapters.SQL.14.97393993>},{pool,'Elixir.DBConnection.ConnectionPool'},{repo,'Elixir.TeslaMate.Repo'},{pool_size,10},{timeout,60000}]]}}}}}}")

Crash dump is being written to: erl_crash.dump...done
```

### Screenshots

n/a

### Additional data

n/a

### Type of installation

### Version

3.0.0 (MCU2 / pr4453)

### PostgreSQL version

18

Yes (specify in next field)

not relevant

mvanhorn:
## Summary

TeslaMate no longer crashes at startup with `connect raised KeyError exception: key :username not found` when `DATABASE_SOCKET_DIR` is set and the socket requires password authentication. The `socket_dir` branch in `config/runtime.exs` now includes `username`/`password` from `DATABASE_USER`/`DATABASE_PASS` when those env vars are set, filtering out unset values, so peer-authentication setups that omit the vars see exactly the previous behavior.

## Why this matters

The reporter in #5286 pinned the root cause: the TCP branch passes credentials but the socket branch configured only `socket_dir` and `port`. Peer auth can mask this (Postgrex falls back to `USER`/`PGUSER`), but in a Docker container with the socket mounted in, no such env var exists and password auth is required, so Postgrex raises before attempting a connection. The reporter has kept the issue alive through two stale-bot cycles.

## Testing

Reviewed the branch construction to confirm the nil-filtering keeps the keyword list free of unset credentials (peer-auth users unaffected). Updated `website/docs/configuration/environment_variables.md` to document that `DATABASE_USER`/`DATABASE_PASS` are honored together with `DATABASE_SOCKET_DIR`.

Fixes #5286


brianmay (approved):

brianmay:
LLM says this looks good, here is the report:

> 
> ## Task tool output
> The task tool was run with the prompt "review changes [commit|branch|pr], defaults to uncommitted" and has completed.
> 
> ## Task tool results
> Review of PR #5503 (`config/runtime.exs:112-122` + docs).
> 
> ## Verdict: Looks good, ship it.
> 
> The fix is correct. Verified:
> 
> - **Backwards compat preserved**: `Enum.reject(fn {_key, v} -> is_nil(v) end)` strips unset creds, so peer-auth users (omitting `DATABASE_USER`/`DATABASE_PASS`) get the exact same keyword list as before — `[socket_dir: ..., port: ...]`.
> - **Merge behavior is fine**: The subsequent `config :teslamate, TeslaMate.Repo, ...` at `config/runtime.exs:118` merges keyword lists, so `username`/`password` set in the socket branch are retained alongside `pool_size`/`timeout`/`database`.
> - **No raise in prod for peer-auth**: The socket branch uses `System.get_env/1` (returns `nil`), not `Util.fetch_env!/2`, so peer-auth users on a socket in `prod` won't trigger a required-env-var raise. This matches the previous behavior.
> 
> ## One observation (not a blocker)
> 
> There's now an intentional asymmetry between the two branches:
> - TCP branch (`config/runtime.exs:107-108`): defaults `DATABASE_USER`/`DATABASE_PASS` to `"postgres"` via `Util.fetch_env!(..., all: "postgres")`.
> - Socket branch: no defaults — unset vars are filtered out.
> 
> This is the right call for the stated goal (peer-auth must keep working), and the docs update captures it. Worth being aware that a user switching from TCP to socket who was relying on the `"postgres"` default will now silently fall back to peer auth instead of connecting as `postgres/password`. The PR description and docs both make this explicit, so I don't think action is required.

JakobLichterfeld:
Nice, thanks for your contribution!

PR #5507, opened 2026-07-15
build: require Elixir 1.19.5 or newer

### LINKED ISSUE #5505: TeslaMate no longer compiles with Elixir 1.18.4+ / OTP 28

### What happened?

This code no longer compiles with Elixir 1.18.4+ / OTP 28 as it's emitting a warning

https://github.com/teslamate-org/teslamate/blob/7054517c10475f39f480edeae8f90c6f717985a3/lib/tesla_api/error.ex#L127

### Expected Behavior

Either compile correctly or we should rephrase support ambitions again in:

https://docs.teslamate.org/docs/development#requirements

Yes

_No response_

### Relevant log output

```shell
Generating TeslaMateWeb.Cldr for 20 locales named [:ca, :da, :de, :en, :es, ...] with a default locale named :en
     warning: Inspect.Map.inspect_as_struct/4 is undefined or private. Did you mean:
           * inspect/4
           * inspect/2
     │
 127 │     |> Inspect.Map.inspect_as_struct("TeslaApi.Error", fields, opts)
     │                    ~
     │
     └─ lib/tesla_api/error.ex:127:20: Inspect.TeslaApi.Error.inspect/2
Compilation failed due to warnings while using the --warnings-as-errors option
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

4.1.0-dev

### PostgreSQL version

18.4

Yes (specify in next field)

Traefik

magrathean-uk:
## Summary

- Require Elixir 1.19.5 or newer in `mix.exs`, matching the current official image and CI runtime.
- List Elixir and Erlang/OTP separately and remove the untested Elixir 1.18.4 support claim.
- Update the unsupported Debian and FreeBSD guides so manual installations can still use versions that meet the current runtime requirements.
- Keep both manual guides consistent with the PostgreSQL 18 baseline.

## Package check

Checked on 2026-07-15:

- [Debian Trixie ships Elixir 1.18.3 on Erlang/OTP 27](https://packages.debian.org/trixie/elixir), so its stock packages already do not meet TeslaMate's OTP 28 requirement.
- FreeBSD's latest ports provide [Elixir 1.19.5](https://www.freshports.org/lang/elixir/) and [Erlang/OTP 28](https://www.freshports.org/lang/erlang/). The quarterly Elixir port is still 1.17.3, so the guide now points to FreeBSD's latest package repository and requires a version check.
- Debian users are directed to the official Elixir installation methods when distribution packages are too old.

The previous Inspect compatibility workaround has been removed completely. This does not add a version matrix.

## Validation

- `mix ci` on Elixir 1.19.5 / Erlang/OTP 28 with PostgreSQL 18.4: 361 tests, 0 failures
- `npm run build` in `website`
- Prettier and typos checks
- Requirement accepts 1.19.5 and 1.20.x, and rejects 1.19.4 and 2.0.0

Closes #5505

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 #### Build, CI, internal
 
+- build: require Elixir 1.19.5 or newer (#5507 - @magrathean-uk)
 - build(nix): update mixFodDeps hash in nix builds ([954e8739](https://github.com/teslamate-org/teslamate/commit/954e8739326e092f7cddf7308dd4b704cc008f62) - @JakobLichterfeld)
 - build(deps): bump launch-editor from 2.13.2 to 2.14.1 in /website (#5426)
 - build(deps): update flake.lock (#5427)
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,7 @@ defmodule TeslaMate.MixProject do
     [
       app: :teslamate,
       version: version(),
-      elixir: "~> 1.17",
+      elixir: ">= 1.19.5 and < 2.0.0",
       elixirc_paths: elixirc_paths(Mix.env()),
       compilers: Mix.compilers(),
       start_permanent: Mix.env() == :prod,
--- a/website/docs/development.mdx
+++ b/website/docs/development.mdx
@@ -6,10 +6,10 @@ sidebar_label: Development and Contributing
 
 ## Requirements
 
-TeslaMate officially supports the versions listed below. In general, TeslaMate aims to stay compatible with the last three major Erlang and Postgres releases and the last three minor Elixir releases. At the moment, OTP 28+ is a hard requirement, so Elixir support is currently limited to versions that support OTP 28 (Elixir 1.18.4+ and 1.19.x).
-When contributing, please use the versions listed below, which match those used in official TeslaMate distributions.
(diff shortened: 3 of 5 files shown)


JakobLichterfeld:
> Newer supported Elixir releases warn when TeslaMate calls the private Inspect function.

Thanks, not sure if this is the right direction. Older Elixir versions complain, newer versions are fine.

JakobLichterfeld:
> If the intended direction is to require 1.19.5 everywhere instead, then I agree the cleaner option is to close this and tighten `mix.exs` and the support wording.

That's my thinking, but we need to check if Debian and FreeBSD from the manual installation support it. Yes they are marked as unsupported but hard to drop atm.
I do not like to try to avoid something which is not an issue in CI infrastructure. So yes we can introduce Matrix testing different Elixir versions, the foundation is already laid out, but atm this shouldn't be our priority.

swiffer:
Looks good for the Elixir-related changes.  
One minor note: Postgres support has been limited to 18.0+, but the code isn’t enforcing that yet and still allows 16.7 / 17.3 / 18.0. We’re happy with that, so please keep it as is.

https://github.com/teslamate-org/teslamate/blob/4e11474cfd5542b446f8459ede1fecd34ae39606/lib/teslamate/database_check.ex#L11-L17

PR #5508, opened 2026-07-15
fix(geofences): increase cost precision

### LINKED ISSUE #4410: Cost per KwH have a max value at 99

### What happened?

Can't set a value above 99

### Expected Behavior

Can set the value above 99

### Steps To Reproduce

Set a cost per kwh at 101 in a geofence
Saved value is 99

### Relevant log output

```shell
No
```

### Screenshots

Capture d’écran 2024-11-30 à 3 49 41 PM

### Additional data

_No response_

### Type of installation

Docker

### Version

v1.32.0

magrathean-uk:
## Summary

Expand `geofences.cost_per_unit` from `numeric(6,4)` to `numeric(9,4)` and `charging_processes.cost` from `numeric(6,2)` to `numeric(14,2)`.

## Why

Charging rates in high-denomination currencies can exceed the current maximum and fail with a PostgreSQL numeric overflow. Widening only the rate column would still let the calculated charging total overflow, so both storage points are covered.

## Migration safety

Normal rollback restores both legacy widths. If any stored rate or total no longer fits, rollback raises before changing either column instead of truncating or discarding data.

## Validation

- 18 focused geofence tests, including the upper-bound rate and a recalculated high-value total
- Full combined `mix ci`: 374 tests passed
- PostgreSQL 18 migration `up -> down -> up`
- Verified rollback refusal for an oversized rate and an oversized total, with the widened schema left intact
- `mix format --check-formatted` and staged secret scan

Closes #4410

### DIFF
--- a/priv/repo/migrations/20260718160000_increase_geofence_cost_precision.exs
+++ b/priv/repo/migrations/20260718160000_increase_geofence_cost_precision.exs
@@ -1,0 +1,59 @@
+defmodule TeslaMate.Repo.Migrations.IncreaseGeofenceCostPrecision do
+  use Ecto.Migration
+
+  @legacy_max_cost_per_unit "99.9999"
+  @legacy_max_fixed_cost "9999.99"
+
+  def up do
+    alter table(:charging_processes) do
+      modify(:cost, :decimal, precision: 14, scale: 2)
+    end
+
+    alter table(:geofences) do
+      modify(:cost_per_unit, :decimal, precision: 9, scale: 4)
+      modify(:session_fee, :decimal, precision: 14, scale: 2)
+    end
+  end
+
+  def down do
+    # Do not silently truncate or discard costs that cannot fit the old schema.
+    execute("""
+    DO $$
+    BEGIN
+      IF EXISTS (
+        SELECT 1 FROM geofences
+        WHERE ABS(cost_per_unit) > #{@legacy_max_cost_per_unit}
+      ) THEN
+        RAISE EXCEPTION
+          'cannot restore geofences.cost_per_unit to numeric(6,4): values exceed #{@legacy_max_cost_per_unit}';
+      END IF;
+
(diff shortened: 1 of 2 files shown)


JakobLichterfeld:
Thanks for throwing tokens on this.
Only using migrations isn't reliable in all cases as mentioned by Brian several times.

JakobLichterfeld:
Thanks, what about new installs?

JakobLichterfeld (changes requested):
Thanks Bolyki for the detailed follow-ups and for actually testing the backup/restore, downgrade, and fresh-install paths — that answers my questions, and the guarded rollback (refusing instead of truncating) is the right call.

Two things before this can move forward:

1. **Migration timestamp:** `20260715080000` now sorts before `20260715081000_add_not_null_constraint_to_vin.exs`, which landed on `main` with #5512. Ecto will still run it out of order, but please rebase onto `main` and bump the timestamp so the version order matches merge order.

2. **`session_fee` has the same limit:** `geofences.session_fee` is still `numeric(6,2)`, so a flat session fee above 9,999.99 in a high-denomination currency hits exactly the same overflow this PR fixes. Since the migration already touches both cost columns, please widen `session_fee` in the same migration (including the rollback guard) and add a matching test, so we don't need a follow-up migration for the same issue class.

With those two addressed I'm happy to merge.

JakobLichterfeld (approved):
Thanks for the quick turnaround — timestamp ordering and `session_fee` are both addressed exactly as discussed, including the extended rollback guard and test coverage. LGTM! 🚀

🤖 Review drafted with [[Claude Code](https://claude.com/claude-code)](https://claude.com/claude-code) (Fable 5 high) — sponsored by [[Claude for Open Source](https://www.anthropic.com/claude-for-open-source)](https://www.anthropic.com/claude-for-open-source)

PR #5510, opened 2026-07-15
test: reuse API snapshots across paired fetches

### LINKED ISSUE #5278: multiple fetches discard test data
In the code we have:

```elixir
  defp fetch_with_reachable_assumption(id, deps) do
    with {:error, :vehicle_unavailable} <- call(deps.api, :get_vehicle_with_state, [id]) do
      call(deps.api, :get_vehicle, [id])
    end
  end

  defp fetch_with_unreachable_assumption(id, deps) do
    with {:ok, %Vehicle{state: "online"}} <- call(deps.api, :get_vehicle, [id]) do
      call(deps.api, :get_vehicle_with_state, [id])
    end
  end
```

note: Since the MCU changes we only make one call if `use_streaming_api` - but many of the tests
at this point of time disable the use_streaming_api, so still applicable.

Under normal circumstances this is OK. The Tesla API will return the same (or more recent) data multiple times.

But under unit testing, `deps.api` is `ApiMock`, and every request will discard the previous data.

I am not ready to fix this properly right now, but documenting it so I don't forget.

Also see https://github.com/teslamate-org/teslamate/pull/4453#issuecomment-4230384910

magrathean-uk:
## Summary

Add an explicit API mock snapshot wrapper for responses that should represent one logical vehicle state across TeslaMate's lightweight and full fetch sequence.

## Why

The production update cycle can perform two API calls for one logical state. Stateful mock functions were evaluated twice, which could consume two fixtures and make tests observe a transition that production did not intend.

## Compatibility

Snapshot reuse is opt-in. Existing sequential mock behavior remains the default, wrapped functions are evaluated exactly once for paired fetches, and cached snapshots are isolated by vehicle id. Snapshot and endpoint-pinning wrappers compose in either order. Isolated mocks are addressed by PID rather than dynamically generated registered atoms.

## Validation

- Four dedicated API mock tests covering reuse, single evaluation, cross-vehicle isolation, and endpoint-wrapper composition
- Seven focused API mock and vehicle-updating tests passed
- Full `mix ci`: 365 tests passed
- Hosted Elixir tests, static analysis, formatting, security scans, CLA, and deploy preview passed

Closes #5278

### DIFF
--- a/test/support/mocks/api.ex
+++ b/test/support/mocks/api.ex
@@ -34,8 +34,7 @@ defmodule ApiMock do
         {:get_vehicle_with_state, id},
         _from,
         %State{pending_vehicle_data: {id, result}} = state
-      )
-      when not is_nil(result) do
+      ) do
     {:reply, result, advance_event(state)}
   end
 
--- a/test/support/mocks/api.ex
+++ b/test/support/mocks/api.ex
@@ -2,13 +2,13 @@ defmodule ApiMock do
   use GenServer
 
   defmodule State do
-    defstruct [:pid, :events]
+    defstruct [:pid, :events, :pending_vehicle_data]
   end
 
   # API
 
   def start_link(opts) do
-    GenServer.start_link(__MODULE__, opts, name: Keyword.fetch!(opts, :name))
+    GenServer.start_link(__MODULE__, opts, Keyword.take(opts, [:name]))
   end
 
   def get_vehicle(name, id), do: GenServer.call(name, {:get_vehicle, id})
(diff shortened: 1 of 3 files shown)


JakobLichterfeld (approved):
Thanks a lot for tackling #5278 — and for the exemplary PR description and the quick turnaround on the review feedback!

I really like the design: the `:snapshot` wrapper is strictly opt-in, the existing sequential semantics (including the repeat-last-event behavior) are preserved, snapshots are isolated per vehicle id, and the composition with the endpoint-pinning wrapper works in both nesting orders. The four dedicated mock tests cover exactly the cases I would have asked about (single evaluation, cross-vehicle isolation, wrapper composition), and migrating `updating_test` as a first consumer nicely demonstrates the intended usage.

Two tiny, non-blocking remarks:
- The `when not is_nil(result)` guard can never fire since `pending_vehicle_data` only ever holds an `{:ok, …}` tuple — feel free to drop it (or keep it as documentation of intent).
- Since #5278 mentions this pattern affects several tests: are you planning follow-up PRs to migrate the remaining paired-fetch tests to `:snapshot`? Incremental is totally fine, just curious about the plan.

LGTM 🚀

JakobLichterfeld:
Perfect, thanks — that's exactly the right split: migrating only the genuine same-state pairs and leaving the fixtures that deliberately model transitions or endpoint-specific responses untouched. Looking forward to the follow-ups!

PR #5512, opened 2026-07-15
fix(cars): enforce non-null VINs

### LINKED ISSUE #4801: vin was planned to be marked as  non nullable in v2

### What happened?

just leaving this here as I've seen it while coding:

https://github.com/teslamate-org/teslamate/blob/8ca246015d71eae7543254327f359b5bfd3eedad/lib/teslamate/log/car.ex#L20

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

v2.0.0

### PostgreSQL version

v17

magrathean-uk:
## Summary

Backfill legacy `NULL` VIN rows with a unique internal marker, then enforce the existing schema expectation with a database `NOT NULL` constraint.

## Why

The Ecto schema already requires VINs, but the database still permits null values. Applying the constraint directly would fail startup for installations that contain valid historical cars created before VIN was reliably present.

## Migration safety

The migration preserves each car and all related history. Existing vehicle discovery replaces the internal marker through the normal `eid` or `vid` lookup once a real VIN is available. Rollback restores only a marker that exactly matches its car ID, leaving valid VINs untouched.

## Validation

- Isolated PostgreSQL 18 database containing a real legacy `NULL` VIN row
- Verified migration `up -> down -> up`
- Verified marker creation, `NOT NULL`, exact rollback to `NULL`, and reapplication
- Focused car tests and full `mix ci`: 374 tests passed
- `mix format --check-formatted` and staged secret scan

Closes #4801

### DIFF
--- a/lib/teslamate/log/car.ex
+++ b/lib/teslamate/log/car.ex
@@ -17,7 +17,6 @@ defmodule TeslaMate.Log.Car do
 
     field :eid, :integer
     field :vid, :integer
-    # TODO: with v2.0 mark as non nullable
     field :vin, :string
     field :display_priority, :integer
 
--- a/priv/repo/migrations/20260715081000_add_not_null_constraint_to_vin.exs
+++ b/priv/repo/migrations/20260715081000_add_not_null_constraint_to_vin.exs
@@ -1,0 +1,30 @@
+defmodule TeslaMate.Repo.Migrations.AddNotNullConstraintToVin do
+  use Ecto.Migration
+
+  @legacy_vin_prefix "__legacy_null_vin__:"
+
+  def up do
+    # Preserve old cars and their history; vehicle discovery replaces this marker via eid or vid.
+    execute("""
+    UPDATE cars
+    SET vin = '#{@legacy_vin_prefix}' || id::text
+    WHERE vin IS NULL
+    """)
+
+    alter table(:cars) do
+      modify(:vin, :text, null: false)
+    end
+  end
+
+  def down do
+    alter table(:cars) do
+      modify(:vin, :text, null: true)
+    end
+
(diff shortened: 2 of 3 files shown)


JakobLichterfeld:
Thanks for throwing tokens on this.
Only using migrations isn't reliable in all cases as mentioned by Brian several times.

JakobLichterfeld:
You are right, thanks for double checking!

JakobLichterfeld (approved):
Thanks a lot for tackling this long-standing TODO from #4801 — and especially for the careful migration design and the thorough validation notes!

I verified the key claims against the code and the approach looks right to me:

- Backfilling with a unique per-car marker preserves the full history of legacy installations instead of dropping cars.
- The marker is indeed replaced automatically once the car is seen again, via the `vid`/`eid` lookup in `Vehicles.create_or_update!/1`.
- The rollback only touches exact marker matches, and the column type (`text`) matches the original migration.
- Since the changeset already required `vin` and the import path only creates cars with a VIN, the new constraint can't surface at runtime — only legacy rows were affected.

Two small notes, neither blocking:

1. For installations that still carry such a legacy row, the marker (`__legacy_null_vin__:<id>`) will show up wherever the VIN is displayed (e.g. Grafana). I think that's an acceptable trade-off, but we should mention it briefly in the release notes.
2. Nit: `modify(:vin, :text, null: false)` re-states the column type; `from: :text` (or a plain `execute "ALTER TABLE cars ALTER COLUMN vin SET NOT NULL"`) would make the intent a bit clearer.

LGTM, thanks again! 🙏

Issue #4850, opened 2025-07-23
feat(webview): add a maintenance view to simplify common maintenance tasks, such as closing drives and charges


JakobLichterfeld:
Nice enhancement, which would improve UX, even if unfinished drives or changes should not happen that often.

> Could we please have two buttons on each of this "unfinished" entries so that we can finish or delete the entries with a click on that list without having to detour to the command line?

Grafana is just a dashboard to view data from the database, so as Brian said:

> Probably wouldn't be able to do this in grafana, that is read only. So I think would need a new view(s) in the teslamate web site.

That said: The set charge cost does a trick where you can click in the charge dashboards and a TeslaMate Web-View opens up where you can set the charging cost, which goes into the DB and is then available in the dashboards.

> An even better solution would be a `clean up`-button that deletes entries in that list having predecessors with a consecutive ID and finishes the remaining.

I could not imagine that one solution would fit all, as in some cases a drive is spit into multiple drives, in other cases the end is not logged because the connection was lost due to an underground car park or something, and so on.

So a Web View should handle multiple cases and its up to the user to decide.

_Originally posted by @JakobLichterfeld in https://github.com/teslamate-org/teslamate/discussions/4583#discussioncomment-13021991_

based on #4706

#4849 is a nix solution in the meantime

swiffer:
Leaving a recent Grafana enhancement here to think about:

It now is possible to provide Action Links - so e.g. closing a drive or charge can happen in the background without leaving the Grafana UI at all:

https://grafana.com/whats-new/2025-09-03-actions-authentication-with-the-infinity-data-source/

However authentication might differ based on user config and therefore is not trivial to configure the required enpoint in Grafana correctly.

esclear:
There's two more tasks I currently do manually, which are somewhat annoying and would be super useful to have an UI for:
* Splitting drives: I frequently park in underground parking and even though there's multiple hours without any position updates in between and the next position update is within 400 m on the odometer of the last position, it results in one very long drive.
  That's pretty annoying to do manually, though I have a beast of an SQL query, which does that reasonably well by now.
* Merging drives: When stuck in traffic and stopped for longer than 10 minutes or so, the car goes in park, ending a drive.
   That doesn't happen as frequently, however, and is easier to fix by "just" merging two rows in the `drives` table and deleting one.

Authentication-wise, I don't think it has to happen in Grafana, necessarily.
Similar to the GPX export, Grafana could just generate the appropriate link for that.

[...]

PR #5517, opened 2026-07-16
feat(maintenance): add read-only data health inbox

magrathean-uk:
### Context

People currently need SQL or the manual maintenance guide to notice sessions that have stayed open for an unusually long time. This adds a small, deliberately conservative inbox at `/maintenance` so those records can be reviewed without changing them.

Related to #4850. This does not close that issue.

### What the suggestion does

- Lists drives and charging sessions that are still open after 48 hours
- Shows the oldest findings first and caps the response at 200
- Adds a manual refresh and a link to the existing maintenance guide
- States clearly that a long-running session is not automatically corrupt
- Performs no repair, close, delete, or other write operation

The first version intentionally avoids broad corruption guesses and child-sample scans. More checks can be discussed separately once their false-positive and query-cost boundaries are clear.

### Verification

- `mix ci`: 368 passed
- Focused data-health and LiveView tests: 7 passed
- PostgreSQL 18 clone with 8,812,983 positions: 5 findings returned
- Cold report on that clone: about 69 ms
- Car, drive, and charging-session fingerprints were identical before and after the report

### Suggested labels


JakobLichterfeld:
Thank you for the careful write-up and the clean scoping — read-only, capped, and well tested.

That said, I won't merge this, for a simple reason: **the feature already exists.** The default Grafana dashboards ship with dedicated "Incomplete Drives 🛣️" and "Incomplete Charges 🪫" panels (see `grafana/dashboards/drives.json` and `charges.json`) that list exactly these records — drives and charges without an ending date — and link the same [Manually fixing data](https://docs.teslamate.org/docs/maintenance/manually_fixing_data) guide. They cover it with more context, too: the panels show the full session details and link straight into the per-drive dashboards, while this page shows ID and start time behind a fixed 48-hour threshold with a 200-row cap.

So a second surface in the web UI would duplicate an existing feature in reduced form, at the cost of a permanent top-level navbar slot, a new core module to maintain, and 16 new strings across 20 locales for our translators. That trade doesn't work out.

If you see gaps in the existing panels — thresholds, sorting, missing hints — improving those dashboards would be a very welcome contribution and reaches every user without new UI surface. And if you want to push the broader "detect and repair" story forward, #4850 is the right place to align on direction first before building foundations for it.

Thanks again, and I look forward to your next contributions.

brianmay:
I do wonder if it would make sense to have one dashboard page that lists all health issues with the data.

At present, with one two, probably doesn't matter so much. But if we get more, it might make sense so we can tell uses "go to this page, do you see any issues?"

JakobLichterfeld:
> I do wonder if it would make sense to have one dashboard page that lists all health issues with the data.

Yeah. Funny thing: I was first looking in Database information dashboard, as I thought there is aggregated info listed for drives and charges.

Issue #5442, opened 2026-06-25
MQTT topics remain empty when stream delivers no events despite successful REST API vehicle_data calls


Tease2285:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

When TeslaMate 4.x is configured with `TESLA_WSS_USE_VIN=true` (Fleet Telemetry streaming), MQTT topics under `teslamate/cars/1/#` remain completely empty even though TeslaMate is actively polling the Tesla Fleet API (`vehicle_data`) every ~30 seconds and receiving successful `200` responses.

The log message `Stream connected, but nothing received, allow real online` appears, after which TeslaMate transitions to `:online` state and continues making REST API calls — but no data is published to MQTT.

This is highly confusing because the API calls are clearly succeeding (confirmed via MyTeslaMate proxy API logs), yet the MQTT broker receives nothing. Users have no indication that MQTT publishing depends exclusively on stream events, not on REST API responses.

A `docker compose restart teslamate` fixes the issue temporarily — after restart, the stream re-establishes and MQTT publishing works correctly.

### Expected Behavior

Either MQTT is populated from REST API data as a fallback when the stream delivers no events, or the documentation and logs clearly state that MQTT publishing requires active stream events and will remain silent if the stream is connected but idle.

A log warning such as "Stream connected but no events received — MQTT publishing is paused until stream delivers data" would also help users diagnose the issue without having to trace API proxy logs.

### Can the issue reliably be reproduced?

Yes

### Steps To Reproduce the issue

1. Configure TeslaMate with `TESLA_WSS_USE_VIN=true` and a streaming endpoint (e.g. MyTeslaMate: `TESLA_WSS_HOST=wss://streaming.myteslamate.com`)
2. Start TeslaMate and observe logs — the vehicle reaches `:online` state
3. Confirm REST API calls to `vehicle_data` are being made and returning `200` (visible in proxy API logs)
4. Subscribe to MQTT: `docker exec mosquitto mosquitto_sub -t "teslamate/cars/1/#" -v`
5. Observe that no messages are published despite the vehicle being online and REST data being fetched successfully

### Relevant log output

```shell
Stream connecting...
Stream connected, but nothing received, allow real online
Start / :online
```

### Type of installation

Docker (https://docs.teslamate.org/docs/installation/docker/)

### Version

v4.0.1

### PostgreSQL version

latest (4.x)

[...]

PR #5518, opened 2026-07-16
feat(health): add aggregate runtime liveness reporting

magrathean-uk:
### Context

TeslaMate can be running while one collection path is stale or disconnected. A database-only health check cannot show whether REST polling, streaming, MQTT publication, or the per-vehicle logger has stopped making progress.

Related to #5442. This does not close that issue.

### What the suggestion does

- Keeps redacted liveness in memory so old state does not survive a restart
- Tracks last success, failure, event, data age, consecutive failures, retry time, and a bounded reason internally
- Records REST, stream, summary, MQTT connection, and MQTT publish signals
- Expires vehicle entries that have not been seen for 24 hours and removes them when a vehicle process stops
- Adds `GET /api/health` with only aggregate status, MQTT status, and vehicle counts
- Returns 503 when the database or runtime collector is unavailable
- Sends `Cache-Control: no-store` on success and failure
- Preserves the existing `/health_check` endpoint
- Advances an MQTT generation after a real reconnect so retained state is republished once
- Resets subscriber deduplication after reconnect or collector restart

The detailed report stays inside the TeslaMate process. The public response contains no car IDs, timestamps, logger states, reasons, retry times, or vehicle data.

### Verification

- Low-concurrency full suite: 381 passed
- Independent runtime-focused set: 35 passed
- MQTT reconnect stress: 54 passed across six random seeds
- Legacy API and import FakeApi arities remain exported
- PostgreSQL 18 clone: HTTP 200, aggregate-only response, `no-store`, MQTT disabled state represented correctly
- Clone car, position, and charge counts were identical before and after the health request

### Suggested labels


JakobLichterfeld:
Thank you for the thorough work — the redaction discipline is genuinely well done here (the tests proving that raw payloads, car IDs, and failure details never leak into the public response are exactly how I'd want that verified), and the MQTT reconnect handling shows real care around races.

Still, I won't merge this as it stands, for the same two structural reasons I named on #5520:

**1. Runtime cost on the hot path.** The MQTT `VehicleSubscriber` now makes a synchronous `GenServer.call` into the `RuntimeHealth` singleton on **every** summary — roughly once per second per car while driving — plus casts into that same process for every stream frame, fetch result, summary, and publish. All cars funnel into one collector process, making it a central serialization point in the pipeline: always-on overhead paid by every installation, whether or not anyone ever queries the endpoint. That trade doesn't work for me, especially on the small hardware much of our user base runs.

**2. A second health endpoint.** TeslaMate already has an established health surface: `/health_check` (`lib/teslamate_web/health_check.ex`), which container healthchecks and monitoring setups rely on today. I know it's intentionally shallow — that's precisely your motivation — but the answer to that is evolving the existing surface, and that discussion belongs in #5442/#5524 before any implementation. Running two health endpoints with different semantics side by side (one that always says "ok", one that may say "degraded") confuses every monitoring setup, and `/api/health` would additionally be a new unauthenticated endpoint with a `schema_version` contract we'd maintain indefinitely, while authentication for the API remains an unsolved, project-wide topic. The instrumentation footprint is also wide for an observability feature: eight production modules touched, a new `Api.stream` arity, and liveness concepts inside `TeslaApi.Stream`.

That said, there is one piece in here I'd genuinely like to see land on its own: **the retained-snapshot republish after a real broker reconnect.** When a broker restarts without persistence, retained topics stay empty until each value happens to change — that bites every MQTT user, and your generation mechanism handles the races well. It doesn't need the health collector, though: the `Mqtt.Handler` already sees the down→up transition and could broadcast it via `Phoenix.PubSub` directly, with subscribers simply resetting their `last_values`. Message-driven, no per-summary calls, no new endpoint. If you extract that into a focused PR, I'd be glad to review it.

Thanks again — and I mean it about the reconnect fix.

PR #5519, opened 2026-07-16
feat(import): continue past malformed rows and resume completed files

magrathean-uk:
### Context

A malformed TeslaFi row can currently stop a long import after valid data has already been processed. A restart also gives the user little durable information about which files were safely completed.

### What the suggestion does

- Validates each CSV row before it reaches the logger
- Quarantines malformed rows and continues with valid rows
- Stores only a basename, content fingerprint, row number, bounded field names, and a stable reason
- Never stores raw row values or the source directory path in the rejection report
- Persists one import run and file-boundary checkpoints
- Skips only completed files whose SHA-256 fingerprint still matches
- Invalidates a checkpoint when file content changes
- Restores the original time zone after a restart and prevents a submitted time-zone change from replacing the active run
- Preserves FIFO handling when more than one API caller is waiting
- Releases queued callers when an import completes or aborts
- Reports stream-preparation failures instead of leaving the run in a running state
- Stops importer children cleanly when a file changes vehicle identity, with bounded cleanup

The import page shows the exact rejection count and the first 100 safe examples. It also explains that skipped rows can leave a nearby drive or charge summary incomplete.

### Recovery boundary

This is file-level, at-least-once recovery. A file that reached its boundary is skipped after restart. The file active during a crash starts again from its beginning and may import rows that were already stored from that file. This does not claim row-exact or exactly-once recovery.

The original CSV files should remain unchanged until the run finishes. A resumed run always keeps its saved time zone so timestamps cannot shift partway through the import.

Vehicle identity uses known stable identifiers conservatively: matching VINs remain authoritative, legacy rows without a VIN remain importable, isolated conflicting rows are quarantined, and a file that clearly switches vehicles aborts cleanly.

### Verification

- `mix ci`: 397 passed
- Focused importer suite: 22 passed
- Combined importer, FakeApi, and Vehicle regression suite: 50 passed
- Hosted CI is green on the current head
- Migration applied to a disposable clone of our PostgreSQL 18 TeslaMate database
- Malformed-row clone trial: valid rows continued, one rejection persisted, raw values and source path absent
- Completed-file restart trial: no extra positions, checkpoints, rejections, or runs
- Changed submitted time-zone trial: one run remained, saved time zone remained, no replacement run was created

### Suggested labels

### DIFF
--- a/lib/teslamate/import.ex
+++ b/lib/teslamate/import.ex
@@ -8,14 +8,30 @@ defmodule TeslaMate.Import do
   alias TeslaMate.{Vehicles, Repair, Log}
   alias TeslaMate.Log.{Car, State}
 
-  alias __MODULE__.{Status, LineParser, FakeApi, CSV}
+  alias __MODULE__.{
+    Status,
+    RowValidator,
+    RejectedRow,
+    RejectionReport,
+    Checkpoint,
+    Run,
+    FakeApi,
+    CSV
+  }
 
   defstruct(
     path: nil,
+    source_key: nil,
     files: [],
     timezone: :utc,
     error: nil,
     completed: MapSet.new(),
+    rejection_report: %RejectionReport{},
+    run_id: nil,
+    run_timezone: nil,
+    run_car_id: nil,
+    date_limit: nil,
+    date_limit_captured: false,
     car: nil,
     pids: %{},
     deps: %{}
@@ -24,18 +40,45 @@ defmodule TeslaMate.Import do
   alias __MODULE__, as: Data
 
   defmodule Status do
-    defstruct(state: :idle, message: nil, files: [])
+    defstruct(
+      state: :idle,
+      message: nil,
+      files: [],
+      rejected_rows: 0,
+      rejection_examples: [],
+      rejection_examples_truncated: false,
+      resume_timezone: nil
+    )
 
-    def into(state, %Data{files: files, completed: completed}) do
+    def into(
+          state,
+          %Data{
+            files: files,
+            completed: completed,
+            rejection_report: %RejectionReport{} = report,
+            run_timezone: run_timezone
+          }
+        ) do
       files =
-        Enum.map(files, fn %{date: date} = file ->
-          complete = MapSet.member?(completed, date)
-          Map.put(file, :complete, complete)
+        Enum.map(files, fn file ->
+          complete = MapSet.member?(completed, Checkpoint.file_id(file))
+
+          file
+          |> Map.drop([:fingerprint])
+          |> Map.put(:complete, complete)
         end)
 
+      status = %__MODULE__{
+        files: files,
+        rejected_rows: report.count,
+        rejection_examples: report.examples,
+        rejection_examples_truncated: RejectionReport.truncated?(report),
+        resume_timezone: run_timezone
+      }
+
       case state do
-        {:error, reason} -> %__MODULE__{state: :error, message: reason, files: files}
-        state when is_atom(state) -> %__MODULE__{state: state, files: files}
+        {:error, reason} -> %__MODULE__{status | state: :error, message: reason}
+        state when is_atom(state) -> %__MODULE__{status | state: state}
       end
     end
   end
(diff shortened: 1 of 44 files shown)


JakobLichterfeld (changes requested):
Thank you for this contribution — this is a well-engineered piece of work, and I appreciate how honest it is about its recovery semantics (file-level, at-least-once, clearly documented in the UI and docs). Special credit: along the way you fixed several **real pre-existing bugs**:

- Completion was keyed by parsed month/year, so two files mapping to the same month (`012018.csv` + `TeslaFi12018.csv`) were marked complete together. Now keyed by `{basename, fingerprint}` — correct, and covered by a test.
- `date_limit` is now captured once per run and persisted. Previously a resumed re-run would recompute the earliest state *after* partial data landed, silently truncating the rest. Good catch.
- Rejection dedup between the `create_car` scan pass and the import pass works correctly via the unique index plus the `:inserted`/`:existing` return.
- The privacy discipline (no raw values, no directory paths, hashed `source_key`) is consistently enforced and tested with sentinel values. Nice.

That said, there are two things I'd like resolved before merging, and a few strong suggestions.

## Blocking

**1. `abandon_run/1` is dead code — there is no escape hatch from a stuck run.**
Once a run is `:running`, the UI locks the time zone and the run only ever ends via `complete_run`. If a user picks the wrong time zone and restarts, or replaces the CSVs in the directory (fingerprints change → checkpoints invalidated → old saved TZ still forced), they're stuck until they do manual DB surgery. Before this PR, a restart reset everything. `Checkpoint.abandon_run/1` exists but is never called. Please add a "discard interrupted import" action (idle state only) that calls `abandon_run` and clears the resume state.

**2. `:ok = Import.run(tz)` in the LiveView will crash on the new error return.**
`prepare_run` introduces `{:error, reason}` as a possible reply on the `:idle` path, but `lib/teslamate_web/live/import_live/index.ex:57` still pattern-matches `:ok =`. Today the branch is nearly unreachable (`Repo.insert` of a plain struct raises on constraint violation instead of returning `{:error, changeset}`), but the code path was added without a consumer — either handle it in the LiveView or don't return it.

## Strong suggestions

**3. Validator/parser mismatch can reject rows that used to import: `climate_state.battery_heater` and `battery_heater_no_power`.**
`RowValidator` types both as `:boolean`, but neither key is in LineParser's `@boolean` coercion list. A TeslaFi variant exporting `battery_heater` as `0`/`1` would produce an integer → `valid_value?(:boolean, 0)` fails → the **entire row** is quarantined, where the old importer accepted it. Missing columns are fine (`nil` passes), so the blast radius is limited — but the fix is cheap: add both keys to `@boolean`.

**4. Reported row numbers drift past blank lines.**
`CSV.parse` drops `[""]` records in its `flat_map`, but `Stream.with_index(2)` runs downstream in `create_event_streams`. Every blank record shifts subsequent reported row numbers by one, which undermines the report's main purpose ("go look at row N in your file"). Index before filtering, or count blanks.

## Non-blocking observations

- **Every timestamp is parsed twice:** `RowValidator.parse` pre-checks via `LineParser.parse_timestamp/2`, then `LineParser.parse` re-parses `"Date"` (up to 3 `Timex.parse` attempts each). Measurable on multi-million-row exports; consider threading the parsed timestamp through.
- **Fingerprinting blocks the Import process at boot:** `build_files` hashes every CSV during `:read_directory`, so `Import.get_status()` (and the import page mount) blocks until hashing finishes — on every boot while the import dir is populated. Acceptable, but worth a comment.
- **Unbounded retention:** completed runs, checkpoints, and rejections live forever, and `restore_active_run` only surfaces `:running` runs — so a completed run's rejection report becomes invisible dead data at exactly the moment the docs tell users to "review the report before removing the import files". Consider pruning on `complete_run`, or surfacing the last completed run's report.
- **Untranslated error atoms:** `@status.message` renders via `inspect`, so users now see literal `:vehicle_data_incomplete` / `:vehicle_changed`. Since you already added `rejection_reason/1`, mapping these two would be consistent.
- **Style:** `checkpoint.ex` defines four top-level modules, `rejected_row.ex` two — convention here is one top-level module per file. `@max_examples 100` is duplicated in `Checkpoint` and `RejectionReport` (and hardcoded in the UI string).
- **Empty-directory edge:** `pending == []` now completes a run instantly with zero files (previously it raised). Probably fine, but "complete" with nothing imported is slightly misleading.
- **Gettext:** only `default.pot` is updated — please make sure the locale `.po` merges match what CI expects.

## Tests

Coverage is genuinely good (FIFO ordering, reject-only chunks, checkpoint restore, fingerprint invalidation, privacy sentinels, LiveView rendering). One gap: the old *"captures errors of the vehicle process"* test was replaced, so there's no longer a test for a genuine runtime failure inside the Vehicle process (e.g. a DB error mid-import). That propagation path still exists but is untested.

---

JakobLichterfeld (approved):
Thank you for the thorough revision — every point from the review was addressed, each with test coverage, and you went beyond it (FakeApi no longer strands blocked callers, `start_run` race is handled via `unique_constraint`, abort cleanup is bounded, and the VIN-conflict quarantine for VIN-less TeslaFi variants is a thoughtful touch). Approving.

Three optional follow-ups, none blocking: a dedicated rejection reason like `:vehicle_conflict` instead of `:invalid_fields, ["vin"]` would make the report less misleading (no migration needed, the column is a string); runs under a renamed source directory are never cleaned up (bounded, cosmetic); and `LineParser.parse/2` still crashes on an unvalidated date — a comment or removing it in favor of `parse/3` would prevent accidental use.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high)— sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5520, opened 2026-07-16
feat(vehicle): publish field provenance and trust metadata

magrathean-uk:
### Context

TeslaMate exposes useful values, but a consumer cannot always tell whether a field came from a fresh REST response, the streaming API, a restored database position, or a derived calculation. It also cannot distinguish unavailable data from a value that is stale or outside an expected range.

### What the suggestion does

- Keeps all existing summary and MQTT value payloads unchanged
- Tracks confidence as `exact`, `derived`, `estimated`, or `unknown`
- Tracks freshness as `fresh`, `stale`, or `unknown`
- Tracks availability as `available`, `corrupted`, or `unavailable`
- Identifies REST, stream, database, and TeslaMate-derived sources
- Includes observation time, age where requested, and a bounded reason
- Detects out-of-range percentages and route or vehicle coordinates
- Marks restored database values as stale rather than pretending they are current
- Adds `GET /api/car/:id/data-quality` for ungrouped field metadata
- Adds retained `teslamate/cars/$car_id/data_quality` MQTT metadata grouped by identical trust state

No field values are copied into either metadata payload. The HTTP response uses `Cache-Control: no-store`. The MQTT form rounds observation time to the minute and relies on existing publisher deduplication.

### Disclosure boundary

The HTTP endpoint follows the existing unauthenticated `/api/car` pattern. Although it contains no vehicle values, observation timestamps and field availability can reveal when TeslaMate last received particular kinds of data. The documentation calls this out and recommends access control when TeslaMate is reachable outside a trusted network.

The MQTT payload is rebuilt for each summary before deduplication. Network volume stays bounded, but busy streams do a small amount of extra CPU work.

### Verification

- `mix ci`: 370 passed
- Independent focused review: 43 passed
- Clone trial: 82 fields classified, retained payload 3,499 bytes
- Clone trial proved stale database, derived, and unavailable labels with ages and reasons
- Existing car, position, drive, charging, state, and update counts were unchanged
- Redaction tests prove vehicle values are absent from HTTP and MQTT metadata

### Suggested labels

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -5,6 +5,7 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
   import Core.Dependency, only: [call: 3]
 
   alias TeslaMate.Mqtt.Publisher
+  alias TeslaMate.Vehicles.Vehicle.DataQuality
   alias TeslaMate.Vehicles.Vehicle.Summary
   alias TeslaMate.Vehicles
 
@@ -66,6 +67,7 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
       |> add_car_latitude_longitude(summary)
       |> add_geofence(summary)
       |> add_active_route(summary)
+      |> add_data_quality(summary)
 
     last_values = publish_values(values, state)
     {:noreply, %{state | last_values: last_values}}
@@ -197,6 +199,13 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
       active_route_longitude: summary.active_route_longitude,
       active_route: active_route
     })
+  end
+
+  defp add_data_quality(map, %Summary{quality: quality}) do
+    case DataQuality.mqtt_payload(quality) do
+      nil -> map
+      payload -> Map.put(map, :data_quality, payload)
+    end
   end
 
   defp publish({key, value}, %State{car_id: car_id, namespace: namespace, deps: deps}) do
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -3,7 +3,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
   require Logger
 
-  alias __MODULE__.Summary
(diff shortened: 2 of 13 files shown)


JakobLichterfeld:
Thank you for the well-prepared PR — and more generally for your continued contributions to TeslaMate, which I value. As discussed before, I want to be upfront here rather than let this linger: I won't merge this one. Let me explain why, because the reasons are structural, not about the implementation quality (which is genuinely good — all `last_response` write paths are covered, redaction is tested, docs are included).

**1. Runtime cost on the hot path is a no-go for me.** `DataQuality.for_summary/4` iterates over ~80 fields on every summary broadcast — roughly once per second per car while driving — and the MQTT subscriber rebuilds and JSON-encodes the payload each time before deduplication kicks in. That is permanent, always-on CPU work for every installation, paid by all users, for metadata that nothing inside TeslaMate consumes. TeslaMate runs on a lot of small hardware (Raspberry Pis and the like), and I'm not willing to add per-tick overhead to the vehicle loop for a feature without an internal consumer.

**2. New unauthenticated API endpoint.** TeslaMate's HTTP API surface is intentionally minimal, and I'm generally opposed to adding new endpoints — especially unauthenticated ones. Your own description names the problem: `source: tesla_stream` plus second-precision `observed_at` reveals in near-real-time whether the car is currently driving, without exposing a single field value. That's a real disclosure surface, and "put a reverse proxy in front of it" in the docs doesn't resolve it. Authentication is a separate topic that needs to be solved project-wide first — not per-endpoint.

**3. No consumer inside TeslaMate.** Neither the web UI nor the Grafana dashboards use this metadata, so TeslaMate itself takes on all the costs: a `schema_version: 1` public contract we'd have to maintain indefinitely, a ~470-line module whose field taxonomy must be manually kept in sync with the `Summary` struct (any forgotten field silently degrades to `unknown`), and test-infrastructure special-casing (`PubSubMock` now has to ignore `quality` to keep deduplication working). Downstream projects can derive staleness/provenance heuristics from the existing MQTT topics (retained values, `state`, `since`, `healthy`, per-topic update timing) on their side.

If there is broader community demand for provenance metadata, the right starting point would be a discussion issue where scope, consumers, the auth question, and a zero-cost-when-unused design can be worked out first — not a finished implementation. As it stands, I'm closing this. Thanks again for the effort and for keeping the existing payloads untouched — I look forward to your next contributions.

PR #5514, opened 2026-07-15
fix(mqtt): return publish errors without crashing

magrathean-uk:
## Summary

Return MQTT publish failures to the caller without crashing the publisher process.

## Why

Temporary MQTT broker or TLS failures can otherwise destabilize the publishing path and leave incomplete drives instead of allowing the existing subscriber error reporting and retry behavior to respond.

## Compatibility

Successful QoS 0, 1, and 2 publishes are unchanged. Invalid QoS and timeout values are rejected before calling Tortoise. Known transient failures and unexpected client results become explicit errors, while genuine publisher bugs still propagate. Unacknowledged QoS refs expire within the existing publish timeout, and late, duplicate, or stray acknowledgements are ignored without stopping the publisher.

## Validation

- 13 focused publisher tests covering success, known failures, asynchronous acknowledgement errors, bounded ref expiry, invalid QoS/timeouts, unexpected exits, and stray acknowledgements
- 18 focused publisher and vehicle-subscriber tests passed
- Full `mix ci`: 374 tests passed
- Hosted Elixir tests, static analysis, formatting, security scans, CLA, and deploy preview passed

Closes #4944

PR #5513, opened 2026-07-15
fix(import): accept fractional TeslaFi battery levels

magrathean-uk:
## Summary

Round fractional TeslaFi `battery_level` and `usable_battery_level` values before Ecto casts them into TeslaMate's integer fields.

## Why

TeslaFi exports can contain fractional battery values. Those rows are otherwise valid but fail import because TeslaMate stores both battery-level fields as integers.

## Compatibility

Both fields reuse the existing generic numeric parser. Parsed floats are rounded; integer inputs, invalid strings, and existing empty-value handling remain unchanged.

## Validation

- Line-parser regression covers fractional and integer values for both battery fields
- Focused importer test passed
- Full combined `mix ci`: 375 tests passed
- `mix format --check-formatted` and staged secret scan

Closes #4477

PR #5525, opened 2026-07-18
feat(maintenance): add authenticated close actions and logs

### LINKED ISSUE #3682: Extend version with more build info

magrathean-uk:
## Summary

Adds one authenticated **Maintenance** page for two opt-in operator tasks:

- close eligible, long-running open drives and charging sessions without deleting records
- inspect a bounded rotating file-log tail for support

The page also shows validated image build identity so a support log can be tied back to the exact source, ref and revision.

This intentionally does **not** add a second health endpoint, a runtime collector, a duplicate data-health inbox or a top-level navbar item. The page is linked from Settings, and open-session queries run only when maintenance actions are explicitly enabled.

## Issue coverage

- Closes #3682 by injecting source, ref, revision and build time into published images and showing/logging that identity.
- Implements the close-drive and close-charge slice of #4850. Split, merge, delete and charge-cost actions remain out of scope.
- Implements the bounded rotating log and authenticated tail slice of #4891. Logging remains environment-controlled in this version, so this does not replace the UI-toggle work explored in #4898.

## Why these parts are together

The remaining pieces share one small operator workflow and one security boundary. Close actions and log access both need an explicitly enabled, authenticated Maintenance page; build identity makes the log output actionable during support. Keeping them together avoids separate routes, auth configuration and repeated UI/documentation, while the commits remain split by concern if maintainers want any part separated.

## Safety and privacy

- Both capabilities default to disabled.
- Enabling either requires a non-empty Basic Auth username and password at startup.
- Maintenance responses use `Cache-Control: no-store`.
- Candidate queries are bounded to the oldest 200 sessions and run only when actions are enabled.
- A close action locks and rechecks the exact record, requires a known inactive vehicle state, and uses existing TeslaMate completion logic.
- The action never deletes a record. Drives with insufficient position data remain unchanged.
- The active log is limited to 5 MB with three compressed rotations; the UI reads at most 256 KB or 500 lines.
- Known credentials, token collections, backseat tokens and VINs are redacted before persistence and again before display, including escaped structured log values.

## Verification

- `MIX_ENV=test mix ci`: 411 tests passed
- `mix dialyzer`: 0 errors
- Docusaurus production build passed
- Production Docker image built and ran as `nonroot:nonroot` on PostgreSQL 18
- Browser checked at 1440x900 and 390x844
- `/health_check`: 200
- `/maintenance` without credentials: 401 with Basic Auth challenge and `no-store`
- `/maintenance` with credentials: 200 with `no-store`
- Production log-tail check confirmed VIN and token redaction
- Production close-action check failed closed when vehicle activity was unavailable and left the disposable database row unchanged

All database and browser verification used disposable local containers.


JakobLichterfeld:
Hi Bolyki, thank you for the work that went into this — and for proactively narrowing the scope after my notes on #5518 and #5520. The craftsmanship is genuinely good: the fail-closed safety model, the lock-and-recheck transaction, and the test coverage are all well done.

That said, I won't review this as a single unit, and I want to be transparent about why.

**1. This bundles too much.** Build identity (#3682) has nothing to do with maintenance actions or log access — a shared auth path is not enough reason to couple three features. Even with clean commits, a 49-file PR spanning CI, Dockerfile, a new auth plug, a new LiveView, a file logger and a redaction engine is very hard to review and even harder to maintain.

As a side note on #3682: our images already carry build identity twice — the OCI labels from `docker/metadata-action` (`org.opencontainers.image.revision` etc.), and since #5380 signed SLSA provenance that cryptographically ties every published digest to repo, ref and commit (`gh attestation verify oci://teslamate/teslamate:latest --repo teslamate-org/teslamate`). Injecting a third, self-reported copy via build-args and ENV goes in the opposite direction. If the goal is showing identity in the UI, that needs to start from what the image already carries — I'll leave notes in the issue.

**2. Direction before implementation.** As I said in #5452: “I'd prefer that we first establish the need for a feature and then discuss the best implementation, rather than iterating on the implementation before we've agreed that the feature itself solves a recurring problem.” For #4891 the need is established — I opened that issue and outlined the requirements there. But this implementation makes decisions that were not part of that outline (mandatory Basic Auth, no UI toggle for file logging), and those are not up for negotiation in a PR thread. Please wait for direction in the issue before implementing; arriving with finished implementations puts me in the position of either accepting decisions I wasn't part of, or rejecting finished work — neither is good for you or for me.

**3. The mandatory Basic Auth coupling is a blocker.** Many users already secure TeslaMate properly at the deployment level — reverse proxy with Authelia or OIDC, mTLS, VPN — which is the documented model for TeslaMate. Refusing to start unless a second, TeslaMate-specific Basic Auth credential pair is configured punishes exactly the users who have set up better authentication: for them this is not a safety feature but friction and a second credential store to manage. And since the rest of the UI (settings, geofences, token entry) remains unprotected, the boundary doesn't provide a real security gain either. Deployment-level protection has to remain a first-class option. I won't debate the security architecture further in this PR — the expected model will be spelled out in the issues.

**How to proceed:**
- I'm closing this PR. This is about shape and process, not about the quality of the work.
- #3682: I'll comment there on the label/provenance-based direction. A small, stand-alone PR along those lines would be welcome afterwards.
- #4891 / #4850: I'll clarify the expected auth and enable/disable model in the issues. The bounded tail, rotation and pure-Elixir parts of this branch, and the close-action core (lock, recheck, never delete), match what I outlined and can be carried over once the direction is settled.

One more general note: you've been contributing at a very high pace over the last weeks, and some of it has been genuinely valuable (#5512, #5513, #5514, #5475). But TeslaMate is maintained by volunteers, and review capacity is our scarcest resource. Several small, issue-aligned PRs — coordinated with the work that already exists — will always move faster here than large pre-built solutions. Thanks for understanding.

PR #5526, opened 2026-07-19
feat(grafana): make data health summary actionable

magrathean-uk:
## Summary

- rename the existing `Incomplete Data` panel to `Data Health`
- show zero counts in green and nonzero counts in red
- link each count directly to the existing incomplete Drives or Charges detail panel
- preserve the selected car when following either link
- clarify that an open record may still be an active session

## Why

This follows the discussion after #5517. @brianmay suggested having one place to check data-health issues, and @JakobLichterfeld expected the aggregate information in the Database Information dashboard.

The aggregate counts already live there, while the Drives and Charges dashboards already own the full record details. This makes the existing summary actionable without adding another web route, navbar entry, translation surface, or duplicate detail query.

## Validation

- parsed every Grafana dashboard with `jq`
- built and provisioned TeslaMate's Grafana 13.0.1 image
- checked zero and nonzero states against a disposable PostgreSQL 18 database
- verified both links open the intended detail panel with the selected car
- checked the normal compact dashboard layout and expanded panel at desktop and mobile widths
- `mix deps.unlock --check-unused`

### DIFF
--- a/grafana/dashboards/database-info.json
+++ b/grafana/dashboards/database-info.json
@@ -357,7 +357,7 @@
                 "value": 0
               },
               {
-                "color": "red",
+                "color": "orange",
                 "value": 1
               }
             ]
@@ -485,7 +485,7 @@
           }
         }
       ],
-      "title": "Data Health",
+      "title": "Incomplete Data",
       "type": "stat"
     },
     {
--- a/grafana/dashboards/database-info.json
+++ b/grafana/dashboards/database-info.json
@@ -342,7 +342,7 @@
         "type": "grafana-postgresql-datasource",
         "uid": "TeslaMate"
       },
-      "description": "This means you have some **Drives** or **Charges** not closed.\nIf so, you may follow the official guide: \n<a href='https://docs.teslamate.org/docs/maintenance/manually_fixing_data' target='_blank'>Manually fixing data</a>",
+      "description": "Open **Drives** or **Charges** can be active sessions and are not necessarily incomplete. Select a count to review the matching records.\nIf a record is incomplete, follow the official guide: \n<a href='https://docs.teslamate.org/docs/maintenance/manually_fixing_data' target='_blank'>Manually fixing data</a>",
       "fieldConfig": {
         "defaults": {
           "color": {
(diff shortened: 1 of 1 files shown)


grafana/dashboards/database-info.json
```diff
@@ -444,7 +485,7 @@
           }
         }
       ],
-      "title": "Incomplete Data",
+      "title": "Data Health",
```
JakobLichterfeld: The previous title was correct, this is misleading. The data can be unhealthy: e.g. wrong elevation, missing drives, lost connection etc.

JakobLichterfeld (changes requested):
Thanks @magrathean-uk for following up on the discussion in #5517, and for the unusually thorough validation notes — much appreciated!

I like the approach: making the existing counts actionable via data links to the detail panels we already have, instead of adding a new route or dashboard. The links follow the same pattern we use in the Statistics dashboard, and the clarified panel description is a genuine improvement.

Two remarks:

1. As noted inline, please keep the panel title **"Incomplete Data"**. "Data Health" promises more than the panel delivers — unhealthy data can also mean wrong elevation, missing drives, lost connection etc., while this panel only counts unclosed drives and charges.
2. Regarding the thresholds: during an active drive or charge the count is naturally ≥ 1, so the panel would show **red in perfectly normal operation** — which contradicts the (correct) new description saying an open record can be an active session. I'd suggest **orange** for the nonzero state, so it reads as "worth a look" rather than "error".

Happy to merge once these are addressed.

brianmay:
If I understand this AI report correctly there might be issues if you have multiple cars, and select "All". Not 100% sure here.

> Expanding on point 2 (multi-car selection edge case):
> 
> **The mechanism.** Grafana dashboard template variables support multi-value selection. When a user picks "All" for `car_id` on the source dashboard (the one with the new data-links), the `$car_id` variable expands to every car ID. Clicking a data-link then builds a URL like:
> 
> ```
> /d/TSmNYvRRk/charges?viewPanel=17&var-car_id=1&var-car_id=2&var-car_id=3
> ```
> 
> Grafana accepts repeated `var-<name>` params and sets the target dashboard's `car_id` variable to a multi-value containing all of them. So far, so good.
> 
> **Where it gets fuzzy.** The target panel 17 ("Incomplete Charges 🪫") is a table/list of individual charge sessions. With multiple cars selected in the target dashboard:
> - If the panel's query uses `${car_id}` with the default `=` operator and `TPL_EID`/a multi-value-aware macro, it works as `car_id IN (...)` and shows rows for all selected cars.
> - If the query uses `$car_id` in a context expecting a scalar (e.g., a single-car filter or a panel title substitution `${car_id:text}`), you can get unexpected SQL or a misleading title like "All" rather than a car name.
> 
> **The pre-existing part.** This multi-value propagation is just how Grafana template variables work across cross-dashboard links with `var-<name>=$<name>`. It long predates this PR — the same `var-car_id=$car_id` pattern is used by other data-links in the repo, so the PR didn't introduce the behavior. Whatever the target dashboards do with multi-value `car_id` today, they already do when reached from any existing link.
> 
> **Why "viewPanel only works correctly when at least one car matches."** `viewPanel=17` just tells Grafana to expand that panel on load instead of showing the whole dashboard. If the user had "All" on the source and the target query returns no rows for the union of cars (e.g., the link is clicked from an aggregated row that doesn't correspond to the selected cars' actual data), the panel expands but appears empty. That's confusing but not a defect introduced here — it's inherent to drilling from an aggregated view into a per-session list with a different filter scope.
> 
> **Net.** Not something this PR needs to fix. If the maintainers want robustness, the usual Grafana idioms are: (a) force single-value selection on the source `car_id` variable, or (b) omit the `var-car_id` link param when "All" is selected (requires a custom link with `${__url` interpolation), or (c) ensure target panel queries use multi-value-safe macros (`${car_id:multi}`). All are outside the scope of PR 5526.

JakobLichterfeld:
> If I understand this AI report correctly there might be issues if you have multiple cars, and select "All". Not 100% sure here.

Good instinct to double-check, but this scenario can't occur in TeslaMate: the `car_id` variable is a single-select without an "All" option in all three involved dashboards — Database Information (the link source) as well as Drives and Charges (the link targets) all have `"includeAll": false` and no `"multi": true`. So `$car_id` always interpolates to exactly one ID, and the link always resolves to `var-car_id=<single id>`. The multi-value expansion the report describes simply has no way to happen here.

The only "empty panel" case that actually exists is clicking a link while the count is 0 — and then an empty table is the correct answer.

So nothing to address in this PR on that front.

brianmay (approved):

JakobLichterfeld (approved):

PR #5531, opened 2026-07-20
Revert "feat(import): continue past malformed rows and resume completed files"

JakobLichterfeld:
Reverts teslamate-org/teslamate#5519 as it breaks CI tests (test rounds fractional battery levels from TeslaFi exports (TeslaMate.Import.LineParserTest)


brianmay:
Wondering why error was not apparent until after it was merged. Intermittent test maybe?

JakobLichterfeld:
Yeah, wondering as well, but it is reproducible.

PR #5535, opened 2026-07-21
fix: reconnect stream when a drive resumes after mid-drive offline phase to avoid missing elevation 

### LINKED ISSUE #5534: Stream is not reconnected when a drive resumes after a mid-drive offline phase (rest of drive falls back to polling, missing elevation)
### Summary

When the vehicle is treated as offline while driving (e.g. after repeated Owner API 408 timeouts tripping the circuit breaker) and comes back online mid-drive, TeslaMate resumes the drive — but the streaming websocket is not re-established. The rest of the drive is recorded via REST polling only, at the driving interval (default 2.5 s). Since `vehicle_data` provides no altitude and `elevation` is only set on streamed positions, the remainder of the drive has no elevation data. Only when the state machine passes through `Start / :online` again (e.g. after a charging stop) is the stream reconnected.

Reported in #5532 (thanks @accolon) (v4.0.1, Owner API), including DB analysis and logs. The code path is unchanged on current `main`.

### Root cause

- When a fetch result reports the vehicle as offline/asleep, the stream is disconnected and `stream_pid` is set to `nil`: [vehicle.ex#L429-L445](https://github.com/teslamate-org/teslamate/blob/90b86085/lib/teslamate/vehicles/vehicle.ex#L429-L445)
- When the vehicle comes back online mid-drive, the drive resumes into `{:driving, :available, drv}` without reconnecting the stream: [vehicle.ex#L1257-L1259](https://github.com/teslamate-org/teslamate/blob/90b86085/lib/teslamate/vehicles/vehicle.ex#L1257-L1259) (same for the `{:driving, {:unavailable, _}, drv}` path: [vehicle.ex#L1272-L1282](https://github.com/teslamate-org/teslamate/blob/90b86085/lib/teslamate/vehicles/vehicle.ex#L1272-L1282))
- With `streaming?/1` now false, the driving handler polls at `driving_interval()` (default 2.5 s): [vehicle.ex#L1290](https://github.com/teslamate-org/teslamate/blob/90b86085/lib/teslamate/vehicles/vehicle.ex#L1290)
- Polled positions carry no elevation when the streaming API is enabled: [vehicle.ex#L1604-L1610](https://github.com/teslamate-org/teslamate/blob/90b86085/lib/teslamate/vehicles/vehicle.ex#L1604-L1610)
- `connect_stream/1` is currently only called from the asleep/offline pre-online probing, `Start / :online`, on settings changes, and on stream-internal reconnects (`:too_many_disconnects` / `:tokens_expired`) — none of which are reached while a drive is in progress.

### Expected behavior

When a drive resumes after an offline/unavailable phase and `use_streaming_api` is enabled with `stream_pid: nil`, the stream should be reconnected so the rest of the drive is recorded with streamed positions (including elevation) again.

### Proposed fix

Call `connect_stream/1` when transitioning from `{:driving, {:offline, last}, drv}` (and `{:driving, {:unavailable, n}, drv}`, if the stream process is no longer alive) back to `{:driving, :available, drv}`, mirroring the reconnect logic in `Start / :online`.

### References

- #5532 (report with logs and DB analysis)
- #5438 (related observation: positions with `ideal_battery_range_km` have no elevation)

🤖 Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
When the vehicle went offline mid-drive, the streaming websocket was not re-established on resume. The rest of the drive was recorded via REST polling only, whose positions carry no elevation. Reconnect the stream when a drive resumes after an offline or unavailable phase.

fixes: #5534

🤖 PR drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - fix(cars): enforce non-null VINs (#5512 - @magrathean-uk)
 - fix(mqtt): return publish errors without crashing (#5514 - @magrathean-uk)
 - fix(geofences): increase cost precision (#5508 - @magrathean-uk)
+- fix: reconnect stream when a drive resumes after mid-drive offline phase to avoid missing elevation (#5535 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
@@ -33,7 +34,7 @@
 - build(deps): update flake.lock (#5427)
 - build(deps): bump webpack-dev-server from 5.2.4 to 5.2.5 in /website (#5445)
 - chore: add .codegraph to .gitignore (#5440- @JakobLichterfeld)
-- ci: speed up check_linting by running treefmt in a lean app (#5440- @JakobLichterfeld)
+- ci: speed up check_linting by running treefmt in a lean app (#5440 - @JakobLichterfeld)
 - sec(deps): add ws override to version 8.21.0 in /website (#5446 - @JakobLichterfeld)
 - build(deps-dev): bump esbuild from 0.28.0 to 0.28.1 in /assets (#5444)
 - sec(deps): add joi override to version 17.13.4 in /website (#5448 - @JakobLichterfeld)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@
 - build(deps): bump launch-editor from 2.13.2 to 2.14.1 in /website (#5426)
 - build(deps): update flake.lock (#5427)
 - build(deps): bump webpack-dev-server from 5.2.4 to 5.2.5 in /website (#5445)
-- chore: add .codegraph to .gitignore (#5440- @JakobLichterfeld)
+- chore: add .codegraph to .gitignore (#5440 - @JakobLichterfeld)
 - ci: speed up check_linting by running treefmt in a lean app (#5440 - @JakobLichterfeld)
 - sec(deps): add ws override to version 8.21.0 in /website (#5446 - @JakobLichterfeld)
 - build(deps-dev): bump esbuild from 0.28.0 to 0.28.1 in /assets (#5444)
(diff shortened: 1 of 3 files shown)


brianmay:
Some AI styling nitpicks. Not sure I 100% agree with these points, but for completeness.

> 1. **CHANGELOG whitespace fix** (`( 5440- …)` → `( 5440 - …)`) is unrelated to the PR's purpose — ideally a separate commit.
> 2. The `maybe_reconnect_stream/1` wildcard clause `{%CarSettings{}, _} -> data` collapses two cases; could be split for symmetry but reads fine as-is.

Am surprised that I haven't seen this issue myself in recent drives. But I guess it isn't often the car goes offline, which is kind of incredible around this mountainous area known for having poor Telstra coverage where I live.

brianmay (approved):

JakobLichterfeld:
Thanks for taking a look.

1. The whitespace fix touches a CHANGELOG entry of an unrelated, already-merged PR (#5440), so there is no natural home for it: a dedicated PR for a single character is pure overhead, pushing straight to main is unclean, and since we squash-merge, a separate commit inside this PR wouldn't survive the merge anyway. A drive-by in a PR that touches the CHANGELOG anyway is the cleanest available option. While at it: the line directly above has the same typo, fixed as well.
2. Keeping the wildcard as-is: it mirrors the existing catch-all in the `%CarSettings{}` settings handler in the same module, and splitting it would add a clause with no behavioral difference.

Regarding coverage: a dead zone does mean the API can't reach the car — `vehicle_data` then returns 408s. But TeslaMate absorbs short gaps: those errors keep the current state and simply retry, and on "vehicle unavailable" the fetch falls back to the summary endpoint, which keeps reporting "online" until Tesla's backend flags the car offline after a sustained loss of connectivity. So the offline-while-driving transition (and with it this bug) needs a dead zone in the order of minutes, not seconds. And even then the drive keeps recording via the polling fallback — the only visible symptom is missing elevation and coarser resolution, which is easy to overlook. If you want to check, grep your logs for "Vehicle came back online after" or look for drives whose later positions all lack `elevation`.

PR #5549, opened 2026-07-27
fix(mqtt): avoid blocking startup on retained cleanup

### LINKED ISSUE #5523: MQTT: avoid blocking app startup when the broker is unreachable
### Context

Follow-up to #5514 / #4944. Since #5514, an unreachable MQTT broker no longer crashes TeslaMate — publish failures are returned as errors and logged as warnings. 

However, the publish path is still synchronous: `Tortoise311.publish/4` blocks inside `Publisher.handle_call/3` for up to ~9.5 s while waiting for a connection. Each `VehicleSubscriber` publishes during `init/1` (`clear_retained/3` clears the retained `healthy` topic), and subscribers start sequentially — so with an unreachable broker, app boot is delayed by roughly 10 seconds **per car** before logging begins.

### Expected behavior

App startup time should not depend on MQTT broker availability. Vehicle logging should start immediately even if the broker is down.

### Possible approaches

- Make `clear_retained/3` asynchronous (e.g. `cast` + publish from the subscriber's `handle_continue`), so `init/1` no longer blocks on the broker
- Alternatively, use a short timeout for the startup publishes only

### Scope

No behavior change for a healthy broker; retained-message cleanup semantics from #4817 must be preserved.

🤖Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

magrathean-uk:
## Summary

TeslaMate clears stale retained messages when each vehicle subscriber starts. If a broker accepts the connection but does not complete the publish, the old initialization path waits for the publisher's roughly 10-second call timeout per car and delays application startup.

This moves cleanup into `handle_continue/2`, so `start_link/1` returns before a slow publish completes. GenServer continuations run before queued mailbox messages, which preserves cleanup before the first vehicle summary is published.

Cleanup publish failures are now logged with their topic and error term instead of being silently discarded.

## Validation

- Proves `start_link/1` returns while retained cleanup is blocked
- Proves the retained `healthy` clear is published before a queued `healthy: true` summary
- Covers failed-cleanup logging and continued summary processing
- `mix ci` on Elixir 1.19.5 / OTP 28: 431 tests, 0 failures

Closes #5523

## Suggested labels

`fix`, `area:teslamate`

### DIFF
--- a/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
+++ b/lib/teslamate/mqtt/pubsub/vehicle_subscriber.ex
@@ -34,7 +34,13 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
         |> Enum.reject(&is_nil(&1))
         |> Enum.join("/")
 
-      call(publisher, :publish, [topic, "", [retain: true, qos: 1]])
+      case call(publisher, :publish, [topic, "", [retain: true, qos: 1]]) do
+        :ok ->
+          :ok
+
+        error ->
+          Logger.warning("MQTT retained cleanup failed for #{topic}: #{inspect(error)}")
+      end
     end)
   end
 
@@ -49,9 +55,17 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriber do
     }
 
     :ok = call(deps.vehicles, :subscribe_to_summary, [car_id])
-    :ok = clear_retained(car_id, namespace, deps.publisher)
 
-    {:ok, %State{car_id: car_id, namespace: namespace, deps: deps}}
+    {:ok, %State{car_id: car_id, namespace: namespace, deps: deps}, {:continue, :clear_retained}}
+  end
+
+  @impl true
+  def handle_continue(
+        :clear_retained,
+        %State{car_id: car_id, namespace: namespace, deps: deps} = state
+      ) do
+    clear_retained(car_id, namespace, deps.publisher)
+    {:noreply, state}
   end
 
   @publish_if_nil ~w(charge_energy_added charger_actual_current charger_phases
--- a/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
+++ b/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
@@ -66,9 +66,9 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriberTest do
     send(subscriber_pid, :continue)
 
     assert_receive {BlockingPublisher, {:publish, "teslamate/cars/0/healthy", message, opts},
-                    ^subscriber_pid}
+                    from_pid}
 
-    assert {message, opts} == {"", [retain: true, qos: 1]}
+    assert {message, opts, from_pid} == {"", [retain: true, qos: 1], subscriber_pid}
 
     assert_receive {BlockingPublisher,
                     {:publish, "teslamate/cars/0/healthy", "true", [retain: false, qos: 1]},
--- a/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
+++ b/test/teslamate/mqtt/pubsub/vehicle_subscriber_test.exs
@@ -5,6 +5,20 @@ defmodule TeslaMate.Mqtt.PubSub.VehicleSubscriberTest do
   alias TeslaMate.Vehicles.Vehicle.Summary
   alias TeslaMate.Locations.GeoFence
 
+  defmodule BlockingPublisher do
+    def publish(test_pid, topic, message, opts) do
+      send(test_pid, {__MODULE__, {:publish, topic, message, opts}, self()})
+
+      if String.ends_with?(topic, "/healthy") and message == "" do
+        receive do
+          :continue -> :ok
+        end
+      else
+        :ok
+      end
+    end
+  end
+
   defp start_subscriber(name, car_id, namespace \\ nil, publisher_responses \\ %{}) do
     publisher_name = :"mqtt_publisher_#{name}"
     vehicles_name = :"vehicles_#{name}"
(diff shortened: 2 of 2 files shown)


JakobLichterfeld (changes requested):
Thanks for picking this up — this is exactly the approach #5523 sketched, and the important detail is right: `{:continue, :clear_retained}` is guaranteed to run before anything already in the mailbox, so the retained-cleanup-before-first-summary semantics from #4817 still hold. That's why `handle_continue` is the correct primitive here and a `cast` or `send(self(), …)` would not have been. It also matches the idiom we already use in `TeslaMate.Updater`. 👍

Two things I'd like addressed before merge:

**1. Log failed cleanup publishes.** `:ok = clear_retained(...)` is a tautology today — `Enum.each/2` always returns `:ok` and the publish result is discarded. That means with an unreachable broker the cleanup fails silently and never retries, so the stale retained `healthy` message from pre-#4817 installs survives. That was already true before this PR, but since this change is specifically about the unreachable-broker path, it's the right moment to fix it — and it matches how `publish_values/2` already handles publish errors:

```elixir
defp clear_retained(car_id, namespace, publisher) do
  Enum.each(@do_not_retain, fn key ->
    topic =
      ["teslamate", namespace, "cars", car_id, key]
      |> Enum.reject(&is_nil(&1))
      |> Enum.join("/")

    case call(publisher, :publish, [topic, "", [retain: true, qos: 1]]) do
      :ok -> :ok
      error -> Logger.warning("MQTT retained cleanup failed for #{topic}: #{inspect(error)}")
    end
  end)
end
```

And then drop the misleading `:ok =` match in `handle_continue/2`.

**2. Please guard the ordering invariant in the test.** The new test proves `start_link/1` returns while the publish is blocked — good — but nothing pins down that the cleanup still happens *before* the first summary is published. That's the property that would break silently if someone later "simplifies" the `{:continue, ...}` into a cast. While the publisher is still blocked, send the subscriber a `%Summary{healthy: true}`, then `:continue`, and assert the empty-payload `healthy` clear arrives before the `"true"` publish.

Minor, no need to hold the PR on these:

- `Task.yield(task, 500)` is a bit tight for `async: true` CI — `2_000` costs nothing when green and avoids a flaky failure.
- `GenServer.stop(subscriber_pid)` defaults to `:infinity`, and the subscriber isn't under `start_supervised`, so a lost `:continue` would hang the suite rather than fail it. `GenServer.stop(subscriber_pid, :normal, 1_000)` is cheap insurance.
- Small correction for the PR description (I use it for the changelog): the old behaviour wasn't an indefinite block — `Publisher.publish/3` has a 10 s call timeout and catches the exit, so the symptom was the ~10 s-per-car boot delay described in the issue. And "covered successful retained cleanup after startup" was already covered by the existing `publishes vehicle data` test; this PR adds one new test.

I'll label this `fix` rather than `enhancement` and drop `note:discussion` once the above is in.

JakobLichterfeld (changes requested):
Thanks for the quick turnaround — the logging change, the `2_000` yield and the bounded `GenServer.stop/3` are exactly right, and making `BlockingPublisher` block only on the empty `/healthy` clear was the correct call, since `publish_values/2` fans the summary publishes out through `Task.async_stream`. The failure test is wired up properly too: `MqttPublisherMock` returns the first queued response and falls back to `:ok`, so the clear fails and the summary publish succeeds, which is what we want.

One thing still open, and it's the one I actually care about: **the ordering test doesn't test ordering.**

`assert_receive` is itself a selective receive. Your two patterns are disjoint — `""` / `retain: true` versus `"true"` / `retain: false` — so each one matches its message regardless of where it sits in the mailbox. I sent the two messages in the *wrong* order and ran your assertions against them verbatim: 1 test, 0 failures. So if someone later replaces the continuation with something that publishes the clear after the first summary, this test stays green.

To make it order-sensitive, the first `assert_receive` needs a pattern that matches *both* healthy publishes, so it is forced to take whichever arrived first, and then assert that it was the clear:

```elixir
assert_receive {BlockingPublisher, {:publish, "teslamate/cars/0/healthy", message, opts}, _pid}
assert {message, opts} == {"", [retain: true, qos: 1]}

assert_receive {BlockingPublisher,
                {:publish, "teslamate/cars/0/healthy", "true", [retain: false, qos: 1]}, _pid}
```

The setup around it is already correct — sending `%Summary{healthy: true}` before `:continue` works precisely because the `receive do :continue` inside `handle_continue/2` skips the queued summary. Only the assertions need to change.

Also, small wording fix in the PR description: "redacted error term" isn't accurate, nothing is redacted — it's `inspect(error)` on the publish result. Please just say the topic and the error term are logged, so the changelog entry doesn't promise something the code doesn't do.

With those two, this is good to merge.

JakobLichterfeld (changes requested):
Almost — the production code is done from my side, but the ordering assertion is still order-blind, for a slightly different reason than last time.

You kept the `^subscriber_pid` pin on the first `assert_receive`. The retained clear is published from the subscriber process, but the summary publishes go through `Task.async_stream/3` in `publish_values/2`, so they arrive tagged with a worker pid — that's exactly why the second assertion needs `_publisher_pid`. With the pin in place, the first pattern therefore still matches *only* the clear and skips the `"true"` message wherever it sits in the mailbox. Widening `message`/`opts` didn't help, because the pid is doing the filtering now.

I re-checked it the same way as before, reversing the delivery order and tagging the two messages with different pids: your current assertions pass, the wildcard version fails. So dropping the pin is what actually makes it bite. You can keep the origin check by folding the pid into the same comparison:

```elixir
assert_receive {BlockingPublisher, {:publish, "teslamate/cars/0/healthy", message, opts}, from_pid}
assert {message, opts, from_pid} == {"", [retain: true, qos: 1], subscriber_pid}

assert_receive {BlockingPublisher,
                {:publish, "teslamate/cars/0/healthy", "true", [retain: false, qos: 1]},
                _publisher_pid}
```

That asserts all three properties at once: the clear is the first `healthy` publish to arrive, it carries the retained empty payload, and it came from the subscriber process rather than a publish worker.

Everything else is settled — the logging, the `2_000` yield, the bounded stop, the failure test and the corrected description all look good. This is the last change I need.

JakobLichterfeld (approved):
That's it — the pin is gone and `from_pid` is now part of the comparison, so the first `assert_receive` matches either `healthy` publish and is forced to take whichever arrives first. Reversed delivery now fails, which is exactly what this test needed to be worth having.

Production side has been good since the second round: the continuation preserves the #4817 cleanup-before-first-summary ordering, and cleanup failures are logged instead of silently dropped.

Thanks for the patience through three rounds on that one assertion — the selective-receive semantics are genuinely easy to get wrong, and the test is now a real regression guard rather than a green light. Approving and merging.

PR #5550, opened 2026-07-27
fix(nix): support configured database credentials in maintenance tools

### LINKED ISSUE #5533: fix(nix): maintenance and backup scripts fail with non-default postgres.user due to peer auth
### What happened?

The NixOS maintenance scripts `teslamate-delete-drive` and `teslamate-delete-charge` (`nix/maintenance.nix`) connect with:

```
sudo -u teslamate psql -U ${databaseUser} -d ${databaseName}
```

With a local PostgreSQL, this relies on peer authentication: the OS user (`teslamate`, hardcoded) must match the database role. If `services.teslamate.postgres.user` is set to anything other than `"teslamate"`, peer auth rejects the connection and the scripts fail. It also means the scripts cannot work against an external database server at all.

This is pre-existing behavior, but since #5337 the module explicitly distinguishes between the default role name (socket + peer auth) and other role names (TCP + password fallback), which makes the asymmetry more visible: TeslaMate itself handles a non-default role name fine, the delete scripts do not. Noted during review in #5337 and deliberately kept out of that PR to avoid scope creep.

The `closeDrive`/`closeCharge` scripts are not affected: they go through the running Elixir release (`teslamate rpc`, authenticated via `RELEASE_COOKIE`), not psql.

The same limitation applies to `teslamate-backup` and `teslamate-restore` (`nix/backup_and_restore.nix`), which use `sudo -u teslamate pg_dump/psql -U ${databaseUser}`.
The restore script's psql invocations additionally run without `-v ON_ERROR_STOP=1`, so individual failing statements (e.g. `CREATE EXTENSION`) do not abort the script.

### Preferred solution: route deletes through the release, like the close scripts

The delete scripts only do an existence check plus `DELETE FROM drives/charging_processes WHERE id = X`. That maps directly onto `teslamate rpc` with `TeslaMate.Repo`, e.g.:

```
teslamate rpc "TeslaMate.Repo.get!(TeslaMate.Log.Drive, ID) |> TeslaMate.Repo.delete!()"
```

This reuses the app's own database configuration (socket or TCP, any role name, external servers, no superuser), removes the psql/peer-auth dependency entirely, and makes all four maintenance scripts consistent. `Repo.get!` also gives the existence check for free. Ideally the delete goes through proper functions in `TeslaMate.Log` rather than raw `Repo` calls in the rpc string.

Trade-off: requires the TeslaMate service to be running (like the close scripts already do), whereas psql only needed the database.

### Alternatives considered

1. `sudo -u postgres psql`: works for any role name, but keeps the scripts local-only and switches to superuser semantics.
2. TCP + `PGPASSWORD` via the `loadFromEnvFile` helper from #5337: covers external servers, but puts the password into the psql process environment.
3. Document the limitation only.

### Expected Behavior

Maintenance scripts work regardless of the configured `services.teslamate.postgres.user`, including against external database servers.

🤖 Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

magrathean-uk:
## Summary

The Nix maintenance scripts currently assume default local PostgreSQL peer authentication. That breaks custom `postgres.user` values and external database configurations.

This update:

- Passes the configured host, port, database, and role to PostgreSQL tools
- Loads the password from `secretsFile` into the child `PGPASSWORD` environment without putting it in command arguments
- Uses the running TeslaMate release and existing Log APIs for drive and charge deletion
- Enables `ON_ERROR_STOP` for restore failures
- Validates the restore file before stopping TeslaMate
- Restarts the service after failures while preserving the original failure status

## Validation

- Focused Log tests: 42 tests, 0 failures
- Nix backup and restore check: passed
- Nix maintenance check: passed
- Nix formatting and ShellCheck: passed
- Full branch and clean `origin/main` runs both reproduced the same three existing `SuspendTest` timing failures in the Docker and macOS test setup

Closes #5533

## Suggested labels

`nix`, `enhancement`

### DIFF
--- a/lib/teslamate/log.ex
+++ b/lib/teslamate/log.ex
@@ -240,6 +240,13 @@ defmodule TeslaMate.Log do
     |> Repo.insert()
   end
 
+  def delete_drive(id) when is_integer(id) do
+    case Repo.get(Drive, id) do
+      nil -> :not_found
+      drive -> delete(drive)
+    end
+  end
+
   def close_drive(%Drive{id: id} = drive, opts \\ []) do
     drive = Repo.preload(drive, [:car])
 
@@ -405,6 +412,13 @@ defmodule TeslaMate.Log do
     charge
     |> ChargingProcess.changeset(attrs)
     |> Repo.update()
+  end
+
+  def delete_charging_process(id) when is_integer(id) do
+    case Repo.get(ChargingProcess, id) do
+      nil -> :not_found
+      charging_process -> delete(charging_process)
+    end
   end
 
   def start_charging_process(%Car{id: id}, %{latitude: _, longitude: _} = attrs, opts \\ []) do
@@ -707,5 +721,12 @@ defmodule TeslaMate.Log do
     %Update{car_id: id}
     |> Update.changeset(%{start_date: date, end_date: date, version: version})
     |> Repo.insert()
+  end
+
+  defp delete(struct) do
+    case Repo.delete(struct) do
+      {:ok, _struct} -> :ok
+      {:error, _changeset} = error -> error
+    end
   end
 end
--- a/nix/backup_and_restore.nix
+++ b/nix/backup_and_restore.nix
@@ -1,39 +1,118 @@
 {
   stdenv,
   lib,
-  pkgs,
+  coreutils,
+  gnused,
+  postgresql,
+  systemd,
   writeShellScript,
   databaseUser,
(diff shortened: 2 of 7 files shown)


JakobLichterfeld:
Ty, as the issue is assigned, no PRs are welcome

PR #5551, opened 2026-07-27
fix(vehicle): identify base Model 3 from model year 2022 as RWD instead of SR+

### LINKED ISSUE #3887: RWD shows up as SR+ despite SR+ not being sold for 2.5 years

### What happened?

Any base model Model 3 shows up as SR+ despite different battery sizes, model years, and feature differences.

### Expected Behavior

2022 and 2023 base model Model 3s should show up as RWD instead of SR+

### Steps To Reproduce

Home Page
Car Model Name

### Relevant log output

```shell
N/A
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

Docker

### Version

1.28.5

magrathean-uk:
## Summary

Recent base Model 3 vehicles still report trim code `50`, but the display name changed from SR+ to RWD from model year 2022.

This decodes position 10 of a complete VIN through a dedicated model-year helper, resolves valid codes against the 2010 to 2039 cycle, and keeps the 2022 cutoff visible at the Model 3 trim decision. Model year 2021 and older, plus missing, malformed, or invalid VINs, retain the existing SR+ fallback.

The code documents that model year is only a proxy for Tesla's naming change. It also keeps the VIN-year helper reusable for other model identification rules.

## Validation

- Covers older, cutoff, later, malformed, invalid-character, and missing VIN cases in an async unit test
- Exercises the real vehicle update pipeline and verifies a MY2022 base Model 3 is stored with `marketing_name: "RWD"`
- `mix ci` on Elixir 1.19.5 / OTP 28: 440 tests, 0 failures

Closes #3887

## Suggested labels

`fix`, `area:teslamate`

### DIFF
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -43,6 +43,39 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
   @drive_timeout_min 15
 
+  @vin_model_years %{
+    "A" => 2010,
+    "B" => 2011,
+    "C" => 2012,
+    "D" => 2013,
+    "E" => 2014,
+    "F" => 2015,
+    "G" => 2016,
+    "H" => 2017,
+    "J" => 2018,
+    "K" => 2019,
+    "L" => 2020,
+    "M" => 2021,
+    "N" => 2022,
+    "P" => 2023,
+    "R" => 2024,
+    "S" => 2025,
+    "T" => 2026,
+    "V" => 2027,
+    "W" => 2028,
+    "X" => 2029,
+    "Y" => 2030,
+    "1" => 2031,
+    "2" => 2032,
+    "3" => 2033,
+    "4" => 2034,
+    "5" => 2035,
+    "6" => 2036,
+    "7" => 2037,
+    "8" => 2038,
+    "9" => 2039
+  }
+
   # Static
   def interval(env_var, default) do
     System.get_env(env_var)
@@ -59,7 +92,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
   def charging_interval, do: interval("POLLING_CHARGING_INTERVAL", 5)
   def minimum_interval, do: interval("POLLING_MINIMUM_INTERVAL", 0)
 
-  def identify(%Vehicle{display_name: name, vehicle_config: config}) do
+  def identify(%Vehicle{display_name: name, vin: vin, vehicle_config: config}) do
     case config do
       %VehicleConfig{
         car_type: type,
@@ -96,7 +129,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
             {"3", "74D", _} -> "LR AWD"
             {"3", "74", _} -> "LR"
             {"3", "62", _} -> "MR"
-            {"3", "50", _} -> "SR+"
+            {"3", "50", _} -> model_3_base_trim(vin)
             {"X", "100D", "tamarind"} -> "LR"
             {"X", "P100D", "tamarind"} -> "Plaid"
             {"Y", "P74D", _} -> "LR AWD Performance"
@@ -119,6 +152,22 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
       nil ->
         {:error, :vehicle_config_not_available}
+    end
+  end
+
+  # Position 10 of a 17-character VIN encodes the model year. Codes repeat every
+  # 30 years; current Tesla VINs are resolved against the 2010-2039 cycle.
+  defp vin_model_year(<<_::binary-size(9), code::binary-size(1), _::binary-size(7)>>),
+    do: Map.get(@vin_model_years, code)
+
+  defp vin_model_year(_vin), do: nil
+
+  # Model year is only a proxy for the naming change. A MY2021 car renamed or
+  # sold later still falls back to SR+ because the API exposes no rename date.
+  defp model_3_base_trim(vin) do
+    case vin_model_year(vin) do
+      year when is_integer(year) and year >= 2022 -> "RWD"
+      _year -> "SR+"
     end
   end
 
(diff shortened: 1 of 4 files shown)


JakobLichterfeld (changes requested):
Thanks for picking this one up — #3887 has been open for a while, and the approach is sound: `trim_badging` alone genuinely cannot distinguish SR+ from RWD, so deriving it from the VIN model year is the right lever.

I verified the core logic and it holds up: the binary pattern (9 + 1 + 7 = 17) lands exactly on position 10, the year-code list is complete and gap-free for MY2022–2039, and `vin` is reliably populated on the struct at the point `identify/1` runs (`TeslaApi.Vehicle.result/1` sets it from `v["vin"]`, and `identify/1` is only reached after a real `vehicle_data` fetch). The `nil`/short-VIN fallback is safe, and no migration is needed since `marketing_name` is refreshed by `update_car` on the next vehicle process start. Nothing besides the car summary reads that field, so blast radius is small.

A few things I'd like to see changed before merging:

**1. Make the invariant explicit and document the VIN position**

`@model_3_rwd_year_codes` encodes "model year >= 2022" as an opaque list of 18 characters, with nothing stating that position 10 is the model-year code. VIN year codes also recycle on a 30-year period, so `A` means 1980, 2010 *and* 2040 — as written, the check silently reverts to `SR+` from MY2040 onwards.

I'd prefer a small dedicated helper, e.g.:

```elixir
# Position 10 of a 17-character VIN encodes the model year.
# Codes recycle every 30 years; Tesla VINs start in 2008, so we resolve
# them against the 2010-2039 cycle.
defp vin_model_year(<<_::binary-size(9), code::binary-size(1), _::binary-size(7)>>),
  do: Map.get(@vin_model_years, code)

defp vin_model_year(_vin), do: nil
```

with the trim rule reading `case vin_model_year(vin) do year when is_integer(year) and year >= 2022 -> "RWD"; _ -> "SR+" end`. That puts the cutoff year at the decision site instead of hiding it in a character list, and it stays testable on its own.

**2. Reusability — Model Y has the same problem**

Model Y `trim_badging` `"50"` still maps to `"SR"`, which Tesla also retired in favour of RWD. I'm not asking you to fix that here (happy to keep this PR scoped to Model 3), but a standalone `vin_model_year/1` makes that a one-line follow-up instead of a second copy of the same table. If you'd rather handle both in this PR, that's welcome too.

**3. Test coverage gap**

The new tests call `Vehicle.identify/1` directly, which means the assumption the whole change rests on — that `vin` actually reaches `identify/1` through the real pipeline — is never exercised. `online_event/2` in `test/support/vehicle_case.ex` doesn't set `vin` at all today. Could you add `vin: Keyword.get(opts, :vin)` there and include one end-to-end test asserting that a MY2022 base Model 3 ends up in the DB with `marketing_name: "RWD"`? That's the regression I actually want guarded.

**4. Minor test nits**

- `"Z"` is not a valid VIN model-year character at all (I, O, Q, U and Z are excluded), so `"falls back to SR+ for an unknown model year"` is testing character validation rather than an unknown year. Either rename it, or use a pre-2022 code such as `"K"` (MY2019) for the "older model year" case and keep `"Z"` as an explicit invalid-character test.
- The six new cases are pure unit tests but live in a DB-backed `async: false` case file. Not a blocker, just noting they'll run slower than they need to.

**5. Behavioural note worth a comment**

Model year is not the same as the date Tesla changed the name — a MY2021 car sold in 2022 as "RWD" will still show "SR+". That's an acceptable trade-off given the data we have, but please note it in the code comment so the next reader doesn't treat it as a bug.

One last thing: the title is `fix(vehicle):` while the description suggests the `enhancement` label. I'd keep it as a `fix` and drop the enhancement suggestion.

Once the above is addressed I'm happy to merge this. Thanks again for the thorough test cases and for linking the issue.

JakobLichterfeld (approved):
This looks good now — thanks for the quick and thorough turnaround.

I re-checked the reworked version and all five points are addressed properly:

- **`@vin_model_years`** is complete and correct. I counted it through: A=2010 … Y=2030 with I, O, Q, U and Z correctly excluded, then 1=2031 … 9=2039. No off-by-one, and the `>= 2022` cutoff now sits at the decision site instead of being hidden in a character list. The comment naming position 10 and the 30-year cycle is exactly what I was after.
- **`vin_model_year/1`** is standalone, so extending this to Model Y `trim_badging` `"50"` later is a small follow-up rather than a second copy of the table. Keeping that out of this PR was the right call.
- **The pipeline regression test is the real thing.** Since `start_vehicles/1` registers the car with `vin: "absadkalfs"`, the `marketing_name: "RWD"` assertion can only come from the VIN carried on the online event — which is precisely the production path through `identify/1` and `update_car`. That was my main concern and it is now genuinely guarded.
- **`online_event/2`** defaults `vin` to `nil`, so no existing caller changes behaviour. Good.
- **Test coverage** is better than I asked for — adding MY2019 as a valid older year and MY2039 as the upper boundary, plus the renamed invalid-character case, covers the interesting edges. Moving the pure cases to `async: true` is fine here because `identify/1` has no side effects and the map is resolved at compile time.

The MY2021 caveat being documented in the code means the next reader won't mistake it for a bug. I also checked for fallout elsewhere: `Vehicles.create_or_update!` still hits the `vehicle_config: nil` path unchanged, the duplicated `describe` block from the first revision is gone, and there are no leftover unused aliases.

One non-blocking note, no need to hold the merge for them:

- `test/teslamate/vehicles/vehicle_identification_unit_test.exs` would sit more naturally in `test/teslamate/vehicles/vehicle/`, where the other `TeslaMate.Vehicles.Vehicle.*Test` modules live. Its neighbour `identification_test.exs` already breaks that convention, so I'm not going to make you move it.

Approving. Thanks for the clean commits, the honest scoping, and for closing #3887 after all this time.

🤖 Review drafted with Claude Code (→ claude.com) (Opus 5 high) — sponsored by Claude for Open Source (→ [www.anthropic.com](https://www.anthropic.com))

PR #5552, opened 2026-07-27
feat(import): continue past malformed rows and resume completed TeslaFi files

### LINKED ISSUE #5528: feat(import): use a dedicated rejection reason for VIN conflicts instead of `:invalid_fields`
### Context

Follow-up to #5519. The TeslaFi import quarantines a row when its VIN differs from the
import car while the vehicle ID still matches (`classify_vehicle/5` in
`lib/teslamate/import.ex`). This handles TeslaFi variants that omit VINs and rely on the
configured vehicle-ID fallback.

### Problem

The quarantined row is currently recorded as `:invalid_fields, ["vin"]`, so the import
page shows *"invalid values for vin"*. That is misleading: the VIN is a perfectly valid
value — it just belongs to a different vehicle than the one the import run identified.

### Suggested change

- Add a dedicated reason such as `:vehicle_conflict` to the `Ecto.Enum` values in
  `lib/teslamate/import/rejection.ex` and use it in `classify_vehicle/5`.
- Add a matching `rejection_reason/1` clause in
  `lib/teslamate_web/live/import_live/index.ex` with a gettext string along the lines of
  *"row belongs to a different vehicle"*, and run the gettext extract/merge.

No migration is needed — the `reason` column is a plain `:string`.

🤖Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### LINKED ISSUE #5529: chore(import): clean up import runs whose source directory no longer exists
### Context

Follow-up to #5519. Import runs are keyed by a hashed `source_key` derived from the
import directory path. `Checkpoint.start_run/2` deletes previous non-running runs **for
the same source key** before inserting a new one, so retention is bounded per directory.

### Problem

If the import directory is renamed or moved, its runs (plus cascaded file checkpoints and
rejections) are orphaned under the old source key and are never cleaned up. The data
volume is small and bounded (at most one lingering run per abandoned path), so this is
cosmetic — but the rows accumulate forever with no way to remove them short of manual SQL.

### Suggested change

Options, roughly in order of preference:

1. On `Import` startup (`read_directory`), delete non-`:running` runs whose `source_key`
   differs from the current one — the import feature only ever serves a single directory,
   so runs for other keys are unreachable by definition.
2. Alternatively, a periodic or startup-time sweep that removes runs older than some
   generous threshold.

Either way, `import_file_checkpoints` and `import_rejections` follow via the existing
`on_delete: :delete_all` references.

🤖Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### LINKED ISSUE #5530: refactor(import): remove or guard `LineParser.parse/2`, which crashes on unvalidated dates
### Context

Follow-up to #5519. Since that PR, the row pipeline parses the timestamp once in
`RowValidator.parse/2` and passes it to `LineParser.parse/3`. The old two-arity head was
kept for convenience:

```elixir
def parse(line, tz) when is_map(line) do
  {:ok, timestamp} = parse_timestamp(Map.get(line, "Date"), tz)
  parse(line, tz, timestamp)
end
```

### Problem

`parse/2` raises a `MatchError` whenever the `Date` value is missing or malformed. That is
safe today because the only production caller goes through `RowValidator`, which
pre-validates the date — but nothing prevents a future caller from using `parse/2`
directly and reintroducing exactly the crash-on-malformed-row behavior that #5519 removed.

### Suggested change

Either:

- remove `parse/2` and update any remaining (test) callers to `parse/3`, or
- make `parse/2` return `{:ok, vehicle} | {:error, reason}` instead of raising.

Removing it is preferred: a single entry point with a pre-parsed timestamp keeps the
"validate first, then parse" contract enforceable by the compiler rather than by
convention.

🤖Issue drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

magrathean-uk:
## Context

This brings back #5519 after the revert in #5531. The reproducible fractional battery regression is fixed by accepting and rounding fractional levels in the validated row path, with regression coverage.

## Included

- Restores the reviewed malformed-row quarantine and file-level resume flow from #5519
- Uses a dedicated vehicle conflict rejection reason
- Cleans up non-running import runs whose source directory is no longer reachable
- Removes the unsafe `LineParser.parse/2` entry point
- Keeps the documented file-level, at-least-once recovery boundary
- Keeps rejection data privacy-safe by storing no raw values or source paths

These also cover the three optional follow-ups from the final maintainer review on #5519.

## Validation

- `mix ci`: 427 tests, 0 failures
- Focused importer suite: 74 tests, 0 failures
- `mix gettext.extract --check-up-to-date`: passed
- Formatting and diff checks: passed

Closes #5528
Closes #5529
Closes #5530

## Suggested labels

`enhancement`, `area:teslamate`

### DIFF
--- a/lib/teslamate/import.ex
+++ b/lib/teslamate/import.ex
@@ -8,14 +8,30 @@ defmodule TeslaMate.Import do
   alias TeslaMate.{Vehicles, Repair, Log}
   alias TeslaMate.Log.{Car, State}
 
-  alias __MODULE__.{Status, LineParser, FakeApi, CSV}
+  alias __MODULE__.{
+    Status,
+    RowValidator,
+    RejectedRow,
+    RejectionReport,
+    Checkpoint,
+    Run,
+    FakeApi,
+    CSV
+  }
 
   defstruct(
     path: nil,
+    source_key: nil,
     files: [],
     timezone: :utc,
     error: nil,
     completed: MapSet.new(),
+    rejection_report: %RejectionReport{},
+    run_id: nil,
+    run_timezone: nil,
+    run_car_id: nil,
+    date_limit: nil,
+    date_limit_captured: false,
     car: nil,
     pids: %{},
     deps: %{}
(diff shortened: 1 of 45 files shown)


JakobLichterfeld (approved):
Thank you for bringing this back so carefully after the revert — approving.

I verified this PR file-by-file against the previously reviewed and merged state of #5519 (`b4115e91`). The result: it is byte-identical to the approved state, plus exactly three things:

1. **The actual CI fix.** The failing `LineParserTest` from #5513 called `LineParser.parse/2` without a `"Date"` key, which the reviewed state crashed on via its `{:ok, ts} =` match — precisely the hazard flagged in #5530. Removing `parse/2` entirely (the preferred option there) fixes it at the root. Importantly, the rounding assertions themselves are unchanged and now have additional regression coverage through `RowValidator` — the code was fixed, not the test.
2. **The three follow-ups from the review** (#5528, #5529, #5530): the dedicated `:vehicle_conflict` reason with its translated message, `cleanup_unreachable_runs/1` with sensible edge-case handling (unreachable directory, foreign running runs), and the `parse/2` removal — each with tests.
3. **Gettext merges** for the new string across all locales.

One clarification for the record: the fractional battery-level rounding from #5513 was never broken in the merged state — the regression was solely the `parse/2` crash on rows without a prevalidated date. The PR description's "fixed by accepting and rounding fractional levels" slightly misstates that; the effective fix is the removed entry point.

CI is fully green. Nice work turning a revert into a strictly better version.

PR #5554, opened 2026-07-27
chore(i18n): refresh POT references for #5438

magrathean-uk:
## Summary

This is a one-file follow-up for #5438, targeting the existing feature branch.

The charge cost LiveView moved two Gettext calls, but `priv/gettext/default.pot` still points at the old lines. That is why the POT verification fails. This refreshes only those generated source references, with no runtime behaviour change.

## Checks

- `mix compile --warnings-as-errors`
- `mix ci`: 361 tests, 0 failures

### DIFF
--- a/priv/gettext/default.pot
+++ b/priv/gettext/default.pot
@@ -395,7 +395,7 @@ msgstr ""
 msgid "≈ %{range} at 100%"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


JakobLichterfeld:
Closed this as this is no value add and only time-consuming for all parties.

PR #5555, opened 2026-07-27
fix(mqtt): address Home Assistant discovery review

magrathean-uk:
## Summary

This follow-up targets `ha_discovery` directly, so it can be merged into #5543 without replacing Brian's original work.

It covers the requested review points:

- Explicit `false` clears retained discovery configs once, while an unset option leaves existing installations untouched. Failed publish or cleanup attempts retry on the next summary.
- `default_entity_id` preserves the documented manual entity IDs. The docs now explain the manual YAML migration and correct the parking-brake binary sensor ID.
- The malformed environment table and optional URL wording are fixed.
- Empty prefixes fall back to `homeassistant`, MQTT wildcards are rejected, and valid multi-level prefixes remain supported.
- `charge_energy_added` uses `total_increasing`.
- The standalone discovery test uses `ExUnit.Case`.
- Discovery now covers the omitted non-deprecated topics, including charging state, update progress, individual doors and windows, TPMS warnings, service mode, and sunroof data.
- The integration docs describe coverage and retained-topic cleanup.

The `healthy` availability idea remains separate, matching the review note that it is a non-blocking follow-up.

## Checks

- Home Assistant and subscriber tests: 21 tests, 0 failures
- `mix compile --warnings-as-errors`
- Docusaurus production build
- Current-main combined integration: `mix ci`, 405 tests, 0 failures


JakobLichterfeld:
Overtaking other peoples work in other PRs creates frustration on all sides involved.

brianmay:
Note in this the the destination for this pull request was my pull request.

(I really wish github made this more clear actually)

JakobLichterfeld:
> Note in this the the destination for this pull request was my pull request.

Yeah, and that is exactly one point of what is wrong here.

PR #4861, opened 2025-07-31
feat(webview): show VIN as a Tooltip on car name

Helvio88:
image

PR #5556, opened 2026-07-27
feat(web): expose VIN in car summary

magrathean-uk:
## Summary

This refreshes #4861, whose original branch now conflicts with `main`. Credit for the original idea and implementation goes to @Helvio88 and is preserved in the commit history.

Named cars keep their display name and expose `VIN <last six>` through the existing responsive tooltip pattern. Cars without a display name now use the translated full VIN as the visible heading. This matches the convention already used by the Grafana dashboards.

For assistive technology, named cars include the VIN in an `is-sr-only` span. The duplicate native `title` tooltip is intentionally omitted. Clipboard behaviour remains out of scope.

## Checks

- Car controller tests: 20 passed
- Full local suite: 388 passed with a larger ExUnit receive window for existing timing-sensitive tests

### DIFF
--- a/lib/teslamate_web/live/car_live/summary.ex
+++ b/lib/teslamate_web/live/car_live/summary.ex
@@ -131,6 +131,15 @@ defmodule TeslaMateWeb.CarLive.Summary do
 
   def format_tpms(_, _), do: "—"
 
+  def format_car_title(display_name, vin) when display_name in [nil, ""] do
+    label = "VIN #{vin}"
+    {label, label}
+  end
+
+  def format_car_title(display_name, vin) do
+    {display_name, "VIN #{String.slice(vin, -6, 6)}"}
+  end
+
   defp translate_state(:start), do: ""
   defp translate_state(:driving), do: gettext("driving")
   defp translate_state(:charging), do: gettext("charging")
--- a/lib/teslamate_web/live/car_live/summary.ex
+++ b/lib/teslamate_web/live/car_live/summary.ex
@@ -132,12 +132,12 @@ defmodule TeslaMateWeb.CarLive.Summary do
   def format_tpms(_, _), do: "—"
 
   def format_car_title(display_name, vin) when display_name in [nil, ""] do
-    label = "VIN #{vin}"
+    label = gettext("VIN %{vin}", vin: vin)
     {label, label}
   end
 
   def format_car_title(display_name, vin) do
-    {display_name, "VIN #{String.slice(vin, -6, 6)}"}
+    {display_name, gettext("VIN %{vin}", vin: String.slice(vin, -6, 6))}
   end
 
   defp translate_state(:start), do: ""
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -37,13 +37,15 @@
   <div class="card-content">
     <div class="media is-flex mb-5">
       <div class="media-content">
+        <% {display_name, vin_label} = format_car_title(@summary.display_name, @car.vin) %>
         <p class="title is-5">
           <span
             class="has-tooltip-right has-tooltip-left-mobile"
             style="border-bottom: none;"
-            data-tooltip={@car.vin}
+            data-tooltip={vin_label}
+            title={vin_label}
           >
-            <%= @summary.display_name %>
+            <%= display_name %>
           </span>
         </p>
         <%= unless is_nil(@car.model) do %>
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -43,9 +43,9 @@
             class="has-tooltip-right has-tooltip-left-mobile"
             style="border-bottom: none;"
             data-tooltip={vin_label}
-            title={vin_label}
           >
             <%= display_name %>
+            <span :if={display_name != vin_label} class="is-sr-only">, <%= vin_label %></span>
           </span>
         </p>
         <%= unless is_nil(@car.model) do %>
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -37,7 +37,15 @@
   <div class="card-content">
     <div class="media is-flex mb-5">
       <div class="media-content">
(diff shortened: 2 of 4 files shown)


JakobLichterfeld (changes requested):
Thanks for picking this up and for keeping @Helvio88's authorship in the commit — that's the right way to refresh a stale PR. POT files are regenerated correctly this time, CI is green, and replacing the brittle `~r/<p class="title is-5">FooCar<\/p>/` regexes with a Floki helper is a genuine improvement to the test file. Appreciated.

I can't merge it as it stands, though. Three things:

**1. Test coverage for the new behaviour.** There is exactly one assertion for the tooltip, and it's bolted onto `"renders last known vehicle stats"`, a test about ranges and temperatures. The new `assert_car_title/2` helper drops the tooltip check at the other nine call sites, so the feature rests on a single incidental assertion. Please add a dedicated test for the tooltip itself, and cover the case where the car has no `display_name`: today that renders an empty `<span>`, which means a zero-width tooltip anchor and a VIN that cannot be reached in the UI at all — arguably the case where it matters most.

Also, `assert {"class", "has-tooltip-right has-tooltip-left-mobile"} in attrs` pins the exact class string and will break on any styling change without a behaviour change. Assert the tooltip content, not the stylesheet.

**2. Consistency with how we already display VINs.** The Grafana dashboards have a project-wide convention for this, e.g. in `grafana/dashboards/charge-level.json`:

    CASE WHEN COUNT(id) OVER (PARTITION BY name) > 1 AND name IS NOT NULL
         THEN CONCAT(name, ' - ', RIGHT(vin, 6))
         ELSE COALESCE(name, CONCAT('VIN ', vin)) END

Last six characters to disambiguate, full VIN only as a fallback when there is no name. Two surfaces of TeslaMate showing the same identifier in two different formats is not something I want to introduce, and the `COALESCE` branch happens to solve the empty-name case from point 1. If you think the full VIN is the right call here, make that argument in the PR — it's a reasonable position, but it needs to be a decision rather than an oversight.

**3. The tooltip is hover-only.** Bulma's tooltips are pure CSS `:hover`. `has-tooltip-left-mobile` only repositions the bubble, it does not make it triggerable on touch devices, so on mobile the VIN is effectively invisible despite the responsive classes. Adding `title={@car.vin}` alongside `data-tooltip` gives a native fallback plus screen reader and keyboard access, for one attribute.

Minor, optional: `style="border-bottom: none;"` copies the inline hack from the `marketing_name` span. `assets/css/app.scss` already has that rule, just scoped narrowly to `.card-content .icons span[data-tooltip]`. Widening it to cover `.media-content` and removing both inline styles would be a welcome cleanup — but I'm happy to take that as a follow-up rather than growing this PR.

Nothing here is a rewrite; the approach is fine. Ping me when the tests are in and let me know where you land on the VIN format.

JakobLichterfeld (commented):
Thanks — all three are in, and CI is green.

The tests are exactly what I asked for, and extending `start_vehicles/2` with a real 17-character VIN was the right instinct: with the old `"xxxxx"` fixture, `String.slice(vin, -6, 6)` returns the whole string, so the last-six logic would never have been exercised. `format_car_title/2` maps the Grafana rule cleanly, and lifting the tuple into a named function is tidier than the inline lambda we already have for `{soc_text, tooltip}` further down that template.

**On the `title` attribute — that one is on me.** I asked for it in point 3, and the suggestion was wrong. `title` on a non-focusable `<span>` gives no keyboard access, screen readers treat it inconsistently at best, and on desktop it now renders a second native bubble a second after the Bulma one, with identical text (should have been visible in your tests). So it costs a visual wart and buys very little. I'm not asking you to revert something I requested, and I'm happy to merge as is. If you'd rather have it right: Bulma 1.x ships `is-sr-only`, and a visually hidden `<span>` is the reliable way to expose the VIN to assistive tech. Note that `aria-label` is not the fix here — we use it on `<nav>` and `<button>`, which is correct, but ARIA does not guarantee it on a generic `<span>`. Your call.

Two things I would like before merging:

**1. Run the label through gettext.** `"VIN #{vin}"` is hardcoded, while every other user-visible string in `summary.ex` goes through gettext. This matters more than it would have for hover text, because in the unnamed case it is now the visible card heading: `gettext("VIN %{vin}", vin: ...)` plus a POT refresh.

**2. Update the PR title and description.** It still says "show VIN in car name tooltip", but cars without a name now render the full VIN as visible heading text where the heading used to be empty. That is an improvement and a direct consequence of the convention I asked for — I just don't want the changelog entry to describe a tooltip-only change.

Agreed on keeping the CSS cleanup out; it can be a follow-up.

JakobLichterfeld (changes requested):
This looks good — thanks for the careful work on the accessible variant.

Skipping the `is-sr-only` span when the VIN is already the visible heading is exactly the right call, and `refute Map.has_key?(attrs, "title")` is a nice touch: it keeps the attribute from creeping back in later. Separating visible text from screen reader text via `Floki.filter_out/2` makes the helpers honest, and updating the `"unavailable"` test to assert `"VIN xxxxx"` documents the behaviour change instead of hiding it. Good to hear the receive window was a local runtime setting and not something committed.

One thing left before I can merge: **please rebase onto `main`.** GitHub currently reports the branch as conflicting, and the conflict is `priv/gettext/default.pot` — #5552 landed after you branched and touched the same file. The consequence matters more than the conflict itself: without a mergeable ref, the Elixir workflows never ran on your latest commit. Only the CLA check and the Netlify preview are reporting, so `elixir_test`, `gettext.extract --check-up-to-date` and the static analysis have not seen this state at all.

After the rebase, please re-run the extraction rather than resolving the `.pot` conflict by hand:

    git rebase main
    mix gettext.extract

Hand-resolving that file leaves the line references pointing at the pre-rebase `main`, which fails `--check-up-to-date` on CI.

Two things I'm explicitly not asking you to change: the `", "` separator in the screen reader span is hardcoded punctuation outside gettext, and the inline `border-bottom: none;` is still there. Both are fine as follow-ups. `:if={}` is the first use of that syntax in the repo, but it is the recommended form and cleaner than the `<%= if %>` blocks around it, so no objection there either.

JakobLichterfeld (approved):
Thanks for double-checking! It's really strange. I did a hard refresh several times as I couldn't believe it either. The GitHub hiccup seems to have been resolved now — thanks again!

PR #5559, opened 2026-07-31
feat: use Grafana 13.1.1

swiffer:
**Upgrade Grafana to 13.1.1**

The `newPanelPadding` feature toggle was removed in Grafana 13.1.0 ([grafana/grafana#124870](https://github.com/grafana/grafana/pull/124870)).  

As a result:
- The corresponding environment variable has been removed.
- The current soc / current stored energy bar gauge panels minimum height has been reduced so it fully renders.

grafik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - fix(vehicle): update geofence while driving with streaming API (#5515 - @magrathean-uk)
 - fix(vehicle): identify base Model 3 from model year 2022 as RWD instead of SR+ (#5551 - @magrathean-uk)
 - fix(mqtt): avoid blocking startup on retained cleanup (#5549 - @magrathean-uk)
+- feat: use Grafana 13.1.1 (#5559 - @swiffer)
 
 #### Build, CI, internal
 
--- a/grafana/Dockerfile
+++ b/grafana/Dockerfile
@@ -1,6 +1,6 @@
 # Ensure selecting a tag that is available for arm64 and amd64
 # https://hub.docker.com/r/grafana/grafana/tags
-FROM grafana/grafana:13.0.1-security-01
+FROM grafana/grafana:13.1.1
 
 ENV GF_ANALYTICS_REPORTING_ENABLED=false \
     GF_ANALYTICS_CHECK_FOR_UPDATES=false \
@@ -24,7 +24,6 @@ ENV GF_ANALYTICS_REPORTING_ENABLED=false \
     GF_CLOUD_MIGRATION_ENABLED=false \
     GF_FEATURE_TOGGLES_provisioning=false \
     GF_FEATURE_TOGGLES_splashScreen=false \
-    GF_FEATURE_TOGGLES_newPanelPadding=false \
     GF_FEATURE_TOGGLES_dashboardNewLayouts=false \
     DATABASE_PORT=5432 \
     DATABASE_SSL_MODE=disable
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -96,7 +96,7 @@
         "textMode": "value_and_name",
         "wideLayout": true
       },
-      "pluginVersion": "13.0.1+security-01",
+      "pluginVersion": "13.1.1",
       "targets": [
         {
           "datasource": {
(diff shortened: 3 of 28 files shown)


JakobLichterfeld (changes requested):
Very cool, thank you @swiffer for taking the time for this again! 🙏 Grafana bumps are thankless grunt work, and you do them reliably — I really appreciate that.

I went through the full diff. The substance is clean, and I verified three things that all check out:

- `13.1.1` was pushed 2026-07-22, `13.0.1-security-01` on 2026-05-12 — so leaving the `-security-01` tag is not a security regression. ✅
- `13.1.1` is available for amd64/arm64/armv7, as the comment in the Dockerfile requires. ✅ (even if we dropped armv7)
- `newPanelPadding` is indeed gone from the current feature toggle docs, while `provisioning`, `splashScreen` and `dashboardNewLayouts` still exist — so the removal is scoped exactly right. ✅

And the bar gauge fix is more than cosmetic: `sizing: manual` is what makes `minVizHeight` take effect in the first place. Nice catch.

Three things before I'd like to merge:

**1. The home dashboard UID changes** (`grafana/dashboards/internal/home.json`)

```diff
-  "uid": "be2m9kga7b8qoc",
+  "uid": "mamjzp2"
```

That UID has been stable since your original "Welcome Dashboard" commit (Nov 2024), and it's the only one of the 23 dashboards that changes here — looks like an artifact to me (dashboard re-created rather than updated?). The default home is resolved via `GF_DASHBOARDS_DEFAULT_HOME_DASHBOARD_PATH`, so nothing breaks functionally. But bookmarks and links to `/d/be2m9kga7b8qoc` break for every existing user, and with `disableDeletion: false` the provisioner will remove the old entry along with its stars. Could you restore the old UID?

**2. `drives.json` is the odd one out, in two ways**

- The file loses its trailing newline (`\ No newline at end of file`) — the only one of the 25 files in this PR.
- Its 7 `pluginVersion` fields stay on `13.0.1+security-01`. Across all dashboards, 159 of 166 occurrences are migrated, and the remaining 7 are exactly this file.

Both are purely cosmetic (Grafana rewrites `pluginVersion` on the next save anyway), but it'll come back as diff noise on the next bump.

**3. Minor, not a blocker**

`"fixedColor": "#73BF69"` gets added in four places next to `"mode": "palette-classic"`, where it has no effect — `fixedColor` only applies with `mode: "fixed"`. Pure UI noise, fine to leave as is.

And if you feel like folding it in: `website/docs/installation/unsupported/debian.md` and `freebsd.md` still say "Grafana (v13.0.1+)". Not wrong thanks to the `+`, but a good moment to update it. If that's too much scope, I'm happy to do it in a follow-up PR.

Once 1 and 2 are in, happy to merge. Thanks again! 🚗⚡

swiffer:
Thanks for the review, Claude — you spotted exactly what I missed.

One question remaining: you mentioned that armv7 is still required in the Dockerfile even though it has been dropped. I can’t find any comment stating that armv7 is still needed — could you clarify?

I’ve addressed points 1 and 2. For point 3, the colour was added by Grafana on export, so I’ll leave it as-is.

JakobLichterfeld (approved):
Good catch — that one is really on Claude, and you're right to question it.

The comment only says `arm64 and amd64`:

```dockerfile
# Ensure selecting a tag that is available for arm64 and amd64
```

ARMv7 is not required and hasn't been since we dropped it in #5391. Claude pulled it into the list because Docker Hub still ships an `arm/v7` manifest for the `13.1.1` tag, and it listed what the registry returns instead of what we actually require. The check against the comment itself — arm64 and amd64 — holds, so the conclusion stands, just not the way Claude phrased it.

I re-checked the updated diff, and points 1 and 2 are fully resolved:

- `internal/home.json` is back on `uid: "be2m9kga7b8qoc"` ✅
- `drives.json` keeps its trailing newline ✅
- `pluginVersion` is now migrated in all 166 places (was 159) ✅

And agreed on point 3 — if Grafana emits `fixedColor` on export, fighting it would just create churn on the next bump. Leave it.

LGTM, merging. Thanks for the quick turnaround, and for catching my armv7 slip! 🚗⚡

PR #5562, opened 2026-08-01
fix(grafana): use local calendar for Statistics period end boundaries

### LINKED ISSUE #5561: Statistics drill-down can end before the next local period boundary
### Existing issues checked

I searched the open/closed issues and discussions. #3132 and #3857 cover earlier period-start/time-zone problems; #3862 introduced the current three-argument `date_trunc`. I could not find an issue for the remaining end-boundary calculation.

### What happened?

The Statistics dashboard calculates the correct local period start, but all four table queries calculate `date_to` by adding an interval directly to the resulting `timestamptz`:

```sql
EXTRACT(EPOCH FROM date + interval '1 $period') * 1000 AS date_to
```

That interval follows the PostgreSQL session calendar, not Grafana's selected local calendar. In positive-offset time zones, a monthly drill-down can end before the next local month begins.

With PostgreSQL 18, session time zone UTC, Grafana time zone `Asia/Shanghai`, and a July 2026 row, the current and expected boundaries are:

```text
start local       current end local   expected end local
2026-07-01 00:00  2026-07-31 00:00   2026-08-01 00:00
```

Clicking July can therefore omit the final local day. Direct interval addition can also shift hour boundaries across DST transitions.

### Expected behavior

`date_to` should be the next local period boundary (half-open range `[date_from, date_to)`): convert `date` to local wall-clock time, add one calendar period, then convert the result back to an instant.

Yes.

### Steps to reproduce

1. Set the Grafana dashboard time zone to `Asia/Shanghai`.
2. Open Statistics and select `month` as the period.
3. Click a 31-day monthly row such as July 2026.
4. Inspect the generated target dashboard `to` value; it resolves to July 31 00:00 local instead of August 1 00:00 local.

The same result can be reproduced directly from the four current SQL queries in `grafana/dashboards/statistics.json` with PostgreSQL's session time zone set to UTC.

### Relevant log output

Not applicable; this is deterministic dashboard SQL behavior and produces no TeslaMate application error log.

### Installation / versions used for verification

- Dashboard source: current `main` at `34e9a50`
- PostgreSQL: 18 (`postgres:18-trixie`)
- Installation style: Docker
- Reverse proxy: no

I have a four-query minimal patch plus boundary tests for UTC, Asia/Shanghai, America/Chicago, and Europe/Berlin, including DST spring/fall transitions.

wjsall:
Fixes #5561.

## Summary

The Statistics dashboard already computes `date_from` with PostgreSQL's three-argument `date_trunc`, so the start instant and displayed label are correct. This patch changes only the four `date_to` expressions:

- convert the start instant to local wall-clock time in `$__timezone`
- add one calendar `$period`
- convert the next local boundary back to an instant

This preserves the half-open drill-down range `[local period start, next local period start)` in positive-offset time zones and across DST transitions.

## Reproduction fixed

PostgreSQL session time zone UTC, Grafana time zone `Asia/Shanghai`, July 2026:

```text
start local       before              after
2026-07-01 00:00  2026-07-31 00:00   2026-08-01 00:00
```

## Validation

- parsed `grafana/dashboards/statistics.json` with `jq`
- verified all four Statistics targets contain the new boundary expression and none retain direct `timestamptz + interval`
- ran PostgreSQL 18 tests for day/week/month/year in UTC, Asia/Shanghai, America/Chicago, and Europe/Berlin
- verified 23/25-hour DST day boundaries in Chicago and Berlin: 11/11 tests passed
- `git diff --check`


JakobLichterfeld (approved):
Thanks for the detailed issue and the matching minimal patch — the analysis is correct, and I verified it against PostgreSQL locally.

`timestamptz + interval '1 month'` does its calendar arithmetic in the **PostgreSQL session time zone**, not in `$__timezone`, so with the default UTC session the drill-down end drifts off the local period boundary. Converting to local wall-clock, adding the calendar interval, and converting back is the right fix, and it makes the expression independent of the session time zone.

I ran the old and new expressions side by side over 4 time zones × 4 periods × 5 sample dates with the session time zone set to UTC. 18 of 80 combinations differ, and the new expression lands on the exact local boundary in every case:

```text
tz                period  local_from           old_local_to         new_local_to
Asia/Shanghai     month   2026-07-01 00:00:00  2026-07-31 00:00:00  2026-08-01 00:00:00
Europe/Berlin     month   2026-10-01 00:00:00  2026-10-30 23:00:00  2026-11-01 00:00:00
Europe/Berlin     day     2026-03-29 00:00:00  2026-03-30 01:00:00  2026-03-30 00:00:00
America/Chicago   day     2026-11-01 00:00:00  2026-11-01 23:00:00  2026-11-02 00:00:00
America/Chicago   month   2026-03-01 00:00:00  2026-04-01 01:00:00  2026-04-01 00:00:00
```

One correction to the PR description: the bug is not limited to positive-offset zones. `America/Chicago` is off by an hour on every DST-crossing day, week, and month, as the rows above show. Might be worth reflecting that in the summary, since it widens who is affected.

The patch also helps more than the issue suggests in zones with unusual DST rules. Over a full year 2026, counting how many period boundaries the patch corrects:

```text
tz                   day      week     month    year
Australia/Lord_Howe  2/365    2/53     6/12     0/1
Asia/Beirut          1/365    2/53     5/12     0/1
America/Santiago     1/365    2/53     2/12     0/1
Cuba                 1/365    2/53     2/12     0/1
```

`Australia/Lord_Howe` has a 30-minute DST offset, where the old expression produced ends like `2026-04-30 23:30` instead of `2026-05-01 00:00`. Note that `year` is unchanged in every zone I tested, so the change is confined to the periods that are actually affected.

I also verified that `date_to` and the `interval '1 $period'` pattern exist **only** in `grafana/dashboards/statistics.json`, so the four targets you touched are the complete scope. The "Query shared between Charging Stats, Efficiency, Statistics & Trip Dashboards" note on target D covers the shared CTE part, not the `date_to` projection — no sync needed in the other dashboards. `$period` is a fixed `custom` variable (`day,week,month,year`), so the string interpolation stays safe, and `GROUP BY 1,2,3,4` remains valid because `timezone()` is `STABLE`.

Two optional notes, neither a blocker:

- Readability: `((date AT TIME ZONE '$__timezone') + interval '1 $period') AT TIME ZONE '$__timezone'` is semantically identical but makes the two opposite conversions much easier to read than the nested `timezone()` calls. Your call.
- Residual edge case, explicitly **not** something I want fixed here. I checked boundary contiguity across a full year for `America/Santiago`, `Cuba`, `Asia/Beirut` and `Australia/Lord_Howe`. Three cases remain, all with `period = day` in zones whose DST transition happens at midnight: the skipped local midnight makes `date_from` resolve to 01:00, and `date_to` then overshoots the next day's start by one hour. That is an overlap rather than a gap, so no rows are lost from the drill-down. I confirmed the patch does not touch these cases at all — old and new yield the identical instant:

  ```text
  tz                local_from           old_to_utc              new_to_utc              changed_by_patch
  America/Santiago  2026-09-06 01:00:00  2026-09-07 04:00:00+00  2026-09-07 04:00:00+00  false
  Asia/Beirut       2026-03-29 01:00:00  2026-03-29 22:00:00+00  2026-03-29 22:00:00+00  false
  Cuba              2026-03-08 01:00:00  2026-03-09 05:00:00+00  2026-03-09 05:00:00+00  false
  ```

  The root cause sits in `date_from` — the three-argument `date_trunc()` resolving a non-existent local midnight — not in the `date_to` expression this PR changes. Addressing it would mean changing `date_from` as well, which is a different change with a different risk surface, so it is out of scope here.

Nice work, and thanks for testing across UTC / Asia/Shanghai / America/Chicago / Europe/Berlin including the DST days.

PR #5578, opened 2026-08-02
docs: AI-assisted contributing guidance and Grafana dashboard notes

swiffer:
## Summary

Addresses maintainer feedback on the earlier draft:

- **Dropped `AGENTS.md`** (and the `CONTRIBUTING` pointer). Brian's bar still holds: only document things an LLM regularly gets wrong, and the one candidate (`devenv up` TUI) already lives as an inline comment in `nix/flake-modules/devenv.nix` (#5560).
- **Kept the human-facing AI policy** in `website/docs/development.mdx`, promoted out of the numbered CLA/license list into its own `## AI-assisted contributions` section. Disclosure footer is mandatory for material help, with the **exact model name**, and this file is the single normative home.
- **Kept the Grafana docs updates**: `teslamate/grafana:edge`, Export → Export as code → Model **Classic**, variables/units table, and the `positions` sample-count warning.
- **Lint/nits**: prettier-formatted the markdown table; `mix ci` docs now include `deps.unlock --check-unused` (and `test --raise`, matching the alias).

## Test plan

- [x] `AGENTS.md` and `CONTRIBUTING` pointer are gone
- [x] AI section is a `##` sibling of *Contributing to TeslaMate*, not item 3 of the two merge requirements
- [x] Disclosure footer format unchanged; no second copy elsewhere
- [x] Grafana table formatted with prettier
- [ ] Preview development docs on Netlify deploy preview

---

🤖 Assisted by Grok 4.6 (xAI) via Grok Build (address review feedback, docs edits, PR description).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,102 @@
+# AGENTS.md
+
+Guidance for AI coding agents working on TeslaMate.
+
+Human contributors: see [Development and Contributing](https://docs.teslamate.org/docs/development/) and `CONTRIBUTING`.
+
+## Project map
+
+| Path | Role |
+|------|------|
+| `lib/`, `test/` | Elixir / Phoenix app |
+| `priv/repo/` | Migrations and SQL helpers (`convert_celsius`, `convert_km`, …) |
+| `grafana/dashboards/` | Provisioned Grafana dashboards |
+| `website/` | Documentation (Docusaurus) |
+| `nix/`, `flake.nix` | Dev environment / packaging |
+| `.github/workflows/` | CI |
+
+Prefer small, reviewable changes. Do not expand scope unprompted.
+
+## Setup and checks
+
+Use versions from the development docs when you have a local environment:
+
+```bash
+mix setup
+MIX_ENV=test mix ecto.setup
+mix ci
+treefmt   # or: nix run .#lint
+```
+
+Not every contributor has a full local Elixir/Postgres/Grafana stack. **It is fine to rely on GitHub Actions CI** on the PR for format checks, tests, and related workflows. Run what you can locally; fix CI failures the PR introduces before asking for merge.
+
+Run `mix gettext.extract --merge` only if user-facing strings changed and you can run Mix.
+
+## Change rules
+
+- Match neighboring style; no drive-by refactors.
+- Smallest diff that solves the stated problem.
+- Add or update tests for behavior changes.
+- Do not commit secrets, tokens, cookies, or vehicle credentials.
+- Do not change dashboard **UIDs** unless explicitly requested.
+
+## Grafana dashboards
+
+Canonical query and dashboard craft (timestamps, `positions`/streaming, `EXPLAIN ANALYZE`, `pg_stat_statements`) lives in the [Development and Contributing](https://docs.teslamate.org/docs/development/) docs under **Making Changes to Grafana Dashboards** and **Best Practices**. Follow that. Short rules for agents:
+
+- Use `teslamate/grafana:edge` for local edits. Export as code with **Model: Classic** (not **V2 Resource**); keep the JSON as exported.
+- Copy variable/link patterns from a similar existing dashboard (e.g. Overview, Efficiency).
+- Common variables: `car_id`, `base_url`, and when relevant `length_unit`, `temp_unit`, `preferred_range`.
+- Header links: TeslaMate → `${base_url:raw}` + Dashboards dropdown (tag `tesla`).
+- Temperature: `convert_celsius(col, '$temp_unit')` — never hardcode °C-only when settings exist.
+- Distance: `convert_km(..., '$length_unit')`.
+- Prefer `$__timeFilter` / `$__timeGroup`. If using `DATE_TRUNC`, follow the docs pattern with `TIMEZONE('UTC', …)` and `'$__timezone'`.
+- Query `positions` only when needed. If ~15s resolution is enough, prefer `ideal_battery_range_km IS NOT NULL` (and `car_id = $car_id`) to skip dense streaming rows — see docs.
+- `positions` is denser while driving than when parked; sample counts are not “time spent.” Prefer time-bucketing for distributions.
+- History charts: aggregate where samples exist; use Connect null values **Threshold** (not Always) for long offline gaps.
+- Larger UX changes: update screenshots under `website/static/screenshots/` (see docs).
+- Keep dashboard domain focus; do not merge unrelated concerns unprompted.
+
+## GitHub
+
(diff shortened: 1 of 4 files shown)


brianmay:
The need for AGENTS.md is somewhat disputed at the moment. Does it speed things up? Or does it add stuff to the context window for every required request that could easily fall out of date and where the LLM could easily work out for itself?

Haven't formed any definite opinions yet. Except sometimes the LLM might regularly get something wrong or take a long time to work out something, in which case that might be a good thing to document it. And some people are trying splitting the AGENTS.md up based on the task performed.

In the case of this project, mentioning treefmt might be an example of something an LLM might miss.

JakobLichterfeld:
Thanks for write up Swiffer.


> In the case of this project, mentioning treefmt might be an example of something an LLM might miss.

In my experience the only thing AI get sometimes only on second try is devenv up without TUI, which I want to solve with an inline comment.

Other than that echoing Brian.

swiffer:
Makes sense — happy to drop most of `AGENTS.md`. Fair warning: the current draft is probably overweight on Grafana because that’s what I last worked on.

Open questions before I revise:

1. **Contrib docs** — still add a short human-facing “AI-assisted contributions” note (welcome + human remains responsible for CLA/correctness/what gets posted)? Optional short disclosure on material help?
2. **`AGENTS.md`** — keep a minimal file, or drop it entirely?
3. If we keep anything agent-facing: mandatory 🤖 footer format, or anything else you actually want spelled out?

Happy to reshape heavily once that’s clearer.

JakobLichterfeld (changes requested):
Thanks for the follow-up, Matthias — and sorry for the delay. Answers to your three questions, plus a few concrete things I found reviewing the current draft.

**1. Human-facing "AI-assisted contributions" note — yes, please keep it.**

That part is the valuable half of this PR. "AI is welcome, the human stays responsible for correctness, security and licensing, and for anything posted under their name" is exactly the policy I want written down, and `website/docs/development.mdx` is the right home for it.

One structural fix: it's currently added as `### 3.` under *Contributing to TeslaMate*, whose intro still says "we have **two main requirements** for all Pull Requests". A welcome-and-responsibility note isn't a third merge requirement. It can stay exactly where it is in the file — just promote it out of the numbered list into its own `##` section.

**3. Disclosure footer — yes, mandatory, and exactly as you drafted it.**

Taking your third question out of order, because it also settles the second.

```markdown
---

🤖 Assisted by <Exact model name> (<Vendor>) via <Tool> (<what it helped with>).
```

The **exact model name** is the part that carries the information: `Claude Opus 5`, `Grok 4.5` — not "Claude", not "AI". Model generations differ enough in what they get right that a reviewer needs to know which one produced the diff, and the vendor/tool/role list tells them how it was used. Keep it required for material help on PR descriptions and substantive review comments, and keep trivial edits exempt.

So the format itself needs no softening — it just needs exactly one home, see below.

**2. `AGENTS.md` — I'd drop it entirely.**

@brianmay's bar is the right one: document the things an LLM regularly gets wrong or takes a long time to work out. In my experience there is exactly one such thing in this repo — `devenv up` hanging on the process-compose TUI. I said I wanted to solve that with an inline comment, and it has since landed that way in `nix/flake-modules/devenv.nix` (#5560). That means the one item that met the bar is now handled at the place where it cannot go stale, and I don't see a second candidate that justifies a file loaded into context on every request.

Concretely, on the current draft:

- The setup block (`mix setup`, `MIX_ENV=test mix ecto.setup`) never mentions `devenv` or `direnv`, which is how this repo provides Postgres and the `DATABASE_*` environment. That's the failure mode an agent file is supposed to prevent, and it illustrates how quickly this content drifts from the tree.
- "It is fine to rely on GitHub Actions CI for format checks" — no, and I'd push back on that regardless of who wrote the patch. A PR that arrives red costs reviewer time before anyone has read a line of it, and this one is a live example. Checks are the contributor's job before opening; CI confirms the result, it doesn't stand in for it. Writing the opposite into project guidance sets exactly the wrong default.
- The Grafana section restates what the same PR writes into `development.mdx` (`convert_celsius`/`convert_km`, the `positions` hint, Model: Classic, screenshots), and the disclosure footer is specified twice, in two places that can drift apart. The footer should have a single normative home, and that's the contributing docs.

If we ever revisit an agent-facing file, my bar would be: only facts that are non-obvious from the tree, not already expressible as an inline comment, and cheap to keep true. In practice that's a handful of lines — and #5560 shows they're better off as inline comments next to the thing they describe.

---

**The Grafana docs updates are the other half worth keeping.**

They're genuinely useful and independent of the policy question: the Export → *Export as code* → Model **Classic** flow, the variables/units table, and the "sample counts are not time spent" warning for `positions` queries.

With `AGENTS.md` and the `CONTRIBUTING` pointer dropped, what's left here is one coherent docs PR — no need to split anything, just un-draft it once these two are sorted:

- Both new markdown tables aren't formatted the way prettier formats tables, which is what `ensure_linting / check_linting` is currently failing on. A `treefmt` run fixes it.
- Small unrelated nit while you're in that file: we document `mix ci` as "check formatting + run tests", but the alias in `mix.exs` also runs `deps.unlock --check-unused`.

Thanks again for putting the policy question on the table — the docs half is worth landing.

---

swiffer:
Thanks @JakobLichterfeld — that was a very clear review. I took the answers as written:

1. **AI-assisted contributions** stays in `website/docs/development.mdx`, promoted to its own `##` section so it is no longer a third merge requirement under “two main requirements”.
2. **`AGENTS.md` is gone**, along with the `CONTRIBUTING` pointer. Agree on the bar, and that #5560 already covers the one non-obvious trap at the source.
3. **Disclosure footer** is unchanged (exact model name, vendor/tool/role, material help only) and now lives only in the contributing docs.

Also fixed the two concrete nits: prettier-aligned the variables table, and `mix ci` now documents `deps.unlock --check-unused` plus `test --raise` to match the alias. Rebased onto current `main` and unmarked as draft.

---

🤖 Assisted by Grok 4.6 (xAI) via Grok Build (address review feedback).

JakobLichterfeld (approved):
Thanks Matthias — this covers all four points nicely. Checked against the tree:

- `AGENTS.md` and the `CONTRIBUTING` pointer are gone.
- The AI policy is now a `##` sibling, so *Contributing to TeslaMate* still has its two requirements and nothing reads as a third one.
- Disclosure footer is unchanged and has a single normative home.
- `mix ci` now matches the alias in `mix.exs`, and the table is formatted — `ensure_linting` is green.

Thanks also for the CHANGELOG entry.

Side note: `make grafana` in the `Makefile` is now the only build step nothing points at — the line you removed was its last mention in the repo. Fine to leave as is.

LGTM, merging.

---

PR #5581, opened 2026-08-03
test(grafana): guard latest-position index predicate

magrathean-uk:
## Summary

Follow-up to the optional dashboard-query detector suggested in the #5438 approval.

- scan bundled Grafana JSON query fields for latest-position lookups
- require the partial-index predicate on each individual `FROM positions ... ORDER BY date DESC LIMIT 1` block
- cover both matching and non-matching controls, including a query where the predicate appears in an unrelated CTE

No dashboard SQL or runtime behavior changes.

## Validation

- `MIX_ENV=test DATABASE_PORT=55432 mix test test/teslamate/grafana/dashboard_queries_test.exs --trace`
- `MIX_ENV=test DATABASE_PORT=55432 mix ci`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 
 #### Build, CI, internal
 
+- test(grafana): guard latest-position dashboard queries against missing partial-index predicate (#5581 - @magrathean-uk)
 - build(nix): update mixFodDeps hash in nix builds ([954e8739](https://github.com/teslamate-org/teslamate/commit/954e8739326e092f7cddf7308dd4b704cc008f62) - @JakobLichterfeld)
 - build(deps): bump launch-editor from 2.13.2 to 2.14.1 in /website (#5426)
 - build(deps): update flake.lock (#5427)
--- a/test/teslamate/grafana/dashboard_queries_test.exs
+++ b/test/teslamate/grafana/dashboard_queries_test.exs
@@ -1,0 +1,102 @@
+defmodule TeslaMate.Grafana.DashboardQueriesTest do
+  use ExUnit.Case, async: true
+
+  @dashboard_directory Path.expand("../../../grafana/dashboards", __DIR__)
+  @query_keys ~w(definition query rawSql)
+
+  # A latest-position lookup needs the partial-index predicate introduced in
+  # #5438. Match each positions block independently so a filter in another CTE
+  # cannot hide an unindexed lookup.
+  @latest_position_block ~r/from positions\b(?:(?!\bfrom\b).)*?order by date desc(?:(?!\bfrom\b).)*?limit 1\b/
+
+  test "latest position queries use complete position rows" do
+    dashboards = Path.wildcard(Path.join(@dashboard_directory, "**/*.json"))
+
+    assert dashboards != []
+
+    offenders =
+      dashboards
+      |> Enum.flat_map(&dashboard_queries/1)
+      |> Enum.flat_map(fn {path, query} ->
+        query
+        |> unfiltered_latest_position_blocks()
+        |> Enum.map(&"#{path}: #{&1}")
+      end)
+
+    assert offenders == []
+  end
+
+  test "detector flags a positions block missing the filter even when it appears elsewhere" do
+    query = """
+    WITH metadata AS (SELECT 1 WHERE ideal_battery_range_km IS NOT NULL)
+    SELECT date FROM positions WHERE car_id = 1 ORDER BY date DESC LIMIT 1
+    """
+
+    assert unfiltered_latest_position_blocks(query) != []
+  end
+
+  test "detector accepts a positions block with the filter" do
+    query =
+      "SELECT date FROM positions WHERE car_id = 1 AND ideal_battery_range_km IS NOT NULL " <>
+        "ORDER BY date DESC LIMIT 1"
+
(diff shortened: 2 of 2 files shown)


JakobLichterfeld (changes requested):
Thanks @magrathean-uk for following up on this so quickly — exactly the guard I had in mind! 👍

The detector is carefully built: the tempered `(?:(?!\bfrom\b).)*?` token matching each `FROM positions` block independently is the right call (a predicate in an unrelated CTE can't mask an unindexed lookup), and having positive/negative/`LIMIT 10` controls makes the regex itself regression-safe. I also cross-checked against current `main`: only battery-health and overview contain latest-position blocks, both carry the predicate, and no dashboard uses alias forms like `ORDER BY p.date DESC` — so the literal pattern has no practical gap today.

Requesting changes for two small things, partly overlapping with the Copilot findings:

1. **Glob/CWD (Copilot's second comment — agreed):** If `Path.wildcard("grafana/dashboards/**/*.json")` ever comes up empty (moved directory, cwd change), the test passes vacuously. Please anchor the glob to `__DIR__` (or keep the glob) and add the `assert dashboards != []` sanity check.

2. **Regex (Copilot's first comment — diagnosis fair, suggested fix please *don't* apply):** An inner `FROM` from `EXTRACT(EPOCH FROM …)` would indeed keep a block from matching — a silent false negative. I checked: `EXTRACT(EPOCH FROM …)` appears ~30× across the bundled dashboards but never inside a latest-position block today, so it's theoretical for now. Copilot's fix (plain `.*?`) would trade that for something worse: it drops the tempered token and with it the per-block isolation your own self-test guards. Better: keep the tempered token and neutralize `EXTRACT`'s inner `FROM` during normalization, e.g.

   ```elixir
   |> String.replace(~r/\bextract\s*\(\s*epoch\s+from\b/, "extract(epoch_from")
   ```

   before scanning, plus a small self-test with `EXTRACT` inside the block.

   And one more nit while you're in there: please add a changelog entry under _Build, CI, internal_.

With those tweaks this is ready to merge from my side. 🚀

JakobLichterfeld (changes requested):
All three points addressed, and nicely so — the anchored glob with the non-empty sanity assert closes the vacuous-pass gap, and the EXTRACT self-test asserting the block is *flagged* (not just matched) proves the false-negative hole is closed while the tempered token keeps its per-block isolation. Thanks for the quick turnaround!

One tiny thing before merge: the changelog entry ended up at the top of the _Build, CI, internal_ section — entries there are chronological, so it belongs at the end. So rebase to latest main. One-line move, then I'll merge. 🚀

swiffer:
Still not a fan of adding this to the elixir testsuite

**Why not put it in the Elixir app testsuite**

Dashboard SQL checks are an observability/infrastructure concern, not application business logic. Folding them into the Elixir suite couples unrelated things, slows down and complicates the main tests (JSON parsing, macro expansion, extra DB fixtures), and makes failures harder to attribute. Dashboard changes and app-code changes should not share the same test surface.

**Better alternatives**

- Keep a dedicated CI job that only depends on the dashboard JSON + schema fixtures.
- Extract the SQL, expand Grafana macros/variables, then run static checks (sqlfluff + anti-pattern rules) and/or `EXPLAIN` / execution against a test database.
- Optionally expose it as a separate Mix task if you want to stay in the Elixir toolchain, but still run it independently of the main suite.

JakobLichterfeld:
@swiffer first of all: an apology. You raised exactly this concern in your #5438 summary — "static SQL linting of dashboard JSON does not belong in the Elixir/ExUnit suite" — and I still steered the follow-up straight into the ExUnit suite without acknowledging your position, let alone discussing it with you first. That wasn't fair to you, and it's on me. So let me engage with it properly now, because your argument deserves that.

I've genuinely gone back and forth on this, and I agree with more of it than you might expect:

- **Separation of concerns:** you're right that this check has zero dependency on the app code, and that dashboard changes and app changes ideally shouldn't share a test surface. That's a sound principle, not a matter of taste.
- **The sqlfluff / macro-expansion / `EXPLAIN` pipeline:** honestly the best long-term idea on the table — it would catch a whole class of problems this regex never will. I'd love to see that happen as its own project, and I'd support it.

Where I ended up landing differently is on what this check *is*. I'd argue it's not linting: it asserts a semantic invariant — "every latest-position lookup must carry the partial-index predicate" — that exists because the app owns the schema (the partial index from #5438's migration) and the dashboards depend on it. That app↔dashboard coupling isn't created by the test; it's real, and the test only makes it visible. Contract tests for schema invariants are what a test suite is for, which is also why `ensure_linting` would be the wrong home — it checks form, not meaning.

On the practical side, two things took the sting out of your cost concerns for me: the implementation is a plain `async: true` ExUnit file — no DB, no fixtures, milliseconds for ~25 JSONs — and `devops.yml` already runs the Elixir suite on every PR including dashboard-only ones, so a dedicated CI job would give us the same coverage at the price of more workflow infrastructure.

So my call: merge it as-is in the suite, and treat your static-analysis pipeline idea as the real follow-up. And again, sorry for running over your objection instead of having this conversation two PRs ago. 🙏

🤖 drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5592, opened 2026-08-07
fix(charging): fall back to charger_power when phase detection fails

### LINKED ISSUE #4206: Unable to display power using three-phase electricity

### What happened?

Unable to display power using three-phase electricity tesla models 2014 P85

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Relevant log output

```shell
teslamate-1  | 2024-09-16 21:25:57.823 [info] GET /
teslamate-1  | 2024-09-16 21:25:57.830 [info] Sent 200 in 6ms
teslamate-1  | 2024-09-16 21:26:25.391 [info] GET /settings
teslamate-1  | 2024-09-16 21:26:25.402 [info] Sent 200 in 10ms
teslamate-1  | 2024-09-16 21:31:17.481 car_id=1 [info] Fetching vehicle state ...
teslamate-1  | 2024-09-16 21:31:17.684 car_id=1 [info] Start / :asleep
teslamate-1  | 2024-09-16 21:31:17.690 car_id=1 [info] Disconnecting ...
teslamate-1  | 2024-09-16 21:40:59.736 car_id=1 [info] Start / :online
teslamate-1  | 2024-09-16 21:40:59.746 car_id=1 [info] Connecting ...
teslamate-1  | 2024-09-16 21:44:03.409 car_id=1 [info] Suspending logging
teslamate-1  | 2024-09-16 21:53:14.281 car_id=1 [info] Fetching vehicle state ...
teslamate-1  | 2024-09-16 21:55:07.322 car_id=1 [info] Fetching vehicle state ...
teslamate-1  | 2024-09-16 21:55:07.400 car_id=1 [info] Start / :asleep
teslamate-1  | 2024-09-16 21:55:07.408 car_id=1 [info] Disconnecting ...
teslamate-1  | 2024-09-16 22:03:49.092 car_id=1 [info] Start / :online
teslamate-1  | 2024-09-16 22:03:49.105 car_id=1 [info] Connecting ...
teslamate-1  | 2024-09-16 22:03:49.923 [info] GET https://nominatim.dhuar.com/reverse -> 200 (814.300 ms)
teslamate-1  | 2024-09-16 22:03:49.939 car_id=1 [info] Charging / SOC: 72% / Home
teslamate-1  | 2024-09-16 22:03:49.939 car_id=1 [info] Disconnecting ...
teslamate-1  | 2024-09-16 22:12:11.399 [info] GET /
teslamate-1  | 2024-09-16 22:12:11.406 [info] Sent 200 in 7ms
teslamate-1  | 2024-09-16 22:47:49.993 [info] GET /
teslamate-1  | 2024-09-16 22:47:50.006 [info] Sent 200 in 10ms
teslamate-1  | 2024-09-16 22:47:56.158 [info] GET /settings
teslamate-1  | 2024-09-16 22:47:56.166 [info] Sent 200 in 7ms
teslamate-1  | 2024-09-16 23:33:08.559 car_id=1 [info] Charging / Complete / 13.52 kWh – 89 min
teslamate-1  | 2024-09-16 23:33:08.561 car_id=1 [info] Start / :online
teslamate-1  | 2024-09-16 23:33:08.570 car_id=1 [info] Connecting ...
teslamate-1  | 2024-09-16 23:36:12.621 car_id=1 [info] Suspending logging
```

### Screenshots

561726593227_ pic
571726593550_ pic

### Additional data

[charges.csv](https://github.com/user-attachments/files/17032683/charges.csv)

### Type of installation

Docker

### Version

1.30.1

### LINKED ISSUE #5586: `charge_energy_used` drops charge rows when phase detection fails: under-reported or NULL on short and mixed sessions
## What happens

`charge_energy_used` is integrated per charge row, and the branch a row takes depends on whether `charger_phases` is set. When [`determine_phases/1`](lib/teslamate/log.ex#L545) returns `nil`, every row that **does** carry `charger_phases` contributes `NULL` and is then filtered out of the sum entirely. The energy is not mis-weighted — it is missing.

Four ways to reach it:

**1. Short sessions.** The guard in `determine_phases/1` is `n > 15` ([log.ex:559](lib/teslamate/log.ex#L559)), where `n` is the number of charge rows in the process. A session with 15 rows or fewer gets `phases = nil` even when **every** row carries phases — so all of them go `NULL` and `charge_energy_used` ends up empty.

**2. Mixed rows.** A charging process containing both rows with and without `charger_phases`. The rows without phases still contribute through `charger_power`, so the result is understated rather than empty — a plausible-looking number instead of an obvious failure.

**3. A single outlier row, above the guard.** When `n > 15`, the tolerance branch decides, and it decides on an unweighted mean over the whole process. One terminal row where current has collapsed while `charger_power` still reports the old value can push the mean past the 0.3 tolerance and null the entire process. The probability falls off as `1/n`, so short sessions fail deterministically and longer ones fail probabilistically — measured crossover around 22 rows. This path requires the vehicle's reported phase count to differ from the measured ratio; where they agree, the first branch matches first and the outlier is harmless.

**4. A phase count of zero.** The tolerance branch has no lower bound, so when `charger_power` reads 0 across most rows, `p` approaches 0, `round(p)` is 0, and 0 is returned as the phase count. Every row then integrates to zero and the process ends with `charge_energy_used = 0.00` — a number, not a NULL. It passes `where: e.energy_used >= 0` and renders as an answer. This is the least visible of the four.

Edit: A measured case for 3 is in the comments. The process originally cited for 4 has been withdrawn: an independent raw-feed comparison showed its zero-power rows were an ingestion defect in the reporting fork, not vehicle data — with the real values the first detection branch would have matched cleanly (see comments). Mode 4 therefore has no field-measured case at present; it rests on the code path — the tolerance branch genuinely has no lower bound — and a synthetic regression test.

## Evidence

**Mode 1 is measured on a production database (686 charging processes, Owner path: streaming plus `vehicle_data`):**

| category | processes | `charge_energy_used` NULL |
|---|---|---|
| all phases, long | 380 | 3 |
| mixed, long | 129 | 0 |
| no phases at all (DC) | 88 | 0 |
| **short, all rows have phases** | **70** | **70** |
| short, mixed | 19 | 10 |

Every process with 15 or fewer rows where all rows carry `charger_phases` has no `charge_energy_used` at all — 70 of 70, deterministic. That is **70 of 686 processes (10.2 %)** losing the value entirely, 80 counting the short mixed ones.

Three further NULLs sit in the unmixed long category and are not explained by this issue.

**Mode 2 is measured on a reporter's session** in #5558: a 47/70 mixture produced `charge_energy_used` of 3.31 kWh where 5.37 kWh was expected — below `charge_energy_added`, which is the visible symptom. After the missing values were supplied, the integration matched a wallbox meter to within 1 %. The production database above does not independently confirm this mode; its `used < added` cases are spread across categories including unmixed ones, so they have at least one other cause.

## Mechanism

```elixir
# lib/teslamate/log.ex:526
c_if is_nil(c.charger_phases) do
  c.charger_power
else
  c.charger_actual_current * c.charger_voltage * type(^phases, :float) / 1000.0
end
```

- `phases` comes from `determine_phases/1`, which returns `nil` at [log.ex:575](lib/teslamate/log.ex#L575) and [log.ex:579](lib/teslamate/log.ex#L579).
- A row with `charger_phases` set takes the else branch and multiplies by a `NULL` parameter → the whole expression is `NULL`.
- The outer query filters with `where: e.energy_used >= 0` ([log.ex:541](lib/teslamate/log.ex#L541)); `NULL >= 0` is not true, so the row is dropped.
- Rows *without* `charger_phases` still contribute through `charger_power`.
- The same `cond` can also return `0` ([log.ex:571](lib/teslamate/log.ex#L571)): there is no lower bound on the tolerance branch, and the outer guard `p > 0` admits arbitrarily small positive values. A returned `0` produces zeros rather than NULLs, which survive the filter.

The function assumes all rows in a charging process are homogeneous. Nothing enforces that.

## How to check your own database

Read-only. This is the query the table above came from.

```sql
SELECT
  CASE
    WHEN with_phases = 0                         THEN 'no phases at all (DC)'
    WHEN charges <= 15 AND with_phases = charges THEN 'short, all rows have phases'
    WHEN charges <= 15                           THEN 'short, mixed'
    WHEN with_phases < charges                   THEN 'mixed, long'
    ELSE                                              'all phases, long'
  END AS category,
  count(*)                                    AS processes,
  count(*) FILTER (WHERE energy_used IS NULL) AS used_null
FROM (
  SELECT cp.id,
         count(c.*)                                           AS charges,
         count(*) FILTER (WHERE c.charger_phases IS NOT NULL) AS with_phases,
         cp.charge_energy_used                                AS energy_used
  FROM charging_processes cp
  JOIN charges c ON c.charging_process_id = cp.id
  GROUP BY cp.id, cp.charge_energy_used
) s
GROUP BY 1
ORDER BY 2 DESC;
```

## Contributing factors

- **`charger_power` is `:integer`** ([charge.ex:18](lib/teslamate/log/charge.ex#L18)), so the nil-phases branch is systematically the coarser one — reported as ~3 % high for an 11 kW column value against an actual 10.66 kW.
- **`charger_power` is written as `... || 0`** in [`vehicle.ex`](lib/teslamate/vehicles/vehicle.ex#L1706), so a missing power becomes 0 kW and contributes zero energy for that interval rather than being visibly absent.
- **The schema default is defeated at the call site.** `charger_phases` is declared `default: 1` ([charge.ex:16](lib/teslamate/log/charge.ex#L16)), but `vehicle.ex` casts the field explicitly, so an explicit `nil` overrides it.

## Scope

Any installation, any source. The production numbers above come from an installation on the Owner path — streaming plus `vehicle_data` — that has never run Fleet Telemetry.

Mixed rows are made more likely by partial-state sources (see #5583) and by a feed stall with a polling resume, but the short-session path needs neither.

Related to #5583 but distinct: that issue is about building a record from one message instead of from known state; this one is about an aggregate over rows that assumes they are homogeneous. Different cause, different fix, different tests.

## Possible directions

Not deciding here.

1. **Per-row fallback.** When `phases` is nil, treat every row as if it had no phases rather than producing NULL — restores uniform treatment and makes the `n > 15` guard harmless.
2. **Per-row phase determination.** Derive phases from `charger_power * 1000 / (charger_actual_current * charger_voltage)` per row instead of once per process.
3. **Revisit the `n > 15` guard.** It exists to avoid inferring from a noisy handful of samples; with a per-row fallback in place it no longer needs to cost the whole integration.
4. **Never drop rows silently.** Rows contributing NULL should be counted, and a partial result should say so rather than presenting a smaller number as a complete one.
5. **A lower guard.** `round(p)` must never be accepted as a phase count when it is 0 — the smallest safe change of the five.
6. **A robust estimator instead of a looser threshold.** A median over the rows absorbs a single terminal outlier where the mean does not.

Whichever direction is taken, existing rows can be recomputed — `calculate_energy_used/1` is already called on an existing process at [log.ex:515](lib/teslamate/log.ex#L515), so a backfill needs no new machinery.

## Acceptance criteria

- A charging process whose rows all carry `charger_phases` produces a non-NULL `charge_energy_used` regardless of how few rows it has.
- A process with a mixture of rows produces a value that includes every row's contribution.
- Regression tests cover three shapes: all rows with phases and ≤ 15 rows, a mixed process, and a process with no phases at all.
- Processes where phase detection succeeds today produce unchanged results.
- Affected historical rows can be recomputed without a schema change.
- `determine_phases/1` never returns a phase count of 0.
- A single outlier row cannot null the result for an entire charging process.

## Source

Mode 2 reported by @kdjkdjkdj in #5558. Mode 1 measured on a maintainer's production database. Verified in code: `lib/teslamate/log.ex:519-582`, `lib/teslamate/log/charge.ex:15,16,18,19`, `lib/teslamate/vehicles/vehicle.ex:1706`.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Opus 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
Fall back to `charger_power` when phase detection fails, and never accept a phase count of 0. Full analysis and measurements in the linked issue.

Fixes #5586
Fixes #4206

🤖drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 max) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 - fix(geocoder): resolve state for Australian territories (#3868 - mattew124)
 - refactor(vehicles): make the geofence name lookup in the charging log total (#5599 - @JakobLichterfeld)
 - feat: use Grafana 13.1.3 (#5587 - @swiffer)
+- fix(charging): fall back to charger_power when phase detection fails (#5590 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 - refactor(vehicles): make the geofence name lookup in the charging log total (#5599 - @JakobLichterfeld)
 - feat: use Grafana 13.1.3 (#5587 - @swiffer)
 - fix(charging): fall back to charger_power when phase detection fails (#5590 - @JakobLichterfeld)
+- fix(charges): enforce positive charger phases at the database (#5590 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [unreleased]
 
+Existing charging processes are recomputed once during the upgrade migration: previously empty or zero `charge_energy_used` values (short or mixed AC sessions) gain values, the first start after the upgrade can take a few minutes longer on databases with years of history and slow HW, and charge costs are deliberately not changed retroactively.
+
 ### New features
 
 - feat: add service mode to webview and reduce log when car is Unlocked at service mode (#5289 - @NirKli)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,8 +41,8 @@
 - fix(geocoder): resolve state for Australian territories (#3868 - mattew124)
 - refactor(vehicles): make the geofence name lookup in the charging log total (#5599 - @JakobLichterfeld)
 - feat: use Grafana 13.1.3 (#5587 - @swiffer)
-- fix(charging): fall back to charger_power when phase detection fails (#5590 - @JakobLichterfeld)
-- fix(charges): enforce positive charger phases at the database (#5590 - @JakobLichterfeld)
+- fix(charging): fall back to charger_power when phase detection fails (#5592 - @JakobLichterfeld)
+- fix(charges): enforce positive charger phases at the database (#5592 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
@@ -110,6 +110,7 @@
 - fix(dashboards): filter latest-value position panels on complete rows so they use the partial index (#5438 - @swiffer)
 - fix(grafana): Battery Health latest SOC/kWh panels pick the newest UNION row and use `usable_battery_level` on charges (#5438 - @swiffer)
 - fix(grafana): use local calendar for Statistics period end boundaries (#5562 - @wjsall)
+- fix(charge-details): keep power panel in sync with the energy integration (#5592 - @JakobLichterfeld)
 
 #### Translations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - feat: use Grafana 13.1.3 (#5587 - @swiffer)
 - fix(charging): fall back to charger_power when phase detection fails (#5592 - @JakobLichterfeld)
 - fix(charges): enforce positive charger phases at the database (#5592 - @JakobLichterfeld)
+- fix(charging): recalculate charge_energy_used for existing processes (#5592 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/grafana/dashboards/internal/charge-details.json
+++ b/grafana/dashboards/internal/charge-details.json
@@ -1239,7 +1239,7 @@
           "editorMode": "code",
           "format": "time_series",
           "rawQuery": true,
-          "rawSql": "SELECT $__timeGroupAlias(date,$__interval),\r\navg(\r\n    case when charger_phases >= 1 then\r\n    cast(nullif(${determine_phases:sqlstring}, '') as numeric) * charger_actual_current * charger_voltage / 1000.0 else charger_power end) as \"Power\" \r\nFROM charges WHERE charging_process_id = $charging_process_id\r\nGROUP BY 1 ORDER BY 1",
+          "rawSql": "SELECT $__timeGroupAlias(date,$__interval),\r\navg(\r\n    case when charger_phases >= 1 then\r\n    coalesce(cast(nullif(${determine_phases:sqlstring}, '') as numeric) * charger_actual_current * charger_voltage / 1000.0, charger_power) else charger_power end) as \"Power\" \r\nFROM charges WHERE charging_process_id = $charging_process_id\r\nGROUP BY 1 ORDER BY 1",
           "refId": "Current",
           "sql": {
             "columns": [
@@ -1891,12 +1891,12 @@
           "text": "2",
           "value": "2"
         },
-        "definition": "with data as (\n\n    select\n        avg(c.charger_power * 1000.0 / nullif(c.charger_actual_current * c.charger_voltage, 0)) p,\n        avg(c.charger_phases) r,\n        avg(c.charger_voltage) v,\n        count(*) as n\n    from charges c\n    where c.charging_process_id = ${charging_process_id}\n\n)\n\nselect\n    case\n        when p is not null and p > 0 and n > 15 then case\n            when r = round(p) then r\n            when r = 3 and abs(p / sqrt(r) - 1) <= 0.1 then sqrt(r)\n            when abs(round(p) - p) <= 0.3 then round(p)\n            else null\n        end\n        else null\n    end as phases\nfrom data",
+        "definition": "with data as (\n\n    select\n        avg(c.charger_power * 1000.0 / nullif(c.charger_actual_current * c.charger_voltage, 0)) p,\n        avg(c.charger_phases) r,\n        avg(c.charger_voltage) v,\n        count(*) as n\n    from charges c\n    where c.charging_process_id = ${charging_process_id}\n\n)\n\nselect\n    case\n        when p is not null and p > 0 and n > 15 then case\n            when r = round(p) then r\n            when r = 3 and abs(p / sqrt(r) - 1) <= 0.1 then sqrt(r)\n            when round(p) > 0 and abs(round(p) - p) <= 0.3 then round(p)\n            else null\n        end\n        else null\n    end as phases\nfrom data",
         "description": "",
         "hide": 2,
         "name": "determine_phases",
         "options": [],
-        "query": "with data as (\n\n    select\n        avg(c.charger_power * 1000.0 / nullif(c.charger_actual_current * c.charger_voltage, 0)) p,\n        avg(c.charger_phases) r,\n        avg(c.charger_voltage) v,\n        count(*) as n\n    from charges c\n    where c.charging_process_id = ${charging_process_id}\n\n)\n\nselect\n    case\n        when p is not null and p > 0 and n > 15 then case\n            when r = round(p) then r\n            when r = 3 and abs(p / sqrt(r) - 1) <= 0.1 then sqrt(r)\n            when abs(round(p) - p) <= 0.3 then round(p)\n            else null\n        end\n        else null\n    end as phases\nfrom data",
+        "query": "with data as (\n\n    select\n        avg(c.charger_power * 1000.0 / nullif(c.charger_actual_current * c.charger_voltage, 0)) p,\n        avg(c.charger_phases) r,\n        avg(c.charger_voltage) v,\n        count(*) as n\n    from charges c\n    where c.charging_process_id = ${charging_process_id}\n\n)\n\nselect\n    case\n        when p is not null and p > 0 and n > 15 then case\n            when r = round(p) then r\n            when r = 3 and abs(p / sqrt(r) - 1) <= 0.1 then sqrt(r)\n            when round(p) > 0 and abs(round(p) - p) <= 0.3 then round(p)\n            else null\n        end\n        else null\n    end as phases\nfrom data",
         "refresh": 1,
         "regex": "",
         "regexApplyTo": "value",
--- a/lib/teslamate/log.ex
+++ b/lib/teslamate/log.ex
@@ -526,7 +526,10 @@ defmodule TeslaMate.Log do
             c_if is_nil(c.charger_phases) do
               c.charger_power
             else
-              c.charger_actual_current * c.charger_voltage * type(^phases, :float) / 1000.0
+              coalesce(
+                c.charger_actual_current * c.charger_voltage * type(^phases, :float) / 1000.0,
+                c.charger_power
+              )
             end *
               fragment(
                 "EXTRACT(epoch FROM (?))",
@@ -568,7 +571,7 @@ defmodule TeslaMate.Log do
 
             :math.sqrt(r)
 
-          abs(round(p) - p) <= 0.3 ->
+          round(p) > 0 and abs(round(p) - p) <= 0.3 ->
             Logger.info("Phase correction: #{r} -> #{round(p)}", car_id: car_id)
             round(p)
 
--- a/lib/teslamate/log/charge.ex
+++ b/lib/teslamate/log/charge.ex
@@ -13,7 +13,7 @@ defmodule TeslaMate.Log.Charge do
     field :usable_battery_level, :integer
     field :charge_energy_added, :decimal, read_after_writes: true
     field :charger_actual_current, :integer
-    field :charger_phases, :integer, default: 1
+    field :charger_phases, :integer
     field :charger_pilot_current, :integer
     field :charger_power, :integer
     field :charger_voltage, :integer
--- a/priv/repo/migrations/20260807090000_enforce_positive_charger_phases.exs
+++ b/priv/repo/migrations/20260807090000_enforce_positive_charger_phases.exs
@@ -1,0 +1,20 @@
+defmodule TeslaMate.Repo.Migrations.EnforcePositiveChargerPhases do
+  use Ecto.Migration
+
+  def up do
+    # A phase count of 0 (or less) was never meaningful; NULL routes those rows
+    # through the charger_power branch of the energy integration.
+    execute("""
+    UPDATE charges
+    SET charger_phases = NULL
+    WHERE charger_phases <= 0
+    """)
+
+    create(constraint(:charges, :positive_charger_phases, check: "charger_phases > 0"))
+  end
+
+  def down do
+    # The 0 -> NULL correction is not reversible.
+    drop(constraint(:charges, :positive_charger_phases))
+  end
+end
--- a/priv/repo/migrations/20260808090000_recalc_charge_energy_used.exs
+++ b/priv/repo/migrations/20260808090000_recalc_charge_energy_used.exs
@@ -1,0 +1,56 @@
+defmodule TeslaMate.Repo.Migrations.RecalcChargeEnergyUsed do
+  use Ecto.Migration
+
+  # charge_energy_used is a stored aggregate; processes completed before the
+  # phase-detection fallback no longer match what the current formula
+  # computes. Recompute the stored values so every installation converges on
+  # the current formula via the upgrade path. Only rows whose value actually
+  # changes are written, an existing value is never overwritten with NULL,
+  # and cost is left untouched since it may have been edited manually.
+  def up do
+    execute("""
+    WITH detection AS (
+      SELECT c.charging_process_id AS cp_id,
+             avg(c.charger_power * 1000.0 / nullif(c.charger_actual_current * c.charger_voltage, 0)) AS p,
+             avg(c.charger_phases)::int AS r,
+             count(*) AS n
+      FROM charges c
+      GROUP BY 1
+    ), phases AS (
+      SELECT cp_id,
+             CASE WHEN p IS NOT NULL AND p > 0 AND n > 15 THEN
+               CASE WHEN r = round(p) THEN r::float
+                    WHEN r = 3 AND abs(p / sqrt(3.0) - 1) <= 0.1 THEN sqrt(3.0)
+                    WHEN round(p) > 0 AND abs(round(p) - p) <= 0.3 THEN round(p)::float
+               END
+             END AS ph
+      FROM detection
+    ), integrated AS (
+      SELECT c.charging_process_id AS cp_id,
+             CASE WHEN c.charger_phases IS NULL THEN c.charger_power::float
+                  ELSE coalesce(c.charger_actual_current * c.charger_voltage * ph.ph / 1000.0, c.charger_power)
+             END *
+             EXTRACT(epoch FROM (c.date - lag(c.date) OVER (PARTITION BY c.charging_process_id ORDER BY c.date))) / 3600 AS energy
+      FROM charges c
+      JOIN phases ph ON ph.cp_id = c.charging_process_id
+    ), recomputed AS (
+      SELECT cp_id,
+             round(sum(energy) FILTER (WHERE energy >= 0)::numeric, 2) AS used
+      FROM integrated
+      GROUP BY 1
+    )
+    UPDATE charging_processes cp
+    SET charge_energy_used = r.used
+    FROM recomputed r
+    WHERE cp.id = r.cp_id
+      AND r.used IS NOT NULL
+      AND cp.charge_energy_used IS DISTINCT FROM r.used
+    """)
+  end
+
+  def down do
+    # The pre-migration values are not preserved; recomputing with the
+    # previous formula would reintroduce the dropped rows.
+    :ok
+  end
+end
--- a/test/teslamate/grafana/dashboard_queries_test.exs
+++ b/test/teslamate/grafana/dashboard_queries_test.exs
@@ -9,14 +9,18 @@ defmodule TeslaMate.Grafana.DashboardQueriesTest do
   # cannot hide an unindexed lookup.
   @latest_position_block ~r/from positions\b(?:(?!\bfrom\b).)*?order by date desc(?:(?!\bfrom\b).)*?limit 1\b/
 
-  test "latest position queries use complete position rows" do
-    dashboards = Path.wildcard(Path.join(@dashboard_directory, "**/*.json"))
+  # The energy integration falls back to charger_power when phase detection
+  # fails (#5558). Power expressions built on the determine_phases variable
+  # replicate that integration and must carry the same fallback.
+  @phases_power_without_fallback ~r/(?<!coalesce\()cast\(nullif\(\$\{determine_phases:sqlstring\}, ''\) as numeric\) \* charger_actual_current/
 
-    assert dashboards != []
+  test "latest position queries use complete position rows" do
+    queries = dashboard_directory_queries()
+
+    assert queries != []
 
     offenders =
-      dashboards
-      |> Enum.flat_map(&dashboard_queries/1)
+      queries
       |> Enum.flat_map(fn {path, query} ->
         query
         |> unfiltered_latest_position_blocks()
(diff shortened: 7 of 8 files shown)


JakobLichterfeld:
@kdjkdjkdj May I ask you to test this PR on your end? 

Your environment is the only one that reproduces A and B, so once it's in, a re-run against your 24 AC sessions would be the check that matters: the nine short ones should come back with values, and process 384 should stop reporting 0.00.

To test, please see https://docs.teslamate.org/docs/development#testing-with-our-ci-which-builds-the-docker-images-automatically-per-pr

you can use this pr with `pr-5592`.

kdjkdjkdj:
Tested. Result below; one thing up front: **process 384 does not stop reporting 0.00 — and it turns out 384 was never a valid example of that failure mode.** Details in the third section.

### How it was tested

Not with the CI image `pr-5592` — that is plain upstream, and this instance runs a fork (Fleet Telemetry ingestion + Supercharger costs). A `git merge` of the PR branch pulled in 68 files / +4,128 −3,508 including the LiveView 0.20 → 1.2.8 major, which would have made any failure unattributable. So the three PR commits were cherry-picked onto our branch instead: 5 files, +96 −6. `dashboard_queries_test.exs` was left out deliberately — it depends on 5438, which our branch doesn't carry. Suite 666/0.

Migration `20260807090000_enforce_positive_charger_phases` applied in 0.0 s; the `CHECK (charger_phases > 0)` constraint is a no-op here — our `charges` distribution is 1 → 102, 2 → 1,932, NULL → 989, no non-positive values.

One practical note for anyone re-running this: `charge_energy_used` is a **stored** column, so the fix changes nothing on existing rows until `Log.update_energy_used/1` is called per process.

### Scope

"Your 24 AC sessions" — I can't reproduce that number, it isn't defined anywhere I can find. Counted here: 373 charging processes total, of which 332 are 2022 imports with no curve rows at all → 41 with curve rows → **28** carrying a phase value, i.e. AC. (The 686 in the issue table are not from this database.)

### Results — all 28 recomputed

- **15 bit-identical** before/after. No regression on the values that were already plausible.
- **10 previously empty now carry values**: 2 (0.06) · 6 (0.03) · 374 · 379 · 380 (0.02) · 387 · 388 (0.66) · 389 · 392 (0.46) · 393 (0.08). In fairness, four of those land on 0.00 because the session added ≤ 0.01 kWh and genuinely has nothing to integrate.
- **361: 0.00 → 0.19** (0.08 kWh added, 5 rows). This is the clean case — a wrong 0.00 demonstrably disappears.
- **391 stays NULL** — structural, not a defect. It has exactly one curve row, and the integration works over `lag(date)`, so that row has no predecessor. A single-point session isn't integrable.
- **384: 0.00 → 0.00.**

### 384 is not a mode-B case, and the evidence is independent

Our fork keeps a raw shadow recording of the vehicle's Fleet Telemetry stream in a separate table, written by a different code path than `charges`. The same calculation `determine_phases/1` performs, applied to both sources over the process window:

| source | n | rows with `charger_power = 0` | 0 kW **while current flowing** | p | mean phases |
|---|---|---|---|---|---|
| raw feed shadow | 35 | **0** | 0 | **0.8259** | 1.00 |
| `charges` — what the code actually saw | 36 | 34 | **33** | **0.0471** | 1.00 |

There were 2 kW at the charge point throughout. Our `charges` rows say 0.

With the real values, `r == round(p)` → `1 == 1` matches the **first** branch and returns 1 phase cleanly, giving ~0.44 kWh against 0.30 added. The tolerance branch would never have been reached. **384 only became a mode-B case through an ingestion defect on our side** (fixed separately in the fork). The code path you describe is real — the tolerance test genuinely has no lower bound — but this particular process does not demonstrate it, and it is in the issue body as evidence. I'd swap it for 361.

Worth separating explicitly: `charger_power = 0` with **no** current flowing is entirely correct. Only 0 kW while current flows is the defect, and that discriminator is what makes the table above load-bearing.

One behavioural note on 384 after the fix, for genuinely broken input: it still reads 0.00, but by a different route. Before, `abs(round(p) − p) ≤ 0.3` matched with `round(p) = 0` → 0 phases → `amp × volt × 0` = 0. Now `round(p) > 0` blocks that branch, the `cond` falls through to `true → nil`, the product is NULL, and `coalesce` lands on `charger_power` — which is itself 0 here. Same output, different origin. NULL would be the more honest result, but distinguishing "power is really 0" from "power is missing" needs more than the fallback has.

### Forward test (2026-08-08)

A fresh AC session under the patched build, started from sleep: process 394, 10 curve rows over 4 minutes, 11 kW at 16 A / 222 V, phases reported as 2, 0.77 kWh added.

`determine_phases/1` returns nil here deterministically — n = 10, below the `n > 15` guard, so the `cond` is never reached. The log carries no `Phase correction` line, which confirms it.

Replaying both formulas over that process's rows:

| | `charge_energy_used` |
|---|---|
| before the PR (`current × voltage × NULL`) | **NULL** |
| with the PR (`coalesce(…, charger_power)`) | **0.7693** → stored as 0.77 |

`charge_energy_added` for the same session is 0.77. They agree to two decimals, so the fallback isn't just producing a non-NULL number, it's producing the right one. Worth stating that our `charge_energy_added` comes from the fork's own AC source rather than stock TeslaMate, so the two figures are close to independent.

No rows with `charger_power = 0` while current was flowing, and no duplicate timestamps.

That closes the forward check. Summary across both halves: no regression on the 15 sessions that were already correct, 10 previously empty ones recovered, 361 fixed from a wrong 0.00 to 0.19, one structurally unfixable single-row session, and a fresh session that would have been NULL before and now reads correctly.

---

🤖 Measured and drafted with [Claude Code](https://claude.com/claude-code) (Opus 5)

JakobLichterfeld:
Thanks KDJ for the detailed report — genuinely useful data, but first a word on how it can be weighed.

**On the test setup:** a cherry-pick onto a fork with its own ingestion path into `charges`, a held-back LiveView major, and the Grafana guard tests left out is not a verification of this PR — none of that combination exists upstream. Verification runs against stock TeslaMate (the `pr-5592` image and a replay over an Owner-API-fed production database). Your run is therefore recorded as a **supplementary field datapoint**, and as such it is a good one: bit-identical results on every session that was already correct, ten empty sessions recovered, and a forward session that would have been NULL now matching your independently sourced `charge_energy_added` to two decimals.

**On 384: agreed and withdrawn** as evidence for the zero-phase mode. Your shadow-feed table is convincing — the tolerance branch was only reached because the stored rows carried 0 kW *while current was flowing*, i.e. an ingestion defect in your fork, and with the real values the first branch would have matched cleanly at 1 phase. That is also exactly why fork data can't serve as evidence here: the one fork-sourced datapoint in this issue turned out to be an artifact of the fork itself.

**One correction to your proposed swap:** 361 cannot take 384's place. With 5 rows it fails the `n > 15` guard, so the `cond` — and with it the tolerance branch — is never reached; 361 is a measured case of the **short-session** mode, not of the zero-phase mode. I'll strike 384 and list 361 under mode 1. Mode 4 then rests on the code path (which your report confirms is real: the tolerance test has no lower bound) plus the synthetic regression test, with no field-measured case at present.

Your "0 kW while current flowing" discriminator is worth keeping — it cleanly separates honest zero power from missing power. That distinction is input-validation territory, though, not integration logic: for such input the fallback yields 0.00 via `charger_power` rather than NULL, and nothing downstream of ingestion can do better.

The "24 AC sessions" figure is not from the issue or the PR — I can't source it either.

🤖 comment drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 max) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
Verification against the production database the issue's numbers came from (686 charging processes, Owner path, stock TeslaMate): a read-only replay of the old and the new formula over the raw `charges` rows, with the old formula validated against the stored values first so the simulation is provably faithful.

| replica faithful | verdict | processes |
|---|---|---|
| yes | unchanged | 593 |
| yes | recovered (was NULL) | 74 |
| yes | recovered (was 0.00) | 9 |
| no | recovered (was NULL) | 7 |
| no | unchanged | 3 |

- **593 unchanged** — every process where phase detection succeeds is bit-identical: no regression.
- **81 recovered from NULL** — matches the 80 short-session NULLs from the issue table plus part of the "three further NULLs".
- **9 recovered from a wrong 0.00** — the short-session cosmetic (a surviving phase-less 0 kW row), same shape as the two short processes in #4206.
- The 10 replica deviations are values stored by older code versions; three of them are exactly the issue's unexplained long-category NULLs — old and new formula agree there today, so a recompute heals them either way.
- A dedicated scan for mode 4 (`n > 15`, `0 < p ≤ 0.3`) found **no occurrence** across all 686 processes: that branch requires corrupted power data to be reached. The `round(p) > 0` guard stays as defense in depth, covered by the synthetic regression test.

Stored values only change on recompute (`Log.update_energy_used/1` per process); the backfill is a separate decision from this PR.

🤖drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 max) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

kdjkdjkdj:
### On 361 — you are right, and thank you for checking that closely

You are right and I was wrong. 361 has five curve rows; at n = 5 the `n > 15` guard fails, `determine_phases/1` returns nil deterministically, and the `cond` — with it the tolerance branch — is never reached. The row count that refutes my proposal was in my own report; I should have applied the guard I had just finished describing. Mode 1 is the correct filing, and striking 384 while leaving mode 4 without a field case is the honest outcome, even though it leaves the issue thinner.

What the rows do show is the shape you named:

| date | `charger_power` | A | V | `charger_phases` |
|---|---|---|---|---|
| 05:12:53.888 | 0 | 1 | 230 | 2 |
| 05:12:59.111 | 2 | 5 | 231 | 2 |
| 05:13:29.307 | 11 | 16 | 229 | 2 |
| 05:13:59.499 | 11 | 16 | 228 | 2 |
| 05:14:29.703 | 0 | 0 | 2 | — |

Four rows carry phases and integrate to NULL under the old formula; the trailing phase-less 0 kW row survives and sums to zero. That is your "short-session cosmetic (a surviving phase-less 0 kW row)" exactly — the same shape as the nine 0.00 recoveries in your own replay. So 361 is a field instance of that and nothing more.

And thank you for the scrutiny in general: two of the three things I put in front of you last week came back corrected, both times against my own data. That is worth more to me than an easy agreement.

### The setup, rebuilt

Your point about verification was fair, so instead of arguing it I rebuilt the test: stock TeslaMate against a copy of this database, no fork code anywhere in the run.

Image `ghcr.io/teslamate-org/teslamate:pr-5592`, digest `sha256:32bd905d…`, revision `275644a1` — the merge of `c7a6ca49` into master `619f9415`, i.e. the current head including the recalc commit. Own Postgres, no published ports, no `TESLA_*` credentials, `private.tokens` emptied after restore so the instance cannot reach the API.

Input is a dump of this production database from 2026-08-07, taken deliberately **before** the manual recompute I reported on 08-08 — otherwise the migration would have had nothing to do. 373 charging processes. Both migrations ran: `20260807090000_enforce_positive_charger_phases` and `20260808090000_recalc_charge_energy_used`.

### Result over all 373 processes

| class | n | of those from the poll path (< 2026-07-13) |
|---|---|---|
| unchanged | **362** | 351 |
| recovered from NULL | **10** | 3 |
| recovered from 0.00 | **1** | 1 |
| **otherwise changed** | **0** | — |

- **"An existing value is never overwritten with NULL" is verified.** The "otherwise changed" class is empty, and the 332 imports from 2022 — no curve rows at all — still stand on NULL after the migration.
- **The migration reproduces my manual recompute of 08-07 exactly**: same ten IDs recovered from NULL, same single one from 0.00.

That also exercises the fourth commit, which your read-only replay by design could not: here the recalc migration actually ran against a real database.

### Scope, stated plainly

Only **4 of the 11** changed processes (2, 6, 361, 374) predate 2026-07-13, when the charge feed on this instance went live; their `charges` rows come from the stock Owner-API poll path. The other seven have fork-written input rows, even though the code that recomputed them here is stock. `charges` carries no column separating poll rows from feed rows, so the date is the only discriminator available — and that ratio is the ceiling of what this environment can offer, not a stage on the way to something better.

### Variant comparison — your draft wins

Two plausible alternatives to the `charger_power` fallback, replayed over the same rows and measured against `charge_energy_added` (battery side, hence a lower bound for grid-side `used`):

| id | `added` | **PR fix** | per-row phases | trapezoid on `charger_power` |
|---|---|---|---|---|
| 361 | 0.08 | 0.19 | 0.13 | 0.19 |
| **388** | 0.60 | **0.66** | **0.43** | 0.66 |
| **392** | 0.44 | **0.46** | **0.30** | 0.46 |
| 393 | 0.07 | 0.08 | 0.07 | 0.08 |

**Using the per-row reported phases instead of falling back to `charger_power` is worse.** At 388 and 392 it lands *below* the battery side — impossible, since `used` is grid-side and charging has losses. So `charger_phases = 2` is demonstrably wrong on those rows while `charger_power` carries the right information. That is exactly the case the fallback exists for.

**And the trapezoid rule changes nothing.** On the three sessions with real energy it is bit-identical to the PR; on the near-zero ones it differs by 0.01–0.02, which is noise at that scale.

**Withdrawing a claim of my own:** I had floated that the integration overcounts at ramp-up, because the current row's power is charged to the *preceding* interval. That holds for a single interval, but it undercounts symmetrically at shutdown — the final 0 kW row erases its own interval — and the two cancel. The trapezoid column above is the measurement that killed it. Retracted.

One caveat on 361 that the table above should not hide: 11 kW reported at 16 A / 229 V, while the battery took 0.08 kWh over 96 s, i.e. about 3 kW average. `charger_power` itself looks overstated on that session, so I would not hold 0.19 up as the correct value — only as no longer wrong in kind. The 0.00 it replaced was wrong for a reason the PR fixes; what the right number is, this database cannot say.

🤖 Measured and drafted with [Claude Code](https://claude.com/claude-code) (Opus 5)

JakobLichterfeld:
Thanks KDJ — this is the rebuilt setup that actually answers the objection: stock image, both migrations executed at boot against a real pre-recompute dump, and the result observed rather than simulated. The empty "otherwise changed" class verifies the never-overwrite-with-NULL guarantee on a live database, and reproducing your manual recompute ID-for-ID closes the one box the read-only replay could not reach by design: the migration's runtime path. Together with the 686-process replay on the Owner-path production database, the verification record for this PR is complete from my side.

361 is settled as a mode-1 filing — and the direct concession is appreciated; checked disagreement is worth more here than easy agreement, in both directions.

On the variant comparison: useful, with one evidentiary note that follows from your own scope statement. 388 and 392 carry feed-written input rows, so the "per-row phases lands below the battery side" result is indicative rather than upstream evidence — held to the same standard as 384 was. The standing upstream justification for the fallback doesn't need it anyway: reported phase counts being unreliable is the reason `determine_phases/1` exists at all. The trapezoid result (bit-identical where energy is real) and your retraction of the ramp-up claim are noted — no change to the integration is warranted.

Your caveat on 361 is fair and worth keeping visible: 0.19 is no longer wrong *in kind*, while `charger_power` coarseness on short ramping sessions remains the known contributing factor from the issue — out of scope for this PR, in scope for the issue's record.

🤖drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 max) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
So ready to review @brianmay if you find the time.

brianmay:
AI summary from Big Pickle (free open code). I haven't yet reviewed any of this yet, but want to copy this before I loose the markdown.

> I've reviewed PR 5592 end-to-end: the runtime formula change in `log.ex`, the schema default removal, both migrations, the dashboard query/variable changes, and all new tests (verifying the test arithmetic by hand). Summary of findings:
> 
> ## Findings
> 
> **1. (Low-moderate, operational) The recompute migration is a full-table operation that runs at startup**
> `recalc_charge_energy_used` scans, groups, and window-sorts the entire `charges` table (`lag() OVER (PARTITION BY charging_process_id ORDER BY date)` + hash/sort group by), then UPDATEs `charging_processes`. On large installs (millions of charge rows) this runs long and delays app boot (migrations run before the app starts, so no concurrent-write risk, but it will extend startup/healthcheck time). Not a correctness issue; worth a release-note caveat or batching for very large DBs.
> 
> **2. (Minor) Elixir `round/1` vs Postgres `round(float8)` disagree at half-integer `p`**
> `lib/teslamate/log.ex:571` and the migration both use `round(p)`, but Elixir rounds half away from zero while Postgres `round(double precision)` rounds half to even. At exactly `p = 0.5` the new `round(p) > 0` guard returns 1 phase at runtime but NULL in the migration; at `p = 2.5` (e.g. a process with equal 2-phase/3-phase rows) the migration's `r = round(p)` branch triggers where runtime falls back to `charger_power`. So the migration's "converge on the current formula" guarantee has a small divergence for contrived inputs. Same applies to `avg(c.charger_phases)::int` truncation vs the SQL cast Ecto emits for `type(:integer)` — equivalent, so fine there.
> 
> **3. (Minor, behavior change) Removing the `charger_phases` default of 1 reclassifies AC/DC in existing dashboards**
> With `field :charger_phases, :integer, default: 1` removed (`log/charge.ex:16`), rows where the API omits `charger_phases` now store NULL instead of 1. The `NULLIF(mode() within group (order by charger_phases), 0)` DC/AC classification in `charges.json`, `trip.json`, `charging-stats.json`, `battery-health.json` will now label such processes as DC where they previously showed AC. Likely intended (the fabricated "1" is the bug being fixed) and arguably more correct, but it's a visible change to existing dashboards that the PR doesn't call out. The MQTT/HA `charger_phases` sensor will also report null more often.
> 
> ## Verified sound
> 
> - **The `coalesce` fallback is consistent** across runtime (`log.ex:528-533`), migration, and dashboard panel — all fall back to `charger_power` when phase detection yields nil, and only for rows where detection failed; rows with nil `charger_phases` keep the power branch. First-row-in-process NULL energy handling matches (`energy >= 0` / `FILTER (WHERE energy >= 0)`).
> - **The `round(p) > 0` guard** only suppresses the `p < 0.5` → 0-phase result; no legitimate 1-phase case is affected (p ∈ [0.7, 1.3] still resolves to 1).
> - **New test expected values check out**: 1.28 (11 × 7/60), 0.32 (19 × 1/60), 2.22 (19 × 7/60); the zero-phase test fails on the old code and the outlier test fails without the coalesce.
> - **Dashboard test regex** uses a fixed-length PCRE lookbehind `(?<!coalesce\()` — valid, correctly rejects the fixed form, catches the unfixed form; the `normalize`/`dashboard_directory_queries` refactor is behavior-preserving.
> - **Migration ordering** is correct (0s nulled before recompute), the `IS DISTINCT FROM` guard prevents redundant writes, and cost is correctly left untouched.

brianmay (approved):

JakobLichterfeld:
Thanks for the thorough pass — the "verified sound" section matches my own verification point for point, including the detail that the migration *ordering* (zeros nulled before the recompute) is load-bearing.

On the three findings:

**1. Startup cost — agreed; it's the one actionable item.** The recompute is a single full pass over `charges`, runs before app start (no concurrent-write exposure), and extends one boot once — minutes on small hardware carrying years of data. The changelog now calls out the one-time recompute. Batching would not reduce the total work and only adds failure states to a one-shot migration, so it stays a single statement.

**2. The half-integer rounding divergence doesn't exist — the premise is off.** In the migration `p` is `numeric`, not `double precision`: `charger_power * 1000.0` starts a numeric chain that survives `avg()`, and Postgres `round(numeric)` rounds half away from zero — the same tie-breaking as Elixir's `round/1`. (Two nits in the finding itself: the `p = 2.5` example conflates `p`, the power ratio, with `r`, the phase average; and `numeric::int` rounds, it doesn't truncate.) The residual theoretical gap — float8 accumulation at runtime vs. numeric in the migration at branch boundaries — is empirically excluded: the migration reproduced an Elixir-side recompute ID-for-ID across 373 real processes, and the same SQL matched stored values on 676/686 processes of a second production database with every deviation accounted for.

**3. The default removal changes no stored value on any real write path.** The struct default only fires when the attrs map lacks the key entirely — and the only caller, `vehicle.ex`, always passes it, so an API-omitted phase count arrived as an explicit `nil` and stored NULL before this PR exactly as after it. That is literally contributing factor #3 in #5586 ("the schema default is defeated at the call site"); there is no second changeset caller (the importer maps the TeslaApi struct). Dashboards therefore see no reclassification, and MQTT is untouched either way — `charger_phases` publishes from the vehicle summary (TeslaApi state), not from the Ecto schema. The new "does not invent a phase count" test pins the missing-key case precisely because no production path takes it.

Issue #4234, opened 2024-09-30
ci: run nix checks


JakobLichterfeld:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Our CI should run `nix flake check`

### Relevant log output

```shell
n.a.
```

### Type of installation

Docker

### Version

v1.30.1

JakobLichterfeld:
Command should be: `nix flake check --override-input devenv-root "file+file://"<(printf %s "$PWD")`

JakobLichterfeld:
Perhabs all-systems is a good addition, at least on darwin: `nix flake check --override-input devenv-root "file+file://"<(printf %s "$PWD") --all-systems`

JakobLichterfeld:
The caching limit which lead to PR #5440 to solve the slow linting check in CI must be considered.

PR #5593, opened 2026-08-08
build(nix): reduce the manually pinned hashes to one and automate it

JakobLichterfeld:
Every mix or npm dependency bump invalidated one or more hashes in nix/flake-modules/package.nix, and each one had to be looked up from a failing build and pasted back by hand. Three changes shrink that to a single hash and take the pasting out of it.

- [x] Fetch the npm dependencies with pkgs.importNpmLock instead of buildNpmPackage and drop the now unneeded hash.
- [x] Derive the cldr revision from the ex_cldr version in mix.lock, this makes the two impossible to drift apart, and a lockfile without ex_cldr now aborts the evaluation instead of falling back to something plausible. 
- [x] Add nix run .#update-nix-hashes for what is left. It builds, reads the hash mismatch Nix reports, writes it back and repeats until the build succeeds.
- [x] CI: Auto update hashes in nix

### DIFF
--- a/.github/workflows/buildx.yml
+++ b/.github/workflows/buildx.yml
@@ -1,7 +1,6 @@
 name: Publish Docker images
 
 on:
-  workflow_dispatch:
   workflow_call:
   schedule:
     - cron: "0 3 * * *"
@@ -13,6 +12,12 @@ on:
       - "!.github/**" # Important: Exclude PRs related to .github from auto-run
       - "!.github/workflows/**" # Important: Exclude PRs related to .github/workflows from auto-run
       - "!.github/actions/**" # Important: Exclude PRs related to .github/actions from auto-run
+  # Pull requests touching .github are excluded from the automatic runs above,
+  # since the workflow definition that would run is the one from the pull
+  # request. This is the deliberate way to build them anyway, by someone with
+  # write access. The job conditions below allow for it. Note that this
+  # publishes: a run from a branch overwrites the edge tag.
+  workflow_dispatch:
 env:
   REGISTRY_IMAGE: teslamate/teslamate
 
@@ -29,7 +34,7 @@ jobs:
 
   teslamate_build:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     strategy:
       fail-fast: false
       matrix:
@@ -62,7 +67,7 @@ jobs:
     needs:
       - check_paths
       - teslamate_build
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     steps:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
@@ -83,7 +88,7 @@ jobs:
             type=edge
   grafana:
     needs: check_paths
-    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule'
+    if: needs.check_paths.outputs.githubfolder == 'false' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-24.04
     timeout-minutes: 10
     steps:
--- a/.github/workflows/devops.yml
+++ b/.github/workflows/devops.yml
@@ -1,7 +1,6 @@
 name: DevOps
 
 on:
-  workflow_dispatch:
   push:
     branches: ["main"]
     paths:
@@ -16,6 +15,11 @@ on:
       - "!.github/**" # Important: Exclude PRs related to .github from auto-run
       - "!.github/workflows/**" # Important: Exclude PRs related to .github/workflows from auto-run
       - "!.github/actions/**" # Important: Exclude PRs related to .github/actions from auto-run
+  # Pull requests touching .github are excluded from the automatic runs above,
+  # since the workflow definition that would run is the one from the pull
+  # request. This is the deliberate way to run the chain for them anyway, by
+  # someone with write access. The job conditions below allow for it.
+  workflow_dispatch:
 
 # Cancel any in-progress runs when a new commit is pushed
 concurrency:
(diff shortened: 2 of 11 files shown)


brianmay:
So if I understand this correctly, it looks like you have found a way to not require the npm hash. Wow! Looks like I am going to have to update my personal projects now :-)

brianmay:
AI feedback (opencode Big Pickle free model).

I think most of these are not relevant in anyway. 1, 2, and 5 are probably not relevant. Maybe point 3 might matter. Don't understand point 4.

> I've completed my analysis, including running the update-hashes script end-to-end against a real Nix flake. Here's the review.
> 
> ---
> 
> ## Review: PR #5593 — build(nix): reduce the manually pinned hashes to one and automate it
> 
> ### Verified correct
> - The `cldrVersion` regex correctly extracts `2.47.5` from the real `mix.lock`; tag `v2.47.5` exists upstream and the pinned SRI `sha256-nGVuv8…=` matches the actual tarball. The `rev`/hash cannot drift anymore.
> - The `update-nix-hashes` script works. I ran it end-to-end on a test flake with real Nix (2.34): it finds pins, fakes each hash, parses the `specified:`/`got:` pair out of multi-error logs, writes the real hash back, restores on failure, and reports "unchanged" when up to date. The awk parser handles interleaved `--keep-going` errors correctly (verified against real output).
> - `importNpmLock`/`buildNodeModules` exist in the pinned nixpkgs (`fd14620`), the lockfile is v3, the only packages without integrity are the three phoenix `file:` deps, and all three are overridden. The `postInstall` bin symlink is needed (buildNodeModules doesn't create `$out/bin`) and runs via `runHook postInstall`.
> - The CI patch-validation logic (apply-to-index → only-M-status → only-`sha256-…`-literal diffs → path under `nix/*.nix`) is sound and appropriately defensive.
> 
> ### Findings
> 
> **1. The biggest functional change is not validated by any CI run (Medium).** Nothing in CI executes `nix build .#default` for this PR: `ghcr_build.yml` builds via the Dockerfile (mix release, not nix), `ensure_linting` only runs `nix run .#lint`, and the new workflow only triggers on `mix.exs`/`mix.lock`. The switch from `buildNpmPackage` to `importNpmLock.buildNodeModules` — with the phoenix `file:` overrides and the npm-install-offline flow — has not been exercised anywhere except (presumably) the author's local build. I reviewed it statically and it looks correct, but I couldn't run the full build here. Please confirm `nix build .#default` passes against the pinned nixpkgs before merging, or this could be the first run that ever builds it.
> 
> **2. The auto-repair loop needs `GH_TOKEN_FOR_UPDATES` provisioned as a Dependabot secret (Medium).** This secret is not used anywhere else in the repo. Without it, the `commit` job fails (cleanly, via the `: "${PUSH_TOKEN:?…}"` guard), but then `verify` fails on every dependency PR with nothing ever pushed to fix it — dependency PRs get permanently blocked with a confusing error. Worth confirming it exists (and that its token has push rights to the repo) before merge.
> 
> **3. `verify`/`compute` now gate every PR that touches `mix.exs`/`mix.lock` on the nix build succeeding (Low, by design).** A Dependabot bump that breaks the nix build for a non-hash reason (compile error, unavailable toolchain) makes `compute` fail with the raw nix error and no actionable hint; the `verify` failure message only mentions stale hashes. Acceptable, but be aware every mix dependency PR becomes dependent on the nix build.
> 
> **4. The script aborts if two distinct pins end up with the same hash value mid-run (Low, theoretical).** I reproduced this: once pin 1 is rewritten, if its new hash equals pin 2's pinned hash, the "pins … on N lines" guard fires and the whole run fails (with restore). `mixFodDeps` and `cldr` can never collide in practice, so this is just a note about the guard's edge.
> 
> **5. Cost of every invocation (Low, documented).** Each run re-fetches every pinned hash and performs up to 3 full `nix build .#default` runs. On a cold cache the 45-minute `compute` budget may be tight; on mix.lock-touching PRs the cache key intentionally misses, so Dependabot mix PRs will be slow.
> 
> ### Nits
> - The `- - docs:` → `- docs:` fix in CHANGELOG.md is unrelated to this PR's purpose, but trivially correct.
> - `update-flake-lock.yml` and `ensure_linting.yml` now bust the nix cache on any `mix.lock` change even when the nix inputs didn't change — consistent with the intent, just a behavior change to be aware of.
> 
> No blocking correctness bugs found in the automation logic; finding 1 is the one I'd want resolved before merge.

brianmay (approved):

brianmay:
For some reason I thought CI did a "nix flake check" somewhere. Which would have addressed point 1. Either i am getting my projects confused or we removed it at some point.

JakobLichterfeld:
> So if I understand this correctly, it looks like you have found a way to not require the npm hash. Wow! Looks like I am going to have to update my personal projects now :-)

Yeah, you are right. Thanks for the kudos. Sumbled across `pkgs.importNpmLock` and gave it a try:

> Fetch the npm dependencies with pkgs.importNpmLock instead of buildNpmPackage. The integrity hashes in assets/package-lock.json are already authoritative, so npmDepsHash is gone entirely and npm bumps no longer touch the Nix code at all. The phoenix, phoenix_html and phoenix_live_view entries are file: links into deps/, which only exists in a working tree after mix deps.get, so packageSourceOverrides points them at mixFodDeps. That also replaces the previous approach of letting npm install them and swapping them for symlinks afterwards, so the substitution is now declarative instead of a correction after the fact.

JakobLichterfeld:
> For some reason I thought CI did a "nix flake check" somewhere. Which would have addressed point 1. Either i am getting my projects confused or we removed it at some point.

In this project we are not yet running `flake check` in CI, mainly due to slowness due to too big caches, see #4234. I will look more into it after this one. I solved it in one of my projects and even run an installation in VM automatically, so perhaps we can adopt it.

brianmay:
> In this project we are not yet running flake check in CI, mainly due to slowness due to too big caches, see https://github.com/teslamate-org/teslamate/issues/4234. I will look more into it after this one. I solved it in one of my projects and even run an installation in VM automatically, so perhaps we can adopt it.

Yes, thought that might be it. nix flake checks tend to be really slow in my projects too.

JakobLichterfeld:
> AI feedback (opencode Big Pickle free model).
> 
> I think most of these are not relevant in anyway. 1, 2, and 5 are probably not relevant. Maybe point 3 might matter. Don't understand point 4.

Thanks for the review — the end-to-end run against real Nix 2.34 is the useful part here: it covers the interleaved `--keep-going` logs, which I had only tested synthetically.

**1** — it isn't unbuilt: `nix build .#default` is green and produces byte-identical asset digests to the previous build. Two things are true though: it has only ever built on `aarch64-darwin`, while the NixOS module and the VM test target Linux; and no CI ran on this PR at all, because `devops.yml` deliberately excludes PRs touching `.github/**`. That is the security rule working as intended, not a gap in job coverage. A follow-up commit makes this workflow honour it too: the same `.github` path exclusions as in `devops.yml`, plus the `check_paths` gate on the `compute` job, which is what covers a PR carrying both a dependency change and a workflow change. Such PRs get a `workflow_dispatch` trigger instead; a manual run only verifies, because the `commit` job's actor check does not hold for it. I'll use it to confirm the Linux build before merging. We don't run `flake check` in CI yet because of cache size (#4234); I'll revisit that separately.

**2** — correct, and a merge precondition: `GH_TOKEN_FOR_UPDATES` has to exist as a **Dependabot** secret, not an Actions secret. I'll set that up before merging, just wanted to get feedback beforehand.

**3** — the mechanism doesn't apply. A non-hash build failure fails `compute` with the script's own message plus the full log; `verify` has `needs: compute` and is therefore skipped, so it never claims "stale hashes". What remains is that mix PRs now depend on the nix build — which is the intent. I do want to have Elixir build and nix build always green no matter what.

**4** — the guard only fires if two pins hold the same literal at the same moment; `mixFodDeps` and `cldr` cannot collide. It then aborts and restores instead of rewriting both lines, which is exactly its job. So imo no issue here.

**5** — not three full builds: the in-loop builds stop at the fixed-output mismatch, because the fetchers are inputs and are built before the package. It's one full build plus the re-fetches per run.

JakobLichterfeld:
Yiha, works as designed:

First failded as Dependabot didn't update the nix hashes:
image

Then correct auto-commit: https://github.com/teslamate-org/teslamate/pull/5570/changes/9f0c0115d1c01a2a79d804bb68db0cd3db6135f3

Check succeeded:
image

PR #5594, opened 2026-08-09
feat(vehicle): allow widening the Sentry Mode polling interval via POLLING_SENTRY_INTERVAL

onevcat:
## Problem

While Sentry Mode keeps the vehicle awake, `try_to_suspend` schedules the next `vehicle_data` fetch at a hard-coded `30 * i` — 30 s in polling-only mode, 60 s with streaming enabled. No `POLLING_*` schedule variable controls this branch.

On the billed Fleet API this makes Sentry parking the single most expensive routine state. At Tesla's ~$0.002/request, a Sentry-parked hour costs ~$0.12 in `vehicle_data` calls; roughly three Sentry-parked hours per day exceeds the entire $10 monthly free tier on their own — more than driving itself under typical polling configs, where streaming carries the drive data.

Measured on a real car (v4.0.1, Fleet Telemetry streaming enabled): the developer billing counter advanced ~1 request/minute while parked with Sentry on, and stopped within minutes of turning Sentry off. Streaming records do not trigger fetches in `:online`, so this polling cadence is the entire Sentry-parked cost.

`POLLING_MINIMUM_INTERVAL` can floor this cadence, but it floors *every* scheduled fetch — the post-wake first fetch, gear/charge-detection fetches and retry paths included — so using it to tame Sentry has global side effects.

## Change

A new `POLLING_SENTRY_INTERVAL` environment variable (seconds, default `0`). The Sentry branch now schedules `max(30 * i, sentry_interval())`:

- Unset or `0`: behavior is exactly as before.
- Set: widens only the Sentry-mode cadence; values below the built-in `30 * i` are ignored, so it can never tighten polling.

The trade-off is documented: a wider interval delays noticing that Sentry Mode was turned off (and therefore the start of the suspend flow) by up to the chosen value.

Includes a test mirroring the existing "does not suspend if sentry mode is active" case with the variable set, and a docs table entry.


onevtail:
I rechecked the state-machine path and found an important caveat in the current documentation. In polling-only mode, increasing `POLLING_SENTRY_INTERVAL` delays the next full `vehicle_data` fetch, so it can delay detection not only of Sentry Mode being disabled, but also of driving and charging transitions. Large values may therefore cause trip or charge logging to start late or miss data.

The implementation still preserves the existing minimum cadence—30 seconds for polling-only and 60 seconds with streaming—even when the variable is unset, zero, or below those thresholds. Before this PR is ready, we should document the broader tradeoff and add regression coverage for both minimum intervals, while restoring any pre-existing environment-variable value during test cleanup.

onevtail - an assistant to @onevcat

PR #5579, opened 2026-08-02
build: use Elixir 1.20.2 OTP 29

swiffer:
## Summary

Full Elixir **1.20.2 / OTP 29** enablement for official runtime, CI, and Nix, superseding the Dependabot-only Dockerfile bump in #5564.

**Status:** Ready for review — rebased on latest `main`, CI green, local validation done.

### Version bump
- `Dockerfile`: builder image `elixir:1.20.2-otp-29`
- `.github/actions/setup-elixir-and-cache-deps`: defaults `1.20.2` / `29`
- `nix/flake-modules/package.nix`: `erlang_29` + `elixir_1_20`
- `mix.exs`: require `~> 1.19` (allows 1.19.x and 1.20.x; official pin is 1.20.2)
- `website/docs/development.mdx`: contributor pin and supported range
- `CHANGELOG.md`: unreleased build note

### Required test fix (`LogMock`)
While running tests on Elixir 1.20, charging-related tests failed with:

```text
** (Protocol.UndefinedError) protocol String.Chars not implemented for
   Ecto.Association.NotLoaded
   #Ecto.Association.NotLoaded<association :geofence is not loaded>
   lib/teslamate/vehicles/vehicle.ex (Enum.join of geofence name in charging log)
```

**Finding:** production `Log.start_charging_process/3` preloads `:address` and `:geofence` (so they are `nil` or a struct). `LogMock` returned a bare `%ChargingProcess{}`, leaving associations as `NotLoaded`.

Vehicle logging does roughly:

```elixir
with(%GeoFence{name: name} <- cproc.geofence, do: name)
# mismatch on NotLoaded returns the NotLoaded value (not nil)
|> Enum.join(...)  # String.Chars → crash
```

**Fix:** set `address: nil` and `geofence: nil` on the mock charging process so it matches preloaded production data.

Related: #5564 (Dependabot Dockerfile-only bump to 1.20.x / OTP 29).

## Test plan
- [x] CI green on this PR (`elixir_test`, static analysis, image build)
- [x] Charging tests previously failing on 1.20.2-otp-29 pass after the LogMock fix (local)
- [x] End-to-end validation on my deployment/setup
- [ ] Confirm Nix package still evaluates with `elixir_1_20` / `erlang_29` on current flake pin

---

🤖 Assisted by Grok 4.5 (xAI) via Grok Build (Elixir 1.20 bump, LogMock fix, PR description).

PR #5598, opened 2026-08-09
fix(test): override meck to 1.2 for OTP 29 compatibility

swiffer:
## Summary

Follow-up to #5579: the test suite fails to compile on **Elixir 1.20 / OTP 29** because transitive `meck 0.9.2` (via `mock`) rejects deprecated bare `catch` syntax.

This was not caught in #5579 CI because not all jobs ran after the GitHub Actions toolchain change (see [comment](https://github.com/teslamate-org/teslamate/pull/5579#issuecomment-5230844985)).

**Fix:** pin/override `meck` to `~> 1.2` for `:test` only until `mock` accepts a newer meck.

Upstream: https://github.com/jjh42/mock/issues/153

## Changes
- `mix.exs`: `{:meck, \"~> 1.2\", only: :test, override: true}`
- `mix.lock`: `meck 0.9.2` → `1.2.0`

## Test plan
- [ ] CI `elixir_test` green on OTP 29
- [ ] `mix deps.compile meck` succeeds locally on 1.20.2-otp-29


JakobLichterfeld (approved):
lgtm, ty for the quick fix! Will merge if CI is green.

PR #5600, opened 2026-08-09
fix(vehicle): shorten the streaming suspend probe interval from 30 to 10 minutes

onevcat:
## Problem

With `use_streaming_api` enabled, both suspend paths hard-code a **30-minute** probe interval (`{3, 30, 2}` in `try_to_suspend`, and the manual `suspend_logging` path), ignoring the stored `suspend_min`. That value was a precaution from the polling-only era, where the suspended-state probe could escalate to a billed `vehicle_data` call that also reset the car's own idle timer.

Under streaming neither concern applies: the suspended-state fetch takes the `allow_vehicle_data? = false` path and issues only `GET /api/1/vehicles/<id>` — free, served from Tesla's cache, and incapable of keeping the car awake. The stream stays subscribed throughout, so drive and charge starts are detected instantly regardless of this timer. The fixed 30 minutes therefore only delays *recording* the sleep transition: a car that falls asleep early in the window is reported `online` for up to ~20 extra minutes, systematically inflating online time in the statistics.

## Change

As suggested in review, simply lower the hard-coded streaming probe interval from 30 to 10 minutes — no new configuration. Both the automatic (`try_to_suspend`) and the manual (`suspend_logging`) paths change; the non-streaming path is untouched and continues to honor the stored `suspend_after_idle_min` / `suspend_min`.

*(An earlier revision of this PR added a `POLLING_SUSPENDED_INTERVAL` environment variable instead; reworked per review feedback — the env var, its docs table entry and its parsing test are gone.)*

## Field results

Before this rework we ran the earlier env-var revision on our own instance (self-hosted Fleet Telemetry + streaming API) with the probe interval set to 5 minutes, continuously since 2026-08-10:

- Sleep transitions are recorded within ~5 minutes of the car actually falling asleep.
- No additional billed traffic appeared on the Fleet API usage/billing page — the probe stays on the free vehicle-list endpoint, as analyzed.
- No effect on the car's ability to reach deep sleep.

10 minutes is the conservative choice suggested in review (144 instead of 48 probes/day/car) and is strictly better than the current 30 minutes for everyone, with no knob to explain.

## Notes

Existing suspend/streaming tests assert state transitions rather than the timer value, so no test changes are needed. CHANGELOG entry included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Improvements and bug fixes
 
+- fix(vehicle): shorten the streaming suspend probe interval from 30 to 10 minutes (#5600 - @onevcat)
+
 #### Build, CI, internal
 
 #### Dashboards
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -342,7 +342,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
 
       suspend_min =
         case {data.car.settings, streaming?(data)} do
-          {%CarSettings{use_streaming_api: true}, true} -> 30
+          {%CarSettings{use_streaming_api: true}, true} -> 10
           {%CarSettings{suspend_min: s}, _} -> s
         end
 
@@ -1736,7 +1736,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
   defp try_to_suspend(vehicle, current_state, %Data{car: car} = data) do
     {suspend_after_idle_min, suspend_min, i} =
       case {car.settings, streaming?(data)} do
-        {%CarSettings{use_streaming_api: true}, true} -> {3, 30, 2}
+        {%CarSettings{use_streaming_api: true}, true} -> {3, 10, 2}
         {%CarSettings{suspend_after_idle_min: i, suspend_min: s}, _} -> {i, s, 1}
       end
 


JakobLichterfeld (changes requested):
Hi Wei Wang, thanks for the well-researched PR — the analysis is spot on: with an active stream the suspended-state probe takes the `allow_vehicle_data? = false` path and only issues `GET /api/1/vehicles/<id>`, so the fixed 30 minutes is purely reporting latency. Drive and charge starts are detected instantly by the stream handlers while suspended, independent of this timer.

That said, I'd like to solve this without a new environment variable. Every knob we add generates support tickets and confusion about what it actually does — and this one would need a long explanation (only applies while streaming, floored to a minute, seconds vs. minutes, …).

Your own analysis points to a simpler conclusion: if the probe is genuinely cheap and cannot keep the car awake, then a shorter interval is better for *everyone*, and the right fix is to just lower the hard-coded value in the streaming path — e.g. from 30 to 10 minutes — with no configuration at all. The 30 minutes was a precaution from the polling-only era; under streaming there's nothing left for the user to tune. The polling (non-streaming) path and its existing settings remain untouched.

Before merging I'd want to be confident the probe is really cost-free across setups (Fleet API quotas/billing, proxy setups), which is also why I'd start conservatively at 10 minutes (144 instead of 48 requests/day/car) rather than 5.

Would you be open to reworking the PR into that one-line change? The docs table entry and the env-var parsing/test would then no longer be needed. Happy to review again.

---

onevcat:
@JakobLichterfeld Thanks for the review — agreed on all points, and reworked in place as suggested: the env var is gone, and the streaming path now simply uses 10 minutes in both `try_to_suspend` and the manual `suspend_logging` path. The docs table entry and the parsing test are dropped; the CHANGELOG entry remains. Title and description are updated to match.

One data point on the cost question: we had been running the previous env-var revision on our own instance (self-hosted Fleet Telemetry + streaming API) with the probe at 5 minutes, continuously since 2026-08-10. Sleep transitions were recorded within ~5 minutes, the car reached deep sleep normally, and the Fleet API usage page showed no additional billed traffic — the probe stayed on the free vehicle-list endpoint as analyzed. So the conservative 10 minutes should be comfortably safe.

Ready for another look whenever you have time.

JakobLichterfeld (approved):
Thanks for the quick rework, exactly what I had in mind — LGTM. 🚀

---

PR #5615, opened 2026-08-15
test: stop the app in test_helper instead of relying on --no-start

swiffer:
`mix test` is green here, but the same suite fails in VS Code via ElixirLS Test Explorer.

ElixirLS starts the full TeslaMate app and cannot pass `--no-start` (our mix alias). Tests that `start_supervised` Endpoint / Repair / MQTT / Terrain then hit already-started processes.

Stop the app in `test_helper.exs` so both runners get the same Repo + PubSub tree. `mix test` is unchanged (476 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - build(deps): bump tzdata from 1.1.3 to 1.1.4 (#5614 - @mews-se)
 - ci: purge orphaned GHCR attestation referrers (#5622 - @swiffer)
+- test: stop the app in test_helper instead of relying on --no-start (#5615 - @swiffer)
 
 #### Dashboards
 
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -1,5 +1,12 @@
-# FIXME Workaround to apply the logger config when running the tests with
-# "--no-start". This looks like a bug in Elixir v1.11.0
+# ElixirLS Test Explorer starts the full app (it cannot pass --no-start).
+# Tear it down so tests get the same Repo + PubSub tree as `mix test`.
+case Application.stop(:teslamate) do
+  :ok -> :ok
+  {:error, {:not_started, :teslamate}} -> :ok
+end
+
+# `mix test --no-start` only loads config; Mix restarts Logger when it starts
+# the app. Bounce it here so config/test.exs (level: :warning) takes effect.
 Application.stop(:logger)
 Application.start(:logger)
 
@@ -9,8 +16,8 @@ for app <- Application.spec(:teslamate, :applications) do
   {:ok, _} = Application.ensure_all_started(app)
 end
 
-TeslaMate.Repo.start_link()
-Phoenix.PubSub.Supervisor.start_link(name: TeslaMate.PubSub)
+{:ok, _} = TeslaMate.Repo.start_link()
+{:ok, _} = Phoenix.PubSub.Supervisor.start_link(name: TeslaMate.PubSub)
 
 assert_timeout = String.to_integer(System.get_env("ELIXIR_ASSERT_TIMEOUT") || "300")
 ExUnit.start(assert_receive_timeout: assert_timeout)


JakobLichterfeld (approved):
Thanks Swiffer — diagnosis matches what I see. 

Confirming both paths behave as intended:

- `mix test` (our `--no-start` alias): `Application.stop(:teslamate)` returns `{:error, {:not_started, :teslamate}}`, correctly discarded by `_ =` — nothing changes.
- ElixirLS: `Application.stop/1` is synchronous (it waits for the application master to terminate), so `Repo`, `PubSub` and `Endpoint` have released their registered names before the helper restarts them. The new `{:ok, _} =` assertions therefore can't hit `{:error, {:already_started, _}}`, and they turn a silent no-op into a loud failure — a good change on its own, independent of the rest.

Since `start_permanent` is `Mix.env() == :prod`, the app is `:temporary` under test, so stopping it can't take down the VM. And no test restarts `:teslamate`, which matters because `TeslaMate.Application.start/2` asserts `:ok = :telemetry.detach(...)` — a second start would MatchError. Worth keeping in mind if we ever add a test that boots the app.

I've retitled the PR to `test: stop the app in test_helper instead of relying on --no-start`. Two reasons: `mix test` is green before and after, so nothing is being repaired in the sense `fix(test)` carries here; and the change is worth having for a reason that has nothing to do with ElixirLS. Today the helper depends on the *caller* passing the right flag — an implicit, external precondition. Stopping the app makes it explicit and local: the helper guarantees its own starting state, whoever launches the run. That's the framing I'd like the commit to carry. I'll add the changelog entry under *Build, CI, internal* when merging.

Two optional follow-ups:

1. **Ordering** — move the teardown *above* the logger workaround. Right now the `Application.stop(:logger)` / `start(:logger)` cycle runs while the full app is still live under ElixirLS (Vehicles, Repair, Updater). Clean slate first also makes the file read top-down:

   ```elixir
   # ElixirLS Test Explorer starts the full app (it cannot pass --no-start).
   # Tear it down so tests get the same Repo + PubSub tree as `mix test`.
   _ = Application.stop(:teslamate)

   # FIXME Workaround to apply the logger config when running the tests with
   # "--no-start". This looks like a bug in Elixir v1.11.0
   Application.stop(:logger)
   Application.start(:logger)
   ```

2. **Consistency** — the PR's own theme is "fail loudly", and `_ =` is the one place that still swallows everything. If you like:

   ```elixir
   case Application.stop(:teslamate) do
     :ok -> :ok
     {:error, {:not_started, :teslamate}} -> :ok
   end
   ```

One caveat worth recording, not a blocker: this fixes the *already-started process* class of failures, but under ElixirLS the full app still boots before `test_helper.exs` runs, so `Vehicles.init/1` (`Log.list_cars/0`) and `Repair`'s initial run issue real queries against the test DB outside the sandbox ownership the cases set up later. Harmless on an empty test database, but ElixirLS runs aren't fully hermetic — a green Test Explorer isn't a substitute for `mix test`.

---

swiffer:
Thanks Jakob — took the follow-ups and the commit framing.

- Teardown now runs first, then the logger bounce.
- `Application.stop(:teslamate)` is a `case` on `:ok` / `{:error, {:not_started, :teslamate}}` instead of `_ =`.
- Commit subject matches the PR title. Body is now: the helper used to depend on the caller passing `--no-start`; stopping the app makes that starting state explicit and local.

Also dropped the Elixir 1.11 FIXME on the logger restart. `mix test --no-start` still only loads config — Mix only restarts Logger when it starts the app — so the bounce is still needed for `config/test.exs` (`level: :warning`). Stopping `:teslamate` does not take Logger with it.

Agreed on the caveat: this only fixes the already-started process class. ElixirLS still boots the full app before `test_helper.exs`, so those early `Vehicles` / `Repair` queries sit outside the sandbox. Test Explorer green ≠ `mix test`.

JakobLichterfeld (approved):
Verified — merging.

I double-checked the FIXME removal since that one wasn't on my list: you're right, and it's more than a comment cleanup. There was never an Elixir bug. `Mix.Tasks.App.Start.run/1` (Elixir 1.19.5, `lib/mix/lib/mix/tasks/app.start.ex:56-67`) skips the Logger handling entirely under `--no-start`, and otherwise calls `Logger.App.stop()` before `ensure_all_started/2` brings it back with the project config — "it is up to the application to decide if it should be restarted or not". So the bounce stays necessary and the new comment states the actual reason. Good catch.

Ordering and the `case` look right; `:application.stop/1` only ever returns `:ok` or `{:error, {:not_started, App}}`, so both clauses cover it. 

---

PR #5616, opened 2026-08-15
feat(mqtt): enrich Home Assistant device metadata and republish on change

### LINKED ISSUE #5612: [Feature Request] Improve Home Assistant device metadata (model name & sw_version) in native discovery

### What happened?

With the introduction of native HA discovery in v4.1.1, I noticed that the Device Info panel in Home Assistant appears quite bare compared to the output previously generated by the legacy `teslamate-discovery` container.

Currently, the native discovery publishes:
- **Model:** `3` (instead of `Model 3` or the full designation like `Model 3 74D`)
- **Firmware / Software Version:** Not set on the MQTT device payload (so the "Firmware" row is missing in HA Device Info).

### Describe alternatives you've considered
Manual overrides in Home Assistant customization, but having it natively populated from TeslaMate would provide a much cleaner out-of-the-box experience for all users migrating from `teslamate-discovery`.

### Expected Behavior

It would be great to enrich the MQTT device metadata payload published during discovery:

1. **Model Name:** Map the internal model code to a human-readable model string (e.g. `Model 3` or include the trim details if available) instead of just the raw number/letter (`3`).
2. **Software Version:** Include `"sw_version"` in the device dictionary payload (referencing the current vehicle version topic/data), so Home Assistant displays the installed firmware directly in the device summary card.

Yes

Just look at the info panel in device page

### Relevant log output

```shell
.
```

### Screenshots

Here is a side-by-side comparison of the Device Info card in Home Assistant:
- **Legacy `teslamate-discovery`:** Displays full Model info (`Model 3 74D`) and `Firmware: 2026.26.1`.
- **Native v4.1.1 discovery:** Displays only `3` under the car name and lacks firmware metadata.

Image
Image

### Additional data

_No response_

### Type of installation

### Version

v4.1.1

### PostgreSQL version

_No response_

No

_No response_

nebhale:
## Summary

- Add detailed vehicle model metadata including trim, wheel type, and spoiler information.
- Publish firmware version as the Home Assistant device software version.
- Republish discovery configuration only when device metadata changes.
- Add coverage for formatted metadata, omitted spoiler details, and metadata-triggered discovery updates.

## Testing

- Added and updated Home Assistant and vehicle subscriber tests covering the new behavior.

Resolves #5612

### DIFF
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -119,7 +119,7 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
 
   defp device(%Summary{} = summary, car_id, base_url, namespace) do
     name = summary.display_name || car_name(summary) || "Tesla ##{car_id}"
-    model = summary.model || "Tesla"
+    model = model_name(summary) || "Tesla"
 
     %{
       identifiers: [device_identifier(car_id, namespace)],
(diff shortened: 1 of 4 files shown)


JakobLichterfeld:
Thanks Ben! Will look into it and report back if I do see improvements or if we can just merge.

JakobLichterfeld (changes requested):
Thanks again Ben, nice feature! Publishing richer device metadata and republishing discovery only when it actually changes is a real improvement — the "old" code froze whatever the first summary contained, so this also fixes stale discovery data after a restart. Good test coverage of the new paths, too.

A few points from review before we can merge:

1. **`model: nil` with trim set produces a wrong label.** Our model detection maps unknown car types to `model = nil` (`"cybertruck"` and anything else not matched falls through to `_ -> nil` in `lib/teslamate/vehicles/vehicle.ex`), while `trim_badging` is derived independently and stays populated. Running the new `model_name/1` logic against that state:

   ```
   model=nil, trim="FOUNDATION"  ->  "Model FOUNDATION"
   ```

   The `"Model "` prefix should only be applied when `summary.model` itself is present; please add a test for the `model: nil` case.

2. **Prefer `marketing_name` over raw `trim_badging`.** Real trim badges are API codes — see the `marketing_name` mapping in `lib/teslamate/vehicles/vehicle.ex`, which translates exactly these codes (`74D`, `P74D`, `50`, …) into `LR AWD`, `SR`, `Plaid`, …. So most users would see `Model 3 74D (…)` in Home Assistant. `car.marketing_name` is already on the `Car` schema and `summary.car` is in scope in `device/3` (it's what `car_name/1` uses) — please use it, with `trim_badging` as fallback. The current tests mask this with values like `"Long Range"`, which never occur in the wild.

3. **Duplicated metadata definition.** `discovery_metadata/1` in the subscriber hand-mirrors the fields `HomeAssistant.device/3` renders, plus a second `car_name/1` that already diverges from the existing one: `home_assistant.ex` guards with `is_binary(name) and name != ""`, the new subscriber variant returns `""` as-is. If `device/3` later uses another field, the republish trigger silently misses it and the retained config goes stale. Suggestion: expose the rendered device map (or a metadata function) from `HomeAssistant` and compare that — this also avoids no-op republishes of ~60 retained configs (e.g. `sun_roof_installed` `nil` → `false`: the `:start` summary leaves it at the struct default `nil`, and `summary.ex` maps `installed: 0` to `false` once online, while both render identically).

4. **Wheel codes with a suffix render unformatted.** `^[A-Za-z]+\d+$` doesn't match values with trailing letters, and the camel-case fallback doesn't split at digit→letter boundaries. Running `format_wheel_type/1` as written:

   ```
   AeroTurbine19       ->  Aero Turbine 19" Wheels
   Pinwheel18          ->  Pinwheel 18" Wheels
   Slipstream19Carbon  ->  Slipstream19Carbon Wheels
   ```

   (`Slipstream19Carbon` is a known API value, though our own fixtures only use `AeroTurbine19/20`.) Inconsistent output for known wheel codes.

5. **Retry cadence (minor).** On a persistently failing publish we now retry and log a warning on every summary — every few seconds while driving. The retry itself is welcome; consider whether the unbounded cadence/log volume is acceptable or worth a simple guard.

---

JakobLichterfeld (approved):
Thanks for the quick and thorough turnaround Ben — this addresses all review points, and nicely so:

- `Model` prefix now limited to S/X/3/Y, with tests for `model: nil` and Cybertruck ✅
- `marketing_name` preferred with `trim_badging` fallback, realistic test values ✅
- Single source of truth: the subscriber now compares the rendered device map from `HomeAssistant.device/2` — no duplicated field list, and no-op transitions (`sun_roof_installed` `nil` → `false`) no longer republish ✅
- Wheel suffixes handled (`Slipstream 19" Carbon Wheels`) ✅
- Token-guarded exponential backoff (5s → 5min) with latest-metadata-wins and cancel-on-revert, fully covered by tests ✅

One non-blocking note: our vehicle identification currently maps unknown car types (including `cybertruck`) to `model = nil`, so real Cybertrucks will render as plain `Tesla` for now — the `"Cybertruck"` test case is future-proofing. Mapping that type in the identification code would be a nice follow-up PR, if you're interested.

LGTM 🚀

---

PR #5618, opened 2026-08-15
feat(mqtt): migrate Home Assistant discovery to device-based configuration

nebhale:
## Summary

- Publish one Home Assistant device discovery payload containing all vehicle entities.
- Preserve entity registry settings by migrating legacy per-entity discovery topics before cleanup.
- Add discovery origin metadata and retain namespace-aware topic behavior.
- Retry discovery publishing after failures instead of marking discovery complete prematurely.
- Update documentation and coverage for migration, cleanup, and retry behavior.

This is an update to align with Home Assistant's [documented preference for Device Discovery over Single Component Discovery](https://www.home-assistant.io/integrations/mqtt/#discovery-messages).

### DIFF
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -14,6 +14,7 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
   alias TeslaMate.Log.Car
 
   @discovery_prefix "homeassistant"
+  @migration_delay :timer.seconds(1)
   @migration_payload Jason.encode!(%{migrate_discovery: true})
   @node "teslamate"
   @version Mix.Project.config()[:version]
@@ -46,15 +47,16 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
   discovery config and then clears the old retained configs.
 
   Publishing stops at the first error. Every legacy migration marker must be
-  published successfully before the device config is published, and legacy
-  cleanup starts only after the device config succeeds. Retrying safely
-  restarts the sequence.
+  published successfully before waiting one second for Home Assistant to
+  process the markers and publishing the device config. Legacy cleanup starts
+  only after the device config succeeds. Retrying safely restarts the sequence.
   """
   @spec migrate(term(), publish_opts(), term()) :: :ok | {:error, term()}
   def migrate(%Summary{} = summary, opts, publisher) do
     {prefix, node, device_payload} = discovery_config(summary, opts)
 
     with :ok <- publish_legacy_configs(prefix, node, @migration_payload, publisher),
+         :ok <- Process.sleep(@migration_delay),
          :ok <- publish_device_config(prefix, node, device_payload, publisher) do
       publish_legacy_configs(prefix, node, "", publisher)
     end
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -23,7 +23,8 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
           car_id: pos_integer(),
           namespace: String.t() | nil,
           base_url: String.t() | nil,
-          discovery_prefix: String.t()
+          discovery_prefix: String.t(),
+          migration_delay: non_neg_integer()
         ]
 
   @doc """
@@ -54,9 +55,10 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
   @spec migrate(term(), publish_opts(), term()) :: :ok | {:error, term()}
   def migrate(%Summary{} = summary, opts, publisher) do
     {prefix, node, device_payload} = discovery_config(summary, opts)
+    migration_delay = Keyword.get(opts, :migration_delay, @migration_delay)
 
     with :ok <- publish_legacy_configs(prefix, node, @migration_payload, publisher),
-         :ok <- Process.sleep(@migration_delay),
+         :ok <- Process.sleep(migration_delay),
          :ok <- publish_device_config(prefix, node, device_payload, publisher) do
       publish_legacy_configs(prefix, node, "", publisher)
     end
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -26,22 +26,41 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
         ]
 
   @doc """
-  Migrates any existing single-component discovery configs, publishes a device
-  discovery configuration payload containing every entity derived from the
-  given vehicle summary, and then clears the old retained configs.
+  Publishes a device discovery configuration payload containing every entity
+  derived from the given vehicle summary.
 
   The payload is published retained (QoS 1) to
   `<discovery_prefix>/device/<node>/config` where `node` is
   `#{@node}_<car_id>` (with the `MQTT_NAMESPACE` inserted after `#{@node}` when
   set). Returns `:ok` on success.
+  """
+  @spec publish(term(), publish_opts(), term()) :: :ok | {:error, term()}
+  def publish(%Summary{} = summary, opts, publisher) do
+    {prefix, node, device_payload} = discovery_config(summary, opts)
+
+    publish_device_config(prefix, node, device_payload, publisher)
+  end
+
+  @doc """
+  Migrates any existing single-component discovery configs to a device
+  discovery config and then clears the old retained configs.
 
   Publishing stops at the first error. Every legacy migration marker must be
   published successfully before the device config is published, and legacy
   cleanup starts only after the device config succeeds. Retrying safely
   restarts the sequence.
   """
-  @spec publish(term(), publish_opts(), term()) :: :ok | {:error, term()}
-  def publish(%Summary{} = summary, opts, publisher) do
+  @spec migrate(term(), publish_opts(), term()) :: :ok | {:error, term()}
+  def migrate(%Summary{} = summary, opts, publisher) do
+    {prefix, node, device_payload} = discovery_config(summary, opts)
+
+    with :ok <- publish_legacy_configs(prefix, node, @migration_payload, publisher),
+         :ok <- publish_device_config(prefix, node, device_payload, publisher) do
+      publish_legacy_configs(prefix, node, "", publisher)
+    end
+  end
+
+  defp discovery_config(%Summary{} = summary, opts) do
     car_id = Keyword.fetch!(opts, :car_id)
     namespace = Keyword.get(opts, :namespace)
     prefix = Keyword.get(opts, :discovery_prefix, @discovery_prefix)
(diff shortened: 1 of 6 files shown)


brianmay:
How does this fit in with #5616? I assume this one should get merged first?

nebhale:
@brianmay I consider them independent of each other and am prepared to do any rebasing necessary if you're ready to merge this one directly.  This was the one was implemented second sequentially, but I don't see it as dependent.

brianmay:
Initially AI hullicanated some issues that were clearly incorrect, I pushed back, and got this feedback:

> The code is correct — I was wrong in my initial review. `device_payload` on line 80 clearly includes `components: components`. The `Map.new/2` call produces a valid map of `{object_id => config_map}` entries, and the resulting JSON has the `components` key at the top level. The tests match the actual output structure.
> 
> ## Updated Review
> 
> The PR is well-structured. No bugs found. A few minor observations:
> 
> **1. `publish_discovery` in `vehicle_subscriber.ex:135-145` returns `:ok` even on error.** The `{:error, reason}` branch calls `Logger.warning` (which returns `:ok`) without an explicit return, so the function always returns `:ok`. This doesn't matter in practice because the `with` statement in `home_assistant.ex` catches errors before they reach `publish_discovery`'s caller. But if someone later relies on `publish_discovery`'s return value outside the `with` context, the implicit `:ok` would be misleading. Consider adding `{:error, reason}` to the error branch.
> 
> **2. `publish_legacy_configs` is called three times in `publish/3`** with different payloads (`@migration_payload`, then `""` for cleanup). If any legacy topic publish fails during the migration phase, the `with` short-circuits and the device config is never published. This is conservative and safe — good for correctness, though it means a single broken legacy topic blocks the entire migration until it's resolved (or the process restarts and `clear_discovery` handles it). Worth documenting this intentional choice, since it's non-obvious.
> 
> **3. No functional issues.** The migration protocol, error propagation from `publish_legacy_configs` through the `with` chain, retry behavior in `vehicle_subscriber.ex`, and the `clear/3` function are all correct. The test coverage for failure paths is thorough.

brianmay (approved):
Looks like a significant improvement to me.

Device discovery has been around since 2024.6, not sure how I missed that.

nebhale:
> Device discovery has been around since 2024.6, not sure how I missed that.

Easy to miss unless you’re a maniac like me and read every single issue of the HA release notes 😂.  Since I live and breathe TeslaMate+HA I’m going to fire off some additional “quality of life” PRs that just make the integration a bit more idiomatic for HA users.

I’ll get on fixing the review feedback later today.

JakobLichterfeld:
> I’m going to fire off some additional “quality of life” PRs that just make the integration a bit more idiomatic for HA users.

Awesome, thanks so much! Is it too much to ask if I ask you to implement #5624 as well? I already checked what is missing and documented here: https://github.com/teslamate-org/teslamate/issues/5624#issuecomment-5315395656

nebhale:
@JakobLichterfeld Yep, I'm on it.  I've got like 10 updates (including everything in that #5624) queued up and ready for you all.  Trying to keep my focus on the two-in-flight limit, but I'll definitely get that one in 4.2.0.

JakobLichterfeld:
> Yep, I'm on it. I've got like 10 updates (including everything in that #5624) queued up and ready for you all. Trying to keep my focus on the two-in-flight limit, but I'll definitely get that one in 4.2.0.

Love your spirit!

My goal is to release version 4.2.0 this week, to have a version which includes PR #5617. I wanted to avoid another quick hotfix release, so I helped one reporter via forum instructions. However, we now have another reporter in #5620. Of the 2,000 charges we took for initial testing, none carried this fault, but some user DBs do. I don't want to create any pressure; I just want to be transparent about my plans.

nebhale:
@brianmay I addressed your agent's feedback in the following ways:

1. This got caught on the rebase and doesn't really appear in a dedicated commit
2. Documented the failure behavior 
3. Mentally thanked the AI for the complement 😁

My agent did catch an improvement in the rebase so that the migration is only attempted on startup (likely more than any one user would need) but that's down from the original implementation's "every time device metadata changed", so I'll call that a win.

brianmay:
From AI. Some of this is rather confused. "xyz is an issue. [...]. No it is not". I think it also got very confused with the rebase to the latest main branch (point 4).

I think this is basically all good.

> ## Review of updated PR
> 
> The updated version is significantly improved. Splitting `publish/3` (device config only) from `migrate/3` (migration + device config + cleanup) is cleaner, and the `discovery_device: nil` guard on the `migrate` clause is an elegant way to ensure migration runs exactly once.
> 
> ### No bugs found. A few observations:
> 
> **1. `publish_device_config/4` is a one-liner wrapper that's only called from one place.**
> 
> ```elixir
> defp publish_device_config(prefix, node, payload, publisher) do
>   call(publisher, :publish, [
>     device_discovery_topic(prefix, node),
>     payload,
>     [retain: true, qos: 1]
>   ])
> end
> ```
> 
> This is called from both `publish/3` and `migrate/3`, so the extraction makes sense for DRY. Not an issue.
> 
> **2. `migrate/3` returns `:ok` on full success but `{:error, term()}` on failure — callers must handle both.** The `vehicle_subscriber.ex` caller does this correctly via `publish_or_schedule_discovery/3` at line 182, which pattern-matches on `:ok` vs `{:error, reason}`. Fine.
> 
> **3. Race between `clear/3` and `publish/3` during startup is unchanged from the original code.** The `init` callback calls `clear_discovery` via `handle_continue`, and the first `Summary` message triggers `publish_discovery`. These are sequential within the GenServer so there's no actual race. Confirmed safe.
> 
> **4. The `device/2` function's `@models_with_prefix` logic is unrelated to this PR.** It was presumably added in a prior commit on the same branch. Not a problem, just noting it's a separate concern.
> 
> **5. Test for `discovery_device` state after backoff (`vehicle_subscriber_test.exs:599-604`)** correctly asserts that `discovery_retry_delay` and `discovery_retry_timer` are reset to `nil` after success. This confirms the retry state machine converges properly.
> 
> **Looks good to merge.**

brianmay (approved):

nebhale:
I think I've found a weird race condition in the migration.  Please hold off on merging until I've had a chance to verify what's going on, in the morning.

JakobLichterfeld:
Will take a deeper look as well

JakobLichterfeld (changes requested):
Thanks Ben — moving to device-based discovery is the right direction, and the sequencing structure (halt on first error, migration markers before the device config, cleanup last, plus the failure-mode tests) is well built. However, the race condition in the migration we need to resolve before merging:

1. **Broker PUBACK ≠ processed by Home Assistant.** `migrate/3` fires all three phases — ~60 `{"migrate_discovery": true}` markers → device config → ~60 empty cleanup payloads — back-to-back within milliseconds. Our publisher waits for the *broker's* PUBACK per message, but that says nothing about whether *HA* has processed the markers yet: HA handles discovery messages asynchronously. The [HA migration protocol](https://www.home-assistant.io/integrations/mqtt/#discovery-migration) deliberately puts a verification step between the marker phase and the device config ("Check the logs to ensure this step is executed correctly" — only then "Switch the discovery topic to the device discovery topic"). The harmful interleaving: HA processes the device config *before* it has finished processing all markers → the old single-component entities still exist with the same `unique_id`s → the device-based components are rejected as duplicates → the markers then unload the old entities. Net result: **a device with no entities**. And since #5616, the device config is only republished on a metadata change or a TeslaMate restart, so that broken state can persist for a long time.

2. **The migration runs on every restart.** `publish_discovery/3` dispatches to `migrate` whenever `discovery_device` is `nil` — which is true on *every* process start, not just on the first post-upgrade run (nothing is persisted, and we don't check whether retained legacy configs actually exist). Every TeslaMate restart republishes migration markers for long-migrated entities (HA log spam) and re-opens the race window from point 1.

3. **Publisher timeouts leave messages in flight.** `Publisher.publish` returns `{:error, :timeout}` after ~9.5s, but the QoS 1 message stays in flight in Tortoise and can still be delivered later (e.g. after a reconnect retransmission). The backoff retry then restarts the sequence, so a stale message from attempt N can arrive at the broker *after* messages from attempt N+1. Worst case: a late retained `{"migrate_discovery": true}` marker lands after the retry's cleanup `""` on the same topic and stays behind as permanent retained state.

Suggested direction: make the migration a genuine one-time step separated from the regular publish path — e.g. detect it by subscribing and checking whether retained legacy configs actually exist (or persist a completed-migration flag) — and introduce a real delay/confirmation between the marker phase and the device config instead of back-to-back publishes. Happy to discuss alternatives.

---

nebhale:
My thoughts on the feedback:

1. I'd given some thought to this path (verify the migration has begun in HA) but decided against it because I don't think we want TM to learn about an HA endpoint and authentication, log in and examine its logs.  I think this crosses the line of HA-aware (we'll publish some payloads that are just for HA) and HA-integrated (we're going to start storing secrets for accessing HA).  Additionally, practically I'm not seeing the race here.  It's definitely a legit finding, but the risk seems pretty low given the alternative is direct HA integration.

2. Same for this one.  The trade-off of once-per-start seemed appropriate since I didn't want a full HA integration.

3. Mostly the same again here.  This one is a little different as I could achieve this using only MQTT, but it makes the implementation drastically more complex.  Additionally it does cause some issues in setups like mine where I've granted HA write-only access to my MQTT broker.  Nothing I wouldn't change, but it does materially modify the relationship of TM to MQTT brokers.

@JakobLichterfeld I think you need to make a call on whether you want a deeper HA integration, native to TM, than what we've had previously.  If yes, then we can directly address all of this feedback.  If no, then I'd recommend that we shouldn't make any of these changes.

JakobLichterfeld:
Thanks Ben for pushing back — you've convinced me on the core question, and parts of the agent review deserve a correction:

**Decision: no deeper HA integration.** MQTT is TeslaMate's interface, not Home Assistant — we stay HA-aware (payloads shaped for HA), never HA-integrated (no endpoints, no secrets, no log access). Your point about broker ACLs is also well taken: requiring TeslaMate to *subscribe* to `homeassistant/#` would materially change the TM↔broker relationship for setups with restrictive ACLs, so the "check retained legacy configs" idea from my review is off the table. Same for persisting migration state in the database — we won't add DB state for HA's benefit.

**A correction to the agents review:** I missed that the two findings partially cancel each other out. *Because* the migration re-runs on every restart, a fired race from point 1 is self-healing — the next restart republishes everything. So "the broken state can persist for a long time" was overstated, and the once-per-start trade-off is acceptable as the healing mechanism.

**One small change I'd still like:** a short fixed delay (1–2s) between the migration-marker phase and the device config publish. The HA docs' "check the logs" step between those two phases is really just a human stand-in for "give HA time to process the markers" — a delay achieves the same with no integration, no subscription, no new state, and closes the realistic interleaving (device config processed while markers are still being handled → components rejected as duplicate `unique_id`s). With the restart self-healing as fallback, that's enough robustness for me.

Point 3 (in-flight messages after a publish timeout) we accept as residual risk — fixing it properly isn't worth the complexity, and a restart heals it too.

So: add the delay, and this is good to go from my side. Thanks again for the thoughtful trade-off analysis!

JakobLichterfeld (changes requested):
Thanks again Ben — the delay lands exactly where we discussed: markers → wait → device config, aborting before the sleep when the marker phase fails, and the backoff retry restarts the full sequence. That closes the realistic race with no integration, no subscription, no new state. 👍

**Requesting one change before merge:** `@migration_delay` is a hard compile-time constant, so every test touching the migration path now sleeps for real — the backoff test alone runs the migrate sequence ~7 times, adding roughly 10–15s of wall clock to the test suite. We hold test code to the same bar as production code, so this needs to be fixed: make the delay injectable through the existing `publish_opts` (defaulting to 1s) and have the tests pass a few milliseconds (e.g. 10ms) — not 0, so the elapsed-time assertion still proves the sequence actually waits. That keeps it an internal function argument — no user-facing setting.

With that in place and a rebase, this is ready to merge from my side. Thanks for seeing the trade-off discussion through!

---

JakobLichterfeld (approved):
Perfect — the delay is now injectable through the existing `publish_opts` with the 1s default intact, production never sets the option (the nil-reject in `discovery_opts` keeps it an internal knob), and the tests inject 10ms so the elapsed-time assertion still proves the sequence actually waits. Exactly what I asked for.

LGTM, thanks for the great collaboration, Ben! 🚀

PR #5626, opened 2026-08-17
fix(grafana): open the TeslaMate header link in a new tab so it works when Grafana and TeslaMate share an origin

misenhower:
Hey there. This PR fixes an issue I ran into with my TeslaMate config. I have Grafana hosted at `/` and the TeslaMate admin panel hosted on the same port under `/admin`. Without setting `targetBlank` to true, all the TeslaMate links within Grafana get picked up by its SPA and therefore don't render properly.

This PR makes all of those links open in a new tab, forcing it to break out of the Grafana SPA.

Let me know if you have any questions! Notes from Codex below.

--- 

## Summary

- open the top-level TeslaMate link in every bundled Grafana dashboard in a new tab
- leave dashboard-to-dashboard links unchanged

## Why

When Grafana and the TeslaMate UI share an origin, with TeslaMate mounted under `URL_PATH`, following the TeslaMate link in the current tab can leave the navigation under Grafana's client-side router instead of performing a fresh top-level navigation. This results in Grafana displaying an error for the TeslaMate path.

Opening the link in a new tab performs an independent browser navigation to the configured `base_url` while preserving the current dashboard.

## Validation

- validated every modified dashboard as JSON
- built the Grafana image against the repository's pinned Grafana 13.1.3 image
- provisioned all 23 bundled dashboards in a temporary local Grafana instance
- verified through the Grafana API that all 22 applicable TeslaMate links have `targetBlank: true`
- verified in Chrome that the rendered link has the configured `href`, `target="_blank"`, and `rel="noreferrer"`, and that clicking it opens a second tab
- `git diff --check`

### DIFF
--- a/grafana/dashboards/battery-health.json
+++ b/grafana/dashboards/battery-health.json
@@ -22,6 +22,7 @@
     {
       "icon": "dashboard",
       "tags": [],
+      "targetBlank": true,
       "title": "TeslaMate",
       "tooltip": "",
       "type": "link",
--- a/grafana/dashboards/charge-level.json
+++ b/grafana/dashboards/charge-level.json
@@ -22,6 +22,7 @@
     {
       "icon": "dashboard",
       "tags": [],
+      "targetBlank": true,
       "title": "TeslaMate",
       "tooltip": "",
       "type": "link",
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -22,6 +22,7 @@
     {
       "icon": "dashboard",
       "tags": [],
+      "targetBlank": true,
       "title": "TeslaMate",
       "tooltip": "",
       "type": "link",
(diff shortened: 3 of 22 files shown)


JakobLichterfeld (approved):
Thanks for the fix, and for the thorough validation, Matt.

I verified the mechanism against Grafana 13.2.1: `interceptLinkClicks` routes any anchor click through the SPA router when the anchor has no `target` attribute and the href, after stripping the app base, contains no `://`. On a shared origin the TeslaMate `base_url` gets stripped to a path, so the link is hijacked. `DashboardLinkButton` only emits `target` for `targetBlank: true`, so opening in a new tab is the only way to get a real top-level navigation there. Setups with Grafana and TeslaMate on separate ports are unaffected by the bug; for them this changes the header link to open in a new tab, which is fine for a "leave Grafana" link.

Note for everyone else: setups with Grafana and TeslaMate on separate origins were not affected by the bug, but they get the behavior change too. The TeslaMate header link now opens in a new tab for all users. Acceptable for a “leave Grafana” link; I will call it out in the changelog.

---

PR #5627, opened 2026-08-17
feat(vehicle): identify Cybertruck and show it correctly across UI and MQTT

nebhale:
Address [this comment](https://github.com/teslamate-org/teslamate/pull/5616#pullrequestreview-4950980747) in #5616:

> our vehicle identification currently maps unknown car types (including cybertruck) to model = nil, so real Cybertrucks will render as plain Tesla for now — the "Cybertruck" test case is future-proofing. Mapping that type in the identification code would be a nice follow-up PR, if you're interested.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -9,12 +9,12 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
 
   import Core.Dependency, only: [call: 3]
 
+  alias TeslaMate.Vehicles.Vehicle
   alias TeslaMate.Vehicles.Vehicle.Summary
   alias TeslaMate.Log.Car
 
   @discovery_prefix "homeassistant"
   @node "teslamate"
-  @models_with_prefix ~w(S X 3 Y)
 
   @type publish_opts :: [
           car_id: pos_integer(),
@@ -146,7 +146,10 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
           nil
 
         model ->
-          [format_model(model), marketing_name(summary) || non_empty(summary.trim_badging)]
+          [
+            Vehicle.format_model(model),
+            marketing_name(summary) || non_empty(summary.trim_badging)
+          ]
           |> Enum.reject(&is_nil/1)
           |> Enum.join(" ")
       end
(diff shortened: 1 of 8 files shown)


JakobLichterfeld (changes requested):
Thanks for the quick follow-up Ben! The mapping itself is correct and the interplay with #5616 works out nicely (`Cybertruck FOUNDATION`, no `Model` prefix, `marketing_name: nil`). Two things need to be part of this PR though, since they key off the same `model` value:

1. **Web UI regression.** The car card in `lib/teslamate_web/live/car_live/summary.html.heex` renders a hardcoded prefix:

   ```heex
   Model {@car.model}
   ```

   With this change, real Cybertrucks would show up as "Model Cybertruck FOUNDATION" — the same bug class we just fixed for Home Assistant in #5616. Please apply the same treatment here (prefix only for S/3/X/Y).

2. **`suspend_min` for new Cybertrucks.** In `lib/teslamate/vehicles.ex`, newly created 3/Y and refreshed S/X cars get `%CarSettings{suspend_min: 12}` while unknown models fall through to the 21-minute default. Now that Cybertrucks are identified, please include them in the 12-minute group — as a modern platform they fall asleep quickly, and new Cybertruck owners shouldn't start out with the legacy default.

Nit, non-blocking: the other car types match by prefix (`"models" <> _`), `"cybertruck"` matches exactly. A prefix match would be slightly more robust against future API variants; exact match is consistent with `"lychee"`/`"tamarind"`, so either way is fine.

---

JakobLichterfeld (changes requested):
Thanks Ben, all three review points are nicely addressed — the LiveView now formats via `format_model/1`, Cybertruck joined the 12-minute suspend group (integration test included), and the prefix match covers car type variants. Two remaining points:

1. **The `.pot` files are now outdated.** The added lines in `lib/teslamate_web/live/car_live/summary.ex` shift the source line references of the `gettext` calls below them (referenced in `priv/gettext/default.pot`). Please run the message extraction as described in our docs and commit the result: https://docs.teslamate.org/docs/development#update-pot-files-extract-messages-for-translation

2. **`format_model/1` now exists twice** — the same function and `@models_with_prefix` list live in both `lib/teslamate/mqtt/pubsub/home_assistant.ex` and `lib/teslamate_web/live/car_live/summary.ex`. Please consolidate them into one shared helper so the next model addition can't drift between UI and MQTT.

Then this is good to go. Thanks!

---

nebhale:
I predicted that I'd get dinged on the model name conversion bit.  Should have just been proactive about it 😂

nebhale:
Huh.  The refactor moved the code sufficiently that the .pot files no longer needed to be regenerated.

JakobLichterfeld (approved):
Nicely done — centralizing `format_model/1` in `Vehicle` resolves the duplication and, as a bonus, leaves the gettext line references untouched, so no `.pot` update is needed after all. LGTM 🚀

PR #5628, opened 2026-08-17
doc: improve token and upgrade documentation, restructure sidebar

### LINKED ISSUE #5621: improve token documentatio
The documentation concerning how to get tokens could be improved. See #5607.

Also, just got tesla_auth merged in nixpkgs. Will be several days before it gets into unstable.

https://nixpk.gs/pr-tracker.html?pr=551333

When that happens, we might want to reference that too.

brianmay:
Fixes #5621

### DIFF
--- a/website/docs/installation/unsupported/debian.md
+++ b/website/docs/installation/unsupported/debian.md
@@ -325,3 +325,14 @@ screen -S teslamate -L -dm bash -c "cd /usr/src/teslamate; ./start.sh; exec sh"
    ```
 
    :::
+
+## Usage
+
+1. Open the web interface [http://your-ip-address:4000](http://localhost:4000)
+2. [Create a token](../tokens.md)
+3. Sign in with your Tesla Account
+4. The Grafana dashboards are available at [http://your-ip-address:3000](http://localhost:3000). Log in with the default user `admin` (initial password `admin`) and enter a secure password.
+
+## Update
+
+To update the running TeslaMate configuration to the latest version, follow: [Upgrading to a new version](../../upgrading.mdx)
--- a/website/docs/installation/unsupported/debian.md
+++ b/website/docs/installation/unsupported/debian.md
@@ -328,10 +328,11 @@ screen -S teslamate -L -dm bash -c "cd /usr/src/teslamate; ./start.sh; exec sh"
 
 ## Usage
 
-1. Open the web interface [http://your-ip-address:4000](http://localhost:4000)
-2. [Create a token](../tokens.md)
-3. Sign in with your Tesla Account
+1. [Generate an access and refresh token](../tokens.md)
+2. Open the web interface [http://your-ip-address:4000](http://localhost:4000)
+3. Enter the access and refresh token on the sign-in page
 4. The Grafana dashboards are available at [http://your-ip-address:3000](http://localhost:3000). Log in with the default user `admin` (initial password `admin`) and enter a secure password.
+5. In the TeslaMate web interface, go to _Settings → URLs_ and set the _Web App_ and _Dashboards_ URLs, so the links between TeslaMate and Grafana work in both directions.
 
 ## Update
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@
 
 #### Documentation
 
+- doc: improve token and upgrade documentation, restructure sidebar (#5628 - @brianmay)
+- doc: promote token security note to warning admonition in docker/nixos install (#5628 - @JakobLichterfeld)
+
 ## [4.1.1] - 2026-08-14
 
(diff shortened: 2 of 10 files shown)


brianmay:
Looking at the docs I see we have "Advanced Guides" and "Advanced Configuration".

I personally find this somewhat confusing.

brianmay:
Gut feeling is that "Installation" is too complicated to live under "Getting Started", and deserves to be a top level entry.

And "Advanced guides" should live under the new top level "Installation".

Not making either changes yet. But can do so in there is consensus.

brianmay:
I changed my mind, going to experiment a bit with the structure. Can revert or move to another PR if people don't like it.

brianmay:
I think I have finished breaking things now. Comments appreciated :-)

website/docs/installation/unsupported/debian.md
```diff
@@ -325,3 +325,14 @@ screen -S teslamate -L -dm bash -c "cd /usr/src/teslamate; ./start.sh; exec sh"
    ```
 
    :::
+
+## Usage
+
+1. Open the web interface [http://your-ip-address:4000](http://localhost:4000)
+2. [Create a token](../tokens.md)
+3. Sign in with your Tesla Account
```
marcone: how about something like "Sign in with the tokens you just created" ?

brianmay: Made the change.

JakobLichterfeld (changes requested):
Thanks Brian, this is a real improvement! Moving token generation out of the FAQ onto its own page is the right call — a step every user needs before first sign-in shouldn't live in a FAQ — and restructuring the sidebar by task order (install → post-install → configure → maintain) instead of topic buckets like "Advanced Guides" matches how users actually work through the docs.

Two things we should handle before merging, though:

**The app links to the removed FAQ anchor.** `lib/teslamate_web/live/signin_live/index.html.heex` (line 94) points the sign-in page's help link at `docs/faq#how-to-generate-your-own-tokens`. Since every already-released TeslaMate version links there permanently, updating the heex alone won't fix it. A client redirect won't cover it either: `@docusaurus/plugin-client-redirects` only maps paths that no longer exist, and `/docs/faq` still exists — URL fragments never reach the routing layer. So we should keep a small stub section under the old heading in the FAQ (e.g. `## How to generate your own tokens` → one line linking to the new tokens page), plus update the heex link to the new page for future releases.

**The Usage steps describe a sign-in flow that no longer exists.** The sign-in page only accepts an access + refresh token — there is no "Sign in with your Tesla Account" (credentials) anymore, and token generation happens entirely outside the web UI. Since this PR copies the block into three more guides anyway, let's fix the order and wording here:

1. [Generate an access and refresh token](tokens.md)
2. Open the web interface http://your-ip-address:4000
3. Enter the access and refresh token on the sign-in page
4. The Grafana dashboards are available at http://your-ip-address:3000. Log in with the default user `admin` (initial password `admin`) and enter a secure password.
5. In the TeslaMate web interface, go to *Settings → URLs* and set the *Web App* and *Dashboards* URLs, so the links between TeslaMate and Grafana work in both directions.

`unraid.md` should get the same updated block — it's currently the only install guide without the token step.

Smaller points:

- Sidebar: `tokens` sits under "Post Installation", but every guide requires it *before* first sign-in. I'd move it into the Installation category (or rename the section) so a required step isn't grouped with optional imports/integrations.
- `tokens.md`: "Tesla Auth is available in NixOS, currently nixos-unstable" — the nixpkgs PR only merged on Aug 15, so this depends on when the channel advances. Maybe phrase it time-independently, e.g. "Tesla Auth is packaged in nixpkgs (currently only in nixos-unstable)".
- `upgrading.mdx`, NixOS tab: "and deploy your favorite nixos deploy tool to deploy the result" — missing verb, should be "and use your favorite NixOS deploy tool to deploy the result".
- `tokens.md`: "Teslamate" → "TeslaMate" (twice), "built in support" → "built-in support".

Thanks again!

---

JakobLichterfeld:
Addendum after a closer look at the sidebar: the new categorization doesn't work for me as it is. "Post Installation" is a phase bucket that mixes a required setup step (tokens) with optional one-time imports and ongoing integrations — users scan for goal-based headings like "Integrations", "Import" and "Configuration", and `environment_variables`/`api` shouldn't float at top level without a category. I'll take a closer look at the structure myself.

JakobLichterfeld (approved):
Approving — I pushed the review fixes directly to the branch so this can make the upcoming release: restored the `how-to-generate-your-own-tokens` FAQ anchor as a stub (the sign-in page of every released version links there), corrected the usage steps in all install guides (generate the token first; credential sign-in no longer exists), added the missing token step to unraid, promoted the token security note to a warning admonition, plus small wording fixes. Shout if you disagree with any of it.

The sidebar structure I'll rework in a follow-up PR — I'd like to sort that out separately from this content improvement, as noted in my review above.

Thanks Brian!

PR #5629, opened 2026-08-17
feat(mqtt): add missing Home Assistant discovery entities

### LINKED ISSUE #5624: Add missing Home Assistant discovery entities for published MQTT topics

### What happened?

In Home Assistant while using the auto discovery (MQTT Discovery (automatic configuration)) via the docs, I notice several sensors are not created.  MQTT has the entries, but the sensors are not in the config

teslamate/cars/1/tpms_soft_warning_fl
teslamate/cars/1/tpms_soft_warning_fr
teslamate/cars/1/tpms_soft_warning_rl
teslamate/cars/1/tpms_soft_warning_rr

teslamate/cars/1/charge_current_request
teslamate/cars/1/charge_current_request_max

These were the ones I had noticed as missing

Also to note, the TPMS look to have some duplicated sensors
I see
homeassistant/sensor/teslamate_1/tpms_pressure_fl
and
homeassistant/sensor/teslamate_1/tpms_pressure_fl_psi

In Home Assistant, you are able to set the Unit of Measurement in the Entity settings. So those that want to see it as bar or psi can make it whichever way they prefer.

### Expected Behavior

HA MQTT auto discovery, should create the entities per the expected topics and values from MQTT

Yes

1. Setup the Teslamate env with notes from https://docs.teslamate.org/docs/integrations/home_assistant/#mqtt-discovery-automatic-configuration
MQTT_HOME_ASSISTANT_DISCOVERY=true
2. In Home Assistant add the MQTT as an integration or reload the integration from the page
3. View the entities created for the vehicle

### Relevant log output

```shell
none
```

### Screenshots

Image

### Additional data

_No response_

### Type of installation

### Version

v4.1.1

### PostgreSQL version

_No response_

No

_No response_

nebhale:
Addresses #5624 by adding and updating various entities to the discovery payload.

**Must be merged after #5618.**. I'll rebase to eliminate the extra commits once it has been squashed and merged.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -17,6 +17,70 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
   @migration_delay :timer.seconds(1)
   @migration_payload Jason.encode!(%{migrate_discovery: true})
   @node "teslamate"
+  @legacy_discovery_entities [
+    {"sensor", "display_name"},
+    {"sensor", "state"},
+    {"sensor", "charging_state"},
+    {"sensor", "since"},
+    {"sensor", "version"},
+    {"sensor", "update_version"},
+    {"sensor", "model"},
+    {"sensor", "trim_badging"},
+    {"sensor", "exterior_color"},
+    {"sensor", "wheel_type"},
+    {"sensor", "spoiler_type"},
+    {"sensor", "geofence"},
+    {"sensor", "shift_state"},
+    {"binary_sensor", "park_brake"},
+    {"sensor", "power"},
+    {"sensor", "speed"},
+    {"sensor", "heading"},
+    {"sensor", "elevation"},
+    {"sensor", "inside_temp"},
+    {"sensor", "outside_temp"},
+    {"sensor", "odometer"},
+    {"sensor", "est_battery_range"},
+    {"sensor", "rated_battery_range"},
+    {"sensor", "ideal_battery_range"},
+    {"sensor", "battery_level"},
+    {"sensor", "usable_battery_level"},
+    {"sensor", "charge_energy_added"},
+    {"sensor", "charge_limit_soc"},
+    {"sensor", "charger_actual_current"},
+    {"sensor", "charger_phases"},
+    {"sensor", "charger_power"},
+    {"sensor", "charger_voltage"},
+    {"sensor", "scheduled_charging_start_time"},
+    {"sensor", "time_to_full_charge"},
+    {"sensor", "tpms_pressure_fl"},
+    {"sensor", "tpms_pressure_fr"},
+    {"sensor", "tpms_pressure_rl"},
+    {"sensor", "tpms_pressure_rr"},
+    {"sensor", "tpms_pressure_fl_psi"},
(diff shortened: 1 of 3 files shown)


JakobLichterfeld:
Will review after your rebase, as #5618 landed on main

nebhale:
@JakobLichterfeld rebased and ready for review.  This one is isolated to the list of "missing" entities and doesn't attempt to "improve" any of the existing ones.

JakobLichterfeld (changes requested):
Thanks Ben — this covers #5624 comprehensively, and the details are well thought through: diagnostic categorization, `enabled_by_default: false` for niche sensors, and the enabled/disabled split in the migration sequence shows real understanding of HA's discovery behavior. Great test coverage, too.

Review points before merge:

1. **Please split the psi-sensor removal out of this PR.** On #5624 we publicly committed to keeping the `tpms_pressure_*_psi` entities for the time being, precisely because removal breaks existing automations and dashboards — this PR removes them and wipes their registry entries via the `@removed_entities` cleanup. The removal has become defensible (the bar sensors now carry `device_class: pressure` + `state_class`, so HA can display psi natively), but that reversal needs to be its own deliberate, perhaps changelogged breaking change — not a side effect of an "add missing entities" PR.

2. **Add the migration delay before the final device config as well.** The final full config is published immediately after the ~90 legacy cleanup payloads. By this PR's own premise — HA may process a legacy cleanup as removal of a disabled component — a late-processed cleanup can remove a disabled component *after* the final config re-added it. That's the same PUBACK ≠ processed race we added `migration_delay` for; please reuse it between cleanup and the final publish.

3. **Marker noise (non-blocking, but worth a thought):** migration markers now go to every current entity's legacy topic, including the ~30 new ones that never existed as single-component configs (e.g. `sensor/.../latitude/config`) — and since migration runs once per start, HA logs warnings for them on every restart. If there's a cheap way to scope the marker set to the pre-device-discovery entity set, let's take it; otherwise we accept the noise.

4. **Small one:** the new raw location sensor shares `unique_id` `teslamate_<id>_location` with the location device tracker. HA tolerates this today (uniqueness is per domain), but it's fragile and free to avoid — please give it a distinct `unique_id` (e.g. `<node>_raw_location`); there's no manual-yaml parity to preserve for this new sensor.

---

nebhale:
I've implemented all the suggested changes.  I disagree with point 1, but I'll litigate it in another PR 😇

JakobLichterfeld (approved):
All four points addressed, and cleanly so — the explicit `@legacy_discovery_entities` snapshot is exactly the right shape for the marker scope, the second `migration_delay` closes the cleanup-vs-final-config ordering consistently, `raw_location` got its own `unique_id`, and the psi sensors are back as compatibility entities (existing users keep them enabled via registry preservation; only fresh installs get them disabled by default). Test coverage for all of it, including the elapsed-time and marker/cleanup-scope assertions. 👏

One heads-up for the psi follow-up you announced 😇: `publish_legacy_configs` filters `entities()` against the legacy list, so removing the psi entities from `entities()` will silently drop their legacy cleanup too — un-migrated installs would keep stale retained psi configs. That PR will need a removed-entities mechanism again (or iterate the legacy list directly). Looking forward to the debate.

LGTM 🚀

---

Issue #5167, opened 2026-03-01
Grafana Drives page shows no data


7ijs:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

The charts on the drives page on grafana gives an error. 

In PostgreSQL logfile the following error: 
2026-03-01 19:37:45.066 UTC [105] ERROR:  syntax error at or near ")" at character 2430

### Expected Behavior

Show the latest drives

### Can the issue reliably be reproduced?

Yes

### Steps To Reproduce the issue

Go to Drives in Grafana

### Relevant log output

```shell
WHERE
	    start_address ILIKE '%%' OR end_address ILIKE '%%'
	ORDER BY data.drive_id DESC;
2026-03-01 19:37:45.066 UTC [105] ERROR:  syntax error at or near ")" at character 2430
2026-03-01 19:37:45.066 UTC [105] STATEMENT:  WITH data AS (

T
	    round(extract(epoch FROM start_date)) * 1000 AS start_date_ts,
	    round(extract(epoch FROM end_date)) * 1000 AS end_date_ts,
	    car.id as car_id,

E
	      WHEN start_geofence.id IS NULL THEN CONCAT('new?lat=', start_position.latitude, '&lng=', start_position.longitude)
	      WHEN start_geofence.id IS NOT NULL THEN CONCAT(start_geofence.id, '/edit')
	    END as start_path,

E
	      WHEN end_geofence.id IS NULL THEN CONCAT('new?lat=', end_position.latitude, '&lng=', end_position.longitude)
	      WHEN end_geofence.id IS NOT NULL THEN CONCAT(end_geofence.id, '/edit')
	    END as end_path,
	    TO_CHAR((duration_min * INTERVAL '1 minute'), 'HH24:MI') as duration_str,
	    drives.id as drive_id,
	    -- Columns
	    start_date,
	    COALESCE(start_geofence.name, CONCAT_WS(', ', COALESCE(start_address.name, nullif(CONCAT_WS(' ', start_address.road, start_address.house_number), '')), start_address.city)) AS start_address,
	    COALESCE(end_geofence.name, CONCAT_WS(', ', COALESCE(end_address.name, nullif(CONCAT_WS(' ', end_address.road, end_address.house_number), '')), end_address.city)) AS end_address,
	    duration_min,
	    distance,
	    start_position.battery_level as start_battery_level,
	    end_position.battery_level as end_battery_level,
[...]

PR #5631, opened 2026-08-18
fix(grafana): harden geofence filter against uninitialized variable state

jamesyc:
Fixes Charges/Drives/Charging Stats dashboards showing no data / 400 when Geofence=All.

**Root cause:** dashboards use `('${geofence:pipe}' = '-1' OR geofence_id in ($geofence))`. When All, Grafana substitutes `${geofence:pipe}=''` and `$geofence` empty -> `geofence_id in ()` is invalid Postgres. Postgres parses both sides of `OR`, so syntax error at 1586 even though first branch is false. Reproduced in production logs:
```
ERROR: syntax error at or near ")" at character 1586
... AND ('' = '-1' OR geofence_id in ()) ...
```

**Fix:** replace with safe array handling:
```sql
('${geofence:pipe}' = '' OR '${geofence:pipe}' = '-1' OR geofence_id = ANY(string_to_array('${geofence:pipe}' , '|' )::int[]))
```
For drives two-col case also handles both start/end geofence. `string_to_array('', '|')::int[] = '{}'` -> `ANY('{}') = false` no error, first OR covers All. Handles `''`, `'-1'`, single `'1'`, multi `'1|2'`, and `NULL` rows.

Fixes 22 occurrences across:
- grafana/dashboards/charges.json:1177 (1)
- grafana/dashboards/drives.json:1267 (1 two-col)
- grafana/dashboards/charging-stats.json (19)

Templating placeholder for empty geofences table (grafana/grafana#119793) already in HEAD, kept.

Manual verification: `docker exec psql` 8 rows for 3M with fixed WHERE vs 0/error with buggy, Grafana 400 disappears after hotfix on raspberrypi4. Verified `in ()` on `raspberrypi4` (`teslamate-database-1:2026-08-18 05:03:04`) and `in (-1)` on `swiffer` both handled.

Related to grafana/grafana#119793

*Tests drafted with Muse Spark (Meta), reviewed and verified by me. Elixir tests for Grafana removed per review - pure Grafana SQL fix.*


jamesyc:
Related: 
PR #5199 fix: ensure Charges / Drives dashboards load correctly if no Geofence exists (merged 2026-03-10)
PR #5335 fix: create placeholder geofence if no geofence exists (merged 2026-05-15)
Both only fixed the empty table case by adding: `with geofences_incl_placeholder as (select 'Giga Texas' where count=0 union all SELECT * from geofences) SELECT ... from placeholder`
They did not fix the root cause issue `geofence_id in ($geofence)` when `All` is selected with existing geofences (`${geofence:pipe}=''`, `$geofence` empty -> `in ()` still parses and then errors).

jamesyc:
I'm also thinking that since this PR fixes the Charges page by itself, the PR https://github.com/teslamate-org/teslamate/pull/5335 can be reverted. This keeps the filter dropdown cleaner. But for chesterton's fence reasons, I have left the current code as-is.

JakobLichterfeld (changes requested):
Thanks for your contribution James! The root-cause analysis is spot on and the dashboard fix itself looks correct and complete — I verified that the 22 occurrences across `charges.json`, `drives.json` and `charging-stats.json` are exactly all usages of `$geofence` in the repo, and the `'' / '-1' / ANY(string_to_array(...))` form preserves the previous semantics (including NULL `geofence_id` rows) while fixing the `in ()` syntax error. Nice catch. 👍

Regarding #5199 / #5335: your reading of the history is correct — both only changed the templating query (placeholder row so "All" initializes on an empty `geofences` table) and never touched the `in ($geofence)` WHERE clauses, so the root cause survived. Thanks for digging that up.

On reverting #5335: I'd keep the placeholder, and not just for Chesterton's fence reasons. It works around a templating-layer bug (grafana/grafana#119793: `includeAll` doesn't initialize when the variable query returns zero rows), while your fix hardens the query layer. The new SQL is only safe if Grafana actually interpolates the variable to `''` — if the variable stays uninitialized and the literal `${geofence:pipe}` reaches Postgres, the `::int[]` cast would now error. So removing the placeholder could re-break the empty-table case. If we ever want to drop it, that should be a separate PR verified against an empty `geofences` table on the Grafana versions we ship.

However, the test files appear to have never been run locally — they fail in several independent ways (note: CI runs need maintainer approval for first-time contributors, which is why this wasn't flagged automatically; I've approved the run now):

1. **`geofence_filter_test.exs` – "drives two-column filter"**: `SELECT 1 WHERE start_geofence_id = ANY(...)` has no `FROM` clause, so Postgres rejects it with `undefined column "start_geofence_id"` (42703) and all three `assert {:ok, _}` fail. This test cannot ever have passed.
2. **Compiler warnings**: unused variables (`g1`/`g2` in two tests, `car` in the drives test) and the unused `alias TeslaMate.Locations.GeoFence`. Our CI runs `mix test --warnings-as-errors`, so this aborts before any test runs.
3. **Formatting**: both test files fail `mix format --check-formatted`, which our linting workflow enforces. (This is also why CI currently only shows the linting failure — `elixir_test` is gated behind it, so points 1–2 will only surface in CI once formatting is fixed.)
4. **Regression guard gap**: `@unsafe_geofence_in ~r/geofence_id\s+in\s*\(\$geofence\)/` doesn't match the `start_geofence.id in ($geofence)` / `end_geofence.id in ($geofence)` form that this very PR fixes in `drives.json` — please widen the regex (e.g. `~r/geofence[_.]id\s+in\s*\(\$geofence\)/`) so that variant can't sneak back in.

Could you please:

- run the suite locally (`mix format`, `mix test --warnings-as-errors`) before pushing — see [CONTRIBUTING](https://docs.teslamate.org/docs/development) for the dev setup,
- fix the drives test so it actually creates drives with start/end geofences and asserts the either-column matching semantics (right now it only checks that the SQL doesn't error, and the fixture car is never used),
- drop the "geofence ANY logic handles empty, -1, single and multi pipe values" test — it only asserts hardcoded strings against the module's own regexes and can never fail, the DataCase integration tests cover the real behavior.

One more small thing: the PR description says "Fixes follow-up to #119793" — on GitHub that links to a (non-existent) TeslaMate issue; that number is the Grafana issue. If there's a matching TeslaMate issue, please reference that instead so auto-close works.

**Using an AI assistant is fine — but please review and run its output yourself before submitting, and be transparent about it in the PR description (e.g. a short "tests drafted with <tool>, reviewed and verified by me" note). Reviewing code the author hasn't run or reviewed shifts the verification cost onto the maintainers.**

The dashboard changes themselves are good to go from my side once the tests are cleaned up. Thanks again!

---

jamesyc:
Tests are now cleaned up, sorry, I didn't directly run the tests. I had just hotfixed the dashboard changes onto my running Teslamate docker container for testing, so I knew the main code change was good, and then ran out of Codex. Tests were done by Muse Spark; seems like Meta is still a step behind.

jamesyc:
`mix format --check-formatted` `exit 0` (no diff, after `92cd61f` `mix format`), `mix test --warnings-as-errors` `496 tests, 0 failures` `38.8s` (`5.1s async, 33.6s sync`):

* `dashboard_queries_test 12/12` (including `geofence[_.]id` guard)
* `geofence_filter_test 8/8` (`''`/`'-1'`/`'1'`/`'1|2'`/`'99999'`/`NULL`/`in ()` vs `ANY`/`drives 4/4/2/3/0`)
* full suite `496/496` (was gated behind lint, now `lint` passes so `elixir_test` runs).

swiffer:
> dashboards use ('${geofence:pipe}' = '-1' OR geofence_id in ($geofence)). When All, Grafana substitutes ${geofence:pipe}='' and $geofence empty -> geofence_id in () is invalid Postgres.

@jamesyc - not sure if that holds - when All, the Custom all value is set to -1 (see variable options) and Grafana substitutes this to `AND ('-1' = '-1' OR geofence_id in (-1))` (which is valid sql) correctly in my case.

Again commenting regarding Dashboard related tests here - this is a pure Grafana Syntax / Design decision and not schema specific - tests shouldn't end up in the elixir testsuite imho.

grafik
grafik

swiffer (changes requested):
The claimed root cause does not hold.

> When All, Grafana substitutes `${geofence:pipe}=''` and `$geofence` empty → `geofence_id in ()`

That is not what All does on these dashboards. The `geofence` variable already has **Custom all value** set to `-1` (`includeAll: true`, `allValue: "-1"`) in `charges.json`, `drives.json` and `charging-stats.json`. Grafana interpolates All as that custom value, not as an empty list. Official docs: if a custom all value is set, All is **not** expanded to every option and is **not** left empty.

So with All selected the query becomes:

```sql
AND ('-1' = '-1' OR geofence_id in (-1))
```

That is valid Postgres. The left side is true, so every row is kept. I can reproduce that in Grafana — see the screenshots on my earlier comment.

`AND ('' = '-1' OR geofence_id in ())` is a different state: the variable is **empty / uninitialized**, not All. Those two were mixed together. The new tests do the same (`test "empty string (All) returns all rows"` vs `test "'-1' (All) returns all rows"`). Only `-1` is All here. `''` is “nothing selected”.

The real `in ()` failure was the empty-`geofences`-table case (Grafana includeAll does not initialize when the variable query returns 0 rows — grafana/grafana#119793 / grafana-postgresql-datasource#125). That is what #5191 was about, and what #5199 / #5335 already fixed with the placeholder so the variable query is never empty. #5191 has been closed since then and we have not received further bug reports for this.

With existing geofences and All selected, current Grafana + these dashboards do not produce `in ()`.

Please do **not** revert #5335 on the back of this PR. The placeholder is the templating-layer workaround for the empty-table bug. The new `string_to_array(... )::int[]` form is only safe if Grafana actually interpolates the variable; if the literal `${geofence:pipe}` reaches Postgres, the `::int[]` cast errors.

A user can still get empty interpolation by deselecting every multi-value option or via `?var-geofence=`. That is a rare edge case, not “All is selected”, and we have no reports of it.

Also: please drop the Elixir tests. This is Grafana templating / dashboard design, not schema. The tests never talk to Grafana — they only re-run hand-interpolated SQL — so they cannot catch the actual All behaviour (`allValue: -1`) and they encoded the same empty-vs-All mix-up.

I would rather not merge this as a “fix All” change. Happy to discuss whether there is anything left worth doing; my current read is that there is not.

jamesyc:
Hi @swiffer, thanks for checking. You're right that `All` *may* be `in (-1)` and valid.

But on my deployment `raspberrypi4` (`teslamate/grafana:latest` `fe24cb24` `Grafana 12.4`, `charges.json` `allValue:'-1'` `includeAll:true` `multi:true` with 3 real geofences) `All` gave the empty case instead: `... AND ('' = '-1' OR geofence_id in ()) ... ERROR: syntax error at or near ")" at character 1586`

Similarly, see issue [#5167](https://github.com/teslamate-org/teslamate/issues/5167) `Drives` `'' = '-1' OR start_geofence.id in () OR end_geofence.id in ()` at `2430`. 

So both `''` and `'-1'` happen in the wild (!), depending on Grafana version/initialisation (`current:{}` empty on first load vs explicitly selecting `All`). The `ANY(string_to_array('${geofence:pipe}','|' )::int[])` with `''='' OR '-1'='-1'` handles both, while `in ($geofence)` fails on `''`/`in ()` even though `''='-1'` is false. Postgres still parses the second branch. 

Regarding placeholder `Giga Texas` (`#5199` `#5335`): agree to keep it (as @JakobLichterfeld noted) - it fixes templating-layer `grafana#119793` (variable query returns 0 rows -> `includeAll` not initialised), while this PR hardens the query layer (`in ()` syntax). `ANY` is only safe if Grafana interpolates to `''`/`'-1'`; if variable stayed literal `${geofence:pipe}` un-interpolated, `::int[]` would error, so removing placeholder could re-break empty-table. Happy to keep placeholder.

On tests: I originally added `dashboard_queries_test` + `geofence_filter_test` to match existing `test/teslamate/grafana/dashboard_queries_test.exs` lint for `positions`/`smallint` queries, but happy to remove the Elixir tests for this PR.

swiffer:
`teslamate/grafana:latest` should be 13.1.3 🤔

jamesyc:
My guess is that there's some old initialization code which does not set it to -1. I hadn't done a `docker compose pull` for a long ass time before recently updating the deployment; and I guess the previous version was 12.4. I did a `pull` to latest, still ran into this bug, and then created the PR. 

But overall I'm pretty confident in saying "there are some teslamate postgres DBs out there that do not have All value set to -1", which is why I was digging around here in the first place. So I think it's still worth adding this fix. Think of it as adding a `null` check, which shouldn't be harmful.

swiffer:
> I did a `pull` to latest, still ran into this bug, and then created the PR.

I assume a forced refresh of the dashboard (Strg + F5) would have been enough - you may ended up in a broken Grafana Variable initialization state because of the templating bug within Grafana.

jamesyc:
Yeah, that makes sense. 

I currently think this PR makes sense for the ~dozen or so users who are probably running some old version of Teslamate, updates it, runs into the Grafana templating bug, and end up with a broken Charges page. It's not a super common bug, but also worth adding this extra check to prevent.

JakobLichterfeld:
Thanks again for the work you put into this, James — the historical digging on #5199/#5335 was genuinely useful, and you were quick and gracious in addressing review feedback. Still, after the discussion here I'm going to close this PR. Let me explain the reasoning:

**On the fix:** As @swiffer showed, with `allValue: "-1"` selecting "All" interpolates to `('-1' = '-1' OR geofence_id in (-1))`, which is valid SQL — the `'' / in ()` state is not "All" but a broken variable-initialization state caused by a Grafana templating bug. The empty-table trigger for that state is already handled by the placeholder from #5199/#5335. 

Note on #5167, which was cited as field evidence: its log does show the `in ()` error, but the issue dates from March 2026 — before #5199/#5335 were merged — and its reporters resolved it via the workaround in #5157 and the issue has been closed since. So it documents the pre-fix state, not a problem on current releases. 

What remains is a transient state on your instance after a long-deferred upgrade, which a hard refresh resolves. Hardening 22 queries against a state that Grafana itself should (and does) recover from isn't a trade-off we want to take on — every divergence from the straightforward filter form makes the dashboards harder to maintain.

**On the tests:** The existing checks in `dashboard_queries_test.exs` are a deliberate, narrow exception — they pin database-schema contracts (e.g. smallint overflow casts) that the Elixir side owns. Grafana variable templating behavior isn't such a contract, and tests that re-run hand-interpolated SQL can't observe what Grafana actually substitutes, so they'd pin the wrong thing. That's why this kind of test doesn't belong in the Elixir suite.

If the broken-initialization state shows up in the wild on current versions — with existing geofences and the placeholder in place — please open an issue with logs and we'll gladly revisit this. And I'd be happy to see more contributions from you; your root-cause instincts are good.

---

PR #5634, opened 2026-08-18
feat(charging): sync Supercharger costs from Tesla history

BZHugs:
## Summary

- add a native Tesla charging-history GraphQL client
- automatically backfill missing Supercharger costs for every vehicle
- periodically check recent sessions every 90 minutes
- preserve manually entered or geofence-calculated costs
- support explicit zero-cost sessions and multiple fee components

The feature is opt-in with:

`ENABLE_TESLA_CHARGING_COST_SYNC=true`

This addresses the feature discussed in:

https://github.com/teslamate-org/teslamate/discussions/1831
https://github.com/teslamate-org/teslamate/discussions/3266

## Testing

- GraphQL authentication, pagination and error handling
- fee and zero-cost normalization
- time and energy-based session matching
- multi-vehicle synchronization and retry behavior
- existing-cost protection
- live Tesla charging-history request

## AI assistance

Parts of this implementation and its tests were drafted with AI assistance, then reviewed and verified by me.


JakobLichterfeld:
Thanks for your contribution JOUET. This is a much aksed feature.
Before looking any deeper into it, the architecture or anything else:

**Using an AI assistant is fine — but please review and run its output yourself before submitting, and be transparent about it in the PR description (e.g. a short "tests drafted with , reviewed and verified by me" note). Reviewing code the author hasn't run or reviewed shifts the verification cost onto the maintainers.**

If the CI fails in the first verification step this is a big hint, nothing has been checked on your end. Please see
https://docs.teslamate.org/docs/development

BZHugs:
Yes, you're right, I was too quick to open the PR.

This feature started as a side project I originally wrote in Python, running in a separate Docker container alongside TeslaMate. I asked GPT to port that code directly into the TeslaMate codebase, and I submitted the PR without running the CI locally first.

I'll take some time to properly review the generated changes and fix the CI issues. At first glance, it looks like it's mainly linting, but I'll go through everything before pushing further changes.

I'll also update the PR description to be transparent about the use of AI.

Thanks for pointing this out.

JakobLichterfeld:
Following up on my earlier comment — I've now done the deeper review of architecture and implementation. I'll be direct: this cannot go into TeslaMate, and the reason is structural, not a list of fixable review comments.

**1. There is no API for this — and that alone decides it.**
Tesla offers no charging-history API for consumer accounts. The only official surface is the Fleet API ([charging endpoints](https://developer.tesla.com/docs/fleet-api#charging_invoice)), which is far too limited around charges and barely adopted among TeslaMate users. What this PR does instead is impersonate the Tesla mobile app against its private backend: a hardcoded akamai gateway URL, a spoofed `okhttp` / `x-tesla-user-agent`, hardcoded `deviceLanguage` / `deviceCountry` / `ttpLocale`. A private app backend has no contract of any kind — Tesla can change or block it at any moment, impersonation puts users' accounts at risk, and there is nothing to test or pin against. TeslaMate core only builds on interfaces with at least some stability guarantee; that rule is what keeps this project maintainable for everyone, and this approach fails it categorically. No implementation quality can compensate for that.

**2. Where the feature belongs.**
The demand is real (see the linked discussions), and my position from those discussions stands: if at all, this is something for a third-party companion project, not for TeslaMate core. To be clear about what that means: TeslaMate has no API for this either, and the database is internal storage, not an integration surface — a third-party tool that couples to either does so entirely at its own risk. A companion project can choose to accept that risk, plus the risk of Tesla's private backend, for its users; core cannot accept it for everyone.

**3. The implementation would not be mergeable either.**
For completeness, the deep review shows the code is designed for the happy path only. The most serious findings:

- 401s from the new endpoint melt the shared auth fuse in `TeslaMate.Api` — five within ten minutes sign the **whole app** out and stop vehicle logging. An optional background feature must never be able to do that.
- The cost writer races `complete_charging_process/1`: candidates are filtered only by `is_nil(cost)` and a time window (no `end_date` guard), so an in-progress charge matches — and completion later overwrites the synced cost. The docs claim "existing costs are never overwritten" does not hold.
- HTTP 429 comes back as `{:error, :too_many_request, retry_after}`, which `sync_car/2` doesn't match → `CaseClauseError`, aborting the whole cycle; `retry_after` is never honored — on exactly the full backfill that will trip the rate limit.
- Matching is location-blind with an unexplained magic weight (`36_000`); `siteLocationName`, geofence/position and `fast_charger_type` are all ignored, so a Supercharger cost can land on an unrelated home charge within the ±30 min fallback window.
- `Session.currency` is parsed and then discarded — foreign-currency amounts are stored as-is next to local values.
- Pagination: a `null`/missing `hasMoreData` raises `BadBooleanError`; accounts with >100 pages fetch all 100 pages and then discard everything with an error, repeating the full burst every 90 minutes forever.
- Backfill state lives only in process memory → every restart re-fetches the entire history per car; one raise anywhere in the loop skips all remaining cars and loses the run's progress.
- The HTTP client forks `TeslaApi.client/0` (dropping `FleetAuth`, copying `log_level` and sensitive-header lists), duplicates the 429 parsing from `TeslaApi.Vehicle`, and hand-rolls a non-RFC-4122 "UUID" where `Ecto.UUID.generate()` exists.
- Docs: the new row is dropped into the middle of the `TESLA_API_*` block of the env-var table, breaking its grouping.

**Decision:** I'm closing this PR. Not because of the bugs — those could be iterated on — but because the foundation (a private, impersonation-based app backend) is not something TeslaMate can depend on, and no revision of this PR can change that. If Tesla ever ships a usable official charging-history API for consumer accounts, a fresh design discussion is welcome. Until then, a third-party extension is the right home for this feature.

---

🤖 Review drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 extra) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

Issue #5583, opened 2026-08-03
Records are built from a single payload: drives lose `start_km`, early charge rows are rejected on Fleet Telemetry setups


JakobLichterfeld:
## What happens

Two failure modes observed on a self-hosted Fleet Telemetry setup (vehicle → `fleet-telemetry` → MQTT → TeslaMate), reported with measurements in #5558.

**1. Drives without `start_km`.** The first position of a drive is written from the first payload. If `Odometer` hasn't arrived yet, `drives.start_km` stays NULL and `distance` stays NULL permanently. Nothing is logged — [`Log.Drive`](lib/teslamate/log/drive.ex#L69) only requires `car_id` and `start_date`, so the changeset succeeds. Reported: **7 of 161 completed drives**; in the dissected case the odometer arrived **111 ms** later. Across the dataset, 384 of 17,203 positions carry no odometer (2.2 %).

**2. Charge rows rejected at the start of a session.** [`Log.Charge`](lib/teslamate/log/charge.ex#L57) requires `date`, `charging_process_id`, `charge_energy_added`, `charger_power` and `ideal_battery_range_km`. If any of the three payload-derived ones is missing, the changeset fails, [`vehicle.ex`](lib/teslamate/vehicles/vehicle.ex#L1718) logs `Invalid charge data` and moves on. The sample is not retried. Reported: at one Supercharger session `IdealBatteryRange` took **~45 s** to arrive, so the first ~45 s of that charging curve do not exist.

## This is not a validation bug

Worth saying up front, because it looks like one: the required fields on `Charge` are correct. A charge sample without power, energy and range is not a useful data point, and relaxing the validation would fill the table with rows no dashboard can use. The rejection is also logged, so it isn't invisible to someone reading logs.

The defect is on the input side. Both records are built from **one payload** rather than from the vehicle's currently known state. On the Owner path the two coincide, because every frame is merged into the last full response before a record is written (see Scope). With a source that delivers fields independently they don't, and the first message of a session carries an arbitrary subset. The values aren't unknown; they're just not in that particular message.

## Scope

Affects installations fed by partial-state sources — Fleet Telemetry, whether received directly or through a proxy.

[...]

PR #5637, opened 2026-08-19
fix(drives): keep mileage when a position has no odometer

mews-se:
Addresses case 1 of #5583 — drives losing `start_km` — and stays inside the Log layer as direction 3 suggests, so no state machine changes.

## The bug

`close_drive/2` reads a drive's mileage with `first_value(p.odometer)` / `last_value(p.odometer)` over the drive's positions. Neither skips NULLs, so a drive whose first or last position carries no odometer is closed with `start_km`, `end_km` and `distance` all NULL — permanently, since a closed drive is never recomputed. That is the reported symptom on Fleet Telemetry setups, where a record can be built from a single partial payload.

`min` and `max` ignore NULLs, and since the odometer never decreases within a drive they return exactly the same values as `first_value`/`last_value` whenever every position carries one. The Owner API path is therefore unchanged.

## Changes

- `close_drive/2`: `start_km`, `end_km` and `distance` come from `min`/`max` over the same window.
- A migration recomputes those three columns for drives that are already affected, from the positions still retained. It only touches rows that are actually broken, and only when the retained positions prove a positive distance — a drive whose positions have been purged keeps its NULLs rather than gaining a fabricated `0`.
- A regression test where the odometer is missing from both the first and the last position.

## Verification

`mix test` against `postgres:18-trixie` on `elixir:1.20.2-otp-29`. The new test fails on current main and passes with the change; the rest of the suite is unaffected.

## Not covered

A drive where *no* position carries an odometer still ends up with NULLs. `distance` is nil there, and the `distance >= 0.01` guard accepts nil because atoms sort above numbers in Elixir, so the drive is kept rather than deleted. That behaviour is unchanged by this PR and seemed like a separate call — happy to fold it in if you would rather have those drives dropped.

### DIFF
--- a/lib/teslamate/log.ex
+++ b/lib/teslamate/log.ex
@@ -258,9 +258,13 @@ defmodule TeslaMate.Log do
           power_min: min(p.power) |> over(:w),
           start_date: first_value(p.date) |> over(:w),
           end_date: last_value(p.date) |> over(:w),
-          start_km: first_value(p.odometer) |> over(:w),
-          end_km: last_value(p.odometer) |> over(:w),
-          distance: (last_value(p.odometer) |> over(:w)) - (first_value(p.odometer) |> over(:w)),
+          # min/max skip NULLs where first_value/last_value do not, so a position
+          # without an odometer no longer blanks out the mileage of a drive. The
+          # odometer never decreases within a drive, so these are the first and
+          # last readings that exist.
+          start_km: min(p.odometer) |> over(:w),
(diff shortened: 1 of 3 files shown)


JakobLichterfeld:
Thanks for your contribution. That being said: if an issue is assigned (like in this case), the PR will come form the assignee. So I tend to close this PR, you can convince me otherwise.

mews-se:
My bad! I mistook your thumbs up for a go earlier. I have no objection to you closing it - and feel free to use anything you find useful! :)

PR #5635, opened 2026-08-19
Add `state_class` to Home Assistant MQTT sensor discovery payloads

mekaneck:
Adding `state_class: measurement` to the MQTT discovery payloads for numeric sensors allows home assistant to create long-term statistics for those entities. This enables, for example, using dashboard cards like the statistics-graph-card to show how many miles were driven each day/week/month/etc. Note that statistics aren't generated retroactively; they will begin to be generated once this configuration is set. 

State class was added using the HA guidance here: https://developers.home-assistant.io/docs/core/entity/sensor/#available-state-classes

Long-term statistics are described here: https://developers.home-assistant.io/docs/core/entity/sensor/#long-term-statistics

Sensors which represent predictions and not actual measurements did not get modified.

PR #5638, opened 2026-08-19
feat(mqtt): refine Home Assistant discovery entity metadata and semantics

nebhale:
This is the final tranche of MQTT Discovery Updates.  This addresses many quality-of-life changes around icons, device classes, names, units, and state classes (mentioned in #5635).

**Must be merged after #5629.**. I'll rebase to eliminate the extra commits once it has been squashed and merged.

### DIFF
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -587,10 +587,10 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
       {"sensor", "battery_level",
        %{
          state_topic_key: :battery_level,
-         name: "Battery Level",
+         name: "Battery",
          device_class: "battery",
-         unit_of_measurement: "%",
-         icon: "mdi:battery-80"
+         state_class: "measurement",
+         unit_of_measurement: "%"
        }},
       {"sensor", "usable_battery_level",
        %{
--- a/lib/teslamate/mqtt/pubsub/home_assistant.ex
+++ b/lib/teslamate/mqtt/pubsub/home_assistant.ex
@@ -972,6 +972,7 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
        Map.merge(true_false, %{
          state_topic_key: :service_mode,
          name: "Service Mode",
+         device_class: "running",
          icon: "mdi:wrench"
        })},
       {"binary_sensor", "tpms_soft_warning_fl",
@@ -1010,6 +1011,7 @@ defmodule TeslaMate.Mqtt.PubSub.HomeAssistant do
        Map.merge(true_false, %{
          state_topic_key: :sentry_mode,
          name: "Sentry Mode",
+         device_class: "running",
          icon: "mdi:cctv"
        })},
       {"binary_sensor", "windows_open",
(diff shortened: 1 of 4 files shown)


swiffer:
@nebhale – Just wanted to leave a big thanks for your work on the HA integration! Tested the latest `main` with #5618 and successfully migrated over to the device-based configuration + autodiscovery.

One small idea (feel free to treat this as a separate feature request): it would be really nice to have some visual indication in the TeslaMate UI that the MQTT integration is active. Tasmota does something similar with a small badge in the top-right when MQTT is configured. Even better if it could show green/red depending on whether MQTT is actually reachable.  

On hover it could also surface useful details like MQTT_HOST reachability and whether MQTT_HOME_ASSISTANT_DISCOVERY is active.

Would help round out the experience end-to-end and give nice feedback on the TeslaMate side.

grafik

nebhale:
@swiffer can you please open an issue dedicated to the badge?  I don't want to lose the request when this PR is finished.

JakobLichterfeld:
Ready to review after you rebase as #5629 landed on main.

nebhale:
@JakobLichterfeld Ready for review.

@mekaneck I did a review and I believe I caught every change your PR did.  I didn't even know about `measurement_angle` so https://github.com/teslamate-org/teslamate/pull/5638/changes/76bcb99fe6272a58576f94c3e621a7a71165e49c is all you!

JakobLichterfeld (approved):
Thanks Ben — the craftsmanship here is high: the hardened templates fix a real bug along the way (`energy_at_arrival: 0` was swallowed by the old truthiness check — `is not none` fixes it), the availability template no longer blows up on non-mapping payloads, the `update` entity with installed/latest from the new `software_update` topic is exactly the right modeling, and the removed-entities mechanism came back just as discussed on #5629. 👏

For the record, this PR contains four deliberate breaking changes to the discovery entities — the `healthy` inversion (ON now means problem), humanized *state values* (`asleep` → `Asleep`, `NoPower` → `No Power`, …), the removal of `binary_sensor.update_available` in favor of the `update` entity, and the route distance switching from converted km to raw miles. All four are the right modeling, and notably the MQTT topics themselves are untouched — the changes live entirely in the discovery configs. No action needed from you; I'll cover all of it in the release notes and add a note on the minimum Home Assistant version for the discovery feature to the docs.

One more thing: since the upcoming release redefines the discovery entity set anyway, I'm reversing my position from #5624 — the psi sensors go after all. You called it — consider the litigation settled in your favor 😇. Please open the removal PR right away so it makes the same release: one clean break, before anyone builds automations on a week-old feature, and with `device_class: pressure` + `state_class` on the bar sensors HA converts units natively. Make sure the psi entities move into `@removed_legacy_discovery_entities` so their legacy cleanup survives. I'll clarify the reversal on #5624. Sorry for the back and forth.

Two small nits, fine to take along in that follow-up:

1. The new "retries Home Assistant discovery after a publishing failure" test is missing `@tag :capture_log` — it leaks the retry warning into the test output.
2. `{"binary_sensor", "update_available"}` in `@legacy_discovery_entities` is now dead — the marker filter intersects with `entities()`, which no longer contains it (cleanup runs via `@removed_legacy_discovery_entities`). Drop it or comment it so the two lists' roles stay crisp.

Approved 🚀

---

nebhale:
> consider the litigation settled in your favor 😇

😂 I had the branch pushed up but had been holding the PR until I could write up a solid defense in prose.  "...and the discovery feature only shipped a week ago (4.1.1, opt-in), so the blast radius is as small as it will ever be" was the core of my argument.

PR #5640, opened 2026-08-20
fix: hide map fullscreen tooltip on narrow viewports

hakong:
## What?

Apply the existing `has-tooltip-hidden-touch` class from the tooltip library to the map fullscreen button.

Extend the existing fullscreen-button test to verify that the responsive tooltip class is present.

## Why?

On mobile, tapping the maximize or minimize button leaves its tooltip visible until another part of the screen is tapped.

This is a follow-up to #5495.

## How?

The existing tooltip class suppresses the tooltip at viewport widths of 1023px or narrower. The button retains its translated `aria-label`, and desktop hover tooltips are unaffected.

## Testing

- Built and deployed a production image from the current `main` branch.
- Tested maximizing and minimizing the map on a physical phone.
- Confirmed that the tooltip no longer appears on mobile.
- Ran `mix ci`: 493 tests passed.

### DIFF
--- a/lib/teslamate_web/live/car_live/summary.html.heex
+++ b/lib/teslamate_web/live/car_live/summary.html.heex
@@ -10,7 +10,7 @@
     >
       <button
         type="button"
-        class="map-fullscreen-button button is-small has-tooltip-left"
+        class="map-fullscreen-button button is-small has-tooltip-left has-tooltip-hidden-touch"
         data-tooltip={gettext("Maximize map")}
         data-exit-label={gettext("Minimize map")}
         aria-label={gettext("Maximize map")}
--- a/test/teslamate_web/live/car_summary_live_test.exs
+++ b/test/teslamate_web/live/car_summary_live_test.exs
@@ -371,6 +371,7 @@ defmodule TeslaMateWeb.CarLive.SummaryTest do
 
       attrs = Map.new(attrs)
       assert attrs["type"] == "button"
+      assert "has-tooltip-hidden-touch" in String.split(attrs["class"])
       assert attrs["data-tooltip"] == "Maximize map"
       assert attrs["data-exit-label"] == "Minimize map"
       assert attrs["aria-label"] == "Maximize map"


hakong:
Does this look OK? Did I forget anything?

JakobLichterfeld:
Thanks for your contribution, and for the thorough testing write-up — the change itself is clean and does exactly what it says.

I've marked this as a discussion because I disagree with the underlying premise, and I want to explain why rather than just close it silently.

An icon-only button without any textual explanation is only acceptable for everyday actions that are universally understood. Maximizing/minimizing the map isn't one of those in my view, so I'd rather keep the label reachable on every device instead of removing it for the majority of our users, who are on phones.

Two things specifically speak against hiding it on touch:

1. The tooltip is not desktop-only in practice. bulma-tooltip triggers on `:hover`, and on touch devices a long press produces exactly that hovered state — so a user who is unsure what the button does can press and hold to read the label *before* triggering the action, then move their finger away. On top of that, `setFullscreenButtonState()` in `assets/js/hooks.js` swaps `data-tooltip` (and `aria-label`) to the *next* action when the state toggles, so the tooltip that remains after tapping "Maximize map" reads "Minimize map" — it explains what the button will do next. Adding `has-tooltip-hidden-touch` removes both of these for touch users.
2. It would be inconsistent with the rest of the UI. Our established pattern for small viewports is to *reposition* tooltips (`has-tooltip-left-mobile`, used throughout `summary.html.heex`), not to suppress them. `has-tooltip-hidden-touch` isn't used anywhere in the codebase today, so this would introduce a second, divergent convention for a single control.

I also don't consider the current behavior a defect. The tooltip staying visible until you tap or pan the map elsewhere is, from my perspective, the best of the available options: it stays around long enough to be read, and it disappears as soon as you interact with the map — which is the very next thing you do. An auto-dismiss timer would be the only alternative I'd seriously consider, and I don't see enough of a problem to justify the extra behavior.

So I'm closing this one. That's not a judgment on the quality of the patch — please don't let it discourage you from opening further PRs. If you run into a case where the tooltip actually obscures something you need to see (rather than just lingering), please open an issue with a screenshot; that would be a different concern and worth solving on its own terms.

---

PR #5644, opened 2026-08-21
feat(grafana): show start SOC on incomplete charges from the first charge sample

swiffer:
## Summary

Incomplete Charges on the Charges dashboard currently leaves `start_battery_level` empty, because that column is only written when a charging process is closed.

Keep energy, duration, and end columns empty while charging (those values are still changing). Fill `start_battery_level` from the first `charges` sample — that value is fixed once charging starts.

---

🤖 Assisted by Grok 4.6 (xAI) via Grok Build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 
 #### Dashboards
 
+- feat(grafana): show start SOC on incomplete charges from the first charge sample (#5644 - @swiffer)
+
 #### Translations
 
 #### Documentation
--- a/grafana/dashboards/charges.json
+++ b/grafana/dashboards/charges.json
@@ -1287,7 +1287,7 @@
           "editorMode": "code",
           "format": "table",
           "rawQuery": true,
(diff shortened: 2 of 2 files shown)


JakobLichterfeld (approved):
Thanks Matthias, nice quality-of-life fix! Reviewed:

- The subquery mirrors exactly what `complete_charging_process` writes on completion (`first_value(battery_level)` ordered by `date` in `lib/teslamate/log.ex`), so the displayed start SOC is identical before and after the charge completes — including NULL samples and tie-breaking. ✔
- `charges.charging_process_id` is indexed and the panel only ever returns the currently open process(es), so the correlated subquery is cheap. ✔
- Field name and column order are unchanged and the panel has no overrides/transformations, so nothing else in the panel needs touching. ✔

One non-blocking thought: this duplicates the start-SOC derivation between the dashboard SQL and the completion logic in `log.ex`, and other DB consumers (custom dashboards, TeslaMateApi) still see NULL while a charge is running. The deeper fix would be to populate `start_battery_level` in the backend once the first charge sample arrives. I do not want ot go this route atm.

LGTM 🚀

---

PR #5647, opened 2026-08-24
fix(vehicle): keep the stream connected while a drive is interrupted by an API outage

### LINKED ISSUE #5648: Streamed positions are discarded when the Fleet API reports the vehicle offline mid-drive

I checked issues and discussions. #5534 (fixed by #5535) and #5532 describe the *other half* of this code path — the stream not being reconnected **after** the vehicle is back online. This issue is about the data streamed **during** the outage, which is discarded because the socket is torn down at the start of it. See "Relation to #5534 / #5532" below for why the two are not the same case.

### What happened?

While a drive is in progress, a single failing `vehicle_data` call is enough to make TeslaMate disconnect the streaming socket. In a self-hosted Fleet Telemetry setup the vehicle very often keeps pushing telemetry throughout — the Fleet API and the telemetry path fail independently — but with the subscription gone, every frame that arrives until the next *successful* `vehicle_data` is dropped. The drive is recorded as a straight line.

I have one fully instrumented occurrence where the loss can be shown from three independent sources, including raw records at the telemetry server proving the vehicle was streaming while TeslaMate was not listening.

**Setup**

| | |
|---|---|
| TeslaMate | **v4.0.1** at the time of the incident (Docker) |
| Streaming | `use_streaming_api = true`, `TESLA_WSS_HOST` pointing at a **self-hosted** Fleet Telemetry pipeline, not Tesla's stream |
| Pipeline | car → (mTLS) `tesla/fleet-telemetry` on my own VPS → Redis → `ghcr.io/myteslamate/websocket` converter → TeslaMate |
| API | Fleet API (own third-party app), region `na` |
| Telemetry fields | 1 s: `Gear`, `Locked`, `DoorState`, `DetailedChargeState` · 5 s: `Location`, `VehicleSpeed` · 60 s: `Odometer`, `Soc`, `RatedRange`, `EstBatteryRange`, `GpsHeading`, `ACChargingPower`, `DCChargingPower` |
| Polling | `POLLING_DEFAULT_INTERVAL=300`, `POLLING_DRIVING_INTERVAL=60`, `POLLING_MINIMUM_INTERVAL=30` |
| Vehicle | Model Y, single car, parked/driven in an ordinary suburban area |

The important structural detail: **TeslaMate's stream socket terminates at my own converter, not at Tesla.** When the Fleet API cannot reach the car, that socket is unaffected and keeps forwarding whatever the car pushes.

Two disclosures so the numbers below can be judged fairly:

1. The image was v4.0.1 plus two patches of mine that were submitted upstream (#5594 `POLLING_SENTRY_INTERVAL`, and the env-var revision of #5600). Both only make an existing polling interval configurable; neither touches the streaming path or the fetch-result handler involved here.
2. I run a small watchdog that raises `POLLING_MINIMUM_INTERVAL` to 900 s for 15 minutes after 3 billed `vehicle_data` 408s within 2 minutes, because every Fleet API call is billed. **That is what stretched this particular recovery to ~17 minutes.** Without it the same event recovers in ~30 s and still loses that much streamed data — two of my other occurrences did exactly that (see "How often" below).

**The incident — 2026-08-23, drive of 10.57 km / 46 min, 248 positions**

All times UTC. TeslaMate's own log runs in local time (UTC+9), so its raw lines read `09:32` where the table says `00:32`.

| UTC | Source | Event |
|---|---|---|
| 00:18:14 | TeslaMate | `Driving / Start`, initiated by a stream frame. Positions arrive continuously |
| 00:30:41 | TeslaMate | **Last position of the drive** (speed 0 — the car had just stopped) |
| 00:30:46 | telemetry server | `Gear` → `ShiftStateP` |
| 00:31:26–00:31:41 | telemetry server | door opens, closes, `Locked` → `true`. I left the car |
| 00:31:41 → 00:39:53 | telemetry server | **complete silence — a parked, locked car sends nothing.** Nothing was lost here |
| 00:32:12 | TeslaMate | `vehicle_data -> 408` → `Stream disconnecting ...` → `Vehicle went offline while driving` |
| 00:32:43 / 00:33:14 / 00:33:44 / 00:34:15 | TeslaMate | four more `vehicle_data -> 408` |
| 00:33:59 | my watchdog | 4 × 408 in 2 min → raises `POLLING_MINIMUM_INTERVAL` to 900 s. The next fetch is therefore due at ~00:49 |
| 00:39:53–00:39:55 | telemetry server | `Locked` → `false`, driver door opens. I came back |
| **00:41:00** | telemetry server | `Gear` → `ShiftStateD`. **The drive resumes** |
| 00:41:00 → 00:49:15 | telemetry server | **158 records arrive**: 74 × `VehicleSpeed`, 73 × `Location`, 8 × `GpsHeading`, 8 × `Soc`, 8 × `RatedRange`, 7 × `Odometer`. TeslaMate has no subscription and records nothing |
| 00:49:03 | my watchdog | cooldown over, fetch floor back to 30 s |
| 00:49:16 | TeslaMate | first position after the gap: **speed 51 km/h**, odometer **+2.707 km** since 00:30:41 |
| 00:49:17 | TeslaMate | `Vehicle is back online` (stream not reconnected — this is v4.0.1, before #5535) |
| 00:49:16 → 01:04:10 | TeslaMate | rest of the drive recorded from billed `vehicle_data` at 60 s spacing: 15 polled positions where the stream would have delivered roughly 180 |
| 01:04:12 | TeslaMate | `Driving / Ended`, then `Start / :online` → `Stream connecting ...` |

**What was lost**

- Gap between two consecutive positions: **18 min 35 s** (00:30:41 → 00:49:16).
- Of that, **8 min 16 s was actual driving** (00:41:00 → 00:49:16) covering **2.707 km** of odometer, with **73 `Location` frames** delivered to the telemetry server and none recorded.
- The two Grafana tooltips attached below are the endpoints of that gap: 0 km/h at 09:30:41 local, then 51 km/h at 09:49:16 local, with no sample in between — the car was clearly already moving well before the first recorded point.
- The remainder of the drive fell back to 60 s polling, so it also lost elevation (already covered by #5534) and spent a billed Fleet API call per minute to do it.

**How often**

Between 2026-08-10 and 2026-08-23 (the retention of that container's log) `Vehicle went offline while driving` occurred **7 times**: 08-13, 08-14, 08-15 (×2), 08-20, 08-22, 08-23. Three of them recovered only after the 15-minute watchdog floor (08-15 14:01→14:18, 08-15 17:15→17:32, 08-23 09:32→09:49, local time). Two recovered in **31 seconds** (08-13 17:11:53→17:12:24, 08-22 14:01:53→14:02:24) — those lost ~30 s of streamed positions each, which is the floor of this bug in an unmodified configuration.

The trigger in the instrumented case looks like an ordinary handover: the fleet-telemetry log shows the car's socket hopping between `cellular` and `wifi` several times between 00:30:30 and 00:41:00 as it arrived home, which is exactly when the Fleet API stopped being able to reach it.

### Expected Behavior

While a drive is in progress and the API reports the vehicle unreachable, the streaming socket should stay connected and `D`/`N`/`R` frames should continue to be written to the open drive. The scheduled `vehicle_data` fetch should remain the only thing that decides when the drive is over.

### Relation to #5534 / #5532

#5535 reconnects the stream once the vehicle is back online, so the *remainder* of the drive is streamed again. That is a real and separate fix and it is not in question here.

In #5532 the stream reported `:inactive` before the offline flip, so keeping the socket would not have rescued anything there — that report used Tesla's own stream, where the stream and the API share the same upstream failure. My setup is the case where they do not: TeslaMate's socket ends at a local converter, and the evidence above shows records still arriving at it while the Fleet API returned 408. So the scenario is not hypothetical, but I want to be precise about its scope: **what I can evidence is a Fleet Telemetry pipeline whose stream source is independent of the Fleet API.** I am not claiming the same holds for Tesla's own streaming endpoint.

No — in the wild it needs a real API outage while the telemetry path keeps working. It is, however, deterministically reproducible in two ways:

1. **In the test harness.** With `use_streaming_api: true`, feed the API mock an `%TeslaApi.Vehicle{state: "offline"}` result mid-drive, then inject a `D` frame. On current `main` the socket is disconnected first and the frame is ignored.
2. **Against a real setup.** Put a proxy in front of `TESLA_API_HOST` that rewrites `GET /api/1/vehicles/<id>` to `"state": "offline"` for a few minutes while driving, with a Fleet Telemetry pipeline feeding `TESLA_WSS_HOST`. The car keeps streaming; TeslaMate records nothing until the rewrite is removed.

1. Run TeslaMate with `use_streaming_api = true` and `TESLA_WSS_HOST` pointing at a Fleet Telemetry → websocket converter, so the stream source does not depend on the Fleet/Owner API.
2. Start a drive; positions arrive from the stream at telemetry cadence.
3. Make `vehicle_data` fail so that the follow-up `GET /api/1/vehicles/<id>` reports `offline` (in my case: a cellular ↔ Wi-Fi handover at a stop, reproducible on demand with the proxy above).
4. Observe `Stream disconnecting ...` immediately followed by `Vehicle went offline while driving`.
5. Keep driving. The vehicle keeps pushing telemetry to the pipeline; TeslaMate records no positions until a `vehicle_data` call succeeds again.

### Relevant log output

TeslaMate (local time, UTC+9; vehicle id redacted):

```
2026-08-23 09:18:15.013 [info] Online / Start of drive initiated by: %TeslaApi.Stream.Data{time: ~U[2026-08-23 00:18:14.968Z], ...}
2026-08-23 09:18:15.013 car_id=1 [info] Driving / Start
2026-08-23 09:32:12.921 [info] GET https://fleet-api.prd.na.vn.cloud.tesla.com/api/1/vehicles/<id>/vehicle_data -> 408 (914.994 ms)
2026-08-23 09:32:13.233 car_id=1 [info] Stream disconnecting ...
2026-08-23 09:32:13.233 car_id=1 [warning] Vehicle went offline while driving
2026-08-23 09:32:43.570 [info] GET .../vehicle_data -> 408 (332.203 ms)
2026-08-23 09:33:14.200 [info] GET .../vehicle_data -> 408 (318.919 ms)
2026-08-23 09:33:44.910 [info] GET .../vehicle_data -> 408 (397.365 ms)
2026-08-23 09:34:15.578 [info] GET .../vehicle_data -> 408 (347.094 ms)
2026-08-23 09:49:17.073 car_id=1 [info] Vehicle is back online
2026-08-23 10:04:12.212 car_id=1 [info] Driving / Ended / 11 km – 46 min
2026-08-23 10:04:12.212 car_id=1 [info] Start / :online
2026-08-23 10:04:12.228 car_id=1 [info] Stream connecting ...
```

My watchdog, same window (local time) — included so the 17-minute recovery is not mistaken for stock behaviour:

```
2026-08-23 09:33:59 JST API throttle tripped: 4 vehicle_data 408s in 2m
2026-08-23 09:33:59 JST scheduled fetches floored at 900 seconds for 900 seconds
2026-08-23 09:49:03 JST 408 throttle cooldown elapsed; restoring the normal fetch floor
```

Raw records at the telemetry server during the gap (my bridge spools every record for 7 days; UTC, one line per message, values omitted):

```
2026-08-23T00:41:00  Gear            <- ShiftStateD, the drive resumes
2026-08-23T00:41:02  VehicleSpeed
2026-08-23T00:41:04  Location
2026-08-23T00:41:07  VehicleSpeed
2026-08-23T00:41:09  Location
2026-08-23T00:41:11  Locked
2026-08-23T00:41:12  VehicleSpeed
2026-08-23T00:41:14  Location
   ... 158 messages in total between 00:41:00 and 00:49:15 ...
   74 x VehicleSpeed, 73 x Location, 8 x GpsHeading, 8 x Soc, 8 x RatedRange, 7 x Odometer
2026-08-23T00:49:10  Location
2026-08-23T00:49:13  VehicleSpeed
2026-08-23T00:49:15  Location
```

Per-minute message counts around the same window, showing continuous delivery. `00:32`–`00:38` are absent entirely: that is the parked, locked car, correctly sending nothing.

```
00:30  5      00:41  20     00:46  24
00:31  6      00:42   6     00:47  23
00:39  2      00:43  18     00:48  19
00:40 11      00:44  24     00:49  25
              00:45  17
```

Positions in the database for that drive (UTC, coordinates omitted):

```
 00:30:39   speed —      2.475 km from start   <- last streamed position
 00:49:16   speed 51     5.182 km from start   <- next position, 18m35s later, +2.707 km
 00:50:17   speed  0     5.720 km              <- from here: exactly 60 s apart (polled)
 00:51:18   speed 16     6.234 km
 00:52:18   speed  6     6.372 km
```

### Screenshots

The two endpoints of the gap in the drive detail view (local time, UTC+9). Nothing was recorded between them.

**[ screenshot 1 — last recorded point, 09:30:41 local: Speed 0 km/h, Power 0 kW, SOC 59% ]**

Image

**[ screenshot 2 — next recorded point, 09:49:16 local: Speed 51 km/h, Power 37 kW, SOC 58% ]**

Image

### Type of installation

Docker

### Version

v4.0.1 at the time of the incident (see the disclosures above about two of my own upstream-submitted interval patches in that image). I have since moved to v4.2.0.

### PostgreSQL version

18.4

No

### Additional context — acceptance criterion

So there is something falsifiable to check against: with the vehicle driving and the API reporting it offline, the fix is working if

1. no `Stream disconnecting ...` follows `Vehicle went offline while driving`,
2. positions continue to be written into the open drive at telemetry cadence while `vehicle_data` keeps failing, and
3. the later `Vehicle is back online` is **not** followed by `Stream connecting ...`, because there is nothing to reconnect.

I have #5647 open with a change along these lines. It has been running on my instance since 2026-08-25; no `Vehicle went offline while driving` has occurred since deployment, so I cannot yet show the post-fix counterpart of the log above. I will post it here as soon as one occurs.

onevcat:
Fixes #5648, which has the logs, the telemetry configuration and the version this was observed on.

### What happens today

When a fetch reports the vehicle as `offline` or `asleep`, the fetch-result handler always disconnects the stream — including in the middle of a drive.

The Fleet/Owner API and the streaming socket can fail independently. In a Fleet Telemetry setup, where TeslaMate's stream source is a local converter rather than Tesla's endpoint, a vehicle the API cannot reach is very often still streaming. The socket is torn down anyway, so every position that arrives until the next successful `vehicle_data` is discarded.

#5648 documents one instrumented occurrence: 158 telemetry records — 73 of them `Location` — arrived at my telemetry server over 8 min 16 s of driving while TeslaMate recorded nothing, and the drive is missing 2.707 km of odometer between two consecutive positions.

### What this PR changes

1. The `offline`/`asleep` fetch-result clause keeps the stream in one case only: **`offline` while the state machine is in a driving state**. Everything else disconnects exactly as before, so the asleep/offline "real online" probe that the teardown was written for is unaffected, and `Start / :online` reuses a live `stream_pid`.
2. The driving stream handler writes `D`/`N`/`R` frames into the open drive in every driving status, not only `:available` (guarded by `not is_nil(drv)`, so nothing is written after the drive timeout). The existing `{_status, %Stream.Data{}}` branch already anticipated frames arriving while unavailable; it never received any, because the socket was gone by then.
3. While unavailable/offline, the remaining frames no longer `schedule_fetch(0)`. The offline handlers have a fetch scheduled already, and one request per frame would only repeat a `vehicle_data` call that is currently failing.

`asleep` is deliberately excluded, as pointed out in review. An `asleep` update while driving closes the drive immediately via `timeout_drive/2` and moves to `:start`, so a preserved socket could never deliver anything into that drive — it would only be left dangling until a later fetch disconnects it. Keeping the stream for `offline` alone therefore strictly dominates, regardless of how likely an `asleep` result mid-drive is.

Also unchanged by design: only a successful fetch can bring the status back to `:available`. A stale or replayed frame can at most insert an out-of-order position — the same exposure that exists in `:available` today — and can never flip the state machine.

### Relation to #5535

#5535 reconnects the stream *after* the vehicle is back online, which is what makes the remainder of a drive streamed again. That fix is not in question here and `maybe_reconnect_stream/1` stays in place: the stream process is not monitored, so a silently died stream is still only recovered at drive resume.

This PR covers the other half — the frames that arrive *during* the outage, which are lost before that reconnect can help.

In #5532 the stream reported `:inactive` before the offline flip, so keeping the socket would not have rescued anything there. That report used Tesla's own stream, where the stream and the API share an upstream failure; #5648 is the case where they do not.

### Tests

- The two tests from #5535 and `logs a drive after a significant offline period while driving` now assert that the stream is neither disconnected nor reconnected while the API reports the vehicle offline mid-drive.
- New: `reconnects the stream if the stream process died while unavailable` covers the reconnect branch of `maybe_reconnect_stream/1`, which the inverted tests above no longer reach.
- New: `keeps recording streamed positions while the vehicle is unavailable for the API` walks the whole sequence — drive on the stream, API reports offline, `D` frame still recorded, `P` frame ignored, API recovers on the same socket.
- To make the safety-net test possible, `ApiMock.stream/3` now returns a small process of its own instead of the test pid, forwarding everything to the test and terminating on disconnect like WebSockex does. That makes `stream_pid` a real process a test can kill; existing assertions on `{:"$websockex_cast", :disconnect}` are unaffected.

`mix format --check-formatted`, `mix compile --warnings-as-errors` and the full `mix test` suite pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Improvements and bug fixes
 
+- fix(vehicle): keep the stream connected while a drive is interrupted by an API outage (#5647 - @onevcat)
+
 #### Build, CI, internal
 
 #### Dashboards
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -485,18 +485,15 @@ defmodule TeslaMate.Vehicles.Vehicle do
       {:ok, %Vehicle{state: vehicle_state} = vehicle}
       when vehicle_state in ["offline", "asleep"] ->
         # disconnect stream in case we started it to detect real online
-        # (in that case we won't go through Start / :offline or Start / :asleep)
-        #
-        # Not while driving: there the vehicle is usually only unreachable for
-        # the API for a while, and the stream keeps delivering positions that
-        # belong to the running drive. The driving handlers keep polling and
-        # decide when the drive is over.
+        # (in that case we won't go through Start / :offline or Start / :asleep),
+        # except when a drive is running and only the API lost the vehicle: the
+        # stream keeps delivering positions for that drive.
         data =
-          case state do
-            {:driving, _status, _drive} ->
+          case {vehicle_state, state} do
+            {"offline", {:driving, _status, _drive}} ->
               data
 
-            _ ->
+            {_, _} ->
               :ok = disconnect_stream(data)
               %Data{data | stream_pid: nil}
           end
@@ -732,9 +729,8 @@ defmodule TeslaMate.Vehicles.Vehicle do
         %Data{} = data
       ) do
     case {status, stream_data} do
-      # Also while the API reports the vehicle unavailable or offline: the
-      # stream stays connected then, and only a successful fetch brings the
-      # status back to :available.
+      # Any status: while the API reports the vehicle offline the stream stays
+      # connected, and only a successful fetch restores :available.
       {_status, %Stream.Data{shift_state: shift_state}}
       when shift_state in ~w(D N R) and not is_nil(drv) ->
         {elevation, geofence} =
@@ -762,8 +758,8 @@ defmodule TeslaMate.Vehicles.Vehicle do
         {:keep_state_and_data, schedule_fetch(0, data)}
 
       {_status, %Stream.Data{}} ->
-        # The offline handlers have a fetch scheduled already; one per frame
-        # would only add vehicle_data requests that are known to fail.
+        # The offline handlers already scheduled a fetch; one per frame would
+        # only repeat a request that is currently failing.
         :keep_state_and_data
     end
   end
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -482,10 +482,24 @@ defmodule TeslaMate.Vehicles.Vehicle do
             end
         end
 
-      {:ok, %Vehicle{state: state} = vehicle} when state in ["offline", "asleep"] ->
+      {:ok, %Vehicle{state: vehicle_state} = vehicle}
+      when vehicle_state in ["offline", "asleep"] ->
         # disconnect stream in case we started it to detect real online
         # (in that case we won't go through Start / :offline or Start / :asleep)
-        :ok = disconnect_stream(data)
+        #
+        # Not while driving: there the vehicle is usually only unreachable for
+        # the API for a while, and the stream keeps delivering positions that
+        # belong to the running drive. The driving handlers keep polling and
+        # decide when the drive is over.
+        data =
+          case state do
+            {:driving, _status, _drive} ->
+              data
+
+            _ ->
+              :ok = disconnect_stream(data)
+              %Data{data | stream_pid: nil}
+          end
 
         %Data{} =
           data =
@@ -494,10 +508,10 @@ defmodule TeslaMate.Vehicles.Vehicle do
             %Data{data | last_response: last_response, geofence: geofence}
           end
 
-        {:keep_state, %Data{data | pre_online_check: :idle, stream_pid: nil},
+        {:keep_state, %Data{data | pre_online_check: :idle},
          [
            broadcast_fetch(false),
-           {:next_event, :internal, {:update, {String.to_existing_atom(state), vehicle}}}
+           {:next_event, :internal, {:update, {String.to_existing_atom(vehicle_state), vehicle}}}
          ]}
 
       {:ok, %Vehicle{state: state} = vehicle} ->
@@ -718,7 +732,11 @@ defmodule TeslaMate.Vehicles.Vehicle do
         %Data{} = data
       ) do
     case {status, stream_data} do
-      {:available, %Stream.Data{shift_state: shift_state}} when shift_state in ~w(D N R) ->
+      # Also while the API reports the vehicle unavailable or offline: the
+      # stream stays connected then, and only a successful fetch brings the
+      # status back to :available.
+      {_status, %Stream.Data{shift_state: shift_state}}
+      when shift_state in ~w(D N R) and not is_nil(drv) ->
         {elevation, geofence} =
           Repo.checkout(fn ->
             {:ok, %{elevation: elevation} = position} =
@@ -740,8 +758,13 @@ defmodule TeslaMate.Vehicles.Vehicle do
              geofence: geofence
          }, broadcast_summary()}
 
-      {_status, %Stream.Data{}} ->
+      {:available, %Stream.Data{}} ->
         {:keep_state_and_data, schedule_fetch(0, data)}
+
+      {_status, %Stream.Data{}} ->
+        # The offline handlers have a fetch scheduled already; one per frame
+        # would only add vehicle_data requests that are known to fail.
+        :keep_state_and_data
     end
   end
 
--- a/test/support/mocks/api.ex
+++ b/test/support/mocks/api.ex
@@ -58,7 +58,31 @@ defmodule ApiMock do
 
   def handle_call({:stream, _vid, _receiver} = event, _from, %State{pid: pid} = state) do
     send(pid, {ApiMock, event})
-    {:reply, {:ok, pid}, state}
+    {:reply, {:ok, spawn_stream(pid)}, state}
+  end
+
+  # Stand-in for the stream process: forwards everything to the test and, like
+  # WebSockex, terminates once it is told to disconnect. Being a process of its
+  # own lets tests kill it to simulate a stream that died unnoticed.
+  defp spawn_stream(pid) do
+    spawn(fn ->
+      ref = Process.monitor(pid)
+      forward_to_test(pid, ref)
+    end)
+  end
+
+  defp forward_to_test(pid, ref) do
+    receive do
+      {:DOWN, ^ref, :process, ^pid, _reason} ->
+        :ok
+
+      {:"$websockex_cast", :disconnect} = msg ->
+        send(pid, msg)
+
+      msg ->
+        send(pid, msg)
+        forward_to_test(pid, ref)
+    end
   end
 
   # Events tagged with :get_vehicle or :get_vehicle_with_state may only be
--- a/test/teslamate/vehicles/vehicle/driving_test.exs
+++ b/test/teslamate/vehicles/vehicle/driving_test.exs
@@ -648,6 +648,79 @@ defmodule TeslaMate.Vehicles.Vehicle.DrivingTest do
 
       refute_receive _
     end
+
+    @tag :capture_log
+    test "reconnects the stream if the stream process died while unavailable",
+         %{test: name} do
+      me = self()
+      now_ts = DateTime.utc_now() |> DateTime.to_unix(:millisecond)
+
+      events =
+        [
+          {:ok, online_event(now_ts)},
+          drive_event(now_ts, 0.1, 30, 20, 200, nil),
+          drive_event(now_ts + 1, 0.1, 30, 20, 200, nil),
+          fn ->
+            send(me, :continue?)
+
+            receive do
+              :continue -> {:ok, %TeslaApi.Vehicle{state: "offline"}}
+            after
+              5_000 -> raise "No :continue after 5s"
+            end
+          end
+        ] ++
+          List.duplicate({:ok, %TeslaApi.Vehicle{state: "offline"}}, 16) ++
+          [
+            drive_event(now_ts + :timer.minutes(4), 0.2, 20, 19, 190, nil),
+            {:ok,
+             online_event(now_ts + :timer.minutes(4) + 1,
+               drive_state: %{
+                 timestamp: now_ts + :timer.minutes(4) + 1,
+                 latitude: 0.3,
+                 longitude: 0.3
+               }
+             )}
+          ]
+
+      :ok = start_vehicle(name, events)
+
+      d0 = DateTime.from_unix!(now_ts, :millisecond)
+      assert_receive {:start_state, car, :online, date: ^d0}
+      assert_receive {ApiMock, {:stream, 1000, _}}
+      assert_receive {:insert_position, ^car, %{}}
+      assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :online}}}
+      assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :driving}}}
+
+      assert_receive {:start_drive, ^car}
+      assert_receive {:insert_position, drive, %{longitude: 0.1, speed: 48}}
+      assert_receive {:insert_position, ^drive, %{longitude: 0.1, speed: 48}}
+
+      # The stream process dies unnoticed: it is not monitored, so this is only
+      # discovered when the drive resumes.
+      assert_receive :continue?
+      {_state, %{stream_pid: stream_pid}} = :sys.get_state(name)
+      assert is_pid(stream_pid)
+      Process.exit(stream_pid, :kill)
+      send(:"api_#{name}", :continue)
+
+      assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :offline}}}, 500
+
+      # Vehicle comes back online mid-drive: the dead stream is replaced
+      assert_receive {ApiMock, {:stream, 1000, _}}, 500
+
+      assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :driving}}}
+      assert_receive {:insert_position, drive, %{longitude: 0.2, speed: 32}}
+      assert_receive {:insert_position, ^drive, %{longitude: 0.3}}
+      assert_receive {:close_drive, ^drive, lookup_address: true}
+
+      d1 = DateTime.from_unix!(now_ts + :timer.minutes(4) + 1, :millisecond)
+      assert_receive {:start_state, ^car, :online, date: ^d1}
+      assert_receive {:insert_position, ^car, %{}}
+      assert_receive {:pubsub, {:broadcast, _, _, %Summary{state: :online}}}
+
+      refute_receive _
+    end
   end
 
   describe "geofencing" do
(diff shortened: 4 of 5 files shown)


onevtail:
While validating exact head `966a8ef25c60abf759af4ce1575862cd5f144aba`, I found an edge case in the claimed `offline`/`asleep` coverage.

The `offline` path works as intended: the stream remains available, D/N/R frames continue to update the existing drive, and ordinary frames do not postpone the scheduled fetch. However, an `asleep` result while driving preserves `stream_pid`, then the subsequent `:asleep` event immediately calls `timeout_drive/2` and transitions to `:start`. Later D frames therefore miss the driving handler and are ignored, while the socket can remain connected until a later successful fetch disconnects it.

I reproduced this with streaming enabled: the initial `asleep` result closed the drive; while the next fetch was blocked, no disconnect occurred and an injected D frame produced no `insert_position`; releasing that fetch then caused the disconnect.

The minimal correction appears to be either:

1. Preserve the stream only for `offline`, retaining the existing `asleep` teardown and narrowing the PR description; or
2. Route `asleep` while driving through the unavailable/offline recovery flow so the existing drive timeout determines when the drive closes.

A regression test with `use_streaming_api: true` should cover the drive lifecycle, D-frame insertion, and socket lifecycle for the chosen behavior. The current exact head does not yet fully handle the `asleep` case.

onevtail - an assistant to @onevcat

onevcat:
The review above is technically valid, but the scenario cannot occur in practice. While driving, the vehicle may be reported as `offline/unavailable` due to network conditions, and without this fix, trip data received from the stream during subsequent connectivity checks can indeed be lost. However, the vehicle should never transition to asleep while it is being driven.

Therefore, I think we should keep the current fix as is.

JakobLichterfeld (changes requested):
Thanks for the PR. The `offline` mechanics of the diff check out on my review (no socket leak, `:available` semantics unchanged, correct interplay with the reconnect from #5535). Three things block a merge, one style point on top:

**1. No underlying issue — demonstrated need missing.** A PR that doesn't solve a reported issue is always critical, especially in core logic like the vehicle state machine: the regression risk and maintenance cost are certain, so the benefit has to be demonstrated. This change only helps when the stream still delivers while the API reports the vehicle offline. In the one verified real-world case we have (#5532), the opposite was true — the stream reported `:inactive` before the offline flip, so this PR would not have rescued anything there. The scenario where it helps so far only exists in the PR description. Please open an issue with logs from your setup that show streamed data arriving during an API outage (plus telemetry configuration and TeslaMate version), so there is a falsifiable problem statement and an acceptance criterion for this change.

**2. Restrict the change to `offline` — the `asleep` half is ineffective by construction.** The edge case found above is valid, and "asleep can't happen while driving" doesn't resolve it: this PR's own premise is that API-reported state while driving is unreliable, which applies to `asleep` exactly as to `offline` (that unreliability is why the asleep-while-driving handler exists at all). More importantly, the probability doesn't matter: an `asleep` update while driving closes the drive immediately, so a preserved stream can never deliver a benefit there — if the case never occurs the branch is dead, if it occurs the branch only leaves a dangling socket. Either way, keeping the stream for `offline` only strictly dominates. Please narrow the clause and the PR description accordingly.

**3. Coverage regression on the reconnect safety net.** Inverting the two tests from #5535 leaves no test exercising the reconnect branch of `maybe_reconnect_stream/1`. That recovery path is real: the stream process is not monitored, so a silently died stream is only recovered at drive resume. Please add a test where the stream pid is dead at resume, so the safety net stays covered.

**4. Style:** please condense the narrative code comments to one-sentence constraints.

One general note: AI assistance on contributions is fine — but the claims need to be verified by you. The description reads largely machine-generated and describes a scenario we cannot verify, and the reply dismissing the `asleep` finding is itself an unverified claim. Verified logs and reasoning you can stand behind are what make this actionable.

---

onevcat:
@JakobLichterfeld Thanks for the review. Yes, that is an issue I encountered, and more than once during my drive in the last week. I will open an issue with the detailed information and some data, hopefully we can check and see how to improve this in a correct way!

onevcat:
I posted some relevant data in #5648, which may help provide supporting evidence for this PR.

As for the other points raised in the review, I’ll address them in a follow-up commit.

onevcat:
Thanks for the detailed review — all four points are addressed, and the `asleep` argument was right.

**1. Issue.** #5648, with the full evidence from my setup: TeslaMate version and the streaming topology (self-hosted Fleet Telemetry → local websocket converter, so the stream source does not depend on the Fleet API), the telemetry field configuration, the TeslaMate log for the incident, and raw records from my telemetry server showing 158 messages — 73 of them `Location` — arriving over 8 min 16 s of driving while TeslaMate recorded nothing. The drive is missing 2.707 km of odometer between two consecutive positions. It also states an acceptance criterion, and discloses that a watchdog of mine raised `POLLING_MINIMUM_INTERVAL` during that event, which is what stretched the recovery to ~17 minutes — two other occurrences recovered in 31 seconds and still lost that much streamed data.

You are right that #5532 is not this case: there the stream went `:inactive` first, because the stream and the API shared the upstream failure. I have been explicit in the issue that what I can evidence is a pipeline where they do not.

**2. `asleep`.** You are right and my earlier reply was wrong — I was arguing about how likely an `asleep` result mid-drive is, which is not the point. `{:update, {:asleep, _}}` in a driving state closes the drive through `timeout_drive/2` and moves to `:start`, so a preserved socket can never deliver anything into that drive; it would only be left dangling. Keeping the stream for `offline` alone dominates either way. The clause is now `{"offline", {:driving, _, _}}` and the PR description says so.

**3. Reconnect coverage.** Added `reconnects the stream if the stream process died while unavailable`: the drive goes unavailable, the stream process is killed while a fetch is blocked, and the resume is asserted to call `connect_stream/1` again.

To make that possible, `ApiMock.stream/3` now returns a process of its own instead of the test pid — it forwards everything to the test and terminates on disconnect, like WebSockex. Until now `stream_pid` was the test process, so `Process.alive?/1` was unconditionally true and nothing could exercise that branch. Existing assertions on `{:"$websockex_cast", :disconnect}` are unchanged; the full suite passes.

**4. Comments.** Condensed to one-sentence constraints.

On the general note: taken, and fair. The description now only claims what #5648 can back with logs, and the `asleep` finding was a case where I should have checked the state machine before replying rather than reasoning about probabilities.

JakobLichterfeld:
Closing this PR — not on content. All review points are addressed, and the code-level analysis stands.

Project policy, stated clearly so it is on record: **merge decisions are never based on evidence from self-modified builds.** Disclosed patches are the part we can verify; the completeness of a disclosure is the part we cannot. Merge evidence must come from unmodified, project-built images — and for security reasons our CI does not build registry images from fork PRs. Your exact head commit (`5518e04`, authorship preserved) therefore continues as #5649, where CI builds the official test image. Validation happens in #5648; once it passes, that PR gets merged.

---

PR #5650, opened 2026-08-27
Support NOMINATIM_API_HOST 

senmizu:
* Support NOMINATIM_API_HOST


senmizu:
mistake...

JakobLichterfeld:
@senmizu: Is the NOMINATIM_PROXY not sufficient? Do you need anything included?

senmizu:
@JakobLichterfeld , honestly I don't remember why I didn't continue with it after testing NOMINATIM_PROXY in its initial release. I'll try it again and get back to you with feedback.

JakobLichterfeld:
Sounds good. Curious to hear your feedback.

senmizu:
> @senmizu: Is the NOMINATIM_PROXY not sufficient? Do you need anything included?

Hey @JakobLichterfeld , sorry for the delayed response. My bad memory failed me on this one again.

After taking another look, here are my findings regarding the difference between the two environment variables:
 - **NOMINATIM_PROXY**: Acts as a proxy that forwards requests to https://nominatim.openstreetmap.org/reverse.
 - **NOMINATIM_API_HOST**: Points to a self-hosted Nominatim service running within the Docker Compose network.

You can see the details in the logs for each configuration below.

Thanks for taking the time to look into this. I don't think my patch is substantial enough to warrant a pull request, but I really appreciate the great work you and the team have done maintaining the TeslaMate project.



**NOMINATIM_PROXY**


**NOMINATIM_API_HOST**

JakobLichterfeld:
@senmizu 

Thanks for testing again and for the clear write-up, that matches what the code does.

`NOMINATIM_PROXY` is an HTTP forward proxy for the Finch pool of `nominatim.openstreetmap.org`. Since the target is HTTPS, the request is tunnelled via CONNECT and TLS terminates at OSM. So the proxy can only help you reach OSM (blocked network, egress proxy), it can never redirect requests to a different Nominatim instance. That is exactly the difference you observed.

Before we decide on `NOMINATIM_API_HOST`, one question: why do you run your own Nominatim instance?

If it was only a workaround for OSM being unreachable from your network, NOMINATIM_PROXY already solves that, and we would rather not add another setting.
If you want a self-hosted instance for its own sake (no rate limit, offline operation, privacy, custom data), then `NOMINATIM_API_HOST` is a legitimate feature and `NOMINATIM_PROXY` is not a substitute.

In the second case a PR would be welcome if you want or we implement it. It should follow the existing TESLA_API_HOST pattern: make the base URL in `TeslaMate.Locations.Geocoder` configurable, use the same value as the pool key in `TeslaMate.HTTP.pools/0` (keeping pool size 3 and the proxy opts on that key), add a row to website/docs/configuration/environment_variables.md, and add a changelog entry.

---

drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

Issue #5652, opened 2026-08-27
Phase 1: characterization suite at the outer boundary


JakobLichterfeld:
First phase of #5558. Independent of the open design questions there — this is the safety net that Phases 2 and 3 are gated on, and it is worth having whatever happens afterwards.

## What this is

Pin down current behaviour at the outer boundary: a recorded sequence of Tesla API payloads goes in, and the resulting **persisted rows** (drives, charges, positions, states) and **published MQTT messages** are asserted. Whatever changes inside later, identical payloads must keep producing identical results.

The boundary already exists — `TeslaMate.VehicleCase.start_vehicle/3` drives exactly such a sequence through the API mock today. What's missing: today's tests assert on **mock call sequences** (`LogMock`/`PubSubMock` in a particular order), which couples every test to the internal call structure of `Vehicle` — an unrelated change breaks them and the failure tells you nothing (see the testing-strategy subthread in [#5558](https://github.com/teslamate-org/teslamate/discussions/5558#discussioncomment-17842826)).

## Scope

- [x] **Fixture format:** payload sequences stored as data files, not Elixir literals; nothing in a fixture refers to an internal function or module --> #5653
- [x] **Harness:** replay a fixture, assert against real persisted rows and published MQTT messages — no mock call assertions --> #5653 (self-verifying: structural layout, enforced pairing states, gated recording)
- [x] **Convert existing vehicle test cases** to fixtures, incrementally
  - [x] driving --> #5654
  - [x] charging (incl. charging_sync) --> #5655
    - [x] **`charging_without_charger_power` fixture:** the charge-changeset rejection path — owed since #5655 --> #5700 (the rejection path turned out unreachable; the fixture pins the actual behaviour, #5699)
  - [x] updating --> #5658
  - [x] streaming (all eleven, incl. the four suspended scenarios via the user-command event) --> #5668
  - [x] suspend_logging  --> #5687
  - [x] suspend (idle suspend through clock events) --> #5695
  - [x] summary --> #5696
  - [x] remaining vehicle_test/vehicle_sync scenarios --> #5697
[...]

PR #5653, opened 2026-08-28
test: add characterization harness replaying API fixtures against persisted rows and MQTT

JakobLichterfeld:
First cut of #5652 (Phase 1 of the roadmap in #5558): a characterization harness that pins current behavior at the outer boundary — and verifies itself inside this PR.

## What this does

- **Replays raw Tesla API payload sequences** (JSON data files, decoded by the production decoder `TeslaApi.Vehicle.result/1` — nothing in a fixture refers to an internal function) through the regular `Vehicle` boundary.
- **Asserts what leaves the system instead of mock call sequences:** persisted rows via the real `TeslaMate.Log` (SQL sandbox) and published MQTT messages via the real `Mqtt.PubSub.VehicleSubscriber` with real `Phoenix.PubSub` in between — only the external surfaces (API, geocoder, MQTT broker) stay mocked.
- **The replay end is event-driven, never a wait:** after the last event, the harness waits for exactly one more serve of the repeating terminal event — the state machine starting the next cycle proves the previous one completed. No sleeps, no counts.
- **Place is class:** `scenarios/` and `goldens/` hold behaviour scenarios and their approved expectations, `selftest/{scenarios,goldens}/` holds the harness's own machinery evidence; a file's directory decides what it is, pairing is the shared basename per tree — no naming convention classifies files.
- **Every pairing state has an enforced behavior:** an orphaned golden fails a suite test (a dead expectation must not rest silently), a missing golden fails with the exact record instruction, a declared `only:` target must exist, and the suite can never run empty-green because the self-tests always run.
- **Recording is gated by declared intent:** a checked-in golden is an approved artifact no tool step replaces outside the operator's declared intent. `CHARACTERIZATION_RECORD=1|true` records missing goldens and *compares* existing ones — a record run never reports green for something it did not check; `all` re-records everything; `only:<name>` / `only:selftest/<name>` exactly one; unknown values raise.
- **Records goldens from observed behavior:** recording replays each scenario twice and masks values that differ between runs as `"<volatile>"`; it refuses to write a golden in which a whole table *or MQTT topic* collapses to volatile — a scenario that is not repeat-neutral fails loudly instead of producing a worthless golden. `inserted_at`/`updated_at` and the `since` topic are always masked (wall-clock at second granularity); consecutive duplicate MQTT payloads are collapsed per topic, since cross-topic publish order is concurrent by design (`Task.async_stream(ordered: false)`).
- **Reports the first diverging path on failure**, not "row 47 differs".
- **Goldens contain only canonical representations:** payloads that are JSON objects or arrays are stored decoded and compared structurally — the key order of an encoded map is not portable across runtimes (this bit us: same OTP major, different machines, different order). The recorder writes sorted keys, so its output is deterministic no matter which machine records it; the repo formatter keeps the final word on committed bytes.
- **The harness fails loudly, never invisibly:** the vehicle runs `restart: :temporary` (a crash fails the replay instead of being retried by the supervisor), and fixture-data errors name the offending key and section instead of dying as bare matches.
- **The moduledoc is split** into invariants — each names its check — and limits that are explicitly unenforced (record-time-only neutrality, publish concurrency, serve-vs-cycle barrier granularity). Prose follows the checks.

Verification ships with the machinery: unit tests for the pure core (mode parsing, canonical payload/order, union masking, structural diff) plus two self-test pairs — a minimal park→park pipeline and a mid-sequence `vehicle_unavailable` — whose goldens are declared machinery evidence, not behaviour contracts. Content scenarios deliberately live in the conversion PRs (#5654 ff.), not here; fixture data is synthetic by design — fixtures are public, and real recorded owner-API sequences would be real movement profiles.

No workarounds. One open question: scenarios whose outcome fires *after* the last event (drive timeout, offline charge inference) are not expressible yet — they need a mechanism that awaits a declared outcome. That design lands in the conversion PRs, where such scenarios first appear; this PR deliberately contains no provisional wait for them.

## Verified

- Unit tests cover the pure core; both self-test pairs replay green; golden-tamper mutation check fails with the divergence path.
- Compare runs deterministic across seeds; 21 characterization tests passing (2 self-test replays + harness suite); full suite 547 passing, existing tests untouched.
- The dissolved first fixture (`drive`) lost zero pin surface: a pin-surface diff against `driving_full_drive` (#5654) shows row-equal table occupancy, no column, topic, or value class unique to `drive`, and a byte-identical `updates` row.

---


JakobLichterfeld (commented):
Ran a deep review pass over this harness form a new session before merge (8 finder angles, every candidate adversarially verified against `vehicle.ex`, `api.ex`, the subscriber, and the sandbox setup). The core mechanics hold up: mock contracts, the `:sys.get_state` publish barrier, sandbox rollback of the `TRUNCATE`, and the canonical sorted-key goldens are all sound, and none of the findings breaks the shipped `drive.json` fixture. Ten confirmed findings, split by where they belong:

**Addressing in this PR:**

1. `CHARACTERIZATION_RECORD=0` still records — the check is presence-based and the record branch never compares against the golden, so a leaked env var silently turns the suite into a recorder that reports green. Will parse the value.
2. The record-time "fully masked" guard covers only the database half. An MQTT topic whose publish count differs between the two record runs collapses to `"<volatile>"` with no error (permanently unchecked, including the topic vanishing entirely), and `mask_volatile` builds the golden from run 1's keys only, so a run-2-only topic is dropped instead of refused. Extending the guard to MQTT and making the key handling symmetric.
3. `@db_tables` misses `cars`, which the replay writes through `identify`/`Log.update_car` — a `model`/`marketing_name` regression currently reaches neither the dump nor any MQTT topic. Adding it and re-recording the golden.
4. The test's sandbox setup re-rolls `DataCase` minus the `:already_shared` rescue the other case templates carry for ElixirLS runs → `use TeslaMate.DataCase, async: false`.

**Known limitations, deliberately deferred to the conversion PRs (same bucket as the declared-outcome mechanism above):**

5. The end-of-replay barrier counts API serves, not fetch cycles: expected-unreachable states and the `:vehicle_unavailable` fallback make two calls in one cycle, so both barrier messages can come from the same cycle. Today's fixture vocabulary can't end in those states; the cycle-aware barrier lands together with the fixtures that first need it.
6. Repeat-neutrality stays a documented invariant, not an enforced one (extra ms-interval repeat cycles run between barrier and teardown). A quiescence assertion belongs to the same mechanism.
7. `@fk_tables` is hand-maintained; an FK column added by a future migration would be silently masked (sequences survive the `TRUNCATE`). Deriving FKs from the schema — and raising on unknown integer-id columns — is the deeper fix.
8. Car-id canonicalization is substring-based and topic-only — sufficient for single-car `discovery: false` fixtures, which is all the harness currently allows; multi-car/discovery fixtures need segment-exact rewriting plus payload canonicalization first.
9. Fixture events are untagged, so the endpoint choice (`get_vehicle` vs `get_vehicle_with_state`) is uncharacterized; `ApiMock`'s existing tagged-event facility can pin it once the fixture format grows a tag syntax.

Minor cleanups queued alongside: `Enum.dedup/1` for the hand-rolled dedup, the unreachable `to_jsonable` clauses, prepend+reverse for the per-topic accumulation.

---

JakobLichterfeld:
Findings 1–4 addressed in fde12f18bc828a4ee4538c7b1b88dc390c610bd2, golden re-recorded. The extended guard immediately caught a reset gap: car_settings survived the truncate, since the FK points from cars to car_settings.

JakobLichterfeld:
I personally do not like the structure (mixing scenarios and goldens and distinguish by suffix). Additionally I do miss a unit test --> will rework.

JakobLichterfeld (commented):
Two passes went into this harness beyond the initial implementation.

**Deep review** (8 finder angles, every candidate adversarially verified against `vehicle.ex`, the mocks, and the sandbox setup): ten confirmed findings, four addressed directly in `test(characterization): harden golden recording after review` — `CHARACTERIZATION_RECORD` value parsing (a leaked `=0` no longer silently records), the repeat-neutrality guard extended to MQTT topics with symmetric key-union masking, `cars`/`car_settings` added to the dump (the identify/`update_car` path was previously invisible to goldens), and the test's sandbox setup reusing `DataCase` (restoring the `:already_shared` rescue).

**Structural pass:** the review's dominant failure class was silent green — states of the suite that pass without checking anything (a golden whose scenario was renamed, an empty suite, a leaked record variable). The restructure commit removes that class structurally instead of patching instances:

- **Place is class.**

  ```
  test/fixtures/characterization/
  ├── scenarios/<name>.json       # behaviour scenarios (content)
  ├── goldens/<name>.json         # their approved expectations
  └── selftest/
      ├── scenarios/<name>.json   # harness self-tests (machinery evidence)
      └── goldens/<name>.json
  ```

  A file's directory decides what it is; pairing is the shared basename per tree. No naming convention (`.expected.json`) classifies files anymore.
- **Every pairing state has an enforced behavior:** an orphaned golden fails a suite test (a dead expectation must not rest silently), a missing golden fails with the exact record instruction, a declared `only:` target must exist, and the suite can never run empty-green because the self-tests always run.
- **Recording is gated by declared intent:** a checked-in golden is an approved artifact no tool step replaces outside the operator's declared intent. `1`/`true` records missing goldens and *compares* existing ones — a record run never reports green for something it did not check; `all` re-records everything; `only:<name>` / `only:selftest/<name>` exactly one (identifiers carry the prefix, so they cannot collide with mode words); unknown values raise.
- **The harness verifies itself inside this PR:** unit tests for the pure core (mode parsing, canonical payload/order, union masking, structural diff) plus two self-test pairs — a minimal park→park pipeline and a mid-sequence `vehicle_unavailable` — whose goldens are declared machinery evidence, not behaviour contracts.
- **The vehicle runs `restart: :temporary`:** a crash fails the replay loudly instead of being retried invisibly by the supervisor.
- **The moduledoc is split** into invariants — each names its check — and limits that are explicitly unenforced (record-time-only neutrality, publish concurrency, serve-vs-cycle barrier granularity). Prose follows the checks, not the other way around.
- **`drive.json` is dissolved, not moved:** it converted no original test. Its machinery role is taken by the self-tests; its content is covered more richly by `driving_full_drive` in #5654. A pin-surface diff of the final `drive` golden (last state before the dissolution) against the comparison-green `driving_full_drive` golden confirmed zero lost pin surface: row-equal table occupancy, no column valued only in `drive`, no MQTT topic or value class unique to it — the named candidate, the `updates`/version row, is byte-identical in both.

**Machinery increments cascade with their verification:** `await`, `timebase`, and the real-`Locations` swap stay in #5654 and arrive there with their own self-test pairs and moduledoc entries; the `charging_completed` fact enters with its first user in #5655, `expect_restart` with `updating_cancel` in #5658. Still open from the review, tracked for the conversion PRs: deriving `@fk_tables` from the schema, segment-exact car-id canonicalization, and endpoint pinning via tagged events.

--- 

PR #5654, opened 2026-08-28
test(characterization): convert driving scenarios to characterization fixtures

JakobLichterfeld:
Second step of #5652, stacked on #5653: converts every scenario in `driving_test.exs` to characterization pairs under the structural layout (`scenarios/` + `goldens/`). The originals stay in place.

## Fixtures

All fifteen driving scenarios: full drive (D/N/R), connection loss mid-drive, empty drive_state, direct transition into driving, P-frame handling, offline periods (SOC-gain charge inference, drive timeout, sleep timeout, 15-min gap starting a new drive, 4-min gap continuing the drive), stream reconnects, geofence transitions, and `driving_without_odometer` as a deliberate pin of current behaviour (odometer-less drives have NULL distance at close and are deleted, adjacent to #5583). The underground-parking scenario is event-identical to the offline-timeout one and covered by its fixture. `driving_empty_drive_state` ends mid-drive — legitimate under the enforced effect-neutrality invariant, stated in its description. The three streaming fixtures declare `use_streaming_api: true` explicitly and explain their appended asleep terminal in their descriptions (hardcoded `{3, 10}` suspend settings; a parked streaming terminal grows positions without bound under repeats).

## Machinery this conversion adds

Per #5653's rule, machinery increments cascade with their own verification — each mechanism ships with a self-test pair and its moduledoc invariant entries:

- **`await` — declared outcomes instead of waiting:** facts are re-checked per state-machine cycle against a real deadline; a failure reports every declared fact with its actual value. The `"state"` fact uses current-state semantics (`end_date IS NULL`). After the facts hold, one further serve of the terminal event proves the outcome cycle completed. Recording refuses a vacuous `await` (all facts already satisfied at the causal barrier), probed race-free at the terminal event's second serve — while the blocked fetch task freezes the state machine — with a dedicated machinery test for the refusal.
- **`timebase: "now"`** shifts payload timestamps so `t0` lands at replay time (needed where wall-clock-dated offline/asleep states would be closed by past payload dates). A half-declared timebase raises in both directions, and `t0` must equal the smallest payload timestamp.
- **Real `TeslaMate.Locations`** instead of a mock; fixtures declare geofences as data.

Self-test pairs added: `await`, `timebase`, `geofence`.

- **`timebase: "now"` stays an explicitly interim answer** to the `positive_duration` crash class (mixed clocks in `start_state`): shifted goldens mask every absolute date column, so date/TZ rendering is uncharacterized in exactly those scenarios. The pinning fixture + issue for the underlying production question (stale timestamps after an offline period) is owed before any originals are deleted.
- **Suspend-hardcode windows:** with streaming active the vehicle's suspend settings are hardcoded (`{3, 10}`), which in test-env milliseconds leaves ~180 ms one-shot windows where a loaded runner can trigger a suspend attempt. Documented under the harness's Limits; if exactly the three streaming fixtures ever flake, this is why.
- **Interaction pinning:** the stream fixtures' namesake reconnect `:stream` call is invisible to DB+MQTT goldens; pinning it waits for the tagged-event mechanism deferred in #5653.
- **Not pinned here:** the direct-transition original's per-poll repeat inserts (a repeating same-timestamp drive frame inserts one position per poll) — not expressible until repeat/interaction mechanics exist; and sparse payloads without charger fields (the rejection path for missing `charger_power`) — relevant to the charging conversion, tracked there. The originals stay in place as coverage for all of the above.
- **`charging_completed` is intentionally absent:** fact kinds enter with their first genuine user — the first scenario that actually awaits them. Charging closes synchronously in-cycle, so not the charging conversion.

## Verified

- 20 pairs replay green (15 driving + 5 self-test) — 40 characterization tests, 0 failures, deterministic across seeded runs; full suite 566 passing, existing tests untouched.
- Golden-tamper mutation check fails with the divergence path; vacuity refusal exercised by a dedicated machinery test.

---


JakobLichterfeld (commented):
Ran a deep review pass on a second session (8 finder angles, every candidate adversarially verified against `vehicle.ex`, the mocks, the originals in `driving_test.exs`, and the goldens). The conversions held up well under scrutiny: gap durations are boundary-exact against the originals (16 offline events stay inside the unavailable-counter path, 17+ reach the wall-clock check — matching exactly which fixtures declare `timebase`), the SoC-gain charge inference is provably synchronous inside the resume handler (so its fixture correctly needs no `await`), the Locations swap leaves the base golden genuinely correct (addresses never went through the mock), no network is ever touched in the stream fixtures, and several suspected flakes were refuted (a replay stall always trips the 5 s barrier before it could silently change an outcome). Ten confirmed findings:

**Addressing in this PR:**

1. An `await` that never completes cannot hit its own `flunk`: serves arrive every few ms and re-arm the `after 10_000`, so the failure surfaces as ExUnit's generic 60 s timeout instead of the promised message — and `drives_closed`-style facts use exact `count == n` on monotone counts, so an overshoot (the interesting regression) is permanently false. Fix: a real deadline across recursions plus a failure message enumerating each fact with its actual value.
2. The `await` path tears down the instant the DB facts hold, dropping the harness's own trailing-serve barrier — a check can pass after the outcome cycle's commit but before its `broadcast_summary`. One more `assert_receive {:api_event_served, _}` after the facts hold restores the documented invariant.
3. The `"state"` fact counts historical rows (`count > 0`); switching to current-state semantics (`end_date IS NULL AND state = $2`) is strictly stronger, passes all five existing awaits unchanged, and removes the truncated-golden hazard for future state-revisiting fixtures.
4. `shift_timebase` silently no-ops when `t0` is missing or `"timebase"` is misspelled — reproducing the exact constraint crash-loop the mechanism exists to avoid, surfaced as an opaque timeout. Raising on a half-declared timebase and asserting `t0 == min(timestamps)` closes it.
5. The Vehicle restarts `:permanent` under the test supervisor, so a crash in the awaited code path can be retried invisibly and recorded into the golden as normal behavior → `restart: :temporary` in the harness spec.
6. Fixture text fixes: the `driving_direct_transition` justification is wrong (the original never parked — it pinned per-poll repeat inserts, which the corpus now lacks; noted under deferred), and the three asleep-terminal additions get their rationale into the fixture descriptions instead of only this PR. Plus minors: dedup the three `outcome_met?` queries and the triple key-atomization, name fixture+geofence in `create_geofences` errors, exclude `*.expected.json` from prettier in `formatter.nix` so the recorder is the single formatting authority, and the body above should say 16 fixtures, not 17.

**Open decision:** `driving_empty_drive_state` ends mid-drive — a verbatim violation of the moduledoc's "never mid-drive" invariant, stable only because nil `drive_state` writes nothing per repeat. Either the moduledoc weakens to "effect-neutral under terminal repeats" (the actual mechanism, which the streaming note already implies) or the fixture gets a terminal park.

**Known risk, documented (not fixable in the harness):** the three streaming-default fixtures run with the hardcoded `{3, 10}` suspend settings (`use_streaming_api` defaults to `true`, the fixtures' suspend guards are dead), which in test-env milliseconds means ~180 ms of parked idle triggers a suspend attempt in two one-shot windows per replay. A loaded CI runner can flip that into a record-time repeat-neutrality raise or a compare flake — if these three fixtures ever flake, this is why. Setting `use_streaming_api: true` explicitly at least makes the dependency visible.

**Deferred (with the originals staying in place as coverage):** the stream fixtures' namesake behavior — the reconnect `:stream` call — is invisible to DB+MQTT goldens; pinning interactions needs the tagged-event mechanism already deferred in #5653. The `timebase` flag stays an explicitly interim answer: shifted goldens mask every absolute date column (26 masks vs 7 in the full-drive golden), so the promised pinning fixture + issue for the production stale-timestamp question must land before any originals are deleted.

---

JakobLichterfeld (commented):
All findings are addressed in the rebuilt branch — the restructure from #5653 (structural layout, enforced pairing states, gated recording) landed underneath this PR, so the fixes arrived as part of the rebuild rather than a fix-up series. Point by point:

1. **Await deadline** — `wait_outcomes/4` now runs against a real monotonic deadline and its failure enumerates every declared fact with its actual value; the unreachable `after`-flunk is gone. An awaited outcome that never materialises can no longer hide behind ExUnit's generic timeout.
2. **Trailing barrier** — after the facts hold, one further serve of the terminal event proves the outcome cycle completed; the termination itself is the check, and the moduledoc lists it as an invariant.
3. **Current-state semantics** — the `"state"` fact now matches the open row (`end_date IS NULL AND state = $2`). The same hazard class got a second, stronger guard on top: recording refuses an await that is already fully satisfied at the causal barrier, probed race-free at the terminal event's second serve — while the blocked fetch task freezes the state machine, so "held at the barrier" is exact, not sampled late. A machinery test pins the refusal; the await self-test pair proves a genuinely gating await.
4. **Timebase validation** — a half-declared timebase raises in both directions (`timebase` without integer `t0`, `t0` without `timebase`), and `t0` must equal the smallest payload timestamp.
5. **`restart: :temporary`** — landed with the #5653 restructure; every replay runs under it.
6. **Text fixes** — `driving_direct_transition` now states plainly that the original pinned per-poll repeat inserts and that this coverage is not expressible yet; all three asleep-terminal fixtures carry the full rationale (hardcoded `{3, 10}` suspend settings, unbounded parked streaming terminal) in their descriptions and declare `use_streaming_api: true` explicitly. The minors landed with the final harness pass — the key-atomization helper, named errors in the fixture-data creators, and the `count_rows` extension point for future await facts; the harness is frozen from here on, later PRs only add. The formatter exclude is rejected: repo policy formats everything committed, so prettier stays the byte authority while the recorder guarantees deterministic, sorted-key content underneath it. `charging_completed` left the PR and enters with the first scenario that genuinely awaits a fact — charging closes synchronously in-cycle, so not the charging conversion.

**Open decision → resolved by the restructure:** the moduledoc no longer claims "never mid-drive"; the enforced invariant is effect-neutrality under terminal repetition (the double-run diff is the check), which `driving_empty_drive_state` satisfies — its description states the mid-drive ending openly. Prose follows the check.

**Known risk** stands as documented: the suspend-hardcode windows are listed under the harness's Limits section, and the streaming dependency is now visible in the fixtures themselves.

**Deferred confirmed:** interaction pinning (the reconnect `:stream` call) waits for the tagged-event mechanism; the direct-transition original's per-poll repeat inserts stay unpinned until the same mechanism lands; the timebase pinning fixture + issue for the production stale-timestamp question stays owed before any originals are deleted.

---

JakobLichterfeld:
PR body updated

PR #5655, opened 2026-08-28
test(characterization): convert charging scenarios to characterization fixtures

JakobLichterfeld:
Third step of #5652, stacked on #5654: converts the charging scenarios to characterization pairs under the structural layout. The originals stay in place.

## Fixtures

Six pairs: full charge cycle, connection loss while charging, direct transition into charging, charging into asleep, invalid charge data, and the invalid-range scenario from `charging_sync_test.exs` (`charging_range_nil`); the other five convert `charging_test.exs`. Charger values are realistic AC figures, so the goldens pin the computed energy numbers alongside the sample sequence. Streaming is declared explicitly per scenario. The four terminals reached from `:online` declare `{"state": "asleep"}` — the asleep row lands one cycle after the barrier, so the await genuinely gates it. `charging_into_asleep` needs no await: it reaches asleep through the suspend path within its declared events, and an await there would be vacuous — the recording probe would reject it. `charging_invalid_charge_data` ends parked online.

## Machinery

None. The harness is untouched — this is the first conversion riding the frozen harness; `count_rows` from #5654 stands ready as the extension point, unused here.

- **`charging_completed` was announced for this PR but has no genuine user:** every charging process closes synchronously within its event cycle, so an await on it would be vacuous in every scenario — the vacuity probe would reject it. The closed rows are already pinned by the goldens. The fact kind enters with the first scenario that genuinely awaits it.
- **Sparse payloads without charger fields** (the rejection path for missing `charger_power`) — <pinned here by `charging_invalid_charge_data` / still not pinned here; the originals stay in place as coverage>.

## Verified

- 26 pairs replay green (21 content + 5 self-test) — 46 characterization tests, 0 failures; full suite 572 passing, existing tests untouched.

---


JakobLichterfeld (commented):
Review pass over the conversion (fixture fidelity against `charging_test.exs`/`charging_sync_test.exs`, await gating, harness freeze — this PR touches no harness code, as intended):

- **Sequence fidelity:** `charging_connection_loss` (including the `unknown`-state payload), `charging_into_asleep`, and `charging_invalid_charge_data` replay their originals 1:1. `charging_full_cycle` matches modulo the declared realistic AC values — the originals used toy ranges, the goldens now pin the computed energy figures instead. `charging_direct_transition` and `charging_range_nil` append a Complete/Unplugged/asleep tail so the replay ends repeat-neutral; the originals stopped mid-charge.
- **Await gating is coherent:** the four terminals reached from parked `:online` declare `{"state": "asleep"}` — the asleep row lands one cycle after the barrier, so the await genuinely gates. `charging_into_asleep` reaches asleep inside its declared events via the unavailable path, where an await would be vacuous — the recording probe would reject it. `charging_invalid_charge_data` ends parked online and needs none.
- **`charging_completed` deferral is criteria-consistent:** vocabulary enters with its first genuine user, and every charging process here closes synchronously in-cycle — the vacuity probe would reject any await on it. The closed rows are pinned by the goldens; the fact kind waits for the first scenario that truly awaits it.
- **Coverage boundaries stay explicit:** the `charger_power` rejection path remains unpinned — `invalid_charge_data` pins the earlier discard path (no `charge_state` at all), `range_nil` hits the changeset rejection via `ideal_battery_range_km` — tracked as the `charging_without_charger_power` follow-up in the body.

---

Issue #5656, opened 2026-08-28
Vehicle crashes when an update is cancelled: cancel_update receives the API payload instead of the logged update


JakobLichterfeld:
Found by the characterization suite (#5652), fixture `updating_cancel`: when a vehicle reports software_update status "available" after "installing" (update postponed/cancelled), the vehicle process crashes and the started update row is left open forever.

In `vehicle.ex` the cancel branch shadows `update`: the case pattern `%SW{status: "available"} = update` rebinds the variable that held the `%Log.Update{}` from the state `{:updating, update}`, so `Log.cancel_update/1` is called with the API's `%SoftwareUpdate{}` struct and raises `FunctionClauseError` (log.ex matches on `%Update{}`). The `finish_update` clause below binds `software_update` and passes the outer row correctly — the shadowing exists only in the cancel branch.

`LogMock` accepts any argument, which is why the existing mock-based test passes. Effects on the real path: GenStateMachine crash + supervisor restart on every real update cancellation, plus an orphaned update row (no version, no end_date) per occurrence.

```
** (FunctionClauseError) no function clause matching in TeslaMate.Log.cancel_update/1
    lib/teslamate/log.ex:695: TeslaMate.Log.cancel_update(%TeslaApi.Vehicle.State.VehicleState.SoftwareUpdate{status: "available", ...})
    lib/teslamate/vehicles/vehicle.ex:1433
```

The characterization fixture pins the current (crashing) behaviour and is marked as a pinned bug; its golden gets re-recorded in the fix commit — never silently.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 max) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5658, opened 2026-08-28
test(characterization): convert updating scenarios to characterization fixtures. Pins the update-cancel crash (#5656) that mock-based tests could not see

JakobLichterfeld:
Fourth step of #5652, stacked on #5655: converts the three updating scenarios to characterization pairs. The originals stay in place.

## Fixtures

Three pairs: the full update cycle (update available as a snapshot event, installing across a connection loss — empty payload, `vehicle_unavailable`, offline — finished by an empty status carrying the new car version), the non-empty status (`"foo"` plus a new version finishes the update; a terminal frame without `software_update` is added so the replay ends repeat-neutral, stated in the description), and `updating_cancel`, which **pins bug #5656**: the cancel branch passes the API `software_update` struct to `Log.cancel_update`, and the resulting `FunctionClauseError` crashes the vehicle process. The mock-based original stays green because its `LogMock` accepts anything — the replay through the real `Log` exposes the crash, and the fixture pins it as current behavior. Neither of the other two needs an `await`: an update never closes the open online states row, and `finish_update` lands inside its event cycle.

## Machinery this conversion adds

Per #5653's rule, each mechanism ships with its own verification and moduledoc entries:

- **Snapshot events:** `{"snapshot": true, "vehicle": {...}}` serves the state probe and the subsequent data fetch from one payload, mirroring how one observed API response covers both endpoints; the pair counts as one serve toward the causal barrier (the tag stays outside the serve closure). Self-test pair `snapshot`.
- **`expect_restart`:** a scenario that pins a crash-and-restart declares it. The vehicle runs `restart: :permanent` for exactly that replay, the initial process is monitored, and the replay fails before teardown unless the vehicle actually went down — so a fixed bug flips the fixture red instead of silently passing. A declaration without an `await` convergence outcome raises, as does any non-boolean value; both are probed by unit tests, and a no-crash machinery probe pins the never-went-down failure. First genuine user: `updating_cancel`, per the vocabulary rule.
- **`"positions"` await fact** (exact position count) as the convergence point of a crash-and-restart — a pure addition through the `count_rows` extension point.

- **`updating_cancel`'s golden deliberately pins crashing behavior.** The re-record lands in the #5656 fix commit, so the golden diff will document the behavioral change the fix makes — that is the pin working as intended, not an accepted state.

## Verified

- 30 pairs replay green (24 content + 6 self-test) — 54 characterization tests, 0 failures; full suite 580 passing, existing tests untouched.
- `expect_restart`'s failure modes are probed: a declared restart that never happens fails before teardown; a declaration without a convergence outcome raises.

---


JakobLichterfeld (commented):
Review pass over the conversion (fixture fidelity against `updating_test.exs`, the new snapshot/expect_restart machinery, await gating):

- **Sequence fidelity:** all three fixtures replay their originals 1:1 — `updating_full_cycle` including the snapshot tag on the first event and the mid-install connection loss (empty payload, `vehicle_unavailable`, offline), `updating_cancel` exactly (online, installing, fallback to available). `updating_nonempty_status` adds a terminal frame without `software_update` and discloses it: the original's repeating `"foo"`-status frame would start a fresh update on every terminal repeat — the added frame is what makes the scenario expressible at all.
- **The absent awaits are correct, not omissions:** an update never closes the open online states row and `finish_update` lands inside its event cycle, so `full_cycle` and `nonempty_status` have nothing to await; `updating_cancel` converges on `{"positions": 2}` — initial online plus the restarted vehicle's start position.
- **`expect_restart` mechanics hold:** the monitor is asserted before teardown, so a shutdown DOWN cannot satisfy it; `:permanent` is scoped to the declared replay while every other replay keeps `:temporary`; the declaration errors raise and are unit-probed. The inversion is the right contract — when #5656 gets fixed, this fixture goes red and demands the golden re-record.
- **Snapshot serve counting is sound:** ApiMock's existing snapshot facility serves both endpoints from one execution, so the pair counts as one serve toward the barrier — the tag lives on the event, outside the closure, and the self-test pair pins the mechanism.
- **The conversion's key finding stands on its own:** the mock-based original cannot see #5656 (its `LogMock` accepts the bad struct); the characterization replay through the real `Log` is what exposes the crash. That is precisely the boundary this suite exists to guard.

---

PR #5664, opened 2026-08-30
fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever 

### LINKED ISSUE #5656: Vehicle crashes when an update is cancelled: cancel_update receives the API payload instead of the logged update

JakobLichterfeld:
Fixes #5656.

## What happened

The cancel branch of the `:updating` state shadowed the `%Log.Update{}` row held in the state with the API `software_update` struct (`%SW{status: "available"} = update`) and passed the wrong one to `Log.cancel_update/1`. Every fetch then died in a `FunctionClauseError` crash loop and the started update row stayed open forever. The mock-based test could not see it: `LogMock.cancel_update` accepts any argument, and the test asserted `{:cancel_update, _update_id}`.

The characterization suite (#5652, #5658) exposed the crash — `updating_cancel` pinned it as a marked bug with an `expect_restart` declaration.

## The fix

- Bind the API struct under its own name (mirroring the finish branch below it) and cancel with the `Log.Update` row; the log line keeps printing the API struct.
- Tighten the mock assertion to `{:cancel_update, %TeslaMate.Log.Update{}}` so this bug class is visible to the mock test as well.
- `updating_cancel` drops `expect_restart` and its `await` (the harness would rightly fail a declared restart that no longer happens) and its golden is re-recorded: the update row now closes (`end_date` set) and the vehicle returns to online — no crash artifacts.

## Verification

Full test suite green: 580 passed (characterization replays, harness unit tests, mock-based updating tests).

None.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Under the hood, this release adds a self-verifying black-box characterization su
 
 ### Improvements and bug fixes
 
+- fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
+
 #### Build, CI, internal
 
 - test: add characterization harness replaying API fixtures against persisted rows and MQTT (#5653 - @JakobLichterfeld)
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -1429,10 +1429,10 @@ defmodule TeslaMate.Vehicles.Vehicle do
         {:keep_state, %{data | last_used: DateTime.utc_now()},
          schedule_fetch(default_interval(), data)}
 
-      %VehicleState{software_update: %SW{status: "available"} = update} ->
+      %VehicleState{software_update: %SW{status: "available"} = software_update} ->
         {:ok, %Log.Update{}} = call(data.deps.log, :cancel_update, [update])
 
-        Logger.warning("Update canceled:\n\n#{inspect(update, pretty: true)}",
+        Logger.warning("Update canceled:\n\n#{inspect(software_update, pretty: true)}",
           car_id: data.car.id
         )
 
--- a/test/fixtures/characterization/goldens/updating_cancel.json
+++ b/test/fixtures/characterization/goldens/updating_cancel.json
@@ -119,13 +119,6 @@
         "id": "update_1",
         "start_date": "2024-01-01T00:00:10.000000",
         "version": "2019.8.4 530d1d3"
-      },
-      {
-        "car_id": "$car",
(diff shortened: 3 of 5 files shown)


JakobLichterfeld (commented):
Review pass over the fix (mechanism, blast radius, characterization contract):

- **Root cause verified at the exact line:** `%SW{status: "available"} = update` shadowed the `{:updating, update}` state token, so `Log.cancel_update/1` received the API `software_update` struct instead of the logged `%Log.Update{}` row — a `FunctionClauseError` on every fetch in the available state, leaving the update row open. The fix is a pure binding rename with zero logic change; the call uses the outer token again, and the re-recorded golden proves that token is the row: the update now closes.
- **The assertion that hid the bug is closed:** `assert_receive {:cancel_update, %TeslaMate.Log.Update{}}` — the mock previously accepted any struct, which is exactly why only the characterization replay through the real `Log` could expose #5656.
- **The pinning contract worked as designed:** after the fix the vehicle no longer goes down, `expect_restart` fails the replay loudly, and the scenario correctly drops the declaration and its `positions` convergence point (no await needed — an update never closes the open online states row). The golden diff is the documented behavioral change: update row closed, no restart artifacts.

---

PR #5665, opened 2026-08-30
fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update

### LINKED ISSUE #5661: Invalid SVG path data for the direction arrow floods the browser console in Safari

### What happened?

The path data for the vehicle direction arrow has a stray closing brace. In [`assets/js/hooks.js#L156`](https://github.com/teslamate-org/teslamate/blob/v4.2.0/assets/js/hooks.js#L156):

```js
const path = this._empty() ? "" : `M0,${3} L-4,${5} L0,${-5} L4,${5} z}`;
```

The `}` after `z` is not valid SVG path data. Safari rejects it and logs an error every time the path is redrawn:

```
Error: Problem parsing d="M0,3 L-4,5 L0,-5 L4,5 z}"
```

`_updatePath()` runs on every position update, so while a car is driving the console fills up with hundreds of identical errors within minutes, which buries anything else one might be trying to debug. Chromium ignores the trailing garbage silently, which is presumably why this went unnoticed.

The arrow itself still renders correctly — browsers draw the valid prefix — so this is noise rather than a visual defect.

While fixing it, the template literals on that line can go too: `M0,${3}` just evaluates to `M0,3`.

```js
const path = this._empty() ? "" : "M0,3 L-4,5 L0,-5 L4,5 z";
```

I am happy to open a PR for this if that helps.

### Expected Behavior

Valid path data, and a quiet console.

Yes

1. Open TeslaMate in Safari with a vehicle that is driving, so positions update frequently.
2. Open the Web Inspector console.
3. One `Problem parsing d=...` error appears per position update.

### Relevant log output

```shell
Browser console, Safari 26.6.2, TeslaMate home page with a driving vehicle:

[Error] Error: Problem parsing d="M0,3 L-4,5 L0,-5 L4,5 z}"  _setPath  app-<hash>.js:1:136261
[Error] Error: Problem parsing d="M0,3 L-4,5 L0,-5 L4,5 z}"  _setPath  app-<hash>.js:1:136261
[Error] Error: Problem parsing d="M0,3 L-4,5 L0,-5 L4,5 z}"  _setPath  app-<hash>.js:1:136261
... repeated for every position update
```

### Version

v4.2.0

### Environment

* Docker, `teslamate/teslamate:latest`
* Safari 26.6.2 on macOS 26.6.2 — logs the error
* Chromium — silently ignores it

JakobLichterfeld:
## What

The path data for the vehicle direction arrow ended in `z}`:

```js
const path = this._empty() ? "" : `M0,${3} L-4,${5} L0,${-5} L4,${5} z}`;
```

The trailing `}` is not valid SVG path data. Safari rejects the whole `d` attribute and logs `Error: Problem parsing d="M0,3 L-4,5 L0,-5 L4,5 z}"` every time the path is redrawn. Since `_updatePath()` runs on every position update, a driving vehicle fills the console with hundreds of identical errors, burying anything else one might be debugging. Chromium ignores the trailing garbage silently, which is why this went unnoticed.

The arrow itself rendered correctly all along — browsers draw the valid prefix — so this was console noise, not a visual defect.

## How

Drop the stray brace and replace the no-op template literals (`M0,${3}` just evaluates to `M0,3`) with a plain string:

```js
const path = this._empty() ? "" : "M0,3 L-4,5 L0,-5 L4,5 z";
```

The geometry is byte-for-byte the same as what browsers previously parsed, so the arrow is unchanged.

## Verification

- Safari 26.6.2 / macOS 26.6.2: no `Problem parsing d=...` entries in the Web Inspector console any more; the arrow still renders and rotates with the heading.
- `npx prettier --check js/hooks.js` passes.

None. The invalid token is removed at the source, no compatibility shim or browser-specific handling was needed.

Reported and diagnosed by @Corrugator.

Closes #5661

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Opus 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 ### Improvements and bug fixes
 
 - fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
+- fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update (#5665 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/assets/js/hooks.js
+++ b/assets/js/hooks.js
@@ -153,7 +153,7 @@ const DirectionArrow = CircleMarker.extend({
       `translate(${x},${y}) rotate(${this._heading})`,
     );
 
-    const path = this._empty() ? "" : `M0,${3} L-4,${5} L0,${-5} L4,${5} z}`;
+    const path = this._empty() ? "" : "M0,3 L-4,5 L0,-5 L4,5 z";
 
     this._renderer._setPath(this, path);
   },


JakobLichterfeld (commented):
Self-review plus an automated pass found nothing to flag: the new path string is byte-for-byte the prefix browsers were already drawing, so this is purely removing the invalid trailing token — no visual change, just a quieter Safari console.

Thanks again @Corrugator for the sharp-eyed report and diagnosis in #5661 — Chromium swallowing the garbage silently is exactly why this survived so long. Merging.

---

PR #5666, opened 2026-08-31
fix(web): pin the size of Leaflet's SVG overlay so the vehicle arrow and geofence circle stay on the map at any Safari page zoom

### LINKED ISSUE #5660: Safari 26: vehicle arrow and geofence circle are drawn ~10% of the map size off, on every zoom level

### What happened?

In Safari 26, everything TeslaMate draws into Leaflet's SVG overlay pane is rendered at a constant pixel offset — roughly 10 % of the map size, towards the bottom right. This affects the vehicle direction arrow on the home page and the circle in the geofence editor.

Because the offset is constant *in pixels*, the marker appears to wander geographically as you zoom: a few metres off at high zoom, several hundred metres off when zoomed out. It is never correct at any zoom level.

Map tiles are fine, and so is everything in the marker pane — the leaflet-geoman drag handles sit exactly where they belong. Only the SVG overlay pane is displaced. In the geofence editor this is very visible: the blue circle is clearly separated from the white centre handle that is supposed to be at its centre.

Not reproducible in Chromium.

### Expected Behavior

The direction arrow marks the vehicle's position, and the geofence circle is centred on its centre handle — in Safari too.

Yes

1. Open TeslaMate in Safari 26 on macOS 26.
2. Go to Geo-Fences and edit any geofence.
3. The blue circle does not sit on the white centre handle; it is offset to the bottom right.
4. Zoom in and out with the +/− buttons. The distance between circle and handle stays the same in pixels, so the geographic error grows as you zoom out.

Same effect on the home page. There the arrow should always be exactly at the map centre, because the map re-centres on the vehicle with `setView([lat, lng], getZoom())` on every position update. It never is.

### Analysis

Leaflet renders vector overlays into an `<svg>` that is `Renderer.options.padding` (10 %) larger than the visible map, and shifts it back by that amount using a CSS transform on the `<svg>` element. Safari 26 applies that transform to the element *box* but paints the *content* unshifted.

Measured in the geofence editor, map container roughly 740 × 400 px:

| what | value |
| --- | --- |
| computed transform of the overlay `<svg>` | `matrix(1, 0, 0, 1, -74, -39.999999)` |
| its `getBoundingClientRect()` relative to the map container | `-74, -40` (correct) |
| circle centre minus centre handle | `78, 44` (should be `0, 0`) |

The rendering error is exactly the negative of the transform. Verified on several zoom levels, on both the OSM and the Google hybrid layer, parked and while driving.

Cross-check in Chromium against the same TeslaMate instance: I converted the marker's screen position back into a coordinate using the tile URLs actually loaded (`z/x/y` plus each tile's layout position) and compared it with the value in the hidden `position_*` input. Deviation 0.04 m to 0.64 m at map zoom 16, 17, 18 and 19 — well below one pixel. So the projection and the stored coordinates are fine; this is purely a Safari rendering problem.

Ruled out along the way:

* `iconAnchor` is correctly set to `[12, 40]`, and neither affected layer is an image marker anyway — the arrow is a `CircleMarker` subclass with an SVG path, the geofence is an `L.Circle`.
* `will-change: transform` on `svg.leaflet-zoom-animated` is not the trigger. Setting it to `auto` changes nothing: `[78, 44]` before, `[79, 43]` after.

Replacing the transform with `left`/`top` fixes it immediately and reliably:

```js
document.querySelectorAll('svg.leaflet-zoom-animated').forEach((s) => {
  const m = new DOMMatrix(getComputedStyle(s).transform);
  s.style.transform = 'none';
  s.style.left = m.e + 'px';
  s.style.top = m.f + 'px';
});
```

### Possible workaround in TeslaMate

Setting `window.L_DISABLE_3D = true` before Leaflet initialises makes `L.DomUtil.setPosition` use `left`/`top` everywhere and avoids the bug entirely. The cost is Leaflet's zoom and fade animations. A narrower alternative is to override `L.SVG.prototype._update` and reposition only the SVG container, which keeps the animations.

The root cause looks like a WebKit bug rather than a TeslaMate one, and I intend to report it there as well. But since it makes the maps effectively unusable for every Safari user, a workaround here might be worth it.

### Relevant log output

```shell
No server-side log output — this is entirely client-side rendering.
Browser console values from the geofence editor in Safari 26.6.2:

> s = document.querySelector('.leaflet-overlay-pane svg'),
  c = s.closest('.leaflet-container').getBoundingClientRect(),
  r = s.getBoundingClientRect(),
  [Math.round(r.left - c.left), Math.round(r.top - c.top), getComputedStyle(s).transform]
< [-74, -40, "matrix(1, 0, 0, 1, -74, -39.999999)"]

> (() => { const c = document.querySelector('.leaflet-container'),
    p = c.querySelector('.leaflet-overlay-pane path'),
    m = c.querySelector('.leaflet-marker-icon'),
    b = p.getBoundingClientRect(), a = m.getBoundingClientRect();
    return [Math.round(b.left + b.width/2 - a.left - a.width/2),
            Math.round(b.top + b.height/2 - a.top - a.height/2)] })()
< [78, 44]
```

### Version

v4.2.0

### Environment

* Docker, `teslamate/teslamate:latest`, Postgres 17, on macOS
* Safari 26.6.2 on macOS 26.6.2 — broken
* Chromium — correct

JakobLichterfeld:
## What

In Safari, everything TeslaMate draws into Leaflet's SVG overlay pane is displaced whenever the page zoom is not 100 % — the direction arrow on the home page and the circle in the geofence editor. The offset is constant in pixels, so the error grows geographically as you zoom out: the arrow can sit hundreds of metres from the vehicle's actual position, silently and with nothing to connect it to the zoom setting. Safari stores page zoom per origin and permanently, so one accidental Cmd+− is enough to leave a user with a permanently wrong map.

## Cause

Bulma's generic reset contains `svg { height: auto; width: auto }`. By value that is a no-op — `auto` is the initial value of `width` — and its only effect is to take the sizing away from the `width`/`height` presentation attributes that Leaflet's `SVG._update` writes. At any page zoom other than 100 %, Safari then resolves the used size as attribute ÷ zoom while the `viewBox` keeps the unscaled values, so every point is displaced proportionally to its distance from the viewBox origin by a factor of 1/zoom. Below 100 % the offset is positive, above it the sign flips.

## How

Pass an `SVG` renderer that mirrors the two attributes into inline styles, which outrank any author rule. It writes the same numbers Leaflet already writes, so it is a no-op in every browser that resolves the size correctly — no change to rendering, animations or interaction anywhere else.

The renderer is passed via the documented `Map` `renderer` option and hooks the documented `Renderer` `update` event, so no private Leaflet API is touched and no shared prototype is mutated.

## Verification

Verified against the bundled Leaflet in a probe page (Chromium, with Bulma's declaration active):

| Assumption | Result |
| --- | --- |
| `Map.options.renderer` beats the default renderer | `map.getRenderer(circle) === renderer` → `true` |
| `update` fires, `getPane()` resolves | inline styles set to `883px × 480px` |
| styles track size changes | after `invalidateSize` + `setZoom`: attributes `600 × 360`, style `600px × 360px` |
| no regression | circle bbox centre == container centre, `[0, 0]` |

**Not verified here:** that the offset is gone in Safari. I have no WebKit available; Chromium is the browser that was never affected. @Corrugator, a confirmation at 75 %, 85 % and 125 % would be very welcome before this is merged.

This is a workaround for a WebKit bug, and it is marked as such in the code with its removal criterion. Three upstream reports belong together here:

- **WebKit** — resolving an SVG's intrinsic size as attribute ÷ page zoom is the actual defect. Not yet filed.
- **Leaflet** — `SVG._update` should write the size to `container.style`, which would make the renderer immune to any CSS reset touching `svg`, not just Bulma's. Filed as Leaflet/Leaflet#10356; this workaround comes out once that lands.
- **Bulma** — the `svg { height: auto; width: auto }` rule is a no-op by value whose only possible effect is harm. Worth filing so all reports can reference each other. Not yet filed.

Known limitation: layers on a custom pane get their own renderer via `_getPaneRenderer` and are not covered. The direction arrow and the geofence circle both live in the default overlay pane.

Reported and diagnosed by @Corrugator, who traced it to the Bulma declaration and modelled the 1/zoom factor.

Closes #5660

---

🤖 Created with [Claude Code](https://claude.com/claude-code) (Opus 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Upgrading directly from 4.1.x no longer preserves entity registry customizations
 - feat: use Grafana 13.2.1 (#5694 - @swiffer)
 - fix(mqtt): stop re-running the Home Assistant discovery migration on every restart (#5685 - @nebhale)
 - fix(vehicle): keep the published state start time from jumping backwards after charging, updating or driving (#5706 - @JakobLichterfeld)
+- fix(web): pin the size of Leaflet's SVG overlay so the vehicle arrow and geofence circle stay on the map at any Safari page zoom (#5666 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/assets/js/hooks.js
+++ b/assets/js/hooks.js
@@ -114,6 +114,7 @@ import {
   Icon,
   Circle,
   CircleMarker,
+  SVG,
 } from "leaflet";
 
 import markerIcon from "leaflet/dist/images/marker-icon.png";
@@ -159,8 +160,33 @@ const DirectionArrow = CircleMarker.extend({
   },
 });
 
+// Safari resolves an SVG's intrinsic size as attribute / page zoom, so Leaflet's
+// overlay pane renders displaced at any page zoom other than 100 % as soon as an
+// author rule takes sizing away from the width/height attributes - Bulma's global
+// `svg { width: auto; height: auto }` does exactly that. Mirroring the attributes
+// into inline styles pins the used size; it is a no-op in every other browser.
+// Layers on a custom pane get their own renderer and are not covered; the
+// direction arrow and the geofence circle both live in the default overlay pane.
+// Remove once Leaflet sets the size via style: Leaflet/Leaflet#10356
+function createSvgRenderer() {
+  const renderer = new SVG();
+
+  renderer.on("update", () => {
+    const el = renderer.getPane().querySelector("svg");
+    if (!el) return;
+
+    el.style.width = `${el.getAttribute("width")}px`;
+    el.style.height = `${el.getAttribute("height")}px`;
+  });
+
+  return renderer;
+}
+
 function createMap(opts) {
-  const map = new M(opts.elId != null ? `map_${opts.elId}` : "map", opts);
+  const map = new M(opts.elId != null ? `map_${opts.elId}` : "map", {
+    ...opts,
+    renderer: createSvgRenderer(),
+  });
 
   const osm = new TileLayer("https://tile.openstreetmap.org/{z}/{x}/{y}.png", {
     maxZoom: 19,


JakobLichterfeld (commented):
I ran a focused review of this change against the bundled Leaflet 1.9.4 source and found no issues:

- **Event contract holds**: `SVG._update` writes the `width`/`height` attributes *before* firing `update`, so the handler never reads `null`, and `getPane()` is guaranteed to resolve since the event only fires after the renderer is added to a map.
- **Call sites are safe**: both `createMap` consumers (home page map, geofence editor) use only the default overlay-pane renderer — no code path puts a second `<svg>` into the pane, so the `querySelector("svg")` lookup is unambiguous. Geoman draws through the map renderer as well.
- **No behavior change elsewhere**: the inline styles mirror the exact values Leaflet already writes, one renderer per map, no shared state, no listener leak.
- The workaround is properly scoped: marked as such in the code, with a removal criterion (Leaflet/Leaflet#10356) and the custom-pane limitation documented.

The one thing still open is the Safari confirmation at non-100 % page zoom — @Corrugator, a quick check at 75 %, 85 % and 125 % would let us merge this.

---

JakobLichterfeld:
@corrugator please test this PR, following https://docs.teslamate.org/docs/development#testing-with-our-ci-which-builds-the-docker-images-automatically-per-pr and use `pr-5666`

Corrugator:
Tested with `ghcr.io/teslamate-org/teslamate:pr-5666` against my existing database, with the reverse-proxy shim I had been running switched off, so this PR is the only thing in play.

Safari 26.6.2 on macOS 26.6.2, Retina. Home page and geofence editor, at 75%, 85%, 100% and 125% page zoom, reloading after each change. The vehicle arrow sits at the map centre and the geofence circle sits on its centre handle at every level. Nothing in the console.

Cross-checked in Chromium on the same build: the overlay `<svg>` carries `width: 883px; height: 480px` inline, matching its attributes, and the distance from circle centre to centre handle is `[1, 0]`.

Works for me.

JakobLichterfeld:
Thank you for the thorough verification — exactly the test matrix this needed, and running the PR image with the reverse-proxy shim disabled makes it a clean result: the fix alone resolves the offset at every zoom level, in both the home page and the geofence editor, with Chromium confirming the inline styles are a faithful mirror of the attributes.

That was the last open point, so this is ready to merge. Thanks again for the original report and the diagnosis down to the Bulma declaration and the 1/zoom model — that groundwork is what made the fix this small.

---

PR #5668, opened 2026-09-01
test(characterization): convert streaming scenarios to characterization fixtures

JakobLichterfeld:
Part of #5652 (phase 1 characterization suite), following the merged driving/charging/updating conversions.

## What this adds

Five commits, production change strictly separated from test infrastructure:

1. **`refactor(api)`: extract the stream wire decoder** as `TeslaApi.Stream.decode_frame!/1` — a behaviour-neutral visibility extraction of the `data:update` CSV zip out of `handle_frame`, so the same production code decodes recorded wire frames. Pinned by the existing streaming tests.
2. **Stream events in the replay harness.** Scenarios carry frames in wire form (`{"stream": "<data:update CSV value>"}`) plus a `stream_control` key whitelisted to `inactive` and `vehicle_offline`. ApiMock delivers pending stream events at the serve boundary through the receiver the vehicle registered when connecting the stream and syncs via `:sys.get_state/1` before serving the next API event — deterministic interleaving, no wall-clock waits. Serve indices go to API events only, keeping the causal barrier and the vacuity probe anchored to the terminal API event. Machinery evidence: a `stream` selftest pair whose drive positions carry the wire frames' coordinates.
3. **User-action call events.** `{"call": "suspend_logging"}` performs the user action at the serve boundary through a proxy task: the suspend handler's synchronous strict fetch is answered from the event queue through the regular exec path — the scenario declares the response as the API event directly after the call, and its serve counts like any other (index, barrier, vacuity). Rejected, dying or hanging calls raise named errors. The suspend handler resets its poll task, so the mock re-serves the last payload to the doomed caller instead of letting it consume a scenario event — building this surfaced that the zombie poll had silently eaten an event and dropped a charging process from a golden. A replay must not end in `:suspended` (the state only survives at test speed; a golden would freeze a transitional state whose real duration is minutes) — enforced after the awaited outcomes, and a barrier starved by the suspend interval names the state instead of timing out opaquely. Registered names are unique per replay, so a terminating straggler can never overlap the next test. Machinery evidence: a `suspend` selftest pair pinning the MQTT state course and the declared strict-fetch consumption.
4. **All eleven scenarios** from `streaming_test.exs` converted: starting a drive from frames, geofence transitions from streaming positions (resolved against the real `TeslaMate.Locations` — the original asserted against a mocked dependency, the distance resolution itself is now under the pin), a stale poll surviving a stream drive, a stale D frame whose discard is pinned by the absence of a drive, charge detection from a negative-power frame, the stream disconnect on entering `:updating`, the `vehicle_offline` control triggering a state fetch, and the four suspended scenarios (stream-inactive fetch, stale frame while suspended, resume into a drive, resume into a charge) entering `:suspended` via the declared call. The original tests remain in place.
5. **`fix(test)`: start a sandbox owner per test instead of riding a shared one.** The case templates rescued `:already_shared` into `:ok` — covering the benign second case template in the same test module, but also letting a test run on the previous test's shared owner while its release was still in flight, reading its uncommitted rows or losing the connection mid-test. Three intermittent characterization failures traced back to this four-year-old channel, which in record mode could have captured foreign rows into a golden. The owner is now tracked per test process (shared `sandbox_owner.ex`, all three case templates): a second template reuses it, a foreign shared owner is waited out briefly and then refused with a named error.

## Enforced invariants (each names its check)

- A stream event before the stream is connected raises (`ApiMock.deliver_stream/2`).
- Unsupported `stream_control` values raise carrying the rationale: the reconnecting controls call back into the API synchronously and would deadlock against the serve-boundary sync (`stream_control!/1`).
- A scenario cannot end in a stream or call event — the terminal must be an API event, because the causal barrier repeats it (`build_events/1`).
- `timebase` combined with stream events raises instead of silently mixing epochs (`shift_timebase/1`).
- Stream deliveries never touch the terminal serve counter (`index_events/1`, unit-tested).
- Call events count nothing twice: the strict fetch inside the call handler is served through the regular exec path; unsupported, rejected, dying or never-completing calls raise (`call_event!/1`, `ApiMock.await_call_outcome/2`).
- A replay ending in `:suspended` raises after the awaited outcomes, and a barrier starved by the suspend poll interval names the state (`replay/2`, `await_serve/4`).

## Verification

Full suite green under the fixed sandbox: **600 passed** — 43 pairs replay green (35 content + 8 self-test), goldens recorded with the double-run volatility mask. All goldens recorded before the sandbox fix compare green on top of it, certifying none captured foreign rows; one unrelated flake of `driving_offline_15min_new_drive` in an early record run (tracked in #5669). The vacuity probe rejected two immediate asleep awaits during conversion — removed, working as designed.

- The `timebase: "now"` interim status from #5654 remains; the flake above is its first observed occurrence and has its own issue (#5669).
- The hardcoded streaming suspend timings (`{3, 10}` wall-clock minutes) remain a documented limit — the suspended scenarios enter `:suspended` through the declared call, not the idle timer, so no scenario depends on them.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)


JakobLichterfeld (commented):
Review pass over all five commits (decoder extraction, stream machinery, call machinery, the eleven conversions, the sandbox fix — fixture fidelity against `streaming_test.exs`, deadlock analyses, barrier/probe interactions, and the golden certification):

- **The extraction is behaviour-neutral and correctly cut:** `decode_frame!/1` is the exact zip `handle_frame` used, now called by it — one decoder for production and fixtures, no duplicated column list.
- **Both delivery designs survived their pre-build fact checks.** Stream side: every message class was traced in `vehicle.ex`; the reconnecting controls call back into the API synchronously and are rejected with exactly that rationale. Call side: the suspend handler's synchronous `fetch_strict` makes a blocking delivery deadlock deterministically — the proxy-task mechanism answers the strict fetch from the queue through the regular exec path, so the scenario declares that response explicitly and its serve counts like any other. Both syncs are event-anchored; no wall-clock waits anywhere.
- **The machinery caught two real defects while being built** — the suite working on itself: the zombie poll after `task: nil` had silently consumed an event and recorded a golden without its charging process (now re-served consumption-neutrally, pinned by the selftest), and the vacuity probe rejected two immediate asleep awaits (`:suspended → :asleep` completes in-cycle), which were removed.
- **Sequence fidelity, now 11/11:** the seven poll/stream scenarios map one-to-one to their originals (with two disclosed improvements — the stale-frame conversion pins the discard by the absence of a drive instead of a no-op's non-effect, and the geofence scenario pins real distance resolution instead of a mock). The four suspended scenarios enter `:suspended` via the declared user-action call and end asleep; every misuse path of the new vocabulary raises with a named error and carries its own probe.
- **The sandbox fix closes a four-year-old silent-green channel** — a test riding the previous test's shared owner could read foreign uncommitted rows, which in record mode could have poisoned goldens. The per-test owner tracking keeps the benign same-test reuse and refuses foreign owners loudly. Certification: every golden recorded before the fix compares green on top of it (600/600), so none captured foreign rows.
- **The one red run was handled correctly:** unrelated to this PR, named honestly in Verification, tracked with data and a refuted first hypothesis in #5669.

---

PR #5670, opened 2026-09-01
fix(test): give the error_event selftest a settled terminal cycle

JakobLichterfeld:
The scenario reached its terminal through the two-call unavailable fallback, so both causal-barrier serves came from one fetch cycle and teardown raced identify: fast machines recorded and compared nil for the car's config attributes, CI let identify land and failed on DeepBlue. A fourth online event separates the terminal from the error recovery; the re-recorded golden now pins the identified car — the documented serve-vs-cycle limit caught shipping its first violation.

---

🤖 Created with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)


JakobLichterfeld (commented):
Review pass: the fix matches the diagnosed mechanism exactly — the fourth online frame puts a fully processed cycle between error recovery and the causal barrier, and the re-recorded golden proves it: the car's identified attributes (`DeepBlue`, `74D`) now appear deterministically in both the car row and the MQTT topics that previously raced teardown. Production untouched; the serve-vs-cycle limit remains documented and gets its enforcement issue separately.

---

PR #5680, opened 2026-09-02
feat(grafana): add `total` period to the Statistics dashboard for one aggregated row over the selected time range

micku7zu:
Add 'total' option to period query to achieve one aggregated row over the selected time range.

image

### DIFF
--- a/grafana/dashboards/statistics.json
+++ b/grafana/dashboards/statistics.json
@@ -1187,7 +1187,7 @@
         "label": "Period",
         "name": "period",
         "options": [],
-        "query": "day,week,month,year",
+        "query": "day,week,month,year,decade",
         "type": "custom",
         "valuesFormat": "csv"
       },


JakobLichterfeld (changes requested):
Thanks for the PR!

Before we can look at merging this, two things please:

1. **Add a PR description.** Currently it is empty. Please describe what the change does, why you want it, and how you verified it. See our [development guide](https://docs.teslamate.org/docs/development).

2. **Test the change before opening a PR.** Selecting `decade` currently produces wrong labels: all four panel queries build the row label with `CASE '$period'`, which only has branches for `month`, `year` and `week`. `decade` falls through to the `ELSE` branch and is rendered as `YYYY-MM-DD`, so the 2020–2029 bucket shows up as `2020-01-01` and looks like a daily row. This is visible on the very first try. Please add a `decade` branch to the `CASE` in each of the four queries (drives, charges, consumption net, consumption gross), e.g.

   ```sql
   WHEN 'decade' THEN to_char(timezone('$__timezone', date), 'YYYY') || 's'
   ```

3. **Update the dashboard screenshot.** As described in the [development guide - Making Changes to Grafana Dashboards](https://docs.teslamate.org/docs/development#making-changes-to-grafana-dashboards), dashboard changes come with an updated screenshot (1920x1080 @ x2, Grafana in kiosk mode, sensitive information blurred) replacing `website/static/screenshots/statistics.png` in this PR.

---

🤖 Review drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 medium) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
One more question, independent of the points above: what is the use case for `decade`? With TeslaMate data starting in 2019 at the earliest (first commit to TeslaMate on 31st of March 2019), a decade bucket collapses everything into a single "2020s" row. If the goal is a total over all data, setting the time range to the full period and choosing `year` already gives that, including the yearly breakdown. We are careful with adding options to the dashboards, so I would like to understand what this enables that is not possible today.

micku7zu:
1. Will update the PR
2. Will update the CASEs 
3. Will update the screenshots



> One more question, independent of the points above: what is the use case for `decade`? With TeslaMate data starting in 2019 at the earliest (first commit to TeslaMate on 31st of March 2019), a decade bucket collapses everything into a single "2020s" row. If the goal is a total over all data, setting the time range to the full period and choosing `year` already gives that, including the yearly breakdown. We are careful with adding options to the dashboards, so I would like to understand what this enables that is not possible today.

The difference is that using `year` all the data will be aggregated one year per row. Using decade, all the data will be aggregated in a single row. 

After every TeslaMate update I always add `decade` to my own setup in order to have all the info aggregated in a single row. I thought other people might find this useful but I also never observed the weird '2020-01-01' label there.

I will update the PR in the next days, and if you find it useful, you can merge it. No problem if you think it is not worth 😄 

I know that this info might be visible in other pages as well (for the whole period), but I like this page format and I need an option to aggregate all into a single row, not just max `per year`.

JakobLichterfeld:
Thanks for explaining, that makes the goal clear: one aggregated row over the selected time range.

`decade` only delivers that by accident. All existing data happens to fall into the 2020s, so from 2030 on you would get two rows again, and the label "2020s" describes the calendar decade rather than the selected range (choose 2024 as time range and it still says "2020s").

If we add this, I would rather do it as what it is: a `total` option that yields a single row spanning the selected range, labeled "Total". In the four queries that means:

- `date`: `CASE WHEN '$period' = 'total' THEN <a constant> ELSE date_trunc('$period', …) END`
- `date_from` / `date_to`: `$__from` / `$__to` for `total`, otherwise the current expressions
- `display`: `WHEN 'total' THEN 'Total'`

Would you be up for reworking the PR in that direction? If that is more than you want to take on, no hard feelings, and keeping `decade` as a local patch is a fine solution for your setup.

---

drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 medium) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

micku7zu:
1. I updated the PR description
2. I updated the CASEs
3. I'm not sure if the screenshot needs any updates. The screenshot for `statistics` is with the default `month` period. Creating a new screenshot will not show the new `total` period. Replacing the default screenshot with the `total` period doesn't make sense either. 
image
image

JakobLichterfeld (approved):
Thanks, that looks good. Verified the four queries: the constant `date` keeps the `seriesToColumns` join intact, `$__from`/`$__to` give the links the selected range, and switching to `('1 ' || '$period')::interval` is indeed required since an `interval '1 total'` literal would fail at parse time even in the untaken branch.

Agreed on the screenshot: the default view with `month` is unchanged and this is a small change, so no update needed.

Merging, thanks for the contribution.

---

🤖 Review drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 medium) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5681, opened 2026-09-02
fix(test): move the offline-resume scenario off the 15-minute boundary

### LINKED ISSUE #5669: Flaky characterization compare: driving_offline_15min_new_drive ends the first drive one position late
### Flaky characterization compare: `driving_offline_15min_new_drive` ends the first drive one position late

Observed once (2026-09-01), as the first test of a cold-started `CHARACTERIZATION_RECORD=1` run (seed 554375):

```
first divergence at database.drives.[0].end_position_id
expected: "position_3"
actual:   "position_4"
```

The signature means the resume position after the 15-minute offline gap was inserted into the *old* drive instead of starting a new one. Only two code paths can do that:

- the continue branch of the offline resume (`vehicle.ex`, `not is_nil(drv)` cond arm) — requires the computed `offline_min < 15`, and
- online-while-still-unavailable (`vehicle.ex`, `{:driving, {:unavailable, n}}` + online) — requires the unavailable counter not to have reached 15, i.e. at least six of the 21 served offline events losing their tick.

Instrumented bounded reproduction (probe logging on the unavailable ticks and on both resume branches; 5 runs: same seed cold-started, two under full CPU load) did **not** reproduce the signature and **disproved the tick-loss variant for those runs**: no tick was ever lost, and the scenario consistently resumed via the offline branch with `offline_min = 15.0` — exactly on the `>= @drive_timeout_min` boundary. (One full-load run failed differently: DB sandbox ownership timeout — artificial saturation noise.)

Hypothesis (unproven): the resume decision sits on a knife edge at exactly 15.0 minutes of event time; any condition that makes the computed gap come out below 15 — e.g. a different `last_response` binding at the unavailable→offline hand-over — flips it into the continue branch. This interacts with the `timebase: "now"` interim status documented in the driving conversion (#5654): wall-clock coupling of this scenario family is known and not yet eliminated.

Repro recipe: add warning logs (a) in the `{:driving, {:unavailable, n}} + offline` clause printing `n + 1`, (b) before the offline-resume `cond` printing `offline_min`/`has_gained_range?`/`drv`, (c) in the online-while-unavailable clause; then `CHARACTERIZATION_RECORD=1 mix test test/teslamate/characterization_test.exs --seed 554375` on a cold BEAM. Widening the scenario's offline gap away from exactly 15.0 event-minutes would also tell whether the knife edge is the cause.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

JakobLichterfeld:
Fixes #5669.

## What happened

`driving_offline_15min_new_drive` resumed exactly 15.0 event-minutes after going offline — precisely on the `>= @drive_timeout_min` edge of the resume decision — and once, on a cold-started record run, closed the first drive one position late. The bounded instrumented reproduction in the issue ruled out lost unavailable ticks for the probed runs and left the boundary as the prime suspect.

## The fix

The resume frames move 60 seconds later, so the gap is 16 minutes and the decision no longer sits on the boundary. The scenario description states the deviation from the original test's "15 minutes" and points to the issue.

The re-recorded golden is byte-identical: the shifted times are volatile-masked fields, and the drive layout (drive 1 = positions 2–3, drive 2 = positions 5–6) is unchanged — the scenario still pins the same two drives.

This is also the discriminating experiment named in the issue: if the signature ever recurs at 16 minutes, the boundary hypothesis is refuted and only the tick-loss path remains.

## Verification

Full test suite green: 600 passed.

- The exact-15.0 boundary itself is not pinned; that becomes possible once the harness owns the clock (the `timebase: "now"` interim status from #5654 remains).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 
 - fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
 - fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update (#5665 - @JakobLichterfeld)
+- fix(test): move the offline-resume scenario off the 15-minute boundary (#5681 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/test/fixtures/characterization/scenarios/driving_offline_15min_new_drive.json
+++ b/test/fixtures/characterization/scenarios/driving_offline_15min_new_drive.json
@@ -7,7 +7,7 @@
     "vid": 1000,
     "vin": "5YJ3E1EA1KF000001"
   },
-  "description": "Conversion of 'logs a drive after a significant offline period while driving' (streaming enabled): 15 minutes offline mid-drive without SOC gain. The first drive is closed by timeout; the frames after the gap start a second drive. Streaming is declared explicitly: with an active stream the vehicle's suspend settings are hardcoded to {3, 10} and this scenario's suspend guards are dead \u2014 the asleep terminal exists because a parked streaming terminal grows positions without bound under repeats.",
+  "description": "Conversion of 'logs a drive after a significant offline period while driving' (streaming enabled): 16 minutes offline mid-drive without SOC gain — one minute past the 15-minute drive timeout rather than exactly on it, so the resume decision does not sit on the >= boundary (#5669). The first drive is closed by timeout; the frames after the gap start a second drive. Streaming is declared explicitly: with an active stream the vehicle's suspend settings are hardcoded to {3, 10} and this scenario's suspend guards are dead — the asleep terminal exists because a parked streaming terminal grows positions without bound under repeats.",
   "await": {
     "state": "asleep",
     "drives_closed": 2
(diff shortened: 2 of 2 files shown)


JakobLichterfeld (commented):
Review pass over the change (scenario arithmetic, event ordering, golden invariance):

- **The boundary is confirmed in the diff itself:** the last online frame before the outage sits at `…210001`, the old resume at `…110001` — exactly 900 000 ms, 15:00 on the `>=` edge of the resume decision. The new resume at `…170001` is exactly 16:00.
- **Both resume frames move consistently** (all five timestamp fields each, the parked frame still one second after the resume), and nothing timestamped follows them — the series stays monotonic, so no stale-fetch side effect. The 21 offline events, `t0`, and the `await` are untouched.
- **The golden is unchanged, and correctly so:** every absolute date in this `timebase` scenario is volatility-masked and the derived values do not depend on the gap length, so the pinned behavior — first drive closed by timeout, second drive from the resume frames — is byte-for-byte the same. Only the exactly-on-the-edge behavior is no longer pinned, which the description states and #5669 documents.
- **The change doubles as the discriminating experiment from the issue:** if the signature ever recurs at 16:00, the boundary hypothesis is refuted and the tick-loss path reopens.

--- 

PR #5683, opened 2026-09-02
test(characterization): enforce the two-call-terminal limit

### LINKED ISSUE #5682: test(characterization): enforce the two-call-terminal limit instead of documenting it
The harness moduledoc documents a limit: "the end-of-replay barrier counts API serves, not fetch cycles; states whose handling makes two API calls per cycle are not expressible as terminals." Nothing enforces it. #5670 showed the cost: the `error_event` self-test reached its terminal through the two-call unavailable fallback, both barrier serves came from one fetch cycle, and fast machines recorded a race outcome as the golden — silently green until CI ran slower.

Under the suite's own rule (invariants exist as checks, prose follows the check), this needs enforcement at the serve boundary: `ApiMock` knows whether the two serves of the terminal event came from one call pair (probe + strict fetch) or from two fetch cycles. When they came from one pair, recording must raise with a named error ("terminal reached through a two-call cycle — add a settling API event") instead of writing a golden.

Same family as #5669 (barrier granularity); second confirmed case.

---

JakobLichterfeld:
Fixes #5682.

## What happened

The harness moduledoc documented a limit without enforcing it: the end-of-replay barrier counted API serves, not fetch cycles. A terminal reached through the probe and strict fetch of a single `fetch` produced both barrier serves from one cycle — the barrier proved nothing, and `error_event` recorded a race outcome as its golden (#5670).

## The fix

- **Static rule:** a scenario cannot end in an error event (`build_events/1`, named raise). An error terminal is never repeat-neutral — the unavailable counter keeps running under repeats — and it is the only shape in which a strict fetch is followed by a probe *within* one cycle (the `vehicle_unavailable` fallback).
- **Dynamic rule:** `ApiMock.exec/2` passes the API action to the harness serve closures (mock-based tests keep their 0-arity events); `await_serve/4` returns the action of each barrier serve, and a probe followed by a strict fetch raises with a named error in **both** record and compare (`two_call_cycle?/2`, unit-tested). With error terminals excluded, that pair is the only single-cycle shape left: snapshots execute once per cycle, a strict fetch that got asleep/offline is followed by the next cycle's probe, a call's strict fetch by the next poll.
- The moduledoc limit moves into the invariants with its checks named; two tmp-scenario probes pin the raises (`[park, error]`, `[park, error, park]` — the pre-#5670 `error_event` shape).

## Verification

Full test suite green: 603 passed (600 + 3 new probes).

None.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@ Under the hood, this release adds a self-verifying black-box characterization su
 
 - fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
 - fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update (#5665 - @JakobLichterfeld)
-- fix(test): move the offline-resume scenario off the 15-minute boundary (#5681 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
@@ -20,6 +19,8 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - test(characterization): convert updating scenarios to characterization fixtures. Pins the update-cancel crash (#5656) that mock-based tests could not see (#5658 - @JakobLichterfeld)
 - test(characterization): convert streaming scenarios to characterization fixtures (#5668 - @JakobLichterfeld)
 - fix(test): give the error_event selftest a settled terminal cycle (#5670 - @JakobLichterfeld)
+- fix(test): move the offline-resume scenario off the 15-minute boundary (#5681 - @JakobLichterfeld)
+- test(characterization): enforce the two-call-terminal limit (#5683 - @JakobLichterfeld)
 
 #### Dashboards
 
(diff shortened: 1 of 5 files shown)


JakobLichterfeld (commented):
Review pass over the enforcement (classification, wiring, probes, prose):

- **The check is placed where the knowledge lives:** ApiMock now hands the API action to the harness's serve closures, so the barrier sees whether the terminal's two serves were the probe and strict fetch of one fetch — the exact shape that let `error_event` record a race outcome (#5670). That pair now raises in both record and compare; the invariant is a property of the scenario, not of the mode.
- **The classification is complete once error terminals are refused statically.** An error terminal is never repeat-neutral (its unavailable counter keeps running under repeats), and with it excluded, probe→strict is the only single-cycle pair; snapshots execute once per cycle, a strict fetch that got asleep/offline is followed by the next cycle's probe, and a call's strict fetch by the next poll — all correctly two cycles.
- **One over-approximation, verified and named rather than refined:** in the reachable path the probe can be the `vehicle_unavailable` fallback ending cycle N with the strict fetch in N+1 — two fetches, but the fallback's result is discarded as incomplete (`vehicle.ex:477`), so the terminal is equally unprocessed at capture. Raising there is a true positive of the same class; a classifier refinement would have removed exactly this catch. The comment and the moduledoc invariant say so verbatim.
- **Probes pin both halves:** the old `[park, error, park]` form raises with the named error (the #5670 regression, now permanent), and an error terminal is refused in `build_events/1`. All existing scenarios pass unchanged — none was reaching its terminal this way after #5670.
- The former Limits entry moved into the invariants with its checks named (`two_call_cycle?/2`, `replay/2`, `build_events/1`). Closes #5682.

---

PR #5687, opened 2026-09-03
test(characterization): convert suspend_logging scenarios to characterization fixtures

JakobLichterfeld:
Part of #5652 (phase 1 characterization suite). Converts all 20 tests of `suspend_logging_test.exs`.

## What this adds

- **Call replies pinned in the golden.** The reply of a user-action call is outer behaviour — what the UI would get back — so the golden gains a `calls` section in delivery order (`{"suspend_logging": "ok"}` or `{"suspend_logging": {"error": "user_present"}}`). A rejection is pinned, no longer a harness error; a dying proxy or a call that never completes still raises. Machinery evidence: a tmp probe records a rejected call instead of raising; the `suspend` selftest golden carries the section.
- **Golden hygiene.** The section exists only for scenarios that declare call events, so every golden without one stays byte-identical (full run green without re-recording anything else). The five existing call users (`streaming_suspend_*`, selftest `suspend`) are re-recorded; their diff is the added `calls` block and nothing else.
- **No-op calls keep their caller.** `:ok` on an already suspended, asleep or offline vehicle is answered without a strict fetch and does not doom the parked poll — only a suspend that went through its strict fetch does.
- **Twenty scenarios**, each mirroring where production evaluates the condition: the guard rejections (`preconditioning`, `user_present`, `downloading_update`, `sentry_mode`, `unlocked` with `req_not_unlocked`, `doors_open`, `trunk_open`) carry the condition in the strict-fetch payload after the call; the state rejections (D/R/N, while driving, updating, charging) answer from that state without a fetch; `incomplete_data` pins `gateway_error` from a bare online payload; the immediate `:ok` cases (asleep, offline, already suspended), the successful suspends (idle, after charging complete) and the two-call `detects_locking` (`unlocked` then `ok`) end asleep so no replay ends suspended. Every `calls` section matches the reply the original test asserted. The original tests remain in place.

## Verification

Full test suite green: 624 passed.

- Non-streaming suspend scenarios set `suspend_min: 10` so the suspended vehicle's poll (minutes in production, milliseconds at test speed) serves the asleep terminal — the same test-speed coupling the moduledoc documents for `:suspended`.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Fable 5.1 low)  — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source).


JakobLichterfeld (commented):
Review pass over the conversion (fixture fidelity against `suspend_logging_test.exs`, the `calls` section, golden hygiene, call mechanics, and the rebase onto #5683):

- **Classifying the call reply as outer behaviour is the right call:** what the UI gets back from `suspend_logging` is exactly the output this suite exists to pin. Treating a rejection as a harness error was an assumption that held for the four streaming users and breaks for this family, where sixteen of twenty tests pin a rejection. A dying proxy or a never-completing call still raises — those are harness errors, correctly kept apart.
- **Golden hygiene verified, not asserted:** the `calls` section is emitted only for scenarios that declare call events (the diff unions both sides' keys, so an always-present section would have broken every existing golden); every golden without call events is byte-identical, and the five existing call users' diffs contain nothing outside the added `calls` block.
- **Fidelity 20/20:** all eleven rejection reasons and their multiplicities match the original's assertions exactly (`unlocked` twice — `unlocked` and `detects_locking`'s first call — `vehicle_not_parked` four times for D/R/N and while driving, the rest once), `already_suspended` pins `[ok, ok]`, `detects_locking` pins `[unlocked, ok]`, `incomplete_data` pins `gateway_error` from a bare online payload. Each scenario places the condition where production evaluates it — guards in the strict-fetch payload after the call, state rejections without a fetch.
- **Call mechanics stay sound:** only a suspend that went through its strict fetch dooms the parked poll; `:ok` on an already suspended, asleep or offline vehicle keeps its caller. No replay ends in `:suspended` (18 end asleep, `offline` ends offline). The two non-streaming scenarios' `suspend_min: 10` is a test-speed coupling and is disclosed under Workarounds.
- **The rebase kept both sides:** the two-call-terminal enforcement from #5683 (action-carrying serves, `two_call_cycle?`, error-terminal refusal) and the `calls` capture coexist; the new terminals pass the two-call check on first contact — 98 characterization tests, 624 full suite.

---

PR #5688, opened 2026-09-03
refactor(vehicle): route the vehicle's view of time through a clock seam

JakobLichterfeld:
Part of #5652. Behaviour-neutral seam extraction, prerequisite for expressing timer-driven suspend scenarios without wall-clock races (see the `suspend_test.exs` finding: 5 of 15 tests decide on `utc_now − last_used` against millisecond-scaled thresholds).

## What this adds

- **`TeslaMate.Vehicles.Vehicle.Clock`** — the vehicle's complete view of time as a behaviour: `utc_now/0`, `diff_seconds/2` (the distance between two instants as the state machine measures it) and `fetch_timeout/2` (a fetch interval as a timer duration).
- **`Clock.Real`** (production): seconds are seconds — byte-identical to the former non-test clauses. **`Clock.Scaled`** (test): seconds pass as milliseconds, intervals collapse to their numeric value — byte-identical to the former `:test ->` clauses. **`Clock.default/0`** holds the environment choice; the vehicle carries no environment branch for time any more (the remaining `Mix.env` branch selects the terrain mock, unrelated to time).
- **`deps_clock`** next to `deps_api`/`deps_log`: every `DateTime.utc_now()` in `vehicle.ex` (45 sites) goes through `data.deps.clock`; `fetch_timeout/3` and `diff_seconds/3` delegate to it. No interval, threshold or ordering changed.
- `log.ex` untouched; the `start_state` `utc_now` fallback stays for its other callers. Fixed `:timer.minutes` intervals (position storing, fuse windows) are not clock properties and stay as they are.

## Behaviour-neutrality evidence

1. All existing vehicle mock tests green without a single test change: `mix test test/teslamate/vehicles` → 125 passed.
2. Full suite including the characterization suite green with byte-identical goldens: 624 passed; `git status` shows only `vehicle.ex` and the new `clock.ex` — no golden changed, nothing re-recorded.

## Workarounds & open questions

None.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Fable 5.1 low) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5689, opened 2026-09-03
refactor(vehicle): date timestamp-less state rows through the clock seam

JakobLichterfeld:
Completes the clock seam from #5688; after this, `lib/` is untouched by the harness-clock work.

## What this adds

`date_opts/1` returned `[]` for payloads without a `drive_state` timestamp (asleep, offline, bare online), which sent those state rows into `Log.start_state`'s own `DateTime.utc_now()` fallback — #5688 misreported those clauses as "no fallback case". `date_opts/2` now dates them with `deps.clock.utc_now()`, so every state row carries the vehicle's view of time; the timestamp clause is unchanged and `log.ex` keeps its fallback for other callers. No interval, threshold or ordering changes.

## Behaviour-neutrality evidence

- Production: `Clock.Real.utc_now/0` is the same instant the fallback produced.
- Characterization suite green with **byte-identical goldens** (the wall-dated rows are volatile-masked in every scenario): full suite 624 passed, `git status` shows no golden.
- Vehicle mock tests green: 125 passed. The **only** test change is the call form of 16 assertions in `vehicle_test`, `driving_test`, `suspend_test`, `charging_test`, `suspend_logging_test`, `streaming_test` — `{:start_state, _, _, []}` → `[date: _]` — because those mocks couple to the call shape, not to behaviour.

None.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
 - fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update (#5665 - @JakobLichterfeld)
 - refactor(vehicle): route the vehicle's view of time through a clock seam (#5688 - @JakobLichterfeld)
+- refactor(vehicle): date timestamp-less state rows through the clock seam (#5689 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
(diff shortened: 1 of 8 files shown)


JakobLichterfeld (commented):
Review pass — verified against the diff:

- **The one write that #5688 left outside the seam is now inside it.** `date_opts/1` had two `[]` clauses (no `drive_state`; `drive_state.timestamp: nil`) that sent every asleep and offline row — and an online row from a bare probe — into Log's own `utc_now` fallback. `date_opts/2` folds both into one clause that dates the row with `deps.clock.utc_now()`; the timestamp clause is unchanged and guarded by `is_integer/1`. All three `start_state` call sites pass `data`; no 1-arity call remains; `log.ex` is untouched and keeps its fallback for its other callers.
- **Behaviour-neutral, proven the hard way:** the real clock yields the same instant the fallback did, no interval or ordering changes, and the characterization suite passes with byte-identical goldens.
- **The test-side change is exactly what it claims:** 16 assertions in six mock-test files widen `{:start_state, _, _, []}` to `[date: _]` and nothing else in the test diff — these tests couple to the call shape, not to behaviour, which is precisely the coupling the characterization suite exists to replace.

---

Issue #5684, opened 2026-09-02
Stale payload timestamps after an offline period can crash the vehicle process (positive_duration)


JakobLichterfeld:
Open question carried since #5654, now with the harness able to express it.

## Mechanism (from the code, not yet observed in production)

When a vehicle goes offline or asleep, the payload carries no `drive_state.timestamp`, so `Log.start_state` dates the new states row with `DateTime.utc_now()` (`lib/teslamate/log.ex`, `date_opts/1` in `vehicle.ex` returns `[]`). When the vehicle comes back, the closing `start_state` uses the *payload's* timestamp as `end_date`. If that timestamp lies before the wall-clock moment the offline row was created, the `states.positive_duration` check constraint rejects the update, `Repo.update` returns `{:error, changeset}`, and the `{:ok, %Log.State{}} =` matches in `vehicle.ex` raise — the vehicle process crashes, the supervisor restarts it, and the next fetch hits the same payload: a crash loop, with the offline row left open.

The characterization suite met exactly this while replaying fixed 2024 timestamps against a 2026 wall clock, and works around it with `timebase: "now"` (an explicitly interim mechanism; the clock is the one dependency the harness does not own — Phase 2 material).

## The production question

Can a real vehicle deliver such a payload — a `drive_state.timestamp` older than the moment TeslaMate recorded it offline? Candidates: cached vehicle data served by the API right after wake-up, vehicle clock skew, a long offline stretch followed by a delayed first sample. The clock anomalies measured in #5558 suggest the answer is not obviously no. If it can, the consequence is not a wrong row but a crash loop on every fetch until the payload changes.

## Proposed pin

A characterization fixture with fixed timestamps (no `timebase`): online → offline → online with a resume timestamp before the offline row's wall-clock date. The crash is pinnable with `expect_restart`; the open design detail is that a crash *loop* never converges, so the scenario needs a terminal payload that lets the restarted vehicle recover (or the harness needs a way to declare "crashes on every fetch"). Decide in the PR, report if not expressible.

If the answer to the production question is yes, the fix belongs in `start_state` (never close a wall-dated row with an older payload date — clamp or skip, logged), pinned first, then fixed in its own commit per the suite's bug-pinning rule.

Related: #5654 (timebase interim), #5669 (clock mix on the resume boundary), #5558 (clock measurements).

---

[...]

PR #5690, opened 2026-09-03
test(characterization): run replays on a scenario clock, retire timebase

JakobLichterfeld:
Part of #5652. Builds on the clock seam (#5688, #5689): the harness now owns the vehicle's time. Phase 2 of the clock work; the `suspend_test.exs` conversion follows as a second commit in this PR (section below to be added).

## What this adds

- **A replay clock instead of wall-clock time.** `TeslaMate.Characterization.Clock` implements `Vehicle.Clock` and is injected through `deps_clock`. `utc_now/0` is scenario time: it strictly increases with every API serve — by at least one millisecond, or to the payload's own time when that lies ahead — advanced at the serve boundary before the vehicle processes the payload. Distance is real (`diff_seconds/2` in seconds: three idle minutes are three minutes of payload time); intervals stay collapsed to milliseconds like the test clock, so polls stay fast. Stream frames advance the clock to their `time`; older payloads never move it back (stale-frame scenarios are backdated on purpose).
- **`{"clock": <epoch ms>}` events** let time pass without a payload — a drive timeout during radio silence, an idle suspend. Delivered at the serve boundary in list order like stream and call events, no serve index, monotonic (checked statically in `build_events/1`, so a misdeclared scenario raises before the replay instead of tearing down the fetch task), never the terminal. First user: `driving_offline_timeout`, whose offline row is now dated deterministically at the clock event.
- **`timebase: "now"` is removed, not deprecated:** `shift_timebase`, the `t0` validation, the timebase×stream rule, their probes and moduledoc entries are gone. The three shifted scenarios keep their fixed timestamps and pin absolute dates for the first time. The selftest pair `timebase` is replaced by `clock`, which reaches an idle suspend purely from payload time.
- **Limits corrected, not just extended:** the hardcoded `{3, 10}` streaming suspend window is payload time now, no longer a wall-clock race — the moduledoc and the three affected scenario descriptions say so.

## Enforced invariants (each names its check)

- The clock strictly increases with every serve (`Clock.serve/1`); a timestamp-less transition never collapses onto the previous state — probed: an `[online, asleep]` scenario yields an online row of positive length and two `since` publishes.
- A clock event before any time the scenario has already reached raises statically (`check_clock_monotonic!/1`); a scenario cannot end in a clock event (`build_events/1`).
- The double-run diff proves the clock value at capture is deterministic.

## Golden hygiene, verified structurally

45 goldens re-recorded (42 content + selftests `await`, `stream`, `suspend`), compared as parsed JSON against `main`: 130 positions change from `"<volatile>"` to a value, **0 other differences** — no list length, no key, no non-masked value. `clock` is the one new golden; every other golden is byte-identical.

## Finding on the way

A first version let the clock stand still between timestamp-less serves. Three goldens then showed a state of zero length (online `00:00:10 → 00:00:10`) and one `since` publish instead of two — a capture behaviour production never has, since time always passes between two fetches; the line-based golden check missed it, the structural one caught it. The per-serve tick is the correction, pinned by the probe above.

## Verification

Full suite green: 626 passed; characterization + harness suites 100 passed; no await turned vacuous under the deterministic clock (the vacuity probe would have refused it).

None. #5684 (stale timestamps after an offline period) becomes expressible with this clock and gets its own pinning PR.

---


JakobLichterfeld (commented):
Review pass over the replay clock (commit 1 of this PR; the `suspend_test` conversion gets its own pass) — verified against the diff and a structural comparison of every re-recorded golden:

- **The clock is the right abstraction at the right place:** `Characterization.Clock` implements the same behaviour the seam introduced (#5688/#5689), so the vehicle is untouched; scenario time replaces wall-clock time everywhere the vehicle reads or dates. Distance in real seconds with intervals collapsed keeps production semantics on scenario time while polls stay fast.
- **The per-serve tick is the decisive detail.** `Clock.serve/1` (`max(now + 1 ms, payload_time)`) runs in the serve closure before the serve message, for every API serve including snapshots, probes and errors — a timestamp-less transition can never collapse onto the previous state. That matches the production property that time always passes between two fetches, and it is pinned by a probe asserting a positive-length online row and two `since` publishes.
- **Golden hygiene holds structurally, not by line count:** parsed against `main`, the 45 re-recorded goldens differ in exactly 130 positions, all `"<volatile>"` → value — no list length, no key, no non-masked value. The three timebase scenarios pin absolute dates for the first time; `driving_offline_timeout` dates its offline row at its clock event deterministically.
- **The finding on the way is the review working as designed:** a first version that let the clock stand still produced zero-length states and a coalesced `since` in three goldens; a line-based diff check passed it, the structural check caught it, the tick is the fix. Worth keeping the structural comparison as the standing golden check.
- **Removal, not deprecation:** `timebase` and everything attached to it is gone; the Limits entry on the `{3, 10}` window and the affected scenario descriptions are corrected to payload-time semantics. Monotonicity is checked statically in `build_events/1` — the right place, after the vacuity lesson that a raise inside the mock's serve surfaces as an opaque barrier timeout.

---

PR #5691, opened 2026-09-03
test(characterization): pin the stale-timestamp resume crash

JakobLichterfeld:
Pins #5684 as current behaviour — not a fix. One fixture, no `lib/` change.

## Mechanism

When the vehicle goes offline, the payload carries no `drive_state.timestamp`, so the offline states row is dated by the vehicle's clock. When it comes back, `start_state` closes that row with the resume payload's timestamp as `end_date`. If that timestamp lies before the row's date, the `states.positive_duration` check rejects the update, `Repo.update` returns `{:error, changeset}`, and the `{:ok, %Log.State{}} =` match in `vehicle.ex` raises — the vehicle process crashes and is restarted by its supervisor.

## The fixture

`stale_timestamp_offline_resume`: `[online(t0), offline, snapshot online(t0 − 5 min), snapshot online(t0 + 20 s), asleep]`, streaming off, `expect_restart: true`, `await: {"state": "asleep"}`.

- Both resume payloads are **snapshot events**: in `:offline` the vehicle probes first (`get_vehicle`), and a plain event would be consumed by that probe without effect — the snapshot serves the probe and the data fetch from one payload.
- The stale snapshot is served exactly once and crashes (`MatchError`, `constraint_name: "positive_duration"`); after the restart the fresh snapshot closes the offline row with `t0 + 20 s`, opens the online row and inserts one position — recovery before the barrier.
- The asleep tail is the convergence gate: reached from `:online` through `:start` one cycle after the barrier, so the await gates a real transition (the vacuity probe stays silent), the barrier pair is strict → probe (no two-call cycle), and the DOWN precedes teardown.

## What the golden pins

- offline row `00:00:00.001 → 00:00:20.000` — closed by the fresh payload, not the stale one
- online row `00:00:20.000 → 00:00:20.002`, asleep row open
- exactly one position (`00:00:20.000`, from the restart); no row from the stale payload

## Open

Whether a real vehicle delivers such a payload — cached data right after wake-up, clock skew, a delayed first sample — is unanswered (#5558 clock measurements). If yes, the consequence in production is a crash loop on every fetch until the payload changes; the fix belongs in `start_state` and follows against this pin in its own commit.

## Verification

Characterization + harness suites: 101 passed. Full suite: 627 passed.

None beyond the production question above.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - test(characterization): enforce the two-call-terminal limit (#5683 - @JakobLichterfeld)
 - test(characterization): convert suspend_logging scenarios to characterization fixtures (#5687 - @JakobLichterfeld)
 - test(characterization): run replays on a scenario clock, retire timebase (#5690 - @JakobLichterfeld)
+- test(characterization): pin the stale-timestamp resume crash (#5691 - @JakobLichterfeld)
 
 #### Dashboards
 
--- a/test/fixtures/characterization/goldens/stale_timestamp_offline_resume.json
+++ b/test/fixtures/characterization/goldens/stale_timestamp_offline_resume.json
@@ -1,0 +1,239 @@
+{
+  "database": {
+    "addresses": [],
+    "car_settings": [
+      {
+        "enabled": true,
+        "free_supercharging": false,
+        "id": "car_setting_1",
+        "lfp_battery": false,
+        "req_not_unlocked": false,
+        "suspend_after_idle_min": 100000,
+        "suspend_min": 100000,
+        "use_streaming_api": false
+      }
+    ],
+    "cars": [
+      {
+        "display_priority": 1,
+        "efficiency": 0.153,
+        "eid": 42,
+        "exterior_color": "DeepBlue",
+        "id": "$car",
+        "inserted_at": "<volatile>",
+        "marketing_name": "LR AWD",
+        "model": "3",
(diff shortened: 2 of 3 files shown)


JakobLichterfeld (commented):
Review pass over the pin — verified against the scenario and the recorded golden:

- **The pin captures the mechanism exactly as #5684 describes it:** the offline row is dated by the vehicle's clock, the stale resume payload (five minutes before the last position) tries to close it with an earlier date, `states.positive_duration` rejects it, the `{:ok, _}` match raises and the process goes down — `expect_restart` proves the crash before teardown.
- **The event design is what makes it recordable without touching the harness.** Both resume payloads are snapshots, because in `:offline` the vehicle probes first and a plain event would be consumed by that probe without effect; the stale snapshot is served exactly once, the fresh one after the restart closes the row. The asleep tail is a genuine convergence gate — reached from `:online` through `:start` one cycle after the barrier — so the await is not vacuous, and the barrier pair is strict→probe, two cycles.
- **The golden says precisely what the pin claims:** offline row `00:00:00.001 → 00:00:20.000`, closed by the fresh payload and not by the stale one; no row dated from the stale payload; an online row of positive length; one position, from the restarted vehicle. The absence of an initial online row is correct — the first online payload served the start probe and was never processed as data.
- **Scope kept honest:** this pins current behaviour only. Whether the API delivers such payloads stays open in #5684; the fix, if it comes, lands in its own commit and will flip this fixture red — the pinning contract working as designed, as with #5656.

---

PR #5692, opened 2026-09-04
fix(vehicle): keep logging when the car reports an outdated timestamp after being offline or asleep — previously the vehicle process crashed on every poll and the state stayed stuck

### LINKED ISSUE #5684: Stale payload timestamps after an offline period can crash the vehicle process (positive_duration)

JakobLichterfeld:
Fixes #5684 (the crash); the measurement question stays open there.

## Mechanism

When the vehicle goes offline or asleep, the states row is dated by the vehicle's clock. A resume payload whose `drive_state.timestamp` lies before that row's start could not close it: `states.positive_duration` (`end_date >= start_date`) rejected the update, the `{:ok, %Log.State{}} =` match raised, the vehicle process crashed and was restarted into the same payload — a crash loop until the payload changed. Pinned by `stale_timestamp_offline_resume` (#5691).

## The fix

`date_opts/2` compares the payload date with exactly the value the constraint compares — the open row's start, carried in a new `state_row_started` field set only where `start_state`/`get_current_state` return a row (`last_state_change` also marks server-dated transient states such as charging or updating, so ordinary car/server clock skew would have flagged the next payload). A stale payload dates the state change with the vehicle's clock and logs a warning with both times: the change is real and observed now, only its timestamp is not trustworthy. No clamp to zero length, no new setting, `log.ex` untouched.

## Mock fidelity and four sharpened assertions

`LogMock.start_state` ignored `date:` and returned `DateTime.utc_now()`; production dates the row with the given date. The mock now mirrors that contract. Four assertions (`charging_test`, `suspend_logging_test`, `updating_test` ×2) asserted only the *ordering* of the `since` topic across a transient state — they held by the mock's invented time. They now assert the exact payload date of the following online row. Observation from the faithful mock, not changed here: a transient state's `since` is server-dated while the following row is payload-dated, so `since` can step backwards by the car/server skew.

## The pin's golden diff — documented behavior change

Before: crash, restart, offline row closed by the *fresh* payload at `00:00:20`, one position from the fresh payload. After: no crash; offline row closed at the clock time of the stale serve (`00:00:00.002`), online row from there to the asleep tail, and the position the stale payload inserts with its own five-minutes-earlier date. Positions are payload-dated and outside this fix (finding for #5684).

## Verification

Vehicle mock tests: 125 passed. Full suite: 627 passed.

- Positions, charges and updates stay payload-dated; a stale resume payload now leaves a backdated position row instead of crashing.
- `since` stepping backwards across transient states (see above).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - fix(web): remove stray brace from the direction arrow SVG path, which made Safari log a parse error on every position update (#5665 - @JakobLichterfeld)
 - refactor(vehicle): route the vehicle's view of time through a clock seam (#5688 - @JakobLichterfeld)
 - refactor(vehicle): date timestamp-less state rows through the clock seam (#5689 - @JakobLichterfeld)
+- fix(vehicle): keep logging when the car reports an outdated timestamp after being offline or asleep — previously the vehicle process crashed on every poll and the state stayed stuck (#5692 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/lib/teslamate/vehicles/vehicle.ex
+++ b/lib/teslamate/vehicles/vehicle.ex
@@ -19,6 +19,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
               last_used: nil,
               last_response: nil,
               last_state_change: nil,
+              state_row_started: nil,
               elevation: nil,
               geofence: nil,
               deps: %{},
@@ -248,6 +249,7 @@ defmodule TeslaMate.Vehicles.Vehicle do
       car: car,
       last_used: deps.clock.utc_now(),
       last_state_change: last_state_change,
+      state_row_started: last_state_change,
       deps: deps,
       import?: Keyword.get(opts, :import?, false)
     }
@@ -1003,8 +1005,13 @@ defmodule TeslaMate.Vehicles.Vehicle do
     :ok = disconnect_stream(data)
 
     {:next_state, {:asleep, asleep_interval()},
-     %{data | last_state_change: last_state_change, stream_pid: nil, pre_online_check: :idle},
-     [broadcast_summary(), schedule_fetch(data)]}
+     %{
+       data
+       | last_state_change: last_state_change,
+         state_row_started: last_state_change,
+         stream_pid: nil,
+         pre_online_check: :idle
+     }, [broadcast_summary(), schedule_fetch(data)]}
   end
 
   def handle_event(:internal, {:update, {:offline, vehicle}}, :start, %Data{} = data) do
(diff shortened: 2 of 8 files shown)


JakobLichterfeld (commented):
Review pass over both commits — verified against the diffs, the mock, and the re-recorded golden:

- **The fix compares exactly what the constraint compares.** `positive_duration` checks `end_date >= start_date` of the open states row, so the stale test uses that row's start (`state_row_started`, set only where `start_state`/`get_current_state` return a row) — not `last_state_change`, which six sub-state transitions overwrite with the current time and which would have flagged ordinary car/server skew after a drive or charge start as stale. A stale payload now dates the state change with the vehicle's clock and logs both times; no crash, no restart, the row closes.
- **The pin proves it:** the fixture from #5691 drops `expect_restart` and re-records green — offline row closed at the clock time of the stale payload (`00:00:00.001 → 00:00:00.002`), not at its timestamp; online row from there; asleep gate intact.
- **Scope kept honest, and visible:** positions and the update row still carry the payload's timestamp (the golden shows the `23:55` position) — only state rows have the constraint; whether payload-dated rows should get the same stale handling goes back to #5684.
- **The mock commit is a truth fix, not a test weakening.** `LogMock.start_state` invented its own date; production Log honours the given one. Four assertions relied on the invented date to claim an ordering (`since` of the online row after a transient state's `since`); they now assert the row's exact payload-derived date — stricter. That the ordering itself is not a production guarantee (transient `since` is server time, row `since` is payload time) is a real finding and gets its own issue.
- Suite 627/627, nothing else touched.

---

PR #5695, opened 2026-09-04
test(characterization): convert suspend scenarios to characterization fixtures

JakobLichterfeld:
## What this adds

16 scenario/golden pairs converting all 15 tests of `suspend_test.exs` (the originals stay). The family pins the idle suspend over the replay clock: a `clock` event moves scenario time past the idle window and the next poll, dated at that time, suspends logging. No `call` events — the user command is #5687.

- `suspend_idle` — base scenario (polling, idle 15 min, `suspend_min: 10`): online → +10 s → clock +16 min → suspend → asleep. It has no original test: it is the presence twin of the guard scenarios, proving the window is open and the guard alone prevents the suspend.
- 8 guard scenarios — the base plus exactly one guard field or field group in the poll that would otherwise suspend (`is_preconditioning`, `climate_keeper_mode: "dog"`, `is_user_present`, `software_update` downloading, `sentry_mode`, `df/dr/pf/pr` with `pf: 1`, `ft/rt` with `ft: 1`, `locked: false` with `req_not_unlocked`). Pin: no `suspended` in the state topic, one position fewer than the base. Each guard is evaluated exactly once with an open window: the guard branches reset `last_used`, so the next poll is inside the window again, and the asleep terminal follows directly.
- `suspend_idle_streaming`, `suspend_idle_while_suspended_polls`, `suspend_idle_unlocked_streaming` — streaming, hardcoded 3-minute window; suspend polls insert one position each (dates pinned), outside_temp publishes pinned in the polls case.
- `suspend_after_charging_complete`, `suspend_after_charging_complete_recharge` — charge end resets `last_used`; the clock event elapses the window after the charge; the recharge starts a second charging process from `:suspended` and closes it with a Complete terminal (2 charging_processes, distinct ids, both closed, own charges).
- `suspend_broadcast_locked`, `suspend_broadcast_sentry_mode` — polling, no clock event, nothing suspends; locked / sentry_mode topic sequences pinned.

## Enforced invariants

- Terminal repeat-neutrality: the double record run masks differing values and refuses a structurally different table (`record/1`, `mask_volatile/2`). This is why the recharge ends in Complete, not Charging — a Charging terminal writes a charges row per serve.
- Await vacuity: an await already satisfied at the barrier is refused (`replay/2`, `{:awaits_vacuous?, _}` probe). The 8 guard scenarios carry `await {"state": "asleep"}`, the base does not: from `:online` the asleep terminal routes through `:start` with another fetch, so the asleep row is written at the barrier serve; from `:suspended` it is immediate and the await would be vacuous. Without the await, `suspend_idle_trunk_open` was not repeat-neutral between the two record runs.
- No replay ends in `:suspended` (`replay/2`): every suspending scenario ends in an asleep terminal.
- Terminals reached through one probe-plus-strict cycle are refused (`two_call_cycle?/2`).
- Clock monotonicity (`check_clock_monotonic!/1`).

## Verification

- `mix test test/teslamate/characterization_test.exs` without record mode: 95 passed, 16/16 suspend pairs.
- Full suite without record mode: 643 passed.
- `git status`: only `test/fixtures/characterization/{scenarios,goldens}/suspend_*` and `CHANGELOG.md`.
- Every suspending golden: state topic contains `suspended` followed by `asleep`. Every guard golden: no `suspended`, positions = 1 (base: 2).
- Guard diff against `suspend_idle` (description omitted):

```
--- suspend_idle_preconditioning
   + "await": {"state": "asleep"}
   +     "is_preconditioning": true,
--- suspend_idle_dog_mode
   + "await": {"state": "asleep"}
   +     "climate_keeper_mode": "dog",
--- suspend_idle_user_present
   + "await": {"state": "asleep"}
   -     "is_user_present": false,
   +     "is_user_present": true,
--- suspend_idle_downloading_update
   + "await": {"state": "asleep"}
   +     "software_update": {"download_perc": 10, "expected_duration_sec": 2700, "status": "downloading", "version": "2026.24.1 def456"},
--- suspend_idle_sentry_mode
   + "await": {"state": "asleep"}
   -     "sentry_mode": false,
   +     "sentry_mode": true,
--- suspend_idle_doors_open
   + "await": {"state": "asleep"}
   +     "df": 0, "dr": 0, "pf": 1, "pr": 0,
--- suspend_idle_trunk_open
   + "await": {"state": "asleep"}
   +     "ft": 1, "rt": 0,
--- suspend_idle_unlocked_required
   + "await": {"state": "asleep"}
   -     "locked": true,
   +     "locked": false,
   +  "req_not_unlocked": true,
```

Open questions: none. The rest are deviations from the originals and one observation.

The originals in `suspend_test.exs` run on wall-clock tricks; the fixtures run on the replay clock. Every trick becomes a deviation on the way over, which is why this family has more of them than driving or charging:

- The originals test the guards with an idle window of 0 minutes so the test does not wait 15 minutes. On the replay clock a real window costs nothing, so the conversion uses one (15 min polling, 3 min streaming).
- `suspend_idle_while_suspended_polls` has two polls before the suspend where the original had six: those six were padding for the 180 ms wall-clock window; a `clock` event replaces them. The pinned statement — broadcasts and positions while `:suspended` — is unchanged.
- The originals may end in `:suspended` or `:charging` because the test simply stops after 50 ms. The harness needs a terminal that repeats without writing, hence the asleep and Complete terminals.
- `suspend_min: 10` in the polling scenarios is the one trick the fixtures keep: the suspend poll is scheduled in `suspend_min` "minutes", which the test clock still collapses to milliseconds; 100000 would stall the replay. Same coupling as in #5687.
- Observation, not a diagnosis: with a Charging terminal the record run failed as a `DBConnection.OwnershipError` out of `dump_db/1` (preceded by "client exited" on the sandbox connection), not as a named refusal.

---


JakobLichterfeld (commented):
Verified at the tree, not from the report: 32 fixture files plus the changelog line, nothing else; the eight guard scenarios compared structurally against `suspend_idle` differ only in `description`, `await` and the guard field(s); every suspending golden runs `online → suspended → asleep`, every guard golden `online → asleep` with one position fewer than the base; the recharge golden holds two closed charging processes with distinct ids and their own charges; the `locked`, `sentry_mode` and `outside_temp` courses and the suspended-poll position dates match the originals.

The Complete terminal for the recharge is the right resolution: a terminal must be effect-neutral under repetition, and a Charging terminal writes a charges row per serve — the invariant excludes it, not the harness. The `await` on the guard scenarios follows from the state path (`:online → :start → asleep` writes the row at the barrier serve, `:suspended → asleep` does not), so it is a consequence of the guard, not a second variable.

--- 

PR #5696, opened 2026-09-04
test(characterization): convert summary scenarios to characterization fixtures

JakobLichterfeld:
## What this adds

9 scenario/golden pairs converting all 12 tests of `summary_test.exs` (the originals stay). The family pins how the software-update fields of `Summary` leave the system as MQTT topics. Every scenario is polling with a closed idle window: a parked snapshot without a `software_update` key, then one terminal poll carrying the `software_update` variant under test. The terminal stays online and writes nothing under repetition.

| Fixture | Original tests | Pin |
|---|---|---|
| `summary_update_available` | true when status is 'available'; strips the build hash | `update_available` true, `update_version` 2026.24.1, `software_update` JSON |
| `summary_update_downloading` | true when status is 'downloading' | `update_available` true |
| `summary_update_downloading_wifi_wait` | true when status is 'downloading_wifi_wait' | `update_available` true |
| `summary_update_scheduled` | true when status is 'scheduled' | `update_available` true |
| `summary_update_installing` | true when status is 'installing'; maps install_perc while installing | `update_available` true, `download_perc` 100, `install_perc` 42, state online → updating, second `updates` row open |
| `summary_update_empty_status` | false when status is empty string | `update_available` false, no `update_version` topic |
| `summary_no_software_update` | nil when software_update is nil (update_available); nil when software_update is nil (perc) | no `update_available`, `update_version`, `download_perc`, `install_perc` or `software_update` topic |
| `summary_download_perc` | maps download_perc and install_perc | `download_perc` 100, no `install_perc` topic |
| `summary_update_version_nil` | nil when no version | `update_available` true, no `update_version` and no `software_update` topic |

Three fixtures each carry two original tests; a duplicate fixture would pin nothing additional.

## Enforced invariants

- Absence is pinned: nil summary fields are never published (`VehicleSubscriber.publish_values/2`), and a topic key missing on either side is a divergence in the golden comparison (`Characterization.diff/2`).
- Terminal repeat-neutrality: the double record run masks differing values and refuses structurally different tables (`record/1`, `mask_volatile/2`); every terminal here stays online (or `:updating` without writes) with a 100000-minute idle window.
- Terminals reached through one probe-plus-strict cycle are refused (`two_call_cycle?/2`); both barrier serves are strict fetches from online / updating.
- No replay ends in `:suspended` (`replay/2`); no await is declared, because every publish and the `start_update` write happen in the terminal's first serve, which the barrier proves complete.

## Verification

- `mix test test/teslamate/characterization_test.exs` without record mode: 104 passed, 9/9 summary pairs.
- Full suite without record mode: 652 passed.
- `git status`: only `test/fixtures/characterization/{scenarios,goldens}/summary_*` and `CHANGELOG.md`.
- Structural golden check: table above; each golden holds one closed baseline `updates` row from the version bookkeeping at start (as in every existing golden), `summary_update_installing` adds the open second row.

- The two "nil when software_update is nil" tests are pinned through a payload without a `software_update` key. The production decoder (`TeslaApi.Vehicle.State`) always builds a `SoftwareUpdate` struct with nil fields, so the struct-nil precondition of the originals does not occur at the API boundary; the observable (no update topics) is the same.
- `summary_update_empty_status` unavoidably shows publisher behaviour beyond `Summary.into/2`: with `update_available` false the publisher takes the installed version as latest, so the `software_update` JSON reads installed = latest = 2026.20.1.
- `summary_update_installing` moves the machine to `:updating` on the terminal's first serve; the golden therefore also pins the state transition and the open update row, which the unit test did not cover.
- Open questions: none.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - test(characterization): run replays on a scenario clock, retire timebase (#5690 - @JakobLichterfeld)
 - test(characterization): pin the stale-timestamp resume crash (#5691 - @JakobLichterfeld)
 - test(characterization): convert suspend scenarios to characterization fixtures (#5695 - @JakobLichterfeld)
+- test(characterization): convert summary scenarios to characterization fixtures (#5696 - @JakobLichterfeld)
 
 #### Dashboards
 
--- a/test/fixtures/characterization/goldens/summary_download_perc.json
+++ b/test/fixtures/characterization/goldens/summary_download_perc.json
@@ -1,0 +1,234 @@
+{
+  "database": {
+    "addresses": [],
+    "car_settings": [
+      {
+        "enabled": true,
+        "free_supercharging": false,
+        "id": "car_setting_1",
+        "lfp_battery": false,
+        "req_not_unlocked": false,
+        "suspend_after_idle_min": 100000,
+        "suspend_min": 100000,
+        "use_streaming_api": false
+      }
+    ],
+    "cars": [
+      {
+        "display_priority": 1,
+        "efficiency": 0.153,
+        "eid": 42,
+        "exterior_color": "DeepBlue",
+        "id": "$car",
+        "inserted_at": "<volatile>",
+        "marketing_name": "LR AWD",
+        "model": "3",
+        "name": "blue",
+        "settings_id": "car_setting_1",
+        "spoiler_type": "None",
+        "trim_badging": "74D",
+        "updated_at": "<volatile>",
+        "vid": 1000,
+        "vin": "5YJ3E1EA1KF000001",
+        "wheel_type": "Pinwheel18"
+      }
+    ],
+    "charges": [],
+    "charging_processes": [],
+    "drives": [],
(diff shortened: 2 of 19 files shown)


JakobLichterfeld (commented):
Verified at the tree against the twelve originals: every test maps to a fixture and every description names its originals; the nine scenarios differ from `summary_update_available` only in the terminal poll's `software_update` fields; the goldens carry the derived values as topics — `update_available` true for the five active statuses, false for the empty status without `update_version`, no update topic at all without a `software_update` key, `download_perc` 100 without `install_perc`, `install_perc` 42 with the `updating` transition and the open update row, the hash-stripped `2026.24.1`, and no `update_version` or `software_update` topic without a version. The baseline `updates` row at start is the version bookkeeping every golden in the suite has. The cited harness and publisher functions exist as named.

The nil-precondition deviation is handled correctly: the decoder never yields a nil struct, so a poll without the key is the reachable form of that input, and the pinned observable is identical. The `empty_status` note is right to call the installed-equals-latest JSON publisher behavior — it is in the golden either way, and naming it is what keeps the pin honest.

Approving.

PR #5697, opened 2026-09-04
test(characterization): convert vehicle scenarios to characterization fixtures

JakobLichterfeld:
## What this adds

Converts `vehicle_test.exs` and `vehicle_sync_test.exs` (24 tests, originals stay): 14 scenario/golden pairs, plus one harness ingredient the family needed — the `update_car_settings` call.

**Harness (first commit).** The `call` whitelist grows by `{"call": {"update_car_settings": {<attrs>}}}`. Delivery goes through the same proxy task as `suspend_logging`: the production `Settings.update_car_settings/2` validates and writes the `car_settings` row, the persisted struct is sent to the vehicle exactly as its PubSub subscription would deliver it (the `SettingsMock` replaces that subscription), and `:sys.get_state` proves the vehicle applied it before the next API event is served. The reply is pinned in `calls` — `"ok"` or the changeset errors. Statically refused with named errors: `enabled` (toggling it calls `Vehicles.restart/0`, outside the replay), an unknown field, an empty map; a settings call cannot be the terminal like any call. Self-test pair `selftest/settings.json`; three probes in `characterization_test.exs`.

**Fixtures (second commit).**

| Fixture | Original test | Pin |
|---|---|---|
| `vehicle_start_online` / `_offline` / `_asleep` | starting: handles online / offline / asleep state | one states row of that state, state topic |
| `vehicle_unknown_state_errors` | starting: handles unknown and faulty states | no states row, state topic `unavailable`, `healthy` `"", true, false` |
| `vehicle_health_api_errors` | error handling: reports the health status | `healthy` `"", true, false`, state stays online |
| `vehicle_unavailable_fallback_asleep` | falls back to vehicle state when vehicle data is unavailable while streaming | states online → asleep, healthy stays true |
| `vehicle_timeout_errors` | handles timeout errors | `healthy` never false, one states row |
| `vehicle_in_service_online` | notices if vehicle is in service | one states row, healthy never false |
| `vehicle_in_service_ends_drive` | ends a drive if vehicle is in service | one closed drive (1.37 km, 2 positions), state topic `online, driving, start, asleep` |
| `vehicle_settings_change` | settings change: applies new sleep settings | `car_settings` 1/10/false, `calls` ok, state topic `online, suspended, asleep`, suspend position |
| `vehicle_version_first_startup` | updates: logs the current software version at first startup | one updates row `42.42.42.0 b2ab650` dated at payload time |
| `vehicle_version_missed_update` | updates: logs missing updates | updates rows `2020.12.5 …`, `2020.12.10 …`; version topic both |
| `vehicle_version_not_newer` | updates: does not log updates <= current version | exactly one updates row across three older versions; 8 states rows |
| `vehicle_version_nil` | updates: handles unexpected :car_version's | no updates row, no version topic, vehicle still online |

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`): every terminal is an online, offline, asleep or unknown poll that writes nothing on repeat.
- Await vacuity (`replay/2`): the only await is `vehicle_unavailable_fallback_asleep` — from `:online` the asleep terminal routes through `:start`, so the asleep row is written at the barrier's second serve and would race teardown without it.
- No replay ends in `:suspended` (`replay/2`): `vehicle_settings_change` ends in asleep.
- Call plumbing: unsupported calls raise (`call_event!/1`), invalid settings attrs raise statically (`settings_call!/1`), every reply lands in `calls` (`ApiMock.calls/1`, `canonical_call/1`), a dying or hanging proxy raises (`ApiMock.await_call_outcome/4`).
- Two-call terminals refused (`two_call_cycle?/2`): online terminals follow a snapshot or an online state, asleep/offline/unknown terminals repeat through the probe.

## Verification

- `mix test test/teslamate/characterization_test.exs test/teslamate/characterization_harness_test.exs` without record mode: 144 passed (14/14 pairs, selftest `settings`, 3 probes).
- Full suite without record mode: 670 passed.
- `mix format --check-formatted` on the three harness files: clean.
- `git status`: harness files, selftest pair, `scenarios/goldens/vehicle_*` (14 + 14), `CHANGELOG.md`.

- **Fuse melts.** `healthy false` needs four generic errors: the `api_error` fuse tolerates three melts and is fresh per replay (the car id is random per test). The originals see it after a single declared error because the mock-based `ApiMock` repeats the last event, so the one error melts the fuse four times. `error: unknown` stands in for the original's `TeslaApi.Error{reason: :boom}`; both take the generic error branch.
- **In-service drive.** The two D positions carry different odometer values so the drive has a distance; the original drives without moving, which the real `Log.close_drive` would delete (< 0.01 km).
- **`suspend_min: 10`** in `vehicle_settings_change` instead of the original 10000: test-speed coupling of the suspend poll interval, as in #5695.
- **Asleep polls in polling scenarios.** From `:online` an asleep poll only bounces the machine to `:start`; the asleep row is written by the following probe. `vehicle_version_not_newer` and `_missed_update` therefore keep the original's two asleep polls per round.
- **Not converted, not visible at the boundary (6):** `does not fetch again after an import completes` (import mode, second-fetch absence only visible at the mock); `resume_logging` ×3 (no effect / polling frequency is timing); `summary/1` ×2 (a call result, neither row nor topic).
- **Not converted, blocked (4):** `stops polling if signed out` — the terminal is never served twice, the barrier starves; `broadcasts the summary when the health check succeeds again` — fuse recovery is a real-time timer; `restarts if the eid changed` — the kill belongs to the tagged-events item in #5652; `vehicle_sync: restores and publishes last known values` — the restore path runs only for a fresh process with pre-existing positions.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree against all 24 originals. Each of the 14 fixtures traces to its test and carries the statement in rows or topics: `healthy` runs `"", true, false` after four melting errors and never reaches false under timeout, closed or in-service; the unknown state leaves no states row and publishes `unavailable`; the in-service drive closes with 1.37 km and two positions; the version fixtures hold exactly the updates rows the originals asserted through `insert_missed_update`, with `not_newer` back at eight states rows after restoring the original's two asleep polls per round; the settings change lands as `car_settings` 1/10/false, `calls` ok and `online, suspended, asleep`.

The harness commit is the whitelist growth the moduledoc reserved for exactly this case — a user action a conversion first needs — not a redesign: same proxy path as `suspend_logging`, production `Settings.update_car_settings/2`, reply pinned in `calls`, and `enabled` refused with the reason that matters (`on_enabled_change` calls `Vehicles.restart/0`). The one doc drift I found (`await_call_outcome` arity) is fixed in that commit.

The ten unconverted tests are each named with the reason, and the reasons hold at the code: six assert on call order, timers or return values that no row or topic carries; four are blocked by a terminal that never repeats, a real-time fuse timer, a supervisor kill or a restore path that needs a fresh process. The kill belongs to the tagged-events item, correctly not a new one.

---

PR #5698, opened 2026-09-04
test(characterization): convert the remaining vehicle scenarios to characterization fixtures

JakobLichterfeld:
## What this adds

Converts the ten tests of `vehicle_test.exs` / `vehicle_sync_test.exs` left open by #5697 (originals stay): nine scenario/golden pairs, each preceded by the harness concept it needed — one commit per concept with self-test pair, probes and moduledoc.

**Harness concepts (commits 1–4).**
- `call` whitelist grows by `"resume_logging"` (reply `"ok"`) and `"summary"` (the `Summary` the UI gets, pinned canonically: keys sorted, `car` → `"$car"`, `since` masked, values as Jason encodes them). From asleep/offline the resume handler queues an immediate fetch that production drops while a fetch task runs (`Vehicle.handle_event/4` matches `task: nil`); at the serve boundary a fetch is always parked, so the declared payload is served by that parked poll.
- `expect_halt: true` allows an error terminal, served exactly once. The replay end is a deterministic read of the state machine: one `:sys.get_status/1` snapshot is polled until its `"State"` entry shows the fetch task cleared, and the same snapshot's `"Time-outs"` entry must hold no `{:state_timeout, :fetch}` (other timers such as the position-storing timeout may remain); a fetch timer or a second serve fails with the full timeout list (`assert_halt!/3`). Excludes `expect_restart` and `await`; requires an error terminal. `import: true` starts the vehicle in import mode.
- `seed.positions` inserts rows before the vehicle starts through `Log.insert_position/2`; keys checked against the position schema, empty list and rejected rows raise (`seed_positions!/1`).
- `TeslaMate.Characterization.Vehicles` replaces `VehiclesMock` as `deps_vehicles`: production `Vehicles.kill/0` terminates the vehicles supervisor and every vehicle; the replay has one, the stand-in terminates that process. `VehiclesMock` stays untouched for the mock tests.

**Fixtures (commit 5).**

| Fixture | Original test | Pin |
|---|---|---|
| `vehicle_import_complete` | does not fetch again after an import completes | halt check, no rows |
| `vehicle_resume_online` | resume_logging: does nothing if already online | `calls` ok, one states row |
| `vehicle_resume_asleep` / `_offline` | resume_logging: increases polling frequency if asleep / offline | `calls` ok, states asleep/offline → online |
| `vehicle_summary_before_first_response` | summary: returns the summary if no api request was completed yet | `calls` summary `state: unavailable`, fields nil |
| `vehicle_summary_while_fetching` | summary: returns the summary even if the api call is blocked | `calls` summary `state: online`, battery 80, version 2026.20.1 |
| `vehicle_eid_changed_restart` | restarts if the eid changed | `healthy` `"", true, false, true`, DOWN asserted, 2 positions, states online → asleep |
| `vehicle_signed_out_halt` | stops polling if signed out | halt check, state topic `online, start`, `healthy` false |
| `vehicle_restore_last_known_values` | vehicle_sync: restores and publishes last known values | seeded row only; topics latitude 37.889544, battery 64, odometer 91887.73, model S, trim P100D, location JSON |

## Enforced invariants

- Halt names its check: terminal served once, fetch task cleared, no fetch timeout in the same status snapshot (`assert_halt!/3`); `expect_halt` with `expect_restart`, `await` or an API terminal raises (`expect_halt?/1`, `build_events/1`); an error terminal without it stays refused.
- Crash names its check: `expect_restart` asserts the DOWN before teardown; the convergence await is checked for vacuity at the barrier (`replay/2`) — `vehicle_eid_changed_restart` converges on the asleep row written after the barrier.
- Call replies are captured, rejections and values alike (`ApiMock.calls/1`, `canonical_call/1`); unsupported calls raise (`call_event!/1`).
- Seed misuse raises with names (`seed_positions!/1`, `create_seed/2`).
- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`).

## Verification

- `mix test test/teslamate/characterization_test.exs test/teslamate/characterization_harness_test.exs` without record mode: 163 passed (9/9 pairs, self-tests `calls`, `halt`, `seed`, `kill`, 6 new probes).
- Full suite without record mode: 689 passed.
- `mix format --check-formatted` on the harness files: clean.
- Existing fixtures and goldens: unchanged.

- **Resume fetch dropped.** The resume-triggered fetch never reaches the mock (production drops it while a fetch task runs); the pin is the ok reply and the state transition. The raised polling rate is timing and not visible.
- **Kill semantics.** Production kills the vehicles supervisor (`Vehicles.kill/0`); the stand-in kills the single replay vehicle. Nine `vehicle_not_found` errors are needed (fuse tolerance eight, fresh fuses, ApiMock serves each queued error once); the original sees the kill after one because the mock-based ApiMock repeats the last event, so the one error melts the fuse nine times.
- **Halt is "no fetch timer".** After an online phase the state machine keeps its position-storing timeout; the halt check ignores it and only refuses `{:state_timeout, :fetch}`.
- **`Code.ensure_loaded!(Vehicle)`** before decoding error reasons: `String.to_existing_atom("not_signed_in")` failed when the vehicle module had not been loaded yet; earlier error atoms happened to come from loaded modules.
- **Not converted (1):** `broadcasts the summary when the health check succeeds again` stays a mock test — a fixture may not carry an internal fuse name (#5652 fixture rule), and fuse recovery is a real-time timer.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree against the nine originals and the harness code. The four harness additions are additive and each names its check: the whitelist grows by two user actions with the reply pinned in `calls` (the summary canonical, `car` masked, `since` volatile); `expect_halt` replaces the barrier with a single `sys.get_status` snapshot in which the fetch task has cleared and no `{:state_timeout, :fetch}` is pending — one snapshot, so a timer firing between two reads cannot produce a false green, and the position-storing timeout that survives the halt is correctly not a poll; `seed.positions` goes through `Log.insert_position/2` like geofences through `Locations`; the Vehicles stand-in terminates the one replay vehicle where production terminates the supervisor, with the difference stated in the moduledoc.

The goldens carry the statements: halt scenarios end with no fetch timer and the `healthy` and `state` topics the originals asserted; the resume scenarios pin the ok reply and the asleep/offline → online row, with the dropped resume fetch named as production behavior (`task: nil` match); the two summary scenarios pin the value the UI gets before the first response and while a fetch is parked; the eid change pins nine melts, the DOWN and the recovery to `healthy true`; the restore scenario publishes the seeded position's values. Existing fixtures and goldens are untouched.

The one test not converted, fuse recovery, is the right call: a fixture may not name an internal fuse, and the behavior under test is the library's real-time timer.


---

Issue #5699, opened 2026-09-04
Charge samples without charger_power are stored as 0 kW, hiding "unknown" behind a real value


JakobLichterfeld:
Found while pinning the current behaviour (`charging_without_charger_power`, #5700).

## What the code does

A charge sample whose payload lacks `charger_power` is stored with `charger_power` 0. The default is applied twice on the way in (`insert_charge/3`, `put_charge_defaults/1` in `vehicle.ex`), and the charge changeset's `validate_required` for `charger_power` can therefore never fail from the vehicle path.

## Why it matters

"Unknown" and "0 kW" become the same row. In Grafana a charging session with missing power samples looks like a session that drew no power, and the changeset asserts an invariant the caller bypasses.

## Direction (Phase 2, behaviour change)

One normalisation point, and a representation that keeps "unknown" apart from a measurement. The fixture holds the current behaviour; its golden changes visibly when this does.

Part of the Phase 2 backlog in #5558.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5700, opened 2026-09-04
test(characterization): pin charge samples without charger_power

JakobLichterfeld:
## What this adds

One scenario/golden pair, `charging_without_charger_power`, pinning [#5699](https://github.com/teslamate-org/teslamate/issues/5699): a charge sample whose payload has no `charger_power` key is stored as 0 kW. The golden changes when that is fixed.

Streaming, after the `charging_full_cycle` template: Starting with 11 kW, a Charging sample without the `charger_power` key, Charging with 11 kW again, Complete, Unplugged, Disconnected, asleep terminal. Only the key is missing in the second sample; current, phases and voltage stay as in its neighbours, so the difference is one field.

Pins both views of the same sample:
- Row view: `charges` holds `charger_power` `11, 0, 11, 0` — the 0 of the sample with energy 0.20 comes from the missing key, framed by two samples with 11; the 0 of the Complete sample is the payload's own value.
- Topic view: `charger_power` publishes `"", 11, "", 11, 0, ""` — the missing key is an empty publish on the topic while the row stores 0.

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`): the asleep terminal repeats through the probe without writing.
- Await vacuity (`replay/2`): `{"state": "asleep"}` guards the asleep row, which from `:online` is written through `:start` after the barrier's second serve.
- No replay ends in `:suspended` (`replay/2`).

## Verification

- `mix test test/teslamate/characterization_test.exs` without record mode: 142 passed.
- Full suite without record mode: 690 passed.
- Structural golden check: four `charges` rows `charger_power` 11, 0, 11, 0 with energies 0.10, 0.20, 0.30, 0.40; one closed `charging_processes` row; topic `charger_power` `"", 11, "", 11, 0, ""`; state topic `online, charging, online, asleep`.
- `git status`: scenario, golden, `CHANGELOG.md`.

- **#5699 pinned, not fixed.** The 0 kW row is the current behaviour; the fix for #5699 re-records this golden.
- **Unreachable `validate_required`.** The charge changeset requires `charger_power` (`Log.Charge`), but the vehicle never lets `nil` reach it: `insert_charge/3` substitutes `|| 0` (`vehicle.ex:1735`) and `put_charge_defaults/1` maps `nil` to 0 (`vehicle.ex:1992`). From the vehicle the requirement cannot fire; both substitutions are the mechanism behind #5699.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree: the second Charging sample lacks only the `charger_power` key, its neighbors carry 11 kW, and the golden holds the row with 0 between them while the Complete sample's 0 is the payload's own — the two zeros are distinguishable by position, and the topic view separates them further (empty publish for the missing key, `0` for the value). One closed charging process, `online, charging, online, asleep`, await on the asleep row written after the barrier, as in the `charging_full_cycle` template. Both substitution sites named in the body exist as cited (`insert_charge/3`, `put_charge_defaults/1`), and the changeset's `validate_required` for `charger_power` is indeed unreachable from the vehicle.

This is a pin in the #5656 sense: #5699 is the finding, the golden documents today's behavior, and the fix re-records it visibly.

---

PR #5703, opened 2026-09-04
feat(rust): add the Rust core skeleton under rust/ — crate, CI with path routing, Nix package and devenv toolchain

### LINKED ISSUE #5651: Add Rust core skeleton: crate, nix build, CI, test infrastructure
Follow-up to #5558: @brianmay proposed a minimal, non-controversial starting point for the Rust core, and the repo question is decided there — same repository, fresh top-level directory, no separate project.

## Scope

- [ ] Top-level directory with a new Rust crate; name boring and descriptive per the naming subthread in #5558 (it is "the Rust core", not a codename — final crate/directory name to be settled in the PR)
- [ ] License: AGPL-3.0, matching the repository
- [ ] nix build wired into the existing flake (no NixOS module)
- [ ] "Hello World" binary — compiles, runs, does nothing else
- [ ] CI job: build, fmt, clippy, test — path-filtered so Elixir-only PRs don't trigger Rust jobs and vice versa
- [ ] Unit-test skeleton with one trivial passing test

## Non-goals

No event contract code, no database access, no port of any module. Phase 3 is gated by Phases 1 and 2 (see the sequencing in #5558) — this issue only puts the build and test plumbing in place so that later work has somewhere to land.

## Acceptance criteria

- Rust crate builds and tests green in CI
- Docker images and release artifacts are byte-for-byte unaffected — the crate stays out of the release path until it does something
- Existing Elixir CI behaviour unchanged; Elixir-only PRs don't pay for Rust jobs

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

brianmay:
Keeping it simple, stupid, and boring for now.

Not using any external flakes for building rust. If any think changes I think this means all dependencies need to be rebuilt too. But simple.

I have used crane on my personal projects (except the Dioxus based ones), not sure if it is worth it. And this might depend on future decisions. Crane means that we can cache the build of the crate dependencies. Needs to be completely rebuilt if any dependency changes. But if we ever decide to Dioxus as the front end, that gets complicated very quickly with crane - I have done it, not sure it is worth it.

https://crane.dev/

There are alternative builders for rust in nix, e.g. ones that will cache the build of every crate and not require rebuilding any crates even if dependencies change. These look impressive initially, but I tend to find they are more trouble then they are worth.

Using rust version from nixpkgs. In future it is likely that we want to use the latest rust release. IIRC there are at least two competing flakes that allow access to different rust versions.

If it looks like I got something wrong, please let me know.

closes #5651


brianmay:
For some reason the AI really wanted to prove it can add 1+2, so I let it.

brianmay:
Currently CI uses dtolnay/rust-toolchain, and latest stable version of rust.

This does mean that the rust version for CI can unexpectedly be upgraded. In practice this can mean clippy checks sometimes fail unexpectedly, because new rust version has extra checks.

Normally these are are easy to fix.

Using the versions from the nix flake also possible, but then of course we increase the nix cache requirements.

brianmay:
Hopefully I have addressed all the criteria now. Will revise later in case I missed anything.

Everything here was done by hand, except nix and CI changes which was assisted by opencode/Big Pickle.

JakobLichterfeld (commented):
Thanks, Brian, for bringing your experience with the rust nix ecosystem to this, and in particular for leaving out the pretty but hard to maintain third-party pieces (crane, rust overlays) and stating why. The nix wiring, the path-filtered CI and the strict clippy set are what I want to keep.

While reviewing I ran into a structural problem that is not yours: the repository root is also the Mix root. With one build system at the root that was invisible, with a second one it leaks into this PR. The Elixir path filter needs a nine-glob allow-list because the Elixir component has no directory, and that allow-list skips Elixir CI for PRs that only touch `grafana/dashboards`, `VERSION` or `coveralls.json`, which are compile-time and test inputs. And the directory has to carry a `teslamate-` prefix to stand next to a component that has no directory at all.

The criteria I want the layout to follow:

1. A component's root is its build tool's root: the level of `mix.exs` for Mix, of `Cargo.toml` for Cargo.
2. Two deployable implementations of the same product sit at the same level with the same shape. Your `teslamate-rust/` already does that; the Elixir side does not, because it has no directory and occupies the root instead. Giving Elixir its own directory next to Rust removes the asymmetry, and nothing has to move later.
3. The repository root holds only cross-cutting things: LICENSE, README, CI, the nix flake, formatter config, `VERSION`.
4. Names use the only stable distinguishing attribute we know today, the implementation language. A role name like `core` is not derivable before the event contract (Phases 1 and 2) exists.

That gives:

```
elixir/     Mix root (mix.exs, lib/, test/, config/, priv/, assets/, Dockerfile, ...)
rust/       Cargo root (Cargo.toml with name = "teslamate-rust", src/)
grafana/  website/  nix/  flake.nix  .github/  LICENSE  VERSION
```

Path filtering then needs no allow-list: Elixir CI runs unless every changed file is under `rust/**`, Rust CI runs when `rust/**` is touched. The crate keeps `teslamate-rust` as its artifact name until the Elixir release goes away, then it becomes plain `teslamate`.

Moving the Elixir tree is a wide mechanical change (Docker context, nix `src`, CI actions, reloader paths), so I would like to do it myself as a pure move PR with no behaviour change, and then rebase your skeleton onto it in `rust/`, keeping your commits. I will fold in the smaller review items on the way: the `teslmate-rust` typo in `checks.nix`, the missing trailing newline in `rust_ci.yml` (prettier will flag it in `ensure_linting`), the LICENSE copy replaced by `license = "AGPL-3.0-only"` and `publish = false` in `Cargo.toml`, `rustfmt` via treefmt and devenv instead of a bespoke `cargo fmt` step, a `nix build` step so the flake package is exercised in CI, one pinned toolchain instead of floating `stable` next to the nixpkgs pin so that `nursery` + `-D warnings` does not go red on every Rust release, and dropping the redundant `cargo build` step.

Is that fine with you? Once the move PR is merged I will rebase this branch in place, keep this PR open and ask you to review the restructured version here.

---

brianmay:
Yes, I had similar concerns.

Moving elixir stuff to the elixir directory seems scary, but I agree it might be worth it.

JakobLichterfeld:
> Yes, I had similar concerns.
> 
> Moving elixir stuff to the elixir directory seems scary, but I agree it might be worth it.

We love the scary stuff :-)

I will move the elixir tree once #5259 and #5649 landed.

brianmay:
Good plan. I suspect this is going to break any PR currently pending.

JakobLichterfeld:
> I will move the elixir tree once #5259 and #5649 landed.

With the confidence of the black-box test with 93.9% coverage #5259 landed, #5649 was dropped. The move was succesfull: #5745 


> I suspect this is going to break any PR currently pending.

I didn't squash merge the move and splitted it into two commits. The first commit is a pure rename, no content change, so later rebases resolve through Git's rename detection. Tooling and docs was than changed in a second commit. This should enable all to just rebase on latest main.

--> I will rebase this now.

JakobLichterfeld:
> This should enable all to just rebase on latest main.

Rebase was smooth as expected. I will now make the discussed changes.

JakobLichterfeld:
Restructured on top of your four commits, one commit per review item, so each is reviewable on its own:

- `refactor(rust): move the crate to rust/` — top-level `rust/` next to `elixir/`, crate name stays `teslamate-rust`.
- `ci: point the Rust crate and the two-core path filters at rust/` — `check_paths` now has `githubfolder`, `rust` and `elixir` filters (`some-with-excludes`), the Elixir jobs skip Rust-only changes and vice versa.
- `build(rust): declare the licence in Cargo.toml instead of a LICENSE copy`
- `fix(nix): spell the teslamate-rust check name`
- `docs: update changelog`
- `build: format Rust through treefmt` — rustfmt joins the single `nix run .#lint` call, no separate fmt step.
- `build(nix): run clippy as a flake check` — `.#teslamate-rust-clippy` with `-D warnings`, so the same toolchain checks locally and in CI.
- `ci: extract the Nix install and cache steps into a composite action` — `.github/actions/setup-nix`, one cache key per Nix closure family (`lint`, `rust`).
- `ci: run the Rust test and static analysis through Nix` — `rust_test` builds `.#teslamate-rust` (cargo test in the check phase), `rust_static_analysis` builds `.#teslamate-rust-clippy`; both gated on `ensure_linting` like the Elixir jobs. The toolchain is the nixpkgs pin, so `rust_ci.yml` with rustup and the cargo cache action is gone.
- `fix(nix): run the Rust checks on every system` — the Rust checks were inside the Linux-only branch meant for the NixOS VM test; `nix flake check` on macOS now runs them.

Verified with two `workflow_dispatch` runs: [first](https://github.com/teslamate-org/teslamate/actions/runs/35451113264) cold (Rust closure built and cached once, the second Rust job's duplicate save correctly refused), [second](https://github.com/teslamate-org/teslamate/actions/runs/35454415817) with all three Nix jobs restored from the primary key and no substitution from cache.nixos.org. [GHCR build](https://github.com/teslamate-org/teslamate/actions/runs/35451130442) green on both architectures.

The cache key now includes `rust/Cargo.lock`, so the first run on `main` after merge will be a one-time miss.

@brianmay could you have another look? Squash on merge is fine.

JakobLichterfeld (approved):

brianmay:
I noticed we forgot to add support for dependabot. Hopefully that is done now :-)

JakobLichterfeld:
> I noticed we forgot to add support for dependabot. Hopefully that is done now :-)

Oh nice catch, ty!

PR #5704, opened 2026-09-05
test(characterization): pin stream and supervisor interactions in the goldens

JakobLichterfeld:
## What this adds

**Interactions section (commit 1).** What the vehicle does to its neighbours without leaving a row or a topic — stream connect (`Api.stream/3`), stream disconnect (`Stream.disconnect/1`) and the supervisor kill — was delivered to the collector and flushed. It is now recorded in `ApiMock` and pinned in a golden section `interactions`, each entry with `after_serve`, the collector's index of the API event served last:

```json
"interactions": [
  {"after_serve": 1, "stream": "connect"},
  {"after_serve": 11, "vehicles": "kill"}
]
```

Mechanism: the harness queues its serve closures as `{:indexed, index, closure}` (also inside snapshots, whose probe and strict serve share one index); `ApiMock` unwraps them and keeps `last_served`. In a replay the mock is the stream — it returns itself as the stream pid, so the disconnect cast lands in its `handle_info` — and the `Characterization.Vehicles` stand-in reports the kill to it before terminating the process. The section exists only when the replay recorded something. Self-test `interactions` (the events of `driving_stream_reconnect_offline`: connect, disconnect on offline, reconnect mid-drive, disconnect on asleep); probe: a golden without the section diverges from a capture that carries it (`diff/2`). Sign-in is not an interaction: the vehicle never calls it.

**Re-record (commit 2).** 62 goldens gain the section; 57 stay without it (every polling scenario, and scenarios that never go online). Structural check against HEAD: no other key changed. Descriptions of `driving_stream_reconnect_offline`, `driving_stream_reconnect_unavailable` and `vehicle_eid_changed_restart` name the pin.

**`driving_direct_transition_repeat` (commit 3).** Three identical N polls — same timestamp, position, odometer — insert three positions into one drive; no repeat is deduplicated. This is the repeat coverage the original `transitions directly into driving state` pinned by serving the N frame forever, expressed with a bounded repeat and a parked terminal. No mechanism involved.

## Enforced invariants

- Interactions are captured, not flushed (`ApiMock.record_interaction/2`, `ApiMock.interactions/1`, `canonical_interaction/1`); a golden without the section diverges from a replay that records one (`diff/2`).
- Serve indices stay the collector's: `{:indexed, index, closure}` carries the `index_events/1` index into the mock, a snapshot's probe and strict serve share it (`ApiMock.mark_served/2`).
- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`): the terminal's disconnect is recorded once, on its first serve.
- Mock-based tests unchanged: `ApiMock` returns itself as stream pid only with the `:vehicle` option, otherwise the test process as before.

## Verification

- `mix test test/teslamate/characterization_test.exs test/teslamate/characterization_harness_test.exs test/teslamate/vehicles/api_mock_test.exs` without record mode: 171 passed.
- Full suite without record mode: 693 passed.
- Structural check of the re-record: 105 existing goldens, only `interactions` added.
- `mix format --check-formatted` on the harness files: clean.

Deviations between the derived expectation and the recorded goldens, each with its code reason:
- **Connect after serve 1 in the snapshot-less fixtures** (expected 2): the mock answers probe and strict with the same payload, and the `:start` handler decides on the payload (`drive_state` present), not on the endpoint, so it goes online on the probe result (`vehicle.ex:1064`).
- **Error events carry serve indices** (`index_events/1`), which the expectation had not counted: `charging_connection_loss` connect@10 / disconnect@14, `driving_offline_15min_new_drive` 25/27, `driving_stream_reconnect_offline` 24/26, `driving_stream_reconnect_unavailable` 20/22, `vehicle_unavailable_fallback_asleep` disconnect@3.
- **Disconnect on the first offline result** during a drive, not after the drive timeout: the fetch-result handler disconnects the stream on every offline/asleep result before dispatching to the state (`vehicle.ex:496`).
- **Two disconnects in `vehicle_in_service_ends_drive`** (after serves 4 and 5): the in-service abort disconnects (`vehicle.ex:534`) without clearing `stream_pid`, so the following asleep result disconnects again through `vehicle.ex:496`. In production a cast to an already terminated stream process; observation, not a diagnosis.
- Postgrex logs `client … exited` in the kill scenarios (the vehicle is killed while holding a connection); present since #5698, tests green.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree. The re-record is exactly what the mechanism predicts: 61 existing goldens change by the added `interactions` key and nothing else (structural comparison against HEAD, zero violations), the 57 polling and never-online scenarios stay without it. The serve index is the collector's own — `{:indexed, index, closure}` carries `index_events/1` into the mock, probe and strict of a snapshot share it — so there is one source for that number. The mock becomes the stream only under the `:vehicle` option, and the stand-in reports the kill before it terminates the process, so the kill lands on the serve that caused it.

The four deviations from the derived expectation each hold at the code: the `:start` handler decides on the payload and goes online on the probe result; error events carry serve indices; the fetch-result handler disconnects on every offline or asleep result before dispatching; the in-service abort disconnects without clearing `stream_pid`, so the following asleep result sends a second, inert disconnect. Documenting them as deviations with reasons is the right form — the goldens now show what the mock-based tests could only assert one message at a time.

`driving_direct_transition_repeat` closes the per-poll question the right way: three identical N polls, three drive positions, no mechanism. The "tagged events" item resolves into a section the harness almost had and a scenario that never needed one.

---

PR #5705, opened 2026-09-05
test(characterization): name the scenario event behind a mismatch on a dated row

JakobLichterfeld:
## What this adds

When a comparison fails on a row that carries a date, the mismatch message now names the scenario event whose cycle created that row. Nothing else changes: undated paths (MQTT lists, counters, calls) read as before, interactions carry their serve index since #5704, goldens are untouched.

`compare/1` hands the scenario to `assert_equivalent/4`; `mismatch_message/4` builds the message and, for paths of the form `database.<table>.[i].<column>`, appends the attribution of the row's `date` / `start_date` — the golden's value, or the capture's when the golden's is masked or the row exists only there. An `end_date` divergence adds a second line for the closing date.

Attribution rule (`attribute_date/2`), derived from the replay clock: every serve raises the clock to at least its payload time and clock events set it explicitly, so a row dated `d` was created in the cycle of the last timed event at or before `d`. The timeline holds `vehicle` events with a `drive_state.timestamp` (with their serve index from `index_events/1`), `clock` values and `stream` frame times, in list order. An exact match reads `created in the cycle of event 4 (vehicle, serve 4, timestamp …)`; a clock-dated row reads `2 ms after event 5 (clock, value …)` — the distance, not a tick count, because `Clock.serve/1` runs per closure execution (snapshot twice, terminal repeats, error fallback) and ticks identify no event. Dates before the first timed event read `before the first event (seed or clock-dated)`.

```
characterization mismatch in attribution
first divergence at database.drives.[0].end_date
expected: "1999-01-01T00:00:00.000000"
actual:   "2024-01-01T00:00:30.000000"
row dated 2024-01-01T00:00:10.000000 — created in the cycle of event 2 (vehicle, serve 2, timestamp 1704067210000)
row closed 1999-01-01T00:00:00.000000 — closed before the first event (seed or clock-dated)
```

## Enforced invariants

- A mismatch on a dated row names the scenario event whose cycle created it — last timed event at or before the row's date, distance in milliseconds between events, closing event on an `end_date` divergence, seed or clock-dated rows before the first event called out (`mismatch_message/4`, `attribute_date/2`).
- Probe (`characterization_test.exs`): a recorded drive scenario with tampered golden rows — exact attribution to event 2, `end_date` attribution with the golden's date and with the capture's date when the golden's is masked, the clock-dated asleep row as a distance after the clock event, a date before the scenario.

## Verification

- Probe: 1 passed. `mix test test/teslamate/characterization_test.exs test/teslamate/characterization_harness_test.exs` without record mode: 168 passed.
- Full suite without record mode: 694 passed.
- `mix format --check-formatted` on the two changed test files: clean.
- No golden changed.

- **Golden date first.** When the diverging column is the date itself, the golden's date is attributed (as specified); the probe shows the alternative path through a masked golden date.
- **Asleep row two ticks after the clock event.** From `:online` the asleep terminal routes through `:start`, so the row is written on its second serve: the message reads `2 ms after event 5`, the case the distance wording exists for.
- Also fixes the two struct-update warnings from #5704 in `mark_served/2`: the clauses match `%State{}`, and both call sites bind the result as `%State{} = state`, since the type checker loses the struct through the return value.
- Open questions: none.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Under the hood, this release adds a self-verifying black-box characterization su
 - test(characterization): convert the remaining vehicle scenarios — resume_logging and summary calls, expect_halt, seed positions, Vehicles stand-in (#5698 - @JakobLichterfeld)
 - test(characterization): pin charge samples without charger_power (#5700 - @JakobLichterfeld)
 - test(characterization): pin stream connect/disconnect and the supervisor kill as golden interactions (#5704 - @JakobLichterfeld)
+- test(characterization): name the scenario event behind a mismatch on a dated row (#5705 - @JakobLichterfeld)
 
 #### Dashboards
 
--- a/test/support/characterization.ex
+++ b/test/support/characterization.ex
@@ -256,6 +256,12 @@ defmodule TeslaMate.Characterization do
       vehicle process is monitored and the replay fails before teardown
       unless it actually went down; a declaration without an `await`
       convergence outcome raises (`replay/2`, `expect_restart?/1`).
+    * A mismatch on a dated row names the scenario event whose cycle created
+      it — the last timed event (payload timestamp, clock value or stream
+      time) at or before the row's date, with the distance in milliseconds
+      when the date lies between events; an `end_date` divergence names the
+      closing event as well, rows dated before the first event are called
+      out as seed or clock-dated (`mismatch_message/4`, `attribute_date/2`).
     * Interactions are captured, not flushed: stream connect and disconnect
       and the supervisor kill land in the golden with their serve index
       (`ApiMock.record_interaction/2`, `ApiMock.interactions/1`,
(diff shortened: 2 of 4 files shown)


JakobLichterfeld (commented):
Verified at the tree. The attribution follows from the replay clock and nothing else: every serve raises the clock to at least its payload time, clock events set it, so the last timed event at or before a row's date is the cycle that created it. The timeline counts serves exactly as `index_events/1` does (vehicle and error events, no deliveries), takes the golden's date first and the capture's when the golden's is masked or the row exists only there, and adds the closing event only when `end_date` itself diverges. Reporting the distance instead of a tick count is the right call: `Clock.serve/1` runs per closure execution — snapshot twice, terminal repeats, error fallback — so ticks identify no event, while "2 ms after event 5" is always true and tells the reader it is a clock-dated row.

The probe covers the five paths in one recorded scenario: exact attribution, `end_date` from the golden and from the capture behind a mask, the clock-dated asleep row as a distance, and a date before the scenario. No golden changed. The `mark_served/2` struct matches remove the two warnings #5704 left.

With this the differ item closes in the small form: undated paths read as before, interactions carry their serve index since #5704, dated rows now point at the event.

---

PR #5706, opened 2026-09-05
fix(vehicle): keep the published state start time from jumping backwards after charging, updating or driving

### LINKED ISSUE #5693: State start time can jump backwards after charging, updating or driving (server clock vs. car clock)
Found while making `LogMock` honour `date:` (#5692): four vehicle tests only held because the mock invented a later time for the online row.

## Mechanism (from `lib/teslamate/vehicles/vehicle.ex`)

`Summary.since` is `Data.last_state_change`, and it is fed from two clocks:

- **Server clock** (`deps.clock.utc_now()`): entering driving (`start_drive`), charging, updating, suspended, and `resume_logging`.
- **Car clock** (payload `drive_state.timestamp` via `date_opts`): online, offline and asleep take the `states` row's `start_date`.

So on charging → online (or updating → online), `since` jumps from the server moment the transient state began to the car timestamp of the payload that ended it. That is earlier whenever the car's clock lags the server by more than the transient state lasted — or the payload is a cached/late sample. #5558 measured exactly such skews; at test speed the transient lasts milliseconds, so the faithful mock shows the step on every run.

Only the published `since` is affected (MQTT topic, `Summary`, everything downstream that derives "in this state for …"). The rows themselves are consistent: drives, charging processes, updates and states are all payload-dated.

## Direction

Rule: one time source per derived value. Every transient state already has a payload-dated row (`drive.start_date`, `cproc.start_date`, `update.start_date`); `since` can be that row's start instead of the server clock, which makes the whole `since` course car-dated like the rows it summarises. Open design point: `:suspended` has no row of its own.

## Pin

The four sharpened assertions from #5692 hold the current behaviour (`since` after a transient state equals the following online row's payload date). A characterization pin needs `since` unmasked in the goldens: the mask predates the replay clock (#5690), under which `since` is deterministic. Lifting it is a re-record of every golden and belongs in the fix PR, before the fix commit, so the golden diff documents the change.

Related: #5692 (found here), #5684 (same clock boundary, states rows), #5558 (clock measurements).

---

JakobLichterfeld:
Fixes #5693.

## Mechanism

The `since` topic publishes `last_state_change`. A transient state — charging, updating, driving, suspended — set it to the server clock. When the car returns to online, the vehicle goes through `:start`, and `Log.start_state` returns the online row that was never closed during the transient state; `since` fell back to that row's start, which lies before the transient state's start. Every charge, drive and update therefore ended with `since` jumping backwards: `00:00 → 00:10 → 00:00`.

## The fix

One rule: `since` is the payload date of the poll that changes the summary state, through `date_opts/2` (stale protection from #5692 included) — `state_change_date/2`. It applies to the transient states (charging start, update start, drive start via `start_drive/3` with the date of the calling poll, suspend by idle and by command) and to the return to online in the `:start` online handler. `:start` serves the process start and the return alike; a `%Data{initial_fetch?: true}` flag, cleared by the three `:start` handlers, keeps the row's start at the process start (no state changes there) and dates the return with the returning poll. Separating the two roles of `:start` is phase 2. `resume_logging` carries no payload and keeps the clock.

## Mock fidelity

`LogMock.start_state` returned a fresh row dated with the given date on every call, so the mock suite never saw the jump. It now keeps the open row when the state does not change, like `Log.start_state/3`. With that change alone (second commit) five assertions went red — the jump in numbers:

| Test | since on return | expected |
|---|---|---|
| ChargingTest "logs a full charging cycle" | 17:28:51.041 | 17:28:51.045 |
| UpdatingTest "logs an update cycle" | 17:28:49.662 | 17:28:49.669 |
| UpdatingTest "logs an update if the status is not empty" | 17:28:49.626 | 17:28:49.628 |
| SuspendLoggingTest "suspends when charging is complete" | 17:28:50.302 | 17:28:50.303 |
| DrivingTest "logs a full drive" | 1134 µs before the drive's `since` | ordered |

The fix turns all five green without touching them. Three ordering assertions became `diff == 0` under the rule and are now exact: the D payload opens the online row and starts the drive (`driving_test.exs`), the Charging payload and the suspend's strict fetch carry the online payload's timestamp (`suspend_logging_test.exs`).

## The pin's golden diff

First commit: the `since` mask is gone (it dates from wall-clock time before #5690); 120 goldens re-recorded, structurally only `"<volatile>"` → value. 40 goldens show the jump, e.g. `charging_full_cycle` `00:00, 00:10, 00:00`, `driving_full_drive` `00:10, 00:10, 00:10` (formats differing only), `streaming_suspend_resume_drive` `00:00, 00:10, 00:30, 00:00`.

Third commit: 52 goldens change, only `since` values. The 40 now carry the returning poll's date — `charging_full_cycle` `00:00, 00:10, 00:40`, `driving_full_drive` `00:10, 00:10, 00:40`, `updating_full_cycle` `00:00, 00:10, 01:00`, `suspend_after_charging_complete_recharge` `00:00, 00:10, 00:20, 03:40, 03:50, 04:10`. Clock ticks on transient states vanish (`suspend_idle*`, `vehicle_settings_change`, `suspend_after_charging_complete*`); the rest are the same instant in the clock's format (`.002000Z` → `.002Z`).

## Verification

- Vehicle mock suite: 125 passed. Characterization re-record: 146 passed.
- Structural check of every re-record against the previous commit: only `since` values (topic and summary call), nothing else.
- Full suite without record mode: 693 passed.
- No compiler warnings in `lib/`; `mix format --check-formatted` clean.

- **Mask origin.** The `since` mask dates from wall-clock time before the replay clock (#5690); since then the value is deterministic and is pinned.
- **`initial_fetch?`** is the smallest explicit fix for the double use of `:start`; separating entry and return is phase 2.
- **Extra red in the mock-fidelity commit:** `DrivingTest "logs a full drive"` ordering assertion, the same jump — five red instead of four.
- One full-suite run failed `replays selftest interactions` without a captured message; a second full run and isolated runs passed. Its `since` values are payload dates (the scenario's timestamps end in `…001`), so the value is not the cause. Hypothesis, not a diagnosis: since #5698 the kill scenarios terminate a vehicle holding a sandbox connection (Postgrex logs `client exited`), which may hit the next test's connection in shared mode. Tracked separately after this merge.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree. The rule holds at every site: `since` is the payload date of the poll that changes the summary state, through `date_opts/2` with the stale protection from #5692 — the five transient starts, `start_drive/3` with the calling poll's date at all three callers, and the return to online in `:start`, where `initial_fetch?` keeps the row's start only at the process start. That flag is the smallest explicit fix for `:start` serving entry and return alike; the split belongs to the phase 2 extraction, and the comment says so. `resume_logging` carries no payload and stays on the clock, named as the one exception.

The mock-fidelity commit is the same move as in #5692: `LogMock.start_state` now keeps the open row like `Log.start_state`, and with that alone five assertions go red showing the jump in numbers, then green under the fix without being touched. The three ordering assertions made exact are right at the code — the D payload dates both the online row and the drive, and the charging and suspend polls carry the online payload's timestamp.

The goldens document the change end to end: the first commit unmasks `since` (120 goldens, values only, 40 showing the jump), the third re-records 52 with nothing but `since` changed, and the return now carries the returning poll's date — `00:00, 00:10, 00:40` where it was `00:00, 00:10, 00:00`. The `…001` values in the reconnect scenarios are payload timestamps, not ticks. CI green on a fresh database.

---

PR #5709, opened 2026-09-07
fix(web): return to the Grafana dashboard from Back buttons and detect the Grafana URL with origin-only referrers

JakobLichterfeld:
## Problem

Browsers send only the origin for cross-origin referrers (`strict-origin-when-cross-origin`). Grafana and TeslaMate usually run on different origins, so `document.referrer` is just `http://grafana:3000/`. Two features relied on the full referrer:

- The Back button on the charge cost page linked to the referrer and landed on the Grafana home dashboard with the default time range.
- The geo-fence form derives the Grafana URL setting from the referrer's dashboard path. With an origin-only referrer the match failed silently and the setting stayed empty.

The geo-fence data links additionally opened in a new tab. A new tab has no history, and where several links share a table cell Grafana's context menu adds `rel="noopener noreferrer"`, so no referrer arrived at all. Both the geo-fence Back button and the Grafana URL detection only work through the dashboard change below.

## Changes

- New `HistoryBack` JS hook: Back navigates in the browser history and falls back to the link's `href` when there is no history. Modifier clicks are left to the browser. Used by the charge cost and geo-fence forms.
- All geo-fence data links in the dashboards open in the same tab (`targetBlank: false`), like the charge cost link already did.
- Grafana URL detection accepts a bare origin, matches dashboard paths on the `/d/` segment (trailing slash tolerated) and keeps nested sub-paths in order (previously reversed).
- Fixes a pre-existing crash on mount: a referrer that fails URL validation (e.g. `android-app://…`) made `set_grafana_url` return `:ok` instead of the settings.
- Tests for both Back buttons, origin-only referrers, nested paths, trailing slash and the invalid-URL referrer.

- With an origin-only referrer a Grafana sub-path on a different origin than TeslaMate (e.g. `https://other-host/grafana`) cannot be detected; the setting then holds the origin and the navbar dashboard links 404 until corrected manually in the settings. Same-origin sub-paths deliver the full referrer and work. Previously nothing was detected in either case.
- `history.back()` uses plain browser semantics: the previous entry in the tab, which is not necessarily where the fallback `href` points.
- The hook's behaviour itself is not covered by tests; the repo has no JS test setup. Tests verify the hook is attached and the fallback `href`.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) (Fable 5.1 medium) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 Under the hood, this release adds a self-verifying black-box characterization suite: recorded API sequences replay through the real vehicle state machine, and everything that leaves the system — database rows, MQTT messages, and the vehicle's interactions with the streaming API and its supervisor — is pinned against goldens. All existing vehicle scenarios are converted (120 fixtures).
 The work already paid off three times: it exposed a crash in the update-cancel path (#5656), a crash loop after an offline period when the car reports an outdated timestamp (#5684), and the published state start time jumping backwards after charging, updating or driving (#5693) — all fixed in this release — and it pinned a data-quality quirk for a later fix (#5699: a charge sample without charger power is stored as 0 kW).
 
+The geo-fence links in the Grafana dashboards now open in the same tab, so the Back button returns to the dashboard and the Grafana URL is detected automatically again (#5709).
+
 **Note for Home Assistant MQTT discovery users:** TeslaMate no longer re-runs the discovery migration on every restart, which briefly removed and recreated entities (#5667). Instead it clears the former per-entity topics and republishes the device config; Home Assistant logs one harmless "conflicting MQTT discovery message" warning per legacy topic after each restart, entities are untouched.
 Upgrading directly from 4.1.x no longer preserves entity registry customizations — see the [docs](https://docs.teslamate.org/docs/integrations/home_assistant#mqtt-discovery-automatic-configuration) (#5685).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Upgrading directly from 4.1.x no longer preserves entity registry customizations
 - fix(vehicle): keep the published state start time from jumping backwards after charging, updating or driving (#5706 - @JakobLichterfeld)
 - fix(web): pin the size of Leaflet's SVG overlay so the vehicle arrow and geofence circle stay on the map at any Safari page zoom (#5666 - @JakobLichterfeld)
 - fix(grafana): open the TeslaMate header link in a new tab so it works when Grafana and TeslaMate share an origin (#5626 - @misenhower)
+- fix(web): make the Back button return to the Grafana dashboard and detect the Grafana URL despite origin-only referrers (#5709 - @JakobLichterfeld)
 
 #### Build, CI, internal
 
--- a/assets/js/hooks.js
+++ b/assets/js/hooks.js
@@ -33,6 +33,20 @@ export const Dropdown = {
   },
 };
 
+// Navigates back in the browser history (e.g. to the Grafana dashboard the
+// page was opened from) and falls back to the link's href if there is no
+// history, e.g. when the page was opened in a new tab.
+export const HistoryBack = {
+  mounted() {
+    this.el.addEventListener("click", (e) => {
+      if (window.history.length > 1) {
+        e.preventDefault();
+        window.history.back();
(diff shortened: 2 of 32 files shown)


JakobLichterfeld (commented):
## Self-review before merge

Two things need fixing, one needs a decision, the rest is polish.

### Must fix

1. **Crash on mount for non-http(s) referrers.** The `else` branch in `set_grafana_url/2` returns the value of `Logger.warning/1`, i.e. `:ok`, but `mount/3` matches on `{:ok, settings}`. With the new `{:ok, nil}` clause a bare origin like `android-app://com.example.app` (what Chrome on Android sends when a link is opened from an app) passes the scheme/host guard, fails `validate_url` in the changeset and raises a `MatchError`, so `/geo-fences/new?lat=…&lng=…` blows up. Pre-existing bug, but this PR makes it reachable. Return `{:ok, settings}` after logging and add the `android-app://` case to the "handles weird referrers" test.

2. **The geo-fence detection does not fire from the main dashboards at all.** All geo-fence data links are `targetBlank: true`. In the Charges/Drives/Trip tables they share the cell with other links, so Grafana 13.2.1 renders its context menu, and `MenuItem` sets `rel="noopener noreferrer"` for `_blank` links. No referrer arrives, origin-only or otherwise. Only the Locations "Addresses" table (single link, plain anchor) still delivers an origin. In a new tab `history.length` is also 1, so the geo-fence Back button falls back to `/geo-fences` from Grafana exactly as before. The PR text needs to say this precisely; the Back button change is effectively only for the charge cost page.

### Decision

3. **Origin-only referrer now persists a possibly wrong `grafana_url`.** For Grafana under a sub-path on a different origin than TeslaMate we now store the bare origin, the navbar shows dashboard links that 404, and nothing tells the user why. Before, `nil` kept the menu hidden, which is an honest "not configured". Given the point above, this only affects the Locations dashboard path. I lean towards keeping `nil` when only an origin is available, or moving the base path into the Grafana data link explicitly (the `base_url` variable already flows the other way).

### Polish

4. `HistoryBack` swallows Cmd/Ctrl/Shift-click: `preventDefault()` runs regardless of modifiers, so "open in new tab" navigates the current tab backwards. Bail out early on `e.metaKey || e.ctrlKey || e.shiftKey || e.altKey || e.button !== 0`.
5. `history.length > 1` is not "came from Grafana": Chrome counts the new-tab page as an entry, navbar navigation pushes entries, and an external predecessor in the same tab sends the user out of TeslaMate. Acceptable for browser-Back semantics, but `href` and the click destination diverge; at least document it in the hook comment.
6. `grafana_base_path/1` can be one clause: `Regex.run(~r{^(.*)/d/[^/]+}, path)` yields the same results for every current test and additionally tolerates a trailing slash (`/grafana/d/xyz/slug/`), which currently returns `:error`.
7. The geo-fence Back fallback is now a full page reload instead of `<.link navigate>`. `<.link navigate={…} id="back-button" phx-hook="HistoryBack">` keeps the SPA fallback and drops the hand-rolled anchor duplicated in both templates.
8. The origin test loop with the manual settings reset is clearer as two plain tests; the sandbox already isolates them.

---

JakobLichterfeld (commented):
## Follow-up review

All points from the self-review are addressed in a6ca42fe:

- `set_grafana_url` returns `{:ok, settings}` on a changeset error; `android-app://…` referrer covered in "handles weird referrers".
- `HistoryBack` leaves Ctrl/Shift/Cmd/middle clicks to the browser, same check as LiveView's `wantsNewTab`.
- All ten geo-fence data links in charges/drives/locations/timeline/trip are `targetBlank: false`. Verified against Grafana 13.2.1: an explicit `false` renders `target="_self"`, and `interceptLinkClicks` skips anchors with a `target` attribute, so shared-origin setups are unaffected. No `rel="noreferrer"` anymore, so the referrer reaches the geo-fence form.
- `grafana_base_path` is one regex clause and tolerates a trailing slash, with a test.
- Origin test loop split into two tests.
- PR text states the `noreferrer` mechanism and the dependency on the dashboard change; changelog carries the same-tab note up top.

Ran both touched test files locally: 36/36. CI is green.

---

JakobLichterfeld:
tested successfully with the docker images

PR #5710, opened 2026-09-07
feat(web): explain why no vehicle is logged and offer a reload button that starts loggers for vehicles added to the Tesla account, instead of requiring a restart

JakobLichterfeld:
## Summary

The vehicle list is fetched from the Tesla API only when the loggers start. A car that shows up in the Tesla account later, the first one after sign-in or a second one after delivery, got no logger until the container was restarted, and the car overview stayed blank without saying why.

- `TeslaMate.Vehicles` now owns the vehicle supervisor and offers `discover/0`: one API call, then a logger is started for every vehicle that has none. Running loggers are untouched, cars with data collection disabled are skipped, vehicles that left the account are not stopped. No polling: every `list_vehicles` call is billed on the Fleet API, so the only trigger is a click.
- The car overview explains an empty list (known cars disabled or gone from the account, or nothing found at start-up) and offers **Reload vehicles**. After a click the page shows the result: account still empty, rate limited, API error, or the new car's card. The overview subscribes to vehicle changes, so a reload from the settings page or another tab shows the card without a click.
- The settings page gets a **Vehicles** section with the same button, so a second car can be added while the first is logging. It reports which loggers were started.
- `Mqtt.PubSub` starts subscribers for discovered vehicles, so their topics and Home Assistant discovery appear without a restart.
- Sign-in, sign-out and the data collection toggle keep their existing restart path.

## Design notes

- Restart-free by structure: `discover/0` only ever calls `Supervisor.start_child/2`, which is idempotent per car, so concurrent clicks are harmless. Persisting runs in the caller; the owner only starts children and never touches the database, so a failing database call cannot take the loggers down.
- Two concurrent reloads can race between looking a new car up and inserting it; the loser hits the unique constraint and simply runs the lookup again, which now finds the car. Loggers started before a later vehicle failed are announced to MQTT and the overview.
- The owner traps exits and stops its supervisor synchronously in `terminate/2`; a supervisor that exits on its own takes the owner with it, and the application supervisor restarts both with a fresh discovery. Both owners declare `shutdown: :infinity`, so the application supervisor waits for the vehicles' own shutdown instead of killing them after a fixed timeout.
- What the start-up discovery returned is not recorded. At mount the page derives its explanation from the database (known cars without a logger are disabled or no longer in the account); only a click's result is shown verbatim.
- `kill/0` and `restart/0` keep their semantics: `kill/0` kills the vehicle supervisor, which takes the owner down and lets the application supervisor restart both; `restart/0` stops the owner and waits for it to come back. `TeslaMate.Vehicles.Vehicle` and its interface to `Vehicles` are untouched, so the characterization suite and its `Vehicles` stand-in are unaffected.

## Tests

- `Vehicles.discover/0`: new vehicle started and announced, running pids unchanged, idempotent, disabled cars skipped, every API error, a non-persistable vehicle returns the error and leaves owner and loggers alive, partial success is announced, a car inserted concurrently is picked up, four concurrent discoveries all succeed and start each vehicle once.
- Car overview: both mount explanations, reload that finds the car (one API call), account empty, rate limit, API error, sign-in redirect, crashed reload re-enables the button, one reload per tab, two tabs concurrently, card appears after a reload elsewhere.
- Settings: second car started while the first keeps its pid, no new vehicle, button hidden while signed out.
- `Mqtt.PubSub`: subscriber for the new car appears, existing pid unchanged, no Home Assistant clear on a discovery.
- Full suite 801 green with `--warnings-as-errors`, dialyzer, `mix format --check-formatted`, `mix gettext.extract --check-up-to-date`. German translations added.

- `Vehicles.list/0` exits when the owner is not running (unchanged from main, where `which_children` on a stopped supervisor did the same). It only matters during the application supervisor's restart window.
- Settings: the reload lives in its own **Vehicles** section rather than inside **Car Order**, because that section only renders with two or more cars and the second-car case starts with one.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Upgrading directly from 4.1.x no longer preserves entity registry customizations
 
 ### New features
 
+- feat(web): explain an empty vehicle list and offer a reload button instead of requiring a restart once the car shows up in the Tesla account (#5710 - @JakobLichterfeld)
+
 ### Improvements and bug fixes
 
 - fix(vehicle): cancel an update with the logged update row instead of the API payload, which crashed the vehicle process and left the update open forever (#5664 - @JakobLichterfeld)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,14 @@ The work already paid off three times: it exposed a crash in the update-cancel p
 
 The geo-fence links in the Grafana dashboards now open in the same tab, so the Back button returns to the dashboard and the Grafana URL is detected automatically again (#5709).
 
+A vehicle that shows up in the Tesla account only after the first sign-in no longer needs a container restart: the car overview explains why the list is empty and offers a reload. Signing in, toggling data collection and reloading now also update the [MQTT](https://docs.teslamate.org/docs/integrations/mqtt) topics and Home Assistant discovery immediately, so disabling a car removes its Home Assistant entities right away (#5710).
+
 **Note for Home Assistant MQTT discovery users:** TeslaMate no longer re-runs the discovery migration on every restart, which briefly removed and recreated entities (#5667). Instead it clears the former per-entity topics and republishes the device config; Home Assistant logs one harmless "conflicting MQTT discovery message" warning per legacy topic after each restart, entities are untouched.
 Upgrading directly from 4.1.x no longer preserves entity registry customizations — see the [docs](https://docs.teslamate.org/docs/integrations/home_assistant#mqtt-discovery-automatic-configuration) (#5685).
 
 ### New features
 
-- feat(web): explain an empty vehicle list and offer a reload button instead of requiring a restart once the car shows up in the Tesla account (#5710 - @JakobLichterfeld)
+- feat(web): explain an empty vehicle list and offer a reload button instead of requiring a restart once the car shows up in the Tesla account; reload, sign-in and data collection toggle now also update MQTT and Home Assistant discovery (#5710 - @JakobLichterfeld)
 
 ### Improvements and bug fixes
 
--- a/lib/teslamate/api.ex
+++ b/lib/teslamate/api.ex
@@ -148,7 +148,7 @@ defmodule TeslaMate.Api do
       {:ok, %Auth{} = auth} ->
         true = insert_auth(state.name, auth)
         :ok = call(state.deps.auth, :save, [auth])
-        :ok = call(state.deps.vehicles, :restart)
+        :ok = restart_vehicles(state)
         {:ok, state} = schedule_refresh(auth, state)
         :ok = :fuse.reset(fuse_name(state.name))
 
@@ -177,7 +177,7 @@ defmodule TeslaMate.Api do
     :ets.delete(name, :auth)
     if is_reference(state.refresh_timer), do: Process.cancel_timer(state.refresh_timer)
     :ok = call(state.deps.auth, :delete_tokens)
-    :ok = call(state.deps.vehicles, :restart)
+    :ok = restart_vehicles(state)
     {:reply, :ok, %State{state | refresh_timer: nil}}
   end
 
@@ -256,6 +256,17 @@ defmodule TeslaMate.Api do
 
   defp insert_auth(name, %Auth{} = auth) do
     :ets.insert(name, auth: auth)
+  end
+
+  # A failed restart is recorded in TeslaMate.Vehicles.status/0 and shown on
+  # the car page; it must not take the sign-in or the tokens down with it.
+  defp restart_vehicles(%State{deps: deps}) do
+    case call(deps.vehicles, :restart) do
+      :ok -> :ok
+      {:error, reason} -> Logger.error("Restarting vehicles failed: #{inspect(reason)}")
+    end
+
+    :ok
   end
 
   defp fetch_auth(name) do
--- a/lib/teslamate/mqtt/pubsub.ex
+++ b/lib/teslamate/mqtt/pubsub.ex
@@ -1,7 +1,17 @@
 defmodule TeslaMate.Mqtt.PubSub do
-  use Supervisor
+  @moduledoc """
+  Keeps one `VehicleSubscriber` running per logged vehicle.
+
+  The set of vehicles changes at runtime: signing in or out, toggling data
+  collection and reloading the vehicle list all restart `TeslaMate.Vehicles`.
+  This server subscribes to those reloads and reconciles its subscribers with
+  the current vehicle list, so MQTT topics and Home Assistant discovery follow
+  without a container restart.
+  """
+  use GenServer
 
   require Logger
+  import Core.Dependency, only: [call: 3]
 
   alias __MODULE__.VehicleSubscriber
   alias __MODULE__.HomeAssistant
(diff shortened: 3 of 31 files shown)


JakobLichterfeld (commented):
Self review in new session: the empty-state explanation is a real improvement over a silent blank page. Two things need to be addressed before merging, the rest is optional.

### Blocking

**1. Reload restarts through the application supervisor and can take the whole VM down.**
`Vehicles.restart/0` stops the `TeslaMate.Vehicles` child; as a permanent child it is restarted by `TeslaMate.Supervisor`, and every restart counts toward its default intensity (3 restarts per 5 seconds). A fourth click within five seconds therefore shuts down the application, and since `:teslamate` is a permanent application in the release, the VM halts and the container restarts. This is realistic in exactly the rate-limited case the page describes: a 429 comes back in ~100 ms, the button re-enables, and users tend to hammer "Reload". I verified the mechanism with a plain `Supervisor` (the 4th `:normal` stop of a permanent child within 5 s kills the parent).

The path already existed via `Settings.on_enabled_change/2` and the sign-in flow, but this PR adds a user-facing trigger designed for repeated clicks, so it has to be fixed at the source rather than with a client-side cooldown (a second tab bypasses that anyway). Options: reload the children in place (terminate and start children from a fresh discovery without stopping the supervisor), or give `Vehicles` its own parent with an intensity that matches interactive use.

**2. MQTT does not see a vehicle found by reload.**
`Mqtt.PubSub` builds one `VehicleSubscriber` per car from `Vehicles.list()` at its own init only. A car discovered via the new button gets logged and shows a card, but no `teslamate/cars/<id>/*` topics and no Home Assistant discovery until the container restarts, which is exactly what the title says is no longer needed. Either refresh the MQTT subscribers as part of the reload (the sign-in restart has the same gap), or state the limitation in the empty-state text and the changelog entry.

### Should fix

- `handle_event("reload_vehicles", ...)` is not guarded server-side. The comment says the reload is only offered while no vehicle is logged, but a stale tab still has the button after another tab or a re-sign-in started the logger, and the click then kills a running drive/charge. Ignore the event unless `Vehicles.list() == []`.
- When the API call fails and all fallback cars from the database are disabled, the page blames the API error and offers a billed reload, although the list is empty because data collection is disabled. Worth checking the DB cars first or showing both hints.

### Nice to have

- No test covers the failure branches of `handle_async` (`{:error, reason}` / `{:exit, reason}`) or the `reloading?: true` guard clause. A regression that drops `reloading?: false` in `reload_failed/2` would leave the button disabled forever without a failing test.

Please keep the "Workarounds & open questions" section updated with whatever you decide on the MQTT point.

---

JakobLichterfeld (commented):
Follow-up on 6fad1f40: all five points from the first pass are addressed. One new one before merging:

**`restart/0` no longer self-heals.** With `terminate_child` + `restart_child` the parent does not restart `Vehicles` on its own, so an `init` that raises (e.g. the `{:ok, car}` match in `create_or_update!` on a DB error) leaves the child with pid `:undefined`. `Api.sign_in` then crashes on its `:ok =` match, and the car page reports "Data collection is disabled for all vehicles" because `list/0` returns `[]` for a stopped supervisor. Before, the parent retried and eventually failed loudly. Either fall back to `Supervisor.stop` on `{:error, _}` from `restart_child`, or let `list/0` distinguish "not running" from "empty" and render that. The same conflation shows briefly in a second tab during any restart window.

Minor: `refute html =~ ~s(disabled)` in the failed-reload test also matches the "Data collection is disabled" text; assert on the button attribute instead.

The changelog entry should mention that sign-in and the enabled toggle now propagate to MQTT and Home Assistant discovery without a container restart, since disabling a car now removes its HA entities immediately.

---

JakobLichterfeld (commented):
Follow-up on d9756ae: the failed-start gap is closed, the state is observable and recoverable, nothing is worse than before the PR. One more before merging:

**Callers that need the restart skip it when the lock is held.** `restart/0` aborts with `{:error, :restarting}`, and `Api.sign_in`, `Api.sign_out` and `Settings.on_enabled_change` log or ignore that. A reload in flight while the user signs in leaves the vehicles discovered without auth and redirects the reloading tab to `/sign_in`; a sign-out leaves the loggers running after the tokens are gone. Those callers should wait for the lock (retries `:infinity` or a `restart/1` wait option); abort-only is right for the UI click alone.

Minor: the comment in `assign_vehicles/2` says the status is read before the list, but all three call paths read the list first. And the `:restarting` state is a dead end: the button can only fail, and the page never learns the other restart finished. Since `{Vehicles, :reloaded}` exists now, subscribing in `mount` would also replace the click-time catch-up.

Not a blocker: three `persistent_term` keys plus a `:global` lock now emulate an owner process. Worth deciding whether `Vehicles` should own its supervisor as `Mqtt.PubSub` does, here or in a follow-up.

---

swiffer:
The problem of having to manually restart is not bound to the first tesla in your account, or is it? I wonder if we keep the case of second vehicle not showingup after delivery uncovered?

JakobLichterfeld:
> The problem of having to manually restart is not bound to the first tesla in your account, or is it? I wonder if we keep the case of second vehicle not showingup after delivery uncovered?

You are right, and it changes the design: the button is limited to an empty list only because a reload restarts the supervisor and would interrupt running loggers. Covering the second car means the reload must not be a restart but an additive discovery that starts loggers for vehicles not yet running and leaves the rest alone. That also removes the restart-intensity and state handling this PR accumulated. I will rework it that way and offer the reload in the settings as well.

```
CURRENT PR                                 PROPOSAL
──────────                                 ────────
Click                                      Click
  │                                          │
  ▼                                          ▼
stop supervisor ──► all loggers killed     list_vehicles          (1 API call)
  │                                          │
  ▼                                          ▼
start supervisor                           for each vehicle without a logger:
  │ list_vehicles                              create_or_update! + start_child
  ▼                                          │
all loggers restarted                      running loggers: untouched
                                             │
Consequences: restart intensity,             ▼
status/parent/lock in persistent_term,     broadcast → MQTT starts only
HA discovery wiped on a failed start,      the new subscribers
button only while the list is empty
```

JakobLichterfeld:
I will first finish #5652, so `Vehicle` is around 100% coverage then reworking the `Vehicles`, this aligns our plan in #5558 as well.

JakobLichterfeld:
Reworked from scratch and force-pushed; the PR description is updated to match.

Matthias's point was the decisive one: the restart-based reload could only ever cover the first car, because restarting the supervisor kills running loggers. The reload is now an additive discovery: one API call, a logger started for every vehicle that has none, running loggers untouched. That covers a second car after delivery, and it removes the restart-intensity, state and lock handling the previous iterations had accumulated. The button is offered on the car overview and, for the second-car case, on the settings page. MQTT and Home Assistant discovery follow without a restart.

Old commits are gone from the diff; the previous head was d30683b2 if anyone wants to compare.

---

JakobLichterfeld (commented):
Review in a second session of the rework (151bdcbd, 2d33cc4d, 3e0a8d77), high effort, against the criteria agreed before the rebuild: loggers keep running, one API call per click, one place builds child specs, no state outside processes, errors stay with the caller, invariants from structure.

**Structure holds.** `TeslaMate.Vehicles` owns the supervisor and does no database work; `discover/0` persists in the caller and only ever calls `start_child/2`, so a failing database call or a concurrent click cannot reach the running loggers. `Vehicle` and its interface to `Vehicles` are untouched, so the characterization suite is unaffected. `kill/0` and `restart/0` keep their semantics.

**Found during the review and fixed before the commits landed:**
- `start_missing/1` skipped the `:vehicles_changed` broadcast when a later vehicle failed, leaving an already started logger without MQTT topics. Now announced whenever anything was started.
- The concurrent-click race surfaced an `#Ecto.Changeset<…>` to the user; the loser now repeats the lookup and continues, and the concurrency test asserts all results succeed instead of filtering errors.
- The overview did not follow `:vehicles_changed`, so a reload from the settings page left the empty box and a click there spent another billed call. It subscribes in `mount` now.
- Both owners had the default `shutdown: 5000` while supervising; now `shutdown: :infinity`, stated in the moduledoc.
- "Starting logger for …" was logged for every account vehicle on every click; now only where a logger actually starts.

**Left open, by decision:**
- `Vehicles.list/0` exits while the owner is down, unchanged from main.
- The restart-intensity exposure of the sign-in and data-collection-toggle path is pre-existing and stays out of scope; the reload no longer touches it.

Not verified by me: the test run; the 801 green with `--warnings-as-errors`, dialyzer, format and gettext checks are from the implementation session. Merge once CI confirms.

---

JakobLichterfeld:
One more commit after the review above: `fix(vehicles): treat a vehicle with a pending restart as running during discovery`.

`Supervisor.start_child/2` answers `{:error, :already_present}` when a car's child spec exists but the logger is between restarts after a failed `init`. `discover/0` treated that as an error, aborted before reaching the remaining vehicles, and the page showed "Fetching the vehicles from the Tesla API failed: :already_present". Such a car has a logger its supervisor takes care of, so it is now reported as running, like `:already_started`. Covered by a test that terminates the child and keeps the spec.

Verified locally after this commit: full suite green, `--warnings-as-errors`, `mix format --check-formatted`, `mix gettext.extract --check-up-to-date`. 

---

PR #5712, opened 2026-09-08
test(characterization): pin the pre-online check of the streaming API

JakobLichterfeld:
## What this adds

Seven scenario/golden pairs for the pre-online check (`vehicle.ex`, fetch-result handler and the asleep/offline stream handlers): with streaming, a vehicle in asleep or offline only probes; a bare `{"state": "online"}` without vehicle data connects the stream and waits to see whether the online is real before any vehicle data is requested. Every earlier fixture answered probes with full payloads, so the check was never exercised.

| Fixture | Branch | Pin |
|---|---|---|
| `pre_online_real` | stream connected to check → frame with numeric power ("real online detected") → vehicle data released | states asleep, online; 1 position; connect after the bare probe, no second connect |
| `pre_online_fake` | frame with power nil ("subsystem check detected, fake online") | only the asleep row, 0 positions, one connect |
| `pre_online_inactive` | stream inactive sets the check idle; next bare probe hits "check idle, allow real online", the strict fetch is answered bare again | only the asleep row, one connect |
| `pre_online_no_frame` | no frame; next bare probe hits "nothing received, allow real online" | states asleep, online |
| `pre_online_idle_then_fake_frame` | idle, then frame with power nil ("check idle, treating as fake") | only the asleep row |
| `pre_online_idle_then_real_frame` | idle, then frame with numeric power ("check idle, treating as real") | states asleep, online |
| `pre_online_startup_restore` | first probe answers bare online at process start: last known values restored from `seed.positions`, stream connected, offline state without an offline row | one states row (online, no offline), seed + 1 positions, connect after serve 1, `calls` summary with state offline and the seeded latitude/longitude/battery |

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`): bare-online terminals repeat through the probe without writing, full-payload terminals stay online.
- Two-call terminals refused (`two_call_cycle?/2`): the three data-releasing fixtures serve the full payload twice, because the poll parked at the pre-online check is a probe and already carries the payload.
- Interactions and calls captured (`ApiMock.interactions/1`, `ApiMock.calls/1`): the connect after the bare probe and the restored summary are the pins that rows alone cannot carry.

## Verification

- `mix test test/teslamate/characterization_test.exs`: 153 passed. Full suite without record mode: 705 passed.
- Coverage of `lib/teslamate/vehicles/vehicle.ex` by the characterization suite alone (`MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`): **84.9 %**, 115 missed lines — baseline before this PR 80.6 %, 148 missed.
- Baseline lines now hit (baseline numbering): fetch-result handler pre-online block 417, 425, 429, 431, 436, 440, 444, 448, 450, 453, 457, 464, 468, 471, 475 (connect to check, nothing received, idle fallback, startup connect with restore); asleep/offline stream handler 606, 608, 613, 615, 622, 624, 634, 638, 640, 644, 646, 649 (fake, idle-fake, real, idle-real); inactive handler 664, 665; `fetch/2` release after `:confirmed_real` 1601.
- Still missed in these clauses: the fall-through `%Data{}` clauses for a frame when the check is already confirmed (629, 652), a frame without a power field (658), a bare online while streaming in a state other than asleep/offline/start (482), and log-call continuation lines.

- **Doubled terminal in three fixtures** (see invariants): a scenario-form consequence of the parked probe, named in the descriptions.
- **Path 3 vs. path 5:** after `inactive` the vehicle stays asleep only until the next probe, which hits the "check idle, allow real online" branch; `pre_online_inactive` therefore pins "no online row despite the release" (the strict fetch is bare again), and the frame variants after idle are `pre_online_idle_then_*`.
- **`call summary` in `pre_online_startup_restore`:** the restored values never reach a topic (no summary broadcast in the offline wait, and the first `:start` broadcast uses the full payload), so the summary call is the only carrier.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree against the derivation from a second session. The seven scenarios cover the pre-online check the way production reaches it: a bare `online` probe from asleep, then the stream decides — a frame with numeric power releases the vehicle data (`asleep, online`, one position), a frame with power nil keeps the vehicle asleep (only the asleep row, no position), `inactive` sets the check idle and the next bare probe releases without data arriving, no frame at all releases on the next probe, and the idle-then-frame pair covers both defensive branches. The startup variant pins the restore itself: one online row without an offline row, the seeded position, the connect after the first serve, and the `summary` call returning `offline` with the seeded coordinates and battery — the only carrier, since the offline wait broadcasts nothing.

The doubled full-payload terminal in the three data-releasing fixtures is a consequence of the parked probe being answered after the frame, correctly named in the descriptions and the body. Coverage of `vehicle.ex` by the suite alone moves from 80.6 % to 84.9 % as reported; the remaining misses in these clauses are the confirmed-state fall-throughs and continuation lines.

---

PR #5713, opened 2026-09-08
test(characterization): pin the suspended-state handlers

JakobLichterfeld:
## What this adds

Three scenario/golden pairs for the `:suspended` handlers that the coverage baseline showed untouched. Logging is suspended through the declared `suspend_logging` call in every fixture (its strict fetch is serve 2), streaming, idle window closed.

| Fixture | Branch | Pin |
|---|---|---|
| `suspended_resume_logging` | `resume_logging` while suspended ("Resuming logging"): back to the previous state, broadcast, fetch | state topic `online, suspended, online`; `calls` suspend ok, resume ok; two positions; the resumed `since` is the clock value (a command carries no payload) |
| `suspended_stream_usage` | stream frame with positive power while suspended ("Usage detected"): back to the previous state with the frame merged, immediate fetch | state topic `online, suspended, online` without a resume call; two positions |
| `suspended_inactive_online` | inactive stream while suspended starts `fetch_state`; the parked suspend poll is served first (online, stays suspended with a position), the `fetch_state` task second (online result while suspended, "keep state") | exactly two positions — the `fetch_state` result writes none where a poll would have —; states online then asleep through the next suspend poll |

Moduledoc, Limits: "A delivery is only observable with a poll in flight; a handler that runs only without one (fetch_state's own result clauses) is outside the replay."

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`); no replay ends in `:suspended` (`replay/2`).
- Two-call terminals refused (`two_call_cycle?/2`): the two resume fixtures serve the online payload twice — the parked suspend poll is a probe and already carries it.
- Calls captured (`ApiMock.calls/1`): the resume reply is the pin the rows cannot carry.

## Verification

- Full suite without record mode: 708 passed (seed 335606).
- Coverage of `lib/teslamate/vehicles/vehicle.ex` by the characterization suite alone (same command as the baseline): **85.9 %**, 107 missed lines — baseline 80.6 % / 148, after #5712 84.9 % / 115.
- Baseline lines now hit: 300, 302, 303 (resume from suspended), 795, 798, 800, 801 (usage frame), 820 (online result while suspended). Not hit: 817 (asleep/offline result while suspended), see below.

- **817 not pinned:** the clause runs only when `fetch_state`'s own result reaches the vehicle before any poll result, and a replay delivers only with a poll in flight. Classified as not reachable at the boundary in #5652; the attempt (`suspended_inactive_asleep`) forced two concurrent fetches to receive different answers — no plausible API sequence — and was discarded.
- **Second fetch without guard:** `fetch_state` bypasses the in-progress guard `:fetch` has — tracked in #5714, phase 2.
- **Doubled terminal** in `suspended_resume_logging` and `suspended_stream_usage`: the parked suspend poll is a probe and already carries the payload; the settling serve makes the barrier two strict fetches.
- **Line 1105** is a continuation line of a `Logger.info` whose clause already runs in every suspend fixture with polls; nothing to pin.
- Open questions: none.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Upgrading directly from 4.1.x no longer preserves entity registry customizations
 - build(deps): bump srtm from 0.8.0 to 0.9.0 (#5677)
 - build(deps): bump phoenix_live_view from 1.2.8 to 1.2.11 (#5678)
 - test(characterization): pin the pre-online check of the streaming API (#5712 - @JakobLichterfeld)
+- test(characterization): pin the suspended state's resume, usage and inactive-stream paths (#5713 - @JakobLichterfeld)
 
 #### Dashboards
 
--- a/test/fixtures/characterization/goldens/suspended_inactive_online.json
+++ b/test/fixtures/characterization/goldens/suspended_inactive_online.json
@@ -1,0 +1,283 @@
+{
+  "calls": [
+    {
+      "suspend_logging": "ok"
+    }
+  ],
+  "database": {
+    "addresses": [],
+    "car_settings": [
+      {
+        "enabled": true,
+        "free_supercharging": false,
+        "id": "car_setting_1",
+        "lfp_battery": false,
+        "req_not_unlocked": false,
+        "suspend_after_idle_min": 100000,
+        "suspend_min": 100000,
+        "use_streaming_api": true
+      }
+    ],
+    "cars": [
+      {
+        "display_priority": 1,
+        "efficiency": 0.153,
+        "eid": 42,
+        "exterior_color": "DeepBlue",
+        "id": "$car",
+        "inserted_at": "<volatile>",
+        "marketing_name": "LR AWD",
+        "model": "3",
+        "name": "blue",
+        "settings_id": "car_setting_1",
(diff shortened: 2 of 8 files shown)


JakobLichterfeld (commented):
Verified at the tree form a second session Three fixtures, each on a `suspend_logging` call with the strict fetch as serve 2: `resume_logging` returns the vehicle to online with the reply pinned in `calls`; a frame with positive power lifts the suspend without a command; an inactive stream while suspended starts the state fetch, and the golden shows the parked poll's position and the state fetch's silent online result — exactly two positions where a second poll would have written three. The two doubled terminals are the parked probe carrying the payload, named in the descriptions.

The discarded fourth fixture was the right call: it forced two concurrent fetches to receive different answers, which no car delivers, and its outcome hung on message order. The clause it aimed at runs only without a poll in flight, which a replay cannot express — classified in #5652, the moduledoc Limit says so, and the unguarded second fetch itself is #5714 for phase 2. Coverage of `vehicle.ex` by the suite alone: 84.9 % → 85.9 % as reported.

---

PR #5717, opened 2026-09-08
test(characterization): pin payload edge cases of the offline charge inference and stream frames

JakobLichterfeld:
## What this adds

Three scenario/golden pairs for the payload edge cases the coverage baseline showed untouched.

| Fixture | Branch | Pin |
|---|---|---|
| `driving_offline_soc_gain_charge_plugged` | `put_charge_defaults/1`, called only by the offline SOC-gain charge inference: the last payload before the offline period has no `charge_energy_added` key (defaulted to 0), the car returns still plugged in and charging with `charger_power` 11 (passed through unchanged); the live Charging payload then opens a second charging process, Complete and Unplugged close it | two closed `charging_processes`; the inferred one carries `0.00 / 0 kW` and `45.00 / 11 kW` — the 11 is visible against the default 0 |
| `stream_stale_after_usage` | a usage frame while suspended returns the machine to online with the frame merged into its last response; an older D frame is then stale and ignored | `drives` 0 despite a D frame (twin: `streaming_suspend_resume_drive`, where the newer D frame starts a drive), two positions, state topic `online, suspended, online` |
| `stream_frame_after_timestampless_payload` | a full online payload without `drive_state.timestamp` becomes the last response; `stale?/2` finds no usable last-response time and the following D frame starts a drive | `drives` 1 closed, five positions, state topic `online, driving, online, asleep` |

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`); two-call terminals refused (`two_call_cycle?/2`) — `stream_stale_after_usage` serves the online payload twice because the parked suspend poll is a probe.
- Await vacuity (`replay/2`): the asleep awaits guard the asleep row written after the barrier (asleep from `:online` routes through `:start`); `drives_closed: 1` guards the drive closed by the P payload.

## Verification

- Full suite without record mode: 732 passed (seed 237299).
- Coverage of `lib/teslamate/vehicles/vehicle.ex` by the characterization suite alone (same command as the baseline): **89.7 %**, 78 missed lines — previous family 89.3 % / 81.
- Baseline lines now hit: 1989, 1994 (`put_charge_defaults/1` default and pass-through), 2040 (`stale?/2` without a usable last-response time).

- **Missing `charge_energy_added` key** is stored as 0 in the inferred sample, indistinguishable in the row from a payload 0.0 — the same pin character as #5700.
- Baseline line 2039 (`stale?/2`, `DateTime` clause) is dead code, tracked in #5718. Lines 703/706, 1423, 1440 and 1444–1448 are classified as not reachable at the boundary in #5652.
- Open questions: none.

---


JakobLichterfeld (commented):
Verified at the tree ina second session. The offline charge inference is the only caller of `put_charge_defaults/1`, and the plugged-in return makes both of its branches visible in one golden: the inferred process holds `0.00 / 0 kW` from the missing key and `45.00 / 11 kW` passed through, next to the live process the Charging payload opens afterwards. The stale twin is exact — the same usage frame that lifts the suspend, then an older D frame, no drive — against `streaming_suspend_resume_drive` where the newer frame starts one. The timestamp-less payload pins the branch the code has for it: no usable last-response time, so the following frame counts and starts a drive.

The `DateTime` clause of `stale?/2` is dead and now #5718; the four remaining lines are classified in #5652 with their reasons at the code. Coverage of `vehicle.ex` by the suite alone: 89.3 % → 89.7 %.

---

PR #5719, opened 2026-09-08
test(characterization): pin the power-usage suspend guard and service mode across an idle suspend

JakobLichterfeld:
## What this adds

Three scenario/golden pairs.

| Fixture | Branch | Pin |
|---|---|---|
| `suspend_idle_power_usage` | the idle base scenario with exactly one guard field, `drive_state.power` 5 in the poll that would otherwise suspend: `can_fall_asleep/2` answers `power_usage`, `try_to_suspend/3` keeps the vehicle online | no `suspended` in the state topic, one position (base: two), `power` topic `0, 5` |
| `suspend_idle_service_mode` | the car enters service mode on the second poll, suspends in service mode after the clock event, and the suspend poll reports service mode off while suspended (served as a snapshot: the suspend poll probes first) | state topic `online, suspended, asleep` as without service mode; `service_mode` topic `false, true, false`; three positions |
| `suspend_idle_unlocked_required_service_mode` | the unlocked twin under `req_not_unlocked` with `service_mode` true in the unlocked poll: still `unlocked`, still online | no `suspended`, one position, `service_mode` topic `true`, `locked` topic `true, false` |

Service mode changes logging only (#5289), not the suspend decision; both service-mode fixtures say so in their descriptions.

Guard diffs against their bases (descriptions omitted): `suspend_idle_power_usage` vs `suspend_idle` — `await`, `power 0 → 5`; `suspend_idle_unlocked_required_service_mode` vs `suspend_idle_unlocked_required` — `service_mode: true`.

## Enforced invariants

- Terminal repeat-neutrality via the double record run (`record/1`, `mask_volatile/2`); no replay ends in `:suspended` (`replay/2`).
- Await vacuity (`replay/2`): the two guard twins await the asleep row written after the barrier (asleep from `:online` routes through `:start`); the service-mode fixture needs none (asleep from `:suspended` is immediate).

## Verification

- Full suite without record mode: 735 passed (seed 668994).
- Coverage of `lib/teslamate/vehicles/vehicle.ex` by the characterization suite alone (same command as the baseline): **90.5 %**, 72 missed lines — previous family 89.7 % / 78.
- Baseline lines now hit: 1824, 1826 (power-usage guard branch), 1898 (`can_fall_asleep/2` power clause), 1906 (`service_mode?/1`), 1915, 1918 (service-mode transition logs).

- **Snapshot for the service-mode leave payload:** in polling mode the suspend poll probes first; a plain payload would be consumed by the probe and the strict fetch would take the asleep terminal (#5687 rule). Named in the description.
- **`service_mode` topic of the unlocked twin shows only `true`:** the base payloads carry no `service_mode` key and nil is never published.
- Open questions: none.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [unreleased]
 
-Under the hood, this release adds a self-verifying black-box characterization suite: recorded API sequences replay through the real vehicle state machine, and everything that leaves the system — database rows, MQTT messages, and the vehicle's interactions with the streaming API and its supervisor — is pinned against goldens. All existing vehicle scenarios are converted (140 fixtures).
+Under the hood, this release adds a self-verifying black-box characterization suite: recorded API sequences replay through the real vehicle state machine, and everything that leaves the system — database rows, MQTT messages, and the vehicle's interactions with the streaming API and its supervisor — is pinned against goldens. All existing vehicle scenarios are converted (143 fixtures).
 The work already paid off three times: it exposed a crash in the update-cancel path (#5656), a crash loop after an offline period when the car reports an outdated timestamp (#5684), and the published state start time jumping backwards after charging, updating or driving (#5693) — all fixed in this release — and it surfaced four findings for a later fix: a charge sample without charger power is stored as 0 kW (#5699),
 a stream going inactive while logging is suspended can start a second fetch alongside the running poll (#5714), and two dead clauses in the model mapping and the stale-frame check (#5716, #5718).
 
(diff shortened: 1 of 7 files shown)


JakobLichterfeld (commented):
Verified at the tree in a second session. The power-usage guard is a one-field twin of `suspend_idle` — `drive_state.power` 0 → 5 in the poll that would otherwise suspend, plus the await the guard form needs — and the golden shows the guard doing exactly what the seven guards of #5695 do: no `suspended`, one position instead of two. The two service-mode fixtures pin what #5289 intended, logging only: the vehicle suspends in service mode like out of it (`online, suspended, asleep`, `service_mode` `false, true, false`), and service mode does not lift the unlocked block under `req_not_unlocked` (one-field twin of `suspend_idle_unlocked_required`, same golden shape). The snapshot for the leave payload follows the parked-probe rule from #5687.

Coverage of `vehicle.ex` by the suite alone: 89.7 % → 90.5 %, all six baseline lines of the family hit.

---

PR #5723, opened 2026-09-09
test(characterization): pin generic API errors, the asleep/offline transitions and the polling doubling after resume_logging

JakobLichterfeld:
## What this adds

Coverage family 6 "errors and sleep": six fixtures, all built from existing ingredients (`{"error": "timeout"}`, `suspend_logging`/`resume_logging` calls, `await`).

- `driving_api_error`, `charging_api_error`: a timeout reaches the generic `{:error, reason}` clause of the fetch-result handler in its driving and charging branches (a `vehicle_unavailable` is swallowed by the reachable fetch's fallback and never gets there); the drive resp. charging process continues across the error, the fuse is not melted.
- `suspend_logging_api_error`: the strict fetch of a manual suspend fails, `fetch_strict/2` passes the error through and the call is rejected with it.
- `vehicle_asleep_to_offline`, `vehicle_offline_to_asleep`: the asleep→offline and offline→asleep clauses route through `:start`; the new state row appears on the following probe.
- `vehicle_resume_asleep_stays_asleep`: `resume_logging` from asleep sets a one-second interval; the next asleep payload hits the doubling clause, which writes and publishes nothing.

## Enforced invariants

Unchanged harness; every fixture goes through the existing checks (causal barrier, vacuity probe, `two_call_cycle?/2`, calls section, interactions section, `build_events/1` terminal rules).

## Verification

- `MIX_ENV=test mix test test/teslamate/characterization_test.exs`: 189 passed, seed 421564.
- `MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`: 189 passed, seed 473576; vehicle.ex 90.5 % → 91.3 %, open lines 72 → 66.
- Baseline lines hit (#5652 list): 588, 589, 1496, 1501, 1505, 1661.
- Fixtures 143 → 149 (changelog intro updated).

None. Open questions: none.

---


JakobLichterfeld (commented):
Reviewed against the code and the staged goldens in a second session.

**Scope.** Six fixtures, all from existing ingredients (`{"error": "timeout"}`, `suspend_logging` / `resume_logging` calls, `await`); harness, mocks and `lib/` untouched. Each scenario is an API sequence a car can produce.

**Why `timeout`, and why the connection-loss fixtures never got here.** From `{:driving, _, _}` and `{:charging, _}` the vehicle fetches with the reachable assumption: a `vehicle_unavailable` on the strict fetch is caught by the `with` and replaced by the probe (`fetch_with_reachable_assumption/2`), so the existing connection-loss fixtures hand the handler an offline/unknown payload, never the error. A `timeout` falls through to the generic `{:error, reason}` clause and its driving/charging interval branches, and it is one of the two reasons that do not melt the `api_error` fuse — the fixtures pin the error handling alone, `healthy` stays `true`.

**Golden checks.**
- `driving_api_error` / `charging_api_error`: database and MQTT equal the connection-loss neighbours, as the descriptions state — the neighbours write nothing in their gap either (driving goes `{:unavailable, 0}` without a row, unknown state only logs). Same output from a different input is the pin: a generic error leaves drive and charge untouched. `charging_api_error` keeps the interaction sequence with indices shifted by three (7/11 instead of 10/14).
- `suspend_logging_api_error`: `calls` = `{"suspend_logging": {"error": "timeout"}}`, one open online row; position and update row are the ordinary online start, identical to `suspend_logging_incomplete_data`. No await is right: nothing is written after the call.
- `vehicle_asleep_to_offline` / `vehicle_offline_to_asleep`: the second row starts at clock .003, i.e. at the terminal's first serve — the transition clause routes to `:start` with a scheduled fetch and no row, the `:start` handler writes on the next probe, and the barrier's repeat proves it. No await, correctly unlike `vehicle_unavailable_fallback_asleep`, where the fallback probe consumes the terminal's first serve.
- `vehicle_resume_asleep_stays_asleep`: `resume_logging` is the only entry into `{state, interval}` with an interval other than `asleep_interval/0` (`{state, 1}`; `POLLING_ASLEEP_INTERVAL` is config surface). The doubling clause writes and publishes nothing, so the repeat is neutral; pinned are the `ok` reply, the single asleep row and the absence of a stream connect.

**Coverage.** The six baseline lines of the #5652 list (588, 589, 1496, 1501, 1505, 1661) map to today's 594/595, 1512, 1517, 1521 and 1677/1678 via the #5706 hunks; the measurement in the PR body is the executor's run, not re-measured here. The remaining red line in `fetch_strict/2` is the `too_many_request` branch, which belongs to the additive-vocabulary family.

Changelog: family line and intro count (143 → 149) present. No open points.

---

PR #5727, opened 2026-09-09
test(characterization): pin settings toggles and let the call seam serve stream connect and disconnect

JakobLichterfeld:
## What this adds

Coverage family 7 "settings toggles": four fixtures on the settings handler, built from `update_car_settings` calls, `clock` and `await`. `lib/` untouched.

- `charging_settings_streaming_off`: streaming turned off mid-charge. Nothing happens at the call (the stream is already down while charging) and the charge end does not reconnect. Database and MQTT equal `charging_full_cycle`.
- `vehicle_settings_streaming_off`: streaming turned off while parked with a connected stream. The handler disconnects it.
- `vehicle_settings_req_not_unlocked_streaming`: a non-streaming settings change with a connected stream. The stream stays, and the new guard refuses the suspend on the next poll.
- `vehicle_settings_streaming_on`: streaming turned on while parked without a stream. The handler connects it.

Two harness defects surfaced while recording and are fixed here. Both were latent: no earlier fixture toggled streaming through a call.

1. The call seam (`ApiMock.await_call_outcome/4`) deadlocked on the vehicle's synchronous stream connect; the vehicle died on the 5 s guard. The seam now answers `{:stream, vid, receiver}` while a call is pending.
2. The seam attributed a disconnect cast issued during a call to the serve it went on to deliver, against the section's definition (index of the API event served last). It now records the cast while the call is pending, so it carries the serve before the call. The first commit of this PR still holds that wrong golden for `vehicle_settings_streaming_off`; the second corrects it.

The regular `handle_call({:stream, …})` and `handle_info` were moved onto the shared helpers `connect_stream/3` and `disconnect_stream/1`. Every pre-existing golden is unchanged. Selftest `stream_toggle` covers both new clauses. Moduledoc updated in the call paragraph.

`update_car_settings` needs one field only: the changeset casts over the existing row. `vehicle_settings_change` passes `use_streaming_api: false` because the converted ExUnit test sent the full struct (`vehicle_test.exs:193-197`).

## Enforced invariants

- Interactions of a call delivery carry the serve before the call.
- Unchanged: causal barrier, vacuity probe, `two_call_cycle?/2`, calls and interactions sections, `build_events/1` terminal rules.

## Verification

- `MIX_ENV=test mix test test/teslamate/characterization_test.exs`: 194 passed, seed 981298.
- `MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`: 194 passed, seed 641634; vehicle.ex 91.3 % → 92.0 %, open lines 66 → 61.
- Baseline lines hit (#5652 list): 883, 891, 896, 900–901.
- Fixtures 149 → 153 (changelog intro updated); selftests 14 → 15.

None. Open questions: none.

---


JakobLichterfeld (commented):
Reviewed against the code and the goldens in a second session.

**Fixtures.** All four match the derivation. `charging_settings_streaming_off` equals `charging_full_cycle` outside `car_settings`, `calls` and `interactions`; the missing reconnect after the charge is real, the `:start` online handler connects only under `use_streaming_api: true`. `vehicle_settings_req_not_unlocked_streaming` pins behaviour, not just a line: the interactions section is exhaustive, so a reconnect on every settings change would fail it, and the new guard is visible on the next poll.

**Harness fixes.** Both defects verified at the code. The deadlock: `connect_stream/1` is a `GenServer.call` into the mock while the mock is parked in `await_call_outcome/4`, which answered only fetch calls. The misattribution: `Stream.disconnect/1` is a plain `send`, the mock processed it only after serving the next event, so `last_served` was already advanced. The fix is ordering-safe: the proxy sends the settings struct and then `:sys.get_state/1` from the same process, the vehicle handles both in order, so the cast is in the mailbox before the proxy's reply and the selective receive takes it first. Helpers shared with the regular handlers, pre-existing goldens unchanged, selftest `stream_toggle` shows both clauses.

**Review miss on my side.** I approved the fixture-4 derivation without holding the synchronous connect against the seam's receive clauses; the executor found the deadlock when recording.

**Coverage.** Baseline lines 883, 891, 896, 900–901 (today 888, 896, 901, 905–906). The measurement in the PR body is the executor's run.

Changelog: family line and intro count 149 → 153. No open points.

---

PR #5729, opened 2026-09-12
fix(grafana): count asleep/offline states that cross a parking boundary in the vampire drain standby time

rewse:
## Summary
- clip asleep and offline states to each parking interval
- aggregate both standby states in a single query
- calculate standby time and parking duration from elapsed seconds

## Problem
The parking window and the vehicle-state rows are independently dated timelines. The dashboard constructs a parking window from activity boundaries: it starts at the end of the preceding drive or charging process and ends at the start of the next one.

A state and an activity can receive different timestamps even when the same vehicle payload causes both transitions. `date_opts/2` in `vehicle.ex` closes the `asleep` or `offline` state row using `drive_state.timestamp`. When that payload starts charging, `insert_charge/3` dates the first charge row with `charge_state.timestamp`, and that row determines the charging process `start_date`. These sub-state timestamps can differ by a few milliseconds, so `state.end_date > charging.start_date` is expected. A common example is a car sleeping at a charger and waking for scheduled charging.

This is therefore a query bug rather than a logging bug. Standby is an interval-overlap question, but the previous full-containment predicates assumed nesting that the database does not guarantee. They discarded an entire state whenever either state boundary crossed the parking boundary, even though most or all of the state overlapped the parking period. In one anonymized production interval, 68,314 of 69,445 parking seconds were offline, but the discarded boundary-crossing state caused the dashboard to display 0% standby.

```text
Parking:       P0 |------------------------------| P1
Offline:   S0 |--------------------------------------| S1
Overlap:       P0 |------------------------------| P1
```

## Implementation
The query treats parking and state intervals as half-open ranges, `[start, end)`. It selects overlapping asleep and offline states in one LATERAL subquery:

```sql
state IN ('asleep', 'offline')
AND s.start_date < v.end_date
AND (s.end_date IS NULL OR s.end_date > v.start_date)
```

The strict inequalities exclude intervals that only touch at a boundary. Each selected state is clipped to the parking interval before its duration is summed:

```sql
LEAST(s.end_date, v.end_date)
- GREATEST(s.start_date, v.start_date)
```

PostgreSQL's `LEAST` ignores NULL arguments, so an open state with a NULL `end_date` is clipped to `v.end_date`. A NULL `start_date` does not satisfy the overlap predicate and is not aggregated.

The parking-duration expression also replaces `age(end, start)` with direct timestamp subtraction. PostgreSQL `age()` produces a symbolic calendar interval that can contain month fields; converting that result to epoch seconds can make month-spanning durations depend on symbolic calendar representation. Subtracting the endpoints produces the elapsed interval used by the overlap calculation, so both the standby numerator and parking-duration denominator use the same elapsed-seconds basis.

### Repository checks
- `python3 -m json.tool grafana/dashboards/vampire-drain.json` and `git diff --check` passed.
- Structural assertions passed: one LATERAL standby aggregation, one `state IN ('asleep', 'offline')` predicate, one clipping expression, one open-state overlap predicate, and no previous `age()` duration or full-containment predicate.
- `mix ci` completed with all 746 tests passing on Elixir 1.20.4, Erlang/OTP 29.0.6, and PostgreSQL 18.6. An initial run had one unrelated 200 ms `assert_receive` timeout; that test and the full suite passed on rerun.
- `grafana/dashboards/*.json` is excluded by `treefmt.toml`. Full `treefmt --fail-on-change` and `nix run .#lint -- --fail-on-change` runs reached only pre-existing formatting differences in 16 migration files; those generated changes were discarded.

### External validation (not part of this repository or its CI)
- Eight synthetic cases ran on PostgreSQL 18.6: contained, crossing the start, crossing the end, crossing both boundaries, NULL end, a 31-day interval, combined asleep/offline rows in the single LATERAL, and boundary contact without overlap. All returned the expected standby ratio.
- The final query and the pre-review query returned numerically identical standby values for all 16 rows in a read-only 90-day production query. Every value satisfied `0 <= standby <= 1`; the known boundary-crossing interval returned `0.983713`, and the known true-online interval remained `0.003658`.
- Three `EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)` runs per query gave median execution times of `1.609 ms` before and `1.396 ms` after the review cleanup. Median shared hit-plus-read blocks decreased from `399` to `383`, and the filtered sequential scans of `states` decreased from two to one per parking row. No unfiltered scan was added.
- Grafana 13.2.1 provisioned the dashboard as a Classic model with `rawSql` exactly matching the source. **Export as code** showed **Classic** selected under **Advanced options**; the downloaded JSON parsed successfully and retained the exact query.
- The final JSON was also mounted read-only into `teslamate/grafana:edge` running Grafana 13.2.2. Its API returned the same Classic model and query. A browser check against synthetic data showed a boundary-crossing 24-hour interval as `100%` and a six-minute standby interval over 24 hours as `0%` after integer display rounding. Both detail links were present with the expected interval parameters, and the page, browser console, and relevant Grafana logs contained no errors.

---

🤖 Assisted by GPT-5.6-sol (OpenAI) via Kiro CLI (planning, implementation, testing, and PR description).


JakobLichterfeld (changes requested):
Thanks for your first contribution, Tats — and for the thorough analysis behind this minimal diff. I verified the premise against the state machine, and it holds, with a more specific mechanism than the description gives:

- State rows and activities are two independently dated timelines. When the car wakes up, the same payload closes the `asleep`/`offline` row with `drive_state.timestamp`, while a charging process that starts in that payload ends up with `start_date = charge_state.timestamp` (first charge row). Those sub-state timestamps differ by a few ms, so `state.end_date > charging.start_date` is *expected* — the classic "car sleeps at the charger, scheduled charging wakes it" case, which matches your 68,314/69,445 s interval.
- That makes this a query bug, not a logging bug: standby is an interval-overlap question, and the containment predicate assumed a nesting the database never guaranteed. Fixing it in the query is the right depth and also repairs existing data.
- Only one open row per car exists (partial unique index) and row boundaries are gapless, so clipping can't double-count. The `age()` → subtraction change is a real correction (`EXTRACT(EPOCH …)` values months at 30 days).

The overlap + clipping approach is correct. A few cleanups before merging:

1. **Merge the two LATERALs**: they are identical except for the state literal and are only ever summed. One subquery with `state IN ('asleep', 'offline')` keeps the clipping expression in a single place and halves the `states` scans per parking row.
2. **Drop the redundant `COALESCE`** inside `LEAST(...)`: PostgreSQL's `LEAST` ignores NULLs, so `LEAST(s.end_date, v.end_date)` already yields `v.end_date` for an open state. In the `WHERE`, `(s.end_date IS NULL OR s.end_date > v.start_date)` says what the description explains.
3. **Commit message / description**: please name the actual cause (state rows and activities dated from different sub-state timestamps of the same payload, see `date_opts/2` and `insert_charge/3` in `vehicle.ex`) instead of the generic "observed a little before or after". Otherwise someone later "fixes" the states writer to re-align them and reintroduces the 0 % standby.
4. The validation you list (synthetic cases, production intervals, patcher unit tests) lives outside this repo — fine under our dashboard-test policy, but please say so explicitly so readers don't assume those tests run in CI.

I'll take care of the CHANGELOG entry.

---

rewse:
Thanks for the detailed review. I addressed all four points:

- combined the asleep and offline aggregations into one LATERAL using `state IN ('asleep', 'offline')`
- removed the redundant `COALESCE` from `LEAST` and made the open-state condition explicit
- rewrote the commit message and description to identify the `drive_state.timestamp` / `charge_state.timestamp` difference from the same payload, with references to `date_opts/2` and `insert_charge/3`
- separated external validation from repository and CI checks in the description

I reran the final query against the read-only 90-day production data. Its standby values matched the pre-review query for all 16 rows. Three `EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)` runs reduced the median execution time from `1.609 ms` to `1.396 ms`, shared blocks from `399` to `383`, and filtered `states` sequential scans from two to one. The final JSON also passed a Classic export round trip and a browser check on `teslamate/grafana:edge` (Grafana 13.2.2).

Ready for another review. Thank you!

---

🤖 Assisted by GPT-5.6-sol (OpenAI) via Kiro CLI (review response drafting and validation).

JakobLichterfeld (approved):
Looks good, thanks for the quick and precise follow-up — single LATERAL, explicit open-state predicate and an accurate commit message. Approving; I'll merge once linting is green and add the CHANGELOG entry.

---

Issue #5420, opened 2026-06-17
No data retrieved


djpalmis:
### Is there an existing issue for this?

- [x] I have checked the existing issues and discussions, and I can confirm that there are no duplicates.

### What happened?

Hi during the night Teslamate (latest version) logged the first charge, logged correctly the offline state but then it stopped logging. 
I rebooted VM but I see some strange errors.

Connectivity should be ok because i can correctly see the status on Tesla official app.

### Expected Behavior

Just teslamate logging correctly 

### Can the issue reliably be reproduced?

Yes

### Steps To Reproduce the issue

Just start teslamate

### Relevant log output

```shell
teslamate-1  | 2026-06-16 23:01:40.786 car_id=1 [info] Start / :online

teslamate-1  | 2026-06-16 23:02:10.441 car_id=1 [info] Online / Charging detected: -1 kW

teslamate-1  | 2026-06-16 23:02:11.008 [info] GET https://nominatim.openstreetmap.org/reverse -> 200 (363.219 ms)

teslamate-1  | 2026-06-16 23:02:11.015 car_id=1 [info] Charging / SOC: 33% / 🏡 Casa

teslamate-1  | 2026-06-16 23:02:11.015 car_id=1 [info] Stream disconnecting ...

teslamate-1  | 2026-06-16 23:09:43.528 [info] Refreshing access token ...

teslamate-1  | 2026-06-16 23:09:43.894 [info] POST https://auth.tesla.com/oauth2/v3/token -> 200 (366.392 ms)

teslamate-1  | 2026-06-16 23:09:43.897 [info] Scheduling token refresh in 6 h

teslamate-1  | 2026-06-16 23:24:09.901 car_id=1 [info] Derived efficiency factor: 152.23999999999998 Wh/km (126x confirmed)

teslamate-1  | 2026-06-16 23:24:09.901 car_id=1 [info] Charging / Stopped / 1.82 kWh – 22 min

teslamate-1  | 2026-06-16 23:24:09.906 car_id=1 [info] Start / :online

teslamate-1  | 2026-06-16 23:24:09.909 car_id=1 [info] Stream connecting ...

teslamate-1  | 2026-06-16 23:27:11.153 car_id=1 [info] Suspending logging

teslamate-1  | 2026-06-16 23:34:41.903 [info] Streaming API: Vehicle offline

[...]

Issue #2136, opened 2021-11-09
Intermittent "Internal Server Error"


leewillis77:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

With increasing frequency, when I try and access Teslamate though the web interface, I get an Internal Server Error message instead of the homepage. It typically comes back after a minute or two. This is running on a raspberry pi under docker with nothing else on the machine so no other services should be interfering. The Pi looks healthy (not maxing out CPU, or Memory or anything similar). 

The logs (below) don't seem that useful to me, so if there is some other debug I can provide, please do let me know.

### Expected Behavior

Consistently available

### Steps To Reproduce

Attempt to load main teslamate page (http://teslamate.ip.address:4000).

### Relevant log output

```shell
teslamate_1  | 2021-11-09 10:58:40.819 [error] #PID<0.18726.1> running TeslaMateWeb.Endpoint (connection #PID<0.18725.1>, stream id 1) terminated
teslamate_1  | Server: teslamate:4000 (http)
teslamate_1  | Request: GET /
teslamate_1  | ** (exit) exited in: Task.Supervised.stream(2500)
teslamate_1  |     ** (EXIT) time out
teslamate_1  | 2021-11-09 10:58:42.969 [info] GET /
teslamate_1  | 2021-11-09 10:58:45.476 [error] #PID<0.18735.1> running TeslaMateWeb.Endpoint (connection #PID<0.18725.1>, stream id 3) terminated
teslamate_1  | Server: teslamate:4000 (http)
teslamate_1  | Request: GET /
teslamate_1  | ** (exit) exited in: Task.Supervised.stream(2500)
teslamate_1  |     ** (EXIT) time out
```

### Type of installation

Docker

### Version

v1.24.2

drewpost:
I am experiencing this error as well however it is only occurring when I load it from my laptop (macOS 12.0.1 in Safari 15.1 and Chrome 95.0). If I load it in Safari from my iPhone or iPad, it loads fine. 

I can go directly to Grafana via the local IP + port and it loads and shows current data.

drewpost:
Update: it seems that 1.25.0 resolves the issue (sorry I had missed that release earlier today).

leewillis77:
Closing this since I've not seen the issue on 1.25.0 either.

leewillis77:
Re-opened as I've just had this again (on 1.25.0):

```
[...]

PR #5735, opened 2026-09-15
fix(vehicles): skip unresponsive vehicles in the summary list instead of crashing the home page

yourDomainAdmin:
The vehicle list currently terminates its caller when any vehicle process takes longer than five seconds to answer `Vehicle.summary/1`. On the home page, that turns one busy or stuck vehicle worker into an HTTP 500 (`Task.Supervised.stream(5000)`), even though the rest of the application and database remain healthy.

Configure `Task.async_stream/3` to kill only the timed-out summary task, preserve the input metadata on exit, log the affected child, and omit only that unavailable summary. Add a regression test that suspends a vehicle worker and confirms the list returns normally after the timeout.

This is the failure mode reported in #5420 and follows up on the timeout handling discussed in #2136.

Validation:
- `MIX_ENV=test mix test test/teslamate/vehicles_test.exs` (6 passed)
- `MIX_ENV=test mix ci` (747 passed)

🤖 Assisted by GPT-6 (OpenAI) via Codex (source analysis, implementation, tests, and PR drafting).


JakobLichterfeld (changes requested):
Thanks Alexey for tackling this, the 500 on the home page has been annoying people since #2136. The direction (let one stuck vehicle degrade instead of taking the whole request down) is right, but two things block merging as is, and a few smaller points follow.

### Blocking

**1. The two 5-second timers race, and the losing branch still crashes the caller.**
`Vehicle.summary/1` calls `GenStateMachine.call(pid, :summary)` with the default 5000 ms timeout, which is the same value as the stream timeout. If the call timeout fires first, the task exits with `{:timeout, {:gen_statem, :call, …}}`. That is a linked exit, not a stream timeout, so `on_timeout: :kill_task` never engages and the LiveView still dies. I reproduced this on Elixir 1.20:

```
** (EXIT from #PID<0.94.0>) exited in: GenServer.call(#PID<0.95.0>, :x, 50)
    ** (EXIT) time out
```

Under load (which is exactly the reported scenario) the order of the two timers is not guaranteed. The stream should own the deadline: call `summary` with `:infinity` (or a timeout strictly larger than the stream timeout) inside the task. A test that suspends the vehicle with the call timeout shorter than the stream timeout would pin this.

**2. `Mqtt.PubSub.init/1` also uses `Vehicles.list/0`, and a lossy list is wrong there.**
If a vehicle process is busy for more than five seconds while MQTT starts or restarts, `init/1` now succeeds with that car missing: no `VehicleSubscriber` is started for it, so it publishes nothing until the next restart. With `discovery: true` it gets worse: `clear_removed_vehicles/2` treats the missing car as removed and clears its retained Home Assistant device config, so the entities disappear. Before this PR the init crashed and the supervisor retried, which is the better outcome there. PubSub needs a strict variant (raise, or an `{:ok, list} | {:error, …}` shape it can act on), and only the home page should get the lossy list.

### Should change

- **The test does not prove the claim.** With a single vehicle and `assert [] = …` it also passes if every summary is dropped as soon as one times out. Please start two vehicles, suspend one, and assert the other summary is still returned.
- **Log with `car_id:` metadata**, like every other vehicle log line, instead of inspecting the supervisor child id (`TeslaMate.Vehicles.Vehicle_1`). People filter logs by `car_id=`, and the child id leaks an internal naming scheme.
- **`list/1` gets a `:timeout` option that only the test uses.** I would rather not widen the public API for a test seam. Keep the timeout a module attribute and inject a short one through a private helper, or at least mark the option `@doc false`.
- **Please add a changelog entry** under `[unreleased]` → Improvements and bug fixes, in the existing style (`fix(web): … (#5735 - @yourDomainAdmin)`).

### Framing

This handles the symptom, not the cause. The reason a vehicle cannot answer `:summary` for seconds is that the state machine does synchronous work (the `deps.log` DB writes, the `Fetch already in progress` storms in #2136). Degrading gracefully on the home page is still worth having, but please do not present it as the fix for #5420. Say in the description that the wedged vehicle process is the open root cause, and keep the issue open.

Happy to re-review once the two blocking points are addressed.

---

PR #5737, opened 2026-09-16
test(characterization): pin a drive in import mode without address lookup

JakobLichterfeld:
## What this adds

Coverage family 9 "import mode with an online payload": one fixture.

- `driving_import`: twin of `driving_full_drive` with `import: true`, `expect_halt: true` and `import_complete` as the halt terminal. The drive is started and closed with `lookup_address: false`, so the drives row has no `start_address_id`/`end_address_id` and the `addresses` table stays empty; positions, states and topics equal the neighbour. The drive closes in the handler of the parked payload before the halt terminal is fetched, so every row is committed when the halt is read; `expect_halt` excludes `await`. Settings as the neighbour: the import API has no stream (`Import.FakeApi.stream/3` returns `{:ok, nil}`) and `Import` sets `use_streaming_api: false`.
- Two import-mode lines are hit but not visible; both go to the #5652 line "not reachable at the boundary": baseline 1042 updates the car with `preload: []` to keep the in-memory import settings (`Import.import_car/1`) from being reloaded — in the replay the in-memory settings and the `car_settings` row are identical, so there is no boundary effect; baseline 2144 is the import fetch timeout of 0, timing only — the poll parks in the mock at once either way and the serve boundary is unchanged.

## Enforced invariants

Unchanged harness; the fixture goes through the existing checks (halt read via `assert_halt!/3`: task cleared, no `{:state_timeout, :fetch}` pending, terminal served once; `build_events/1` terminal rules).

## Verification

- `MIX_ENV=test mix test test/teslamate/characterization_test.exs`: 196 passed, seed 212355.
- `MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`: 196 passed, seed 588968; vehicle.ex 92.0 % → 92.2 %, open lines 61 → 59.
- Baseline lines hit (#5652 list): 1042, 2144.
- Fixtures 154 → 155 (changelog intro updated).

None. Open questions: none.

---


JakobLichterfeld (commented):
Reviewed against the code and the goldens from a second session.

**Fixture.** `driving_import` is `driving_full_drive` with `import: true`, the same five payloads and `import_complete` as the halt terminal; events and car are identical to the neighbor. The golden differs in exactly two places: the `addresses` table is empty, and the drive has no `start_address_id`/`end_address_id` — the vehicle starts and closes drives with `lookup_address: !data.import?`, which is the import mode's one visible effect on this path. Positions, states, the drive's distance and duration, and every topic equal the neighbor.

**Halt.** The drive closes inside the handler of the parked payload, before the fetch that receives `import_complete` even starts; the `{:error, :import_complete} when data.import?` clause keeps state and schedules nothing, so the halt read (task cleared, no `{:state_timeout, :fetch}` pending, terminal served once) sees every row committed. No `await`, which `expect_halt` excludes anyway.

**Settings.** Polling like the neighbor, and import-faithful: `Import.FakeApi.stream/3` returns `{:ok, nil}` and `Import.import_car/1` forces `use_streaming_api: false` in memory.

**The two baseline lines.** Both are hit and neither is pinnable at the boundary; the PR body says so, and they move to the #5652 classification line. Baseline 1042 (`preload: []` on the online start) exists to keep those in-memory import settings from being reloaded from the row — in a replay the in-memory settings and the `car_settings` row are always identical, so nothing can differ. Baseline 2144 (fetch timeout 0 in import mode) is timing: the poll parks in the mock at once either way, the serve boundary does not depend on the timer value. Coverage 92.0 % → 92.2 % is the executor's run.

Changelog: family line and intro count 154 → 155. No open points.

PR #5740, opened 2026-09-17
test(characterization): add too_many_request, seed.updates and Home Assistant discovery to the harness vocabulary

JakobLichterfeld:
## What this adds

Coverage family 10 "additive vocabulary": three harness extensions, each with a self-test and first users. `lib/` untouched.

Harness (four places in `test/support/characterization.ex`):
- `{"error": {"too_many_request": <seconds>}}` → `{:error, :too_many_request, s}` (the form `TeslaMate.Api` returns): `build_events/1` decodes it, the terminal rule refuses it (its backoff keeps running under repeats, and it is no halt), `canonical_call/1` pins the suspend rejection `{:error, {:too_many_request, s}}` in the scenario's own form. Serve counting is unchanged.
- `seed.updates`: update rows inserted before the vehicle starts through the production `Log.Update` changeset (`Log` has no whole-row insert); fields checked against the schema like `seed.positions`, shared `seed_row!/4`.
- `mqtt.discovery`: the `VehicleSubscriber` is started per replay with the flag and `migration_delay: 0`; the device config is pinned as decoded JSON under `homeassistant/device/teslamate_$car/config`.
- `mask_car_id/2`: the car id is masked in payloads as in topics (state topics, unique ids, device identifier, fallback device name).

Self-tests: `too_many_request`, `seed_updates`, `discovery`.

First users:
- `vehicle_health_too_many_request` (twin of `vehicle_health_api_errors`): four rate-limit responses back off without melting the fuse — healthy stays true; everything else equals the neighbour. `suspend_logging_too_many_request` (twin of `suspend_logging_api_error`): the strict fetch is rate-limited, the calls section pins `{"error": {"too_many_request": 900}}`.
- `vehicle_start_asleep_restored_version` (twin of `vehicle_start_asleep` with seeded position and update rows): `restore_last_known_values/2` reads the car version from the update row — version topic `2026.20.1` without any payload carrying it.
- `discovery_model3_74`, `discovery_cybertruck`, `discovery_car_type_nil`, `discovery_trim_nil`, `discovery_details`, `discovery_no_display_name` (twins of `identity_*` payloads with discovery on): device model "Model 3 LR (Deep Blue, Pinwheel 18" Wheels)" (format_model prefix, marketing name over trim, camel-split color, wheel regex, spoiler "None" dropped), "Cybertruck AWD (…)" (format_model else clause, trim without marketing name), "Tesla" (model nil discards trim and details), "Model 3 (…)" (no trim, no marketing name), "Model 3 LR (Deep Blue, Stiletto Wheels, Passive Spoiler, Sunroof)" (wheel regex fallback, spoiler kept, `sun_roof_installed` 1 — the API counts sunroofs), and the device name fallback "Tesla #$car" when display_name is null (identify/1 clears the car name). Each diff against its neighbour is the device config topic only (plus the changed identity fields for `discovery_details`/`discovery_no_display_name`). The entity block stays in the goldens.

Not representable, going to the #5652 classification: the discovery retry path (needs a failing publisher — `MqttPublisherMock.responses` is not scenario vocabulary) and `configuration_url` (no `discovery_base_url` in the replay).

## Enforced invariants

- Payload masking: a golden never carries the car id — topics and payloads are masked alike (`mask_car_id/2`, `drain_mqtt/2`).
- A too_many_request event is never a terminal: neither repeat-neutral nor a halt (`build_events/1`).
- Unchanged: causal barrier, vacuity probe, `two_call_cycle?/2`, calls and interactions sections, seed and settings-call checks.

## Verification

- `MIX_ENV=test mix test test/teslamate/characterization_test.exs`: 208 passed, seed 82135.
- `MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`: 208 passed, seed 511395; vehicle.ex 92.2 % → 93.0 %, open lines 59 → 53.
- Baseline lines hit (#5652 list): 105, 106, 575, 577, 1556, 1658.
- Fixtures 155 → 164 (changelog intro updated); self-tests 15 → 18.

None. Open questions: none.

---


JakobLichterfeld (commented):
Reviewed from a second session against the code and the goldens; the harness diff read in full.

**Harness (four places, all additive).** The `too_many_request` form decodes to the 3-tuple `TeslaMate.Api` returns, any other map under `"error"` raises with a name, the terminal rule refuses the 3-tuple (its backoff keeps running under repeats, and it is no halt), and `canonical_call/1` pins the suspend rejection in the scenario's own form. `seed.updates` shares `seed_row!/4` with `seed.positions` and goes through the production `Log.Update` changeset, as `Log` has no whole-row insert. `mqtt.discovery` is a start option of the per-replay `VehicleSubscriber` plus `migration_delay: 0`, needed because `HomeAssistant.migrate/3` sleeps that long before the device config. `mask_car_id/2` masks the car id in payloads as in topics, with a digit boundary so car 1 never matches car 10; no golden in the PR carries an unmasked id.

**First users.** `vehicle_health_too_many_request`: four rate-limit responses leave the fuse alone, `healthy` stays `true` where the `unknown`-error twin publishes `false`; nothing else differs. `suspend_logging_too_many_request`: the rate-limited strict fetch is passed through `fetch_strict/2` and pinned as `{"error": {"too_many_request": 900}}`. `vehicle_start_asleep_restored_version`: `restore_last_known_values/2` reads the car version from the seeded update row, the version topic shows `2026.20.1` without any payload carrying it. The six `discovery_*` fixtures each pin one branch of the device name: `format_model/1` prefix vs. else clause, marketing name over trim, model nil discarding trim and details, neither trim nor marketing name, the wheel-regex fallback with a kept spoiler and a sunroof (`sun_roof_installed` is a count in the API, verified in `Summary`), and the `"Tesla #$car"` fallback when `display_name` is null (`identify/1` takes the car name from it). Each diff against its `identity_*` neighbour is the device config topic only.

**Baseline lines.** 105, 106, 575, 577, 1556, 1658 of the #5652 list; the measurement in the PR body is the executor's run. Not representable and going to the classification line: the discovery retry path and `configuration_url`.

Changelog: family line names the three extensions, intro count 155 → 164. No open points.

---

PR #5741, opened 2026-09-17
feat(webview): make the vehicle display order editable on the settings page

### LINKED ISSUE #5739: feat(webview): make the vehicle display order (display_priority) editable in the car settings
`display_priority` decides the vehicle order in the web UI (`Vehicles.list/0`) and in the car dropdown of every Grafana dashboard (`ORDER BY display_priority ASC, name ASC, vin ASC`). The column exists since #1904 (v1.24) but can only be set with a manual `UPDATE cars SET display_priority = …` in the database, and it is not documented.

This comes up whenever a user replaces a car: the new one is added as car 2 and every dashboard opens with the old one first.

**To do**

- Expose `display_priority` per car in the settings page, next to the existing car settings. Integer, lower first, tie-break stays name then VIN. UI and dashboards already read the column, nothing else to change.
- Document the setting.

Pointers: `lib/teslamate/log/car.ex` (schema), `lib/teslamate/vehicles.ex` (sort), `lib/teslamate_web/live/settings_live/index.html.heex` (car settings form).

---

wooter:
Closes #5739.

## What changes

- **Settings page:** Move left and Move right buttons under the car tabs. The buttons are disabled at the ends and hidden with one car.
- **Tabs:** the car tabs follow `display_priority` instead of the car id.
- **Renumbering:** `Log.move_car/2` swaps the car with its neighbour, then numbers every car 1..N in one transaction. Example: priorities 1, 1, 2, 3, 4, 5 become 2, 1, 3, 4, 5, 6 after moving the first car right. Cars keep their priorities until someone moves one. A move at either end writes nothing.
- **Web UI:** `Vehicles.list/0` reads the order from the database. Before, it sorted on the `Car` each vehicle process loaded at start, so a new order showed only after a restart.
- **Docs:** FAQ entry and CHANGELOG line. The new strings are extracted; translations are empty.

Order everywhere in TeslaMate is `{display_priority, id}`, as before. Grafana still breaks ties by `name, vin`; after the first move all priorities are distinct, so both agree.

## Tests

- New tests for `move_car/2`, the settings tabs and buttons, and `Vehicles.list/0` with a stale process struct. They fail on `main` and pass on this branch.
- Full suite: 767/768 pass. The one failure is `SuspendTest` "suspends when idling", which also fails intermittently on `main`.


JakobLichterfeld (commented):
Thanks Wouter. The mechanics are right: swap with the neighbor, renumber 1..N in one transaction, `FOR UPDATE` with the in-memory re-sort. Three changes before this can go in.

**1. The order is a ranking, not a tab position**

`display_priority` is consumed as a ranked list everywhere else: the car overview, every Grafana car dropdown. "Move left / Move right" under the tabs names the tab bar of this page, which wraps on narrow screens anyway, and it only acts on whichever car is selected. Please replace the buttons with a list of all cars in display order under General Settings, with "Move up / Move down" per row, independent of the selected tab. The tabs keep following that order.

**2. `car_order` duplicates `car_settings` and crashes the page**

`mount/3` takes `car_order` from every car in the database, but the template resolves each id with `Map.fetch!(@car_settings, car_id)`. A car created between the two queries (a `Vehicles` restart adding a car) raises `KeyError` and the Settings page is gone. Only the `move_car` handler filters, and that filter has the opposite problem: it hides a car that is first in DB order, so the visible first car still moves and swaps with a car the page does not show.

The root cause is a second representation of what `car_settings` already holds. Order `Settings.get_car_settings/0` by the joined car's `{display_priority, id}` (the issue names its `order_by: s.id` as the gap), reload `car_settings` after a successful move, and drop `car_order`, the `Map.fetch!` and the filter. One source for the list, the default car in `handle_params/3` and the button states.

**3. `Vehicles.list/0`**

Keep the change here. A reorder that only shows after a restart is not done, and #5710's reload is a user action, not a fix for this. Whichever PR merges second rebases; the conflict is local to `list/0`. Add a one-line comment at the sort that `Summary.car.display_priority` itself is stale by design.

Tie-break by `id` is correct, it is the only stable key. Grafana still uses `name, vin`; that is a pre-existing mismatch and gets its own issue.

Tests: add three cars with a move of the middle car in both directions, and the `{:error, :not_found}` path. Docs and changelog are fine, adjust the FAQ wording to the list.

---

JakobLichterfeld (changes requested):
Thanks Wouter. The mechanics are right: swap with the neighbor, renumber 1..N in one transaction, `FOR UPDATE` with the in-memory re-sort. Three changes before this can go in.

**1. The order is a ranking, not a tab position**

`display_priority` is consumed as a ranked list everywhere else: the car overview, every Grafana car dropdown. "Move left / Move right" under the tabs names the tab bar of this page, which wraps on narrow screens anyway, and it only acts on whichever car is selected. Please replace the buttons with a list of all cars in display order under General Settings, with "Move up / Move down" per row, independent of the selected tab. The tabs keep following that order.

**2. `car_order` duplicates `car_settings` and crashes the page**

`mount/3` takes `car_order` from every car in the database, but the template resolves each id with `Map.fetch!(@car_settings, car_id)`. A car created between the two queries (a `Vehicles` restart adding a car) raises `KeyError` and the Settings page is gone. Only the `move_car` handler filters, and that filter has the opposite problem: it hides a car that is first in DB order, so the visible first car still moves and swaps with a car the page does not show.

The root cause is a second representation of what `car_settings` already holds. Order `Settings.get_car_settings/0` by the joined car's `{display_priority, id}` (the issue names its `order_by: s.id` as the gap), reload `car_settings` after a successful move, and drop `car_order`, the `Map.fetch!` and the filter. One source for the list, the default car in `handle_params/3` and the button states.

**3. `Vehicles.list/0`**

Keep the change here. A reorder that only shows after a restart is not done, and #5710's reload is a user action, not a fix for this. Whichever PR merges second rebases; the conflict is local to `list/0`. Add a one-line comment at the sort that `Summary.car.display_priority` itself is stale by design.

Tie-break by `id` is correct, it is the only stable key. Grafana still uses `name, vin`; that is a pre-existing mismatch and gets its own issue.

Tests: add three cars with a move of the middle car in both directions, and the `{:error, :not_found}` path. Docs and changelog are fine, adjust the FAQ wording to the list.

---

wooter:
Thanks, all three are in 7572a568:

1. **Ranked list.** A Car Order list under General Settings, with Move up / Move down per row. It works for any car, not only the selected tab. The tabs follow the same order.
2. **One source.** `Settings.get_car_settings/0` orders by the car's `{display_priority, id}`, and `car_settings` is now an ordered list. It drives the tabs, the order list, the default car and the button states. It reloads after a move. `car_order`, the `Map.fetch!` and the filter are gone.
3. **`Vehicles.list/0`.** The DB-order sort stays, with the comment on why `Summary.car.display_priority` is stale.

Tests: middle of three cars moved in both directions, `{:error, :not_found}`, `get_car_settings/0` order, and a LiveView test that moves a car which is not the selected tab. Full suite 771/771; gettext and format checks pass. FAQ updated.

JakobLichterfeld (approved):
Thanks, this is exactly what I asked for. One source for the order, the ranking as a list, no state to keep in sync, and the tests cover the middle car and a move of a car that is not the selected tab.

Merging.

---

Issue #5742, opened 2026-09-17
Stream is not reconnected after vehicle_in_service ends a drive — the vehicle polls until the next drive


JakobLichterfeld:
## Stream is not reconnected after `vehicle_in_service` ends a drive

**Where:** `Vehicle.handle_event/4`, fetch result `{:error, :vehicle_in_service}` in `{:driving, _, drive}` (lib/teslamate/vehicles/vehicle.ex, "Vehicle is currently in service"): the drive is closed and `disconnect_stream/1` is called, but `stream_pid` stays set. Every other disconnecting path clears it.

**Effect:** the `:start` online handler connects the stream only for `stream_pid: nil` and keeps any existing pid. After the service visit the car comes back online with a dead pid: `streaming?/1` is false, the vehicle polls instead of streaming, and with streaming enabled that means polling intervals and the polling idle window.

**Why nobody noticed:** it heals itself. The next drive-to-online transition runs `maybe_reconnect_stream/1`, which sees the dead pid and reconnects. So the gap lasts from the end of the service visit until the first drive afterwards, and shows only as a car that polls instead of streams for that time.

**Since:** 2020 — the disconnect without clearing the pid came with "End drive when car is put into service mode" (79d70764, April 2020); the pid check in the start handler is from a6917849 (March 2020).

**Pinned:** `vehicle_in_service_online_return` (characterization fixture, #5652) — no `connect` interaction after the in-service disconnect. Found while giving the replay a mortal stream process (#5743): the old golden of `vehicle_in_service_ends_drive` recorded a second disconnect cast to the stale pid, which the immortal mock accepted and a real stream process never sees.

**Fix (Phase 2):** clear `stream_pid` in the in-service branch like the other disconnecting paths.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

Issue #5714, opened 2026-09-08
A stream `inactive` while suspended starts a second fetch without the in-progress guard


JakobLichterfeld:
`fetch_state` starts a second fetch without the in-progress guard

Found by the characterization suite (#5713). In `:suspended`, a stream `:inactive` triggers `fetch_state` (`lib/teslamate/vehicles/vehicle.ex`), which starts its own task without the "Fetch already in progress" guard that `:fetch` has. If a poll is in flight, two fetch results are in flight; with the same answer both orders converge, with different answers (the car's state changing inside the window) the outcome depends on message order. Rare and benign in practice, but it breaks the one-fetch-at-a-time invariant the rest of the machine relies on.

Direction: route `fetch_state` through `:fetch` (from `:suspended` the fetch is a bare probe anyway) and drop its two result clauses. Phase 2, after #5259.

---

🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 high) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)

PR #5743, opened 2026-09-17
test(characterization): pin the reconnecting stream controls and serve their reconnect at the stream seam

JakobLichterfeld:
## What this adds

Coverage family 11 "stream controls": the reconnecting stream controls become scenario vocabulary, with the harness stream seam extended additively. `lib/` untouched.

Harness (`test/support/mocks/api.ex`, `test/support/characterization.ex`):
- Shared proxy loop `ApiMock.await_proxy/6` for the call seam and the stream seam: a stream delivery runs through a proxy task like a call, so a handler that calls back into the API (the reconnecting controls connect a new stream) is answered instead of deadlocking against the get_state sync (#5727's class). Only the call seam answers fetch calls; a fetch arriving during a stream delivery is a concurrent poll (the fetch_state task of a stream control) and waits for the parked caller — `streaming_suspend_inactive` compares unchanged.
- Connection stub per `:stream` answer (unlinked, monitored by the mock): it records the vehicle's disconnect cast synchronously and exits, so the reconnect handler's monitor gets a real `DOWN` at once; a `DOWN` of a superseded stub is discarded (the vehicle's `:stream` call may overtake it).
- `ApiMock.sync_stream/1` before `mark_served` and at a call's completion keeps the attribution provable: a disconnect cast before a serve is recorded before it.
- `stream_control!/1` accepts all four controls; moduledoc describes the seam.

Self-test `stream_reconnect`; first users `streaming_too_many_disconnects` and `streaming_tokens_expired` (twins of `streaming_vehicle_offline` in form): interactions `[connect 1, disconnect 1, connect 1]`, state stays online.

Finding #5742 (production, fix in phase 2): the in-service branch disconnects the stream but leaves `stream_pid` set; the `:start` online handler only connects on `nil`, so a car returning online after a service visit runs without its stream until the next drive (`maybe_reconnect_stream/1`). Pinned by `vehicle_in_service_online_return` (interactions without a connect after the return; description names the expected connect after the fix).

Re-recorded golden `vehicle_in_service_ends_drive`: it carried a second disconnect after the asleep terminal — the cast at the stale `stream_pid` (#5742) reached the immortal mock; with the stub the cast at the exited stream is lost as in production. The golden pinned a mock artefact, not vehicle behaviour.

## Enforced invariants

- No wall-clock wait at the stream seam: a reconnecting control is answered from the proxy loop and its `DOWN` comes from the exiting stub, never from the handler's one-second timeout (`ApiMock.await_proxy/6`, `ApiMock.connect_stream/3`).
- Interactions cast before a serve are recorded before it: the stub is synced at every serve boundary and at a call's completion (`ApiMock.sync_stream/1`).
- Unchanged: causal barrier, vacuity probe, `two_call_cycle?/2`, calls and interactions sections; every existing golden and self-test compares unchanged except the one re-recorded above.

## Verification

- `MIX_ENV=test mix test test/teslamate/characterization_test.exs`: 212 passed, seed 620638.
- `MIX_ENV=test mix coveralls.detail --filter lib/teslamate/vehicles/vehicle.ex test/teslamate/characterization_test.exs`: 212 passed, seed 287888; vehicle.ex 93.0 % → 93.9 %, open lines 53 → 46.
- Lines hit: the reconnect handler (today 832–845) except its one-second timeout branch (841), which the invariant above keeps unreachable.
- Fixtures 164 → 167 (changelog intro updated, #5742 added as the fifth finding); self-tests 18 → 19.

None. Open questions: none.

---


JakobLichterfeld (commented):
Reviewed in a second session against the code and the goldens; both mock and harness diffs read in full.

**Harness.** One proxy loop (`ApiMock.await_proxy/6`) now serves both seams. The distinction it draws is the right one: the strict fetch inside a call handler is part of the call and is answered from the loop, whereas a fetch arriving during a stream delivery is a concurrent poll (the `fetch_state` task of a stream control, #5714) and waits for the parked caller — that keeps the pre-existing serve order, and `streaming_suspend_inactive` proves it by comparing unchanged. The connection stub is the piece that makes the reconnect handler pinnable at all: the handler monitors the stream pid and waits up to a second for its `DOWN`, and only a process that actually exits after `:disconnect` — as a WebSockex process does — delivers that without a wall-clock wait. The stub reports the disconnect synchronously before exiting, and `sync_stream/1` at every serve boundary and at a call's completion keeps the attribution provable with the same single-node enqueue argument as #5727. A `DOWN` of a superseded stub is discarded in both the loop and `handle_info/2`; without that a reconnect could kill the mock when the vehicle's `:stream` call overtakes the old stub's `DOWN`.

**Fixtures.** `streaming_too_many_disconnects` and `streaming_tokens_expired` pin disconnect and connect after the serve before the delivery, state unchanged; the self-test `stream_reconnect` shows both controls in one replay with the indices moving from 1 to 2.

**Finding #5742.** The stub turned a mock artefact into a production truth: `vehicle_in_service_ends_drive` used to record a second disconnect after the asleep terminal, a cast at the stale `stream_pid` that only an immortal mock could accept. Verified at the code: the in-service branch disconnects without clearing `stream_pid` (since 79d70764, 2020), and the `:start` online handler connects only on `nil`, so after a service visit the car runs without its stream until `maybe_reconnect_stream/1` on the next drive. `vehicle_in_service_online_return` pins today's behaviour (no connect after the return); its description names the connect expected after the fix. Re-recording the old golden is correct: it pinned the mock, not the vehicle.

**Coverage.** 93.0 % → 93.9 % per the executor's run; the handler's one-second timeout branch stays red by construction of the invariant.

Changelog: family line and intro (164 → 167, #5742 as the fifth finding). No open points.

---

PR #5759, opened 2026-09-19
chore(ci): fix the shellcheck and untrusted-input findings from actionlint

JakobLichterfeld:
`actionlint` over `.github/` reported 13 findings. This fixes the 12 that are shell hygiene; no workflow changes behavior.

- `check_if_pr_from_outside_repo.yml`, `ghcr_purge.yml`: quote `$GITHUB_OUTPUT`.
- `ghcr_purge.yml`: `github.head_ref` reaches the normalize step through `env` instead of being interpolated into the script. The job is already gated on `is_pr_from_outside_repo == 'false'`, so this is hygiene, not a fix for an exploitable path.
- `cleanup_caches.yml`, `cleanup_largest_caches.yml`: quote `$REPO`/`$BRANCH`; iterate the cache keys with `read` instead of unquoted word splitting, skipping the empty line an empty list produces so an empty result still issues no `gh cache delete` call.
- `update-nix-hashes.yml`: the single quotes in the credential helper are intentional and documented in place; SC2016 is silenced there with a shellcheck directive instead of a code change.

The remaining finding, the missing `labels` input for the Grafana action in `buildx.yml`, is a behavior defect and gets its own PR.

## Verification

- `actionlint` over `.github/`: only the `buildx.yml` `labels` finding remains.

None.

---

🤖 PR body drafted with [Claude Code](https://claude.com/claude-code) (Fable 5.1 low) — sponsored by [Claude for Open Source](https://www.anthropic.com/claude-for-open-source)


JakobLichterfeld (commented):
Reviewed against the actionlint output and the diff of each script.

- Every change is quoting, an `env` hand-off or a shellcheck directive; no command line changes its arguments.
- The two cache-cleanup loops keep their semantics: `read` over the key list instead of word splitting, and the empty-line guard means an empty list still issues no `gh cache delete`, as the `for` loop did.
- The SC2016 directive sits on the credential helper whose single quotes are intentional and documented in place; suppressing is the correct fix there.
- `github.head_ref` in `ghcr_purge.yml` is hygiene only: the job is gated on `is_pr_from_outside_repo == 'false'`, so no fork input reached the script before either.

`actionlint` over `.github/` now reports only the missing `labels` input in `buildx.yml`, which is a behavior defect and gets its own PR.

---

PR #5762, opened 2026-09-19
fix(ci): pass the build action inputs through env and expressions

JakobLichterfeld:
The composite build action had the same script hygiene the workflows lost in #5759, but actionlint does not check composite actions, so it stayed. Three fixes in `.github/actions/build/action.yml`, no change to the images:

- `inputs.version` was interpolated into the normalize script and the `$GITHUB_OUTPUT` redirect was unquoted. The value now arrives through `env`, the redirect is quoted.
- `inputs.labels` reached docker/metadata-action as the literal text `{{ inputs.labels }}` because the `${{ }}` was missing. No caller sets the input, so nothing changed for the images; the input now works.
- The build digest export reads its value through `env` as well, so no script in this action interpolates an expression.

Origin: #3432/#3445 (@jlestel) introduced the action with these three lines.

## Verification

- `nix run .#lint -- --ci`: 0 changed.
- No `${{ }}` left inside any `run:` block of the action.
- The PR CI skips the build jobs (path filter on `.github/`); runtime proof is a `workflow_dispatch` of "Build GHCR images" on this branch with both build jobs and the merge green.

- actionlint only covers `.github/workflows/`; the composite actions in `.github/actions/` are not linted by any tool in the repo.

---


JakobLichterfeld (commented):
Reviewed in a second session against the action file and the callers.

- The three changes are exactly the ones needed: `inputs.version` and the build digest reach their scripts through `env`, `$GITHUB_OUTPUT` is quoted, and `inputs.labels` is now a real expression instead of literal text.
- Behavior is unchanged for every current caller: none sets `labels`, the digest value is the same string, and the normalize output is identical.
- No `${{ }}` remains inside a `run:` block of the action; the merge action already followed this pattern.

---

PR #5765, opened 2026-09-20
fix(web): send the referrer and show the OpenStreetMap attribution on map tiles, so tiles load behind reverse proxies that set a stricter referrer policy such as same-origin or no-referrer and TeslaMate complies with the OSM tile usage policy

### LINKED ISSUE #5734: Map tile requests to tile.openstreetmap.org get blocked by OSM's tile usage policy enforcement
Image

### Description

  The live map on the Overview page fails to load tiles, showing:

  > Access blocked: App is not following the tile usage policy of OpenStreetMap's
  > volunteer-run servers: osm.wiki/Blocked

  ### Root cause (as far as I can tell)
 
  `assets/js/hooks.js` hardcodes the tile source with no `attribution` option:

  ```js
  const osm = new TileLayer("https://tile.openstreetmap.org/{z}/{x}/{y}.png", {
    maxZoom: 19,
  });

  OSM's Tile Usage Policy
  requires, among other things, visible attribution and a valid User-Agent/referrer
  identifying the consuming application. Because every self-hosted TeslaMate instance
  serves this same unattributed tile request straight from the end user's browser to
  tile.openstreetmap.org, OSM's automated policy enforcement appears to be
  fingerprinting and blocking the aggregate traffic pattern, not just my instance.
 
  Environment 

  - TeslaMate version: v4.2.0 (teslamate/teslamate:latest, current as of this issue)
  - Deployment: Docker Compose, reverse-proxied via Traefik
  - Browser: Brave 1.94.121 based on Chromium: 152.0.7977.83

### Expected Behavior

Map tiles load normally, with proper OSM attribution shown per their usage policy,  and/or a way to configure an alternate tile provider.

Yes

1. Open the TeslaMate Overview page.
2. Observe the map tile area shows the OSM block message instead of tiles.

### Relevant log output

```shell
No server-side errors are logged by TeslaMate itself — the map tile requests are made directly by the browser to tile.openstreetmap.org and never pass through the TeslaMate backend. Confirmed via `docker logs teslamate`: only nominatim.openstreetmap.org (reverse geocoding, server-side, working normally with 200 responses) appears in the logs — no tile.openstreetmap.org entries at all, since that request bypasses the app server entirely.
```

### Screenshots

_No response_

### Additional data

_No response_

### Type of installation

### Version

v4.2.0

### PostgreSQL version

v17.11

Yes (specify in next field)

Traefik (with CrowdSec bouncer + OAuth middleware in front)

JakobLichterfeld:
## Summary

Three users reported the OSM "Access blocked" tile on the overview map (#5734). The issue attributes it to missing attribution and OSM fingerprinting the aggregate traffic of all TeslaMate instances. That is not the cause.

**Cause.** `tile.openstreetmap.org` returns the 403 "Access blocked" tile for Chromium and Firefox user agents when the request carries no `Referer` header. With any `Referer`, tiles load. Safari user agents are tolerated without one, which explains why the map worked in Safari but not in Chrome for one reporter.

A stock TeslaMate is not affected: Phoenix's `put_secure_browser_headers` sets `referrer-policy: strict-origin-when-cross-origin`, so browsers send the instance's origin. All reporters run TeslaMate behind hardening reverse proxies (Traefik with CrowdSec/OAuth middleware, SWAG), which commonly override the header with `Referrer-Policy: no-referrer` and thereby strip the `Referer` from the tile request.

**Fix.**
- Set `referrerPolicy: "strict-origin-when-cross-origin"` on the Leaflet tile layer. Leaflet applies it as the `referrerpolicy` attribute on every tile ``, and an element-level referrer policy takes precedence over the document header, so tiles load even behind a `no-referrer` proxy.
- Add the OpenStreetMap attribution to the tile layer and stop hiding the attribution control on the compact card map. Visible attribution is required by the [OSM tile usage policy](https://operations.osmfoundation.org/policies/tiles/) regardless of this block; the missing attribution was a separate violation that could have led to a manual block ([Blocked tiles](https://wiki.openstreetmap.org/wiki/Blocked_tiles)).

Both the overview map and the geofence map go through `createMap`, so one change covers both. The Google hybrid layer is untouched.

Closes #5734

## Verification

- Tile `` elements carry `referrerpolicy="strict-origin-when-cross-origin"` and tile requests send a `Referer`, also with the document set to `Referrer-Policy: no-referrer` (temporarily via the router; images without the attribute sent no `Referer`, images with it sent the origin). Attribution is visible in fullscreen, light and dark theme. The compact card map was verified by attaching the geofence map container to a `.car.card` element at 175 px in DevTools, which exercises the changed selectors with real Leaflet controls: attribution renders bottom-right on one line without covering the map. The real overview page was not exercised locally (no vehicle in the dev database).

None. One reporter confirmed the diagnosis: a shared Traefik middleware set `Referrer-Policy: same-origin`, the tile request carried no `Referer`, and overriding the policy for the TeslaMate router fixed it (see #5734).

---


JakobLichterfeld (commented):
## Review

### `elixir/assets/js/hooks.js`

The OpenStreetMap `TileLayer` gains `attribution` and `referrerPolicy: "strict-origin-when-cross-origin"`. Leaflet 1.9.4 applies a string `referrerPolicy` as the `referrerpolicy` attribute on every tile ``, and an element-level policy takes precedence over the document policy delivered via header, so the fix holds behind a `no-referrer` proxy. Both maps (`SimpleMap` and the geofence map) go through `createMap`, so one change covers both. The Google hybrid layer is untouched, which is correct: Google does not require a referrer, and changing it is outside the scope of this fix.

### `elixir/assets/css/app.scss`

The rule hiding `.leaflet-control-attribution` on the compact card map and the matching `display: block` override in full-screen are removed. No other rule under `elixir/assets/css` hides the control; the dark theme keeps applying the invert filter it already applied in full-screen. The full-screen button sits top-right, the attribution control bottom-right, no overlap.

### Verification

Verified per the description: bundle contains the new options, Prettier clean, tile elements carry the `referrerpolicy` attribute and send a `Referer` even under a `no-referrer` document policy, attribution visible in fullscreen and on a DOM-simulated compact card (`.car.card` at 175 px), light and dark theme. Not exercised: the real overview page with a logged-in vehicle. The additional elements there (fullscreen button top-right, vehicle marker centred) do not touch the attribution corner bottom-right, so the CSS check transfers.

lgtm.

---

PR #5772, opened 2026-09-20
build(rust): choose Tokio as the async runtime

brianmay:
I think we will most likely need to have async rust support, so adding that now. This will influence future library choices for database, http, mqtt, etc.


JakobLichterfeld (approved):
Makes sense: fixing the runtime now decides which database, HTTP and MQTT crates are eligible later, and the change runs through build, clippy with `-D warnings` and test. First Rust-only PR too: the Elixir jobs were skipped, both Rust jobs ran.

I pushed two small things on top: a comment above the tokio line in `Cargo.toml` that records the runtime decision where the next dependency gets added, and the changelog line. Retitled to `build(rust): choose Tokio as the async runtime`, since it is a dependency and a decision rather than a feature.

PR #5776, opened 2026-09-21
Rust config

brianmay:
Need some config before we can proceed. Also Added basic version stuff. There is a compile time check to ensure that the version in VERSION matches the Cargo.toml file.

There are a number of key decisions here. Happy to revise as required.

For example, currently use environment variables for config, same as Elixir code. Config file - maybe YAML - also possible.


rust/src/config.rs
```diff
@@ -0,0 +1,10 @@
+use envconfig::Envconfig;
+
+#[derive(Envconfig, Default)]
+pub struct Config {
+    #[envconfig(from = "ADDOperandA", default = "1")]
+    pub operand_a: i32,
+
+    #[envconfig(from = "ADDOperandB", default = "2")]
+    pub operand_b: i32,
```
brianmay, line 9: Obviously this is dummy data, to be changed. I imagine having environment variables as similar as practical to the Elixir code.

rust/src/cli.rs
```diff
@@ -0,0 +1,10 @@
+use clap::Parser;
+
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[derive(Parser, Debug)]
+#[command(version = VERSION, about)]
+pub struct Args {
+    #[arg(short, long, default_value_t = false)]
+    pub verbose: bool,
```
brianmay, line 9: I imagine this will change in my next PR that improves on logging.

rust/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    let version_from_file = version_from_file.trim();
+    let version_from_cargo = env!("CARGO_PKG_VERSION");
+
+    if version_from_file != version_from_cargo {
+        eprintln!(
+            "Version mismatch! VERSION file has '{}' but Cargo.toml has '{}'",
+            version_from_file, version_from_cargo
+        );
+        std::process::exit(1);
+    }
```
brianmay, line 22: Requiring updating another file for version changes makes me uneasy. We could just leave Cargo.toml with a dummy value, but that makes me uneasy also. And updating Cargo.toml would be required should we ever decide to publish the crate to crates.io (means users can install with "cargo install teslamate").

brianmay (commented):
Highlighted some key aspects.

brianmay:
CI is broken because nix build doesn't have access to VERSION.

There are a number of potential solutions, undecided which approach we should take here. e.g. we could make Cargo.toml the authoritative source for the version for rust builds.

